PR #2, opened 2015-09-15
Updating README with information on how to configure a separate OPs E…

ewolinetz:
…S/Kibana cluster; renamed ops template to be more appropriate


ewolinetz:
@jcantrill PTAL

README.md
```diff
@@ -10,3 +10,7 @@ oadm policy add-cluster-role-to-user cluster-reader system:serviceaccount:defaul
 ```
 
 At this point you can create the stack using logging.yml.
+
+
+## Operations logging scenario
+If you would like to configure your EFK stack where logs from the openshift* and default namespaces and Syslog are sent to a separate ES cluster from application logs, you can use the same steps as above to create your example keys and secrets.  You will still need to configure your K8S_HOST_URL for fluentd, as you will have a single fluentd pod to service the two Elasticsearch/Kibana clusters.  However, you can use the `ops-logging.yml` file to create the additional services and replication controllers for you.  You can also create your OPs Elasticsearch and Kibana RC/Services using `logging.yml` separately, you will need to ensure that you specify the correct OPS_HOST and OPS_PORT for fluentd to connect to.
```
jcantrill: How about something like:

>The logging stack can be configured to send operations logs to a different ElasticSearch cluster then the one aggregating the application logs.  The operations logs are sourced from syslog and the openshift* and default namespaces.  A single fluentd pod services both ElasticSearch clusters and is required to be configured with the K8S_HOST_URL environment variable pointing to the master API server.  The `ops-logging.yml` template can be used to create the additional services and replication controllers to support this alternate deployment.  The `logging.yml` template can be used separately to create the resources (e.g. services, resource controllers) to support the opeations ElasticSearch cluster and Kibana instance.  Fluentd requires the OPS_HOST and OPS_PORT in order to send logs to the operations ElasticSearch instance.

There is still something amiss about the ops-logging.yml and the logging.yml.  It is not clear to me what the differences is and when to use one or the other or both.  It kind of reads that they do the same thing.

ewolinetz:
@jcantrill I realized I wasn't communicating well what I was trying to at all.  I rephrased the operations section so that it is hopefully more clear

jcantrill:
@ewolinetz I like this version much better

ewolinetz:
@jcantrill are there any other changes you would like to see made as part of this PR?  I don't have permissions to merge it myself :)

jcantrill:
please squash and  can merge

ewolinetz:
Squashed commits

PR #3, opened 2015-09-21
Add in origin-kibana plugin with main.less precompile

jcantrill:
@sosiouxme @ewolinetz  @themurph  please review and comment.  I was unable to figure out how to install dependencies and compile the LESS file without install/uninstalling nodejs...booo

### DIFF
--- a/kibana/Dockerfile
+++ b/kibana/Dockerfile
@@ -1,19 +1,33 @@
-FROM openshift/base-centos7
+FROM openshift/base-centos7 
 
 MAINTAINER Eric Wolinetz <ewolinet@redhat.com>
 
-ENV KIBANA_VER 4.1.1
+EXPOSE 5601
+
+ENV KIBANA_VER=4.1.1 \ 
+    ES_HOST=localhost \ 
+    ES_PORT=9200
 
 LABEL io.k8s.description="Kibana container for querying Elasticsearch for aggregated logs" \
   io.k8s.display-name="Kibana" \
   io.openshift.expose-services="5601:http" \
   io.openshift.tags="logging,elk,kibana"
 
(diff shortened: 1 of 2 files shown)


kibana/Dockerfile
```diff
@@ -1,19 +1,33 @@
-FROM openshift/base-centos7
+FROM openshift/base-centos7 
 
 MAINTAINER Eric Wolinetz <ewolinet@redhat.com>
 
-ENV KIBANA_VER 4.1.1
+EXPOSE 5601
```
ewolinetz: Do we want to expose port 5601 from the container?  Currently we let the template handle this and it maps it to port 80

sosiouxme: What's in the docker file is really only used as guidance in forming the pod and service. The pod won't expose anything unless so configured.

kibana/kibana.yml
```diff
@@ -19,8 +19,8 @@ kibana_index: ".kibana"
 # used by the Kibana server to perform maintence on the kibana_index at statup. Your Kibana
 # users will still need to authenticate with Elasticsearch (which is proxied thorugh
 # the Kibana server)
-kibana_elasticsearch_username: kibana
-kibana_elasticsearch_password: changeme
+#kibana_elasticsearch_username: kibana
```
ewolinetz: Uncomment out these two lines please, at least until we change how we do auth throughout the logging stack

ewolinetz:
@jcantrill one nit then LGTM

kibana/Dockerfile
ewolinetz: Ah, I misunderstood it's effect in the Dockerfile

PR #4, opened 2015-09-22
HA changes

ewolinetz:


ewolinetz:
@jcantrill @sosiouxme PTAL, this encompasses changes for https://trello.com/c/CWjwr3YI plus some cleaning up of Dockerfiles + the fluentd config

jcantrill:
@ewolinetz LGTM.  We should consider adding tests for the ruby code that is included here.  Maybe even breaking it into additional mixins.  @sosiouxme any opposition to merging?

elasticsearch/run.sh
```diff
@@ -1,7 +1,11 @@
 #! /bin/bash
 
-sed --in-place=.bak 's/searchguard.ssl.transport.http.enabled: true/searchguard.ssl.transport.http.enabled: false/' /usr/share/elasticsearch/config/elasticsearch.yml
-sed -i 's/searchguard.allow_all_from_loopback: false/searchguard.allow_all_from_loopback: true/' /usr/share/elasticsearch/config/elasticsearch.yml
+mkdir -p /elasticsearch/$CLUSTER_NAME
+ln -s /etc/elasticsearch/keys/searchguard.key /elasticsearch/$CLUSTER_NAME/searchguard_node_key.key
+
+sed --in-place=.bak 's/enabled: true/enabled: false/' /usr/share/elasticsearch/config/elasticsearch.yml
```
sosiouxme: @ewolinetz  What's the purpose of this - to disable clustering during prehook? Seems like it could be a bit too broad.

ewolinetz: To disable SSL during seeding.  Yeah, this is probably too generic, I'll change the config file and the script so it is more explicit of what this line updates.

fluentd/fluent.conf
```diff
@@ -135,18 +134,36 @@
     bearer_token_file /var/run/secrets/kubernetes.io/serviceaccount/token
     ca_file /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
   </filter>
+  <filter **kibana**>
```
sosiouxme: Is this filter just cruft?

fluentd/fluent.conf
```diff
@@ -135,18 +134,36 @@
...
 
   <match **default** **openshift**>
-    @type elasticsearch
+    @type elasticsearch_dynamic
     host "#{ENV['OPS_HOST']}"
     port "#{ENV['OPS_PORT']}"
     scheme https
-    logstash_format true
-    logstash_prefix operations
+    index_name operations.${Time.at(time).getutc.strftime(@logstash_dateformat)}
```
sosiouxme: Could it be `.operations` so it won't conflict with logs from project `operations`?

fluentd/fluent.conf
```diff
@@ -180,13 +196,22 @@
...
+  </filter>
   <match **>
-    @type elasticsearch
+    @type elasticsearch_dynamic
     host "#{ENV['OPS_HOST']}"
     port "#{ENV['OPS_PORT']}"
     scheme https
-    logstash_format true
-    logstash_prefix operations
+    index_name operations.${Time.at(time).getutc.strftime(@logstash_dateformat)}
```
sosiouxme: here too

fluentd/fluent.conf
ewolinetz: This filter is for changes specific to Kibana containers.  It does funky formatting of its messages, so this is to normalize it to be more like the other containers.

fluentd/fluent.conf
ewolinetz: Should work, I'll verify

fluentd/fluent.conf
```diff
@@ -135,18 +134,36 @@
...
     type flatten_hash
     separator _
   </filter>
+  <filter kubernetes.**>
+    type record_transformer
+    enable_ruby
+    <record>
+      hostname ${kubernetes_host}
+      message ${log}
+      @version 3.1
```
sosiouxme: Is this version at all significant?
Seems like something that ought to be set as an env var in the Dockerfile.

fluentd/fluent.conf
sosiouxme: Not sure I follow. Why does the message format matter? We're not going to reformat logs for arbitrary applications... can't, really. Just because it says it's kibana doesn't mean it is.

fluentd/fluent.conf
ewolinetz: This is part of the ECL initiative.  The idea for this is that if we make further fluentd config changes we would increment this version so that if there is an ES instance that persists multiple fluentd config versions we could use the @version tag so that we correlate data between versions (in the chance that there is no backwards compat).

fluentd/fluent.conf
ewolinetz: This is also part of the ECL initiative.  The filter in this case matches on the tag, which for this path is the [container] file name.  The thing we gain from this is that any log messages will all end up with the same key otherwise we would have 'message', 'msg', 'log', and 'err' that all contained messages we would want to correlate with.  

I'm not sure if there would be any more restrictive of a match that we can put here since anyone could name their container *kibana* and be matched here, I'll verify whether or not anything negative happens to application logs that aren't coming from kibana are filtered through here (e.g. loss of data).

ewolinetz: There was a loss of logs if someone had a container with 'kibana' in the name, addressed it in latest commit.

ewolinetz:
Addressed comments above, I'll squash changes once there are no more to make to this PR.

fluentd/fluent.conf
sosiouxme: OK, so it's kind of arbitrary. This is a totally minor nit but can it track
Origin's version? Since it's kinda upstream.

fluentd/fluent.conf
sosiouxme: We certainly want a consistent key; I just don't understand why kibana gets
specific treatment. Why doesn't this apply for every container's logs? They
are all coming out of Docker in the same format.

ewolinetz: I think the issue is how Kibana writes into its container logs.
{"log":"{\"name\":\"Kibana\",\"hostname\":\"kibana-1-1lezg\",\"pid\":8,\"level\":30,\"msg\":\"No existing kibana index found\",\"time\":\"2015-09-29T20:04:06.713Z\",\"v\":0}\n","stream":"stdout","time":"2015-09-29T20:04:06.716406777Z"}

It has a bunch of extra fields in there that we don't see with the other containers.
  ES: {"log":"[2015-09-29 19:43:28,333][INFO ][node                     ] [Vashti] initializing ...\n","stream":"stdout","time":"2015-09-29T19:43:28.334237592Z"}

  Fluentd:  {"log":"2015-09-29 19:57:01 +0000 [info]: reading config file path=\"/etc/fluent/fluent.conf\"\n","stream":"stdout","time":"2015-09-29T19:57:01.143011924Z"}

fluentd/fluent.conf
ewolinetz: Makes sense to me, I can do that

fluentd/fluent.conf
sosiouxme: So kibana is logging in JSONL format. Does fluentd detect that and handle
it differently? I would have expected fluentd to just pass it through as a
string. Or is the extra stuff coming from elasticsearch analysis of the
message?

jimmidyson: The kubernetes metadata plugin likes JSON logs because it (me) likes structured logs (events). Anything logged in JSON will be merged into the fluentd record so that fields will show up in Elasticsearch as separate fields rather than just a single string.

If you want to disable it add `merge_json_log false` to the kubernetes filter plugin config.

jimmidyson: It's not *kibana* that's being treated differently, it's log messages in JSON format.

sosiouxme: Neat. How does it detect when the logs are JSONL? Look for { at beginning?
Try to parse it?

I don't object to extracting more metadata, but the actual original line
should be available as the standard field for log message (whatever
docker's "log" gets translated to normally).

jimmidyson: See [here](https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/blob/master/lib/fluent/plugin/filter_kubernetes_metadata.rb#L170-L178) - checks first & last chars, parses.

It also removes the `log` field [here](https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/blob/master/lib/fluent/plugin/filter_kubernetes_metadata.rb#L175) - could remove that if you'd like.

sosiouxme: IMHO we just want to keep the 'log' key then.

PR #5, opened 2015-09-23
Improving ES installation process

ewolinetz:


ewolinetz:
@jcantrill @sosiouxme PTAL, reducing layers in ES install and making the ACL seeding smarter by checking if ES is up instead of just `sleep`ing

elasticsearch/Dockerfile
```diff
@@ -21,20 +21,17 @@ RUN rpm --import https://packages.elastic.co/GPG-KEY-elasticsearch && \
   yum install -y -- setopt=tsflags=nodocs \
   java-1.8.0-openjdk \
   elasticsearch && \
```
sosiouxme: add in `yum clean all` here so the downloaded RPMs are removed from the layer

sosiouxme:
Otherwise, LGTM!

ewolinetz:
added `yum clean all` to clean up after yum install completes

PR #7, opened 2015-10-11
Updating multiline regex - was not correctly matching single digit da…

ewolinetz:
…ys of the month

Examples of lines to match:
Jan 26 20:20:39 localhost kernel: Booting paravirtualized kernel on Xen HV

Oct  1 23:02:28 ip-10-69-169-224 systemd: Started Session 1 of user centos.

@jcantrill @sosiouxme  PTAL

### DIFF
--- a/fluentd/fluent.conf
+++ b/fluentd/fluent.conf
@@ -111,7 +111,7 @@
   pos_file /var/log/node.log.pos
   tag system.var.log
   format multiline
-  format_firstline /^[A-Z][a-z]{2} [0-3]+[0-9] [0-2][0-9]:[0-5][0-9]:[0-5][0-9] /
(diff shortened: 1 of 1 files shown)


jcantrill:
@ewolinetz looks good to me.  Do we have any way of creating tests for the various bits of this repo?

ewolinetz:
@jcantrill I'm not sure if there is a particular way to test the different parts [in an automated way] - that might be dependent on the deployer which unfortunately would require most of this work to be built into their respective RPMs.  We might be able to do the following:
1. Smoke tests - make sure everything starts up and can connect with the other components
2. Verify that the log messages we have in /var/log/message* and /var/log/containers/* end up in Kibana with no losses of lines
3. Other than that, a lot of additional functionality we have is coming from our plugins..

Let me know if this blocks merging this, otherwise please feel free.

jcantrill:
@ewolinetz not a blocker for me

PR #9, opened 2015-10-14
Updating the matching in the fluentd conf file

ewolinetz:
to match our namespaces more explicitly @sosiouxme @jimmidyson PTAL

### DIFF
--- a/fluentd/fluent.conf
+++ b/fluentd/fluent.conf
@@ -158,7 +158,7 @@
     remove_keys log,stream
   </filter>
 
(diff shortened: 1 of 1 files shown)


ewolinetz:
We want to be more deliberate with what we are matching so that we do not grab logs for a pod with a name like 'abcdefaultxyz'.  Tag reads as podname_namespace_**

sosiouxme:
@portante no, log entry is locked down to only our fluentd agent.

LGTM, merge away

PR #14, opened 2015-10-28
Updating version of openshift-elasticsearch-plugin and removing seedi…

ewolinetz:
…ng from run.sh

@jcantrill @sosiouxme PTAL for after the new version of openshift-elasticsearch-plugin is published to maven central


jcantrill:
LGTM merging

sosiouxme:
LGTM 2

PR #16, opened 2015-10-30
logging and plugins

sosiouxme:


sosiouxme:
@jcantrill @ewolinetz PTAL

elasticsearch/logging.yml
```diff
@@ -6,9 +6,9 @@ logger:
   action: WARN
   # reduce the logging for aws, too much is logged under the default INFO
   com.amazonaws: WARN
-  #io.fabric8.elasticsearch: DEBUG
-  #io.fabric8.kubernetes: DEBUG
-  #io.fabric8.elasticsearch.plugin.acl: DEBUG
+  io.fabric8.elasticsearch: ${PLUGIN_LOGLEVEL}
```
ewolinetz: Setting the log level for this package should also set the log level for io.fabric8.elasticsearch.plugin.acl; we should either let you set them separately, or remove line 11

ewolinetz:
one nit otherwise LGTM

elasticsearch/logging.yml
sosiouxme: Thanks. Durrrr :)

sosiouxme:
I added on an improvement to the RAM allocation calculator. Now it won't bork on 1G or under-allocate odd Gigs.

ewolinetz:
LGTM

PR #21, opened 2015-11-13
Adding flags to keep time field and update to match multiple spaces i…

ewolinetz:
…n time field

@sosiouxme @jcantrill PTAL, not sure how I let this slip past me...

### DIFF
--- a/fluentd/fluent.conf
+++ b/fluentd/fluent.conf
@@ -104,6 +104,27 @@
   log_level warn
 </system>
 
+#
+#  This <source> looks to match messages from /var/log/messages*.  
+#
+#  The following are examples of what should be matched:
+#  1. Sep 10 22:19:28 localhost kernel: Blkfront and the Xen platform PCI driver have been compiled for this kernel: unplug emulated disks.
+#     You might have to change the root device
+#     from /dev/hd[a-d] to /dev/xvd[a-d]
+#     in your root= kernel command line option
+#
+#  2. Nov 6 11:24:03 dhcp-32-77 dbus[1236]: [system] Activating via systemd: service name='org.freedesktop.hostname1' unit='dbus-org.freedesktop.hostname1.service'
+#
+#  3. Aug 17 08:11:18 tcp9-54-255-137-97.example.com NetworkManager[1257]: <info>  VPN service 'libreswan' started (org.freedesktop.NetworkManager.libreswan), PID 27699
+#
+#
+#  The following are examples of what should NOT be matched:
+#  1. December 24 08:11:35 dhcp-22-55 fprintd: ** Message: No devices in use, exit
+#
+#  2. Jan 22  6:1:2 dhcp-36-79 kernel: IPv4 over IPSec tunneling driver
+#
+#  3. Apr8 15:20:01 dhcp-42-42 systemd: Created slice user-0.slice.
+#   
 <source>
   @type tail
   @label @SYSTEM
@@ -111,9 +132,13 @@
   pos_file /var/log/node.log.pos
   tag system.var.log
   format multiline
-  format_firstline /^[A-Z][a-z]{2} [ |0-3]+[0-9] [0-2][0-9]:[0-5][0-9]:[0-5][0-9] /
-  format1 /^(?<time>[^ ]* [^ ]* [^ ]*) (?<host>[^ ]*) (?<ident>[a-zA-Z0-9_\/\.\-]*)(?:\[(?<pid>[0-9]+)\])?[^\:]*\: *(?<message>.*)$/
+  # Begin possible multiline match: "Mmm DD HH:MM:SS "
+  format_firstline /^[A-Z][a-z]{2}\s+[0-3][0-9]\s+[0-2][0-9]:[0-5][0-9]:[0-6][0-9]\s/
+  # extract metadata from same line that matched format_firstline
+  format1 /^(?<time>\S+\s+\S+\s+\S+)\s+(?<host>\S+)\s+(?<ident>[\w\/\.\-]*)(?:\[(?<pid>[0-9]+)\])?[^\:]*\:\s*(?<message>.*)$/
+  time_format %b %d %H:%M:%S
   read_from_head true
+  keep_time_key true
 </source>
 
 <source>
@@ -201,6 +226,7 @@
     enable_ruby
     <record>
       hostname ${host}
+      time ${Time.at(time).to_datetime.to_s}
       version 1.0.6
     </record>
     remove_keys host


ewolinetz:
@portante That is the case, the issue was that the format1 wasn't accounting for multiple spaces in the time like how format_firstline was.  

The `keep_time_key true` line is to address the issue that richm and @thoraxe noticed where we weren't seeing logs from the nodes appear in Kibana (we weren't sending a time field with these logs..)

jcantrill:
It's all regex to me.... but LGTM

ewolinetz:
@portante this regex is a little different than the normal format that syslog has; the syslog regex actually specifies its fields to read in are between '^' and '$' and I removed the trailing '$'.  With the multiline match this allows me to pull in anything past the initial columns (time, pid, hostname, etc) and allow 'message' to pull from multiple lines.

What format_firstline tells fluentd is what the very first line of a message looks like, after that it will pull in fields based on the format# lines that follow it.  That's why the format_firstline is more specific of a patten; I could update my format1 regex, but decided to keep it closer to the one that fluentd provides for easier comparison of the two.  This was my way to get around multi-line messages that only had one timestamp.

E.g.
Sep 10 22:19:28 localhost kernel: Blkfront and the Xen platform PCI driver have been compiled for this kernel: unplug emulated disks.
You might have to change the root device
from /dev/hd[a-d] to /dev/xvd[a-d]
in your root= kernel command line option

portante:
@ewolinetz, that seems a bit fragile, but maybe this will work fine.  It seems really hard to maintain, since a fix to one regex could invalidate the other.

What is your rsyslog.conf file on these hosts?  I do not think multi-line log messages should happen from the kernel if they are flowing through the journal.

ewolinetz:
@portante it probably is too fragile, one of the things we are looking to do to harden our solution is keeping the logs in json format.  When we do that this configuration file will be updated accordingly.

portante:
@ewolinetz that is a good idea.  Rsyslog can do that today, trivially if you are interested in trying that out to make it easier to have fluentd parse them.

portante:
For example of a complete rsyslog.conf that dumps everything from the journal as JSON:
```
module(load="imjournal" StateFile="/var/lib/rsyslog/imjournal.state")
global(workDirectory="/var/lib/rsyslog" preserveFQDN="on")
template(name="MyTempl" type="list") {
    property(name="$!all-json")
    constant(value="\n")
}
action(type="omfile" template="MyTempl" file="/var/log/messages.json")
```

portante:
@ewolinetz, why not have a `format2` which matches anything and a `format1` that matches a timestamp line?

Won't that work with single lines too?

ewolinetz:
@portante I'm not sure if that would work with single lines, it'd be something to test.  
At the same time, I don't think we'd gain anything by having a format2 since we aren't parsing out anything different on another line, we just want to extend what is picked up for `message`.

portante:
@ewolinetz I think we gain clarity of how the configuration is intended to work.  It will likely be easier to maintain for folks later who have to pick this up when there is a problem or new consideration.

ewolinetz:
@portante there is an example for this configuration that I borrowed from the fluentd documentation (http://docs.fluentd.org/v0.10/articles/in_tail).  I apologize, there isn't an anchor for this example so I'm pasting it below.  It's also worth noting that if/when we update fluentd's log collection strategy, we would no longer need to use multiline.

One more example, you can parse Java like stacktrace logs with multiline. Here is a configuration example.
```
format multiline
format_firstline /\d{4}-\d{1,2}-\d{1,2}/
format1 /^(?<time>\d{4}-\d{1,2}-\d{1,2} \d{1,2}:\d{1,2}:\d{1,2}) \[(?<thread>.*)\] (?<level>[^\s]+)(?<message>.*)/
```

If you have a following log:
```
2013-3-03 14:27:33 [main] INFO  Main - Start
2013-3-03 14:27:33 [main] ERROR Main - Exception
`javax.management.RuntimeErrorException: null
    at Main.main(Main.java:16) ~[bin/:na]
2013-3-03 14:27:33 [main] INFO  Main - End
```

It will be parsed as:
```
2013-03-03 14:27:33 +0900 zimbra.mailbox: {"thread":"main","level":"INFO","message":"  Main - Start"}
2013-03-03 14:27:33 +0900 zimbra.mailbox: {"thread":"main","level":"ERROR","message":" Main - Exception\njavax.management.RuntimeErrorException: null\n    at Main.main(Main.java:16) ~[bin/:na]"}
2013-03-03 14:27:33 +0900 zimbra.mailbox: {"thread":"main","level":"INFO","message":"  Main - End"}
```

portante:
@ewolinetz, thanks, that helps, so that `format_firstline` pattern matching the `format1` in that example is what I was looking at, and why I was suggesting that if you don't want to have patterns that are the same between `format_firstline` and `format1` that a `format2` might add clarity.  I did not realize that that would then mean you cannot then append those following lines to the message.  Thanks!

fluentd/fluent.conf
```diff
@@ -112,8 +112,10 @@
   tag system.var.log
   format multiline
   format_firstline /^[A-Z][a-z]{2} [ |0-3]+[0-9] [0-2][0-9]:[0-5][0-9]:[0-5][0-9] /
-  format1 /^(?<time>[^ ]* [^ ]* [^ ]*) (?<host>[^ ]*) (?<ident>[a-zA-Z0-9_\/\.\-]*)(?:\[(?<pid>[0-9]+)\])?[^\:]*\: *(?<message>.*)$/
+  format1 /^(?<time>[^ ]*[ ]*[^ ]*[ ]*[^ ]*) (?<host>[^ ]*) (?<ident>[a-zA-Z0-9_\/\.\-]*)(?:\[(?<pid>[0-9]+)\])?[^\:]*\: *(?<message>.*)$/
```
sosiouxme: Could you include a comment indicating examples of what you do and don't want to match here? It would help evaluate whether the regex is doing the right thing.

Also \s (space) and \S (non-space) would be easier to read than the bracket notation (for one versed in regex at least). And I wonder if you want + instead of * ?

ewolinetz: I definitely want + instead of * for the white space... or were you thinking more replacing every instance of * with +?
I also should update line 114 to remove +

I can also update to use \s and \S.  I wasn't sure if it would make it easier for most to read or not

sosiouxme: On Tue, Nov 17, 2015 at 10:43 AM, Eric Wolinetz <notifications@github.com>
wrote:

> I definitely want + instead of * for the white space... or were you
> thinking more replacing every instance of * with +?
>
I don't really know, that's part of why I wanted a comment with examples :)
Using a * means it could match even if it's not there; using them in
sequence "smells" funny because it's not very specific about what you're
matching. Using a + asserts that you do actually want to make sure it
matches something.

> I also should update line 114 to remove +
>
> I can also update to use \s and \S. I wasn't sure if it would make it
> easier for most to read or not
>
It also has the benefit of considering tabs as spaces, in case anything
uses them.

>

ewolinetz:
@sosiouxme added some examples and cleaned up regex patterns

fluentd/fluent.conf
```diff
@@ -111,9 +132,11 @@
   pos_file /var/log/node.log.pos
   tag system.var.log
   format multiline
-  format_firstline /^[A-Z][a-z]{2} [ |0-3]+[0-9] [0-2][0-9]:[0-5][0-9]:[0-5][0-9] /
-  format1 /^(?<time>[^ ]* [^ ]* [^ ]*) (?<host>[^ ]*) (?<ident>[a-zA-Z0-9_\/\.\-]*)(?:\[(?<pid>[0-9]+)\])?[^\:]*\: *(?<message>.*)$/
+  format_firstline /^[A-Z][a-z]{2} [ |0-3][0-9] [0-2][0-9]:[0-5][0-9]:[0-5][0-9] /
```
sosiouxme: We're making some assumptions about the format here. For instance that the month will be 3 letters - not sure that's true in all locales. Also I'm pretty sure timestamp can take quite a variety of formats. But I guess for now it's enough to cover the default. Still, a couple of nits:

1. Whenever I match space in a regex I prefer \s+ so I don't have to think about how much or what kind of whitespace to match. The only problem is if this ends up matching newlines, but I wouldn't think in this context that would happen.
2. Leap seconds look like `:60` :)

So might I propose this:
```
# Begin possible multiline match: "Mmm DD HH:MM:SS "
format_firstline /^[A-Z][a-z]{2}\s+[0-3][0-9]\s+[0-2][0-9]:[0-5][0-9]:[0-6][0-9]\s/
```

fluentd/fluent.conf
```diff
@@ -111,9 +132,11 @@
...
+  format1 /^(?<time>\S*\s+\S*\s+\S*) (?<host>\S*) (?<ident>[a-zA-Z0-9_\/\.\-]*)(?:\[(?<pid>[0-9]+)\])?[^\:]*\: *(?<message>.*)$/
```
sosiouxme: This is reasonably clear if you know how the multiline format works, not so much if you don't :) Also echoing the above concerns re `\s+` as well as noting that `\S*` could theoretically match nothing (we know it will work because format_firstline matched, but still... best to be explicit). So I'd suggest another brief comment as well as minor adjustments to the regex:

```
# extract metadata from same line that matched format_firstline
format1 /^\s*(?<time>\S+\s+\S+\s+\S+)\s+(?<host>\S+)\s+(?<ident>[\w\/\.\-]*)(?:\[(?<pid>[0-9]+)\])?[^\:]*\:\s*(?<message>.*)$/
```

ewolinetz: Since we are denoting that the first line begins `^[A-Z][a-z]{2}` should we start format1 with `^\s*` ?

sosiouxme: Oh, good point. They should be consistent.
If we're assured the line doesn't start with whitespace then we can leave it out. I just don't like having to check assumptions like that.

ewolinetz:
addressed nits re: firstline and format1 regex patterns

ewolinetz:
squashed changes

Issue #25, opened 2015-11-18
Fluentd will blindly use hostname from syslog


ewolinetz:
Currently Fluentd will just use the value of the third column to populate the value of 'hostname', even if it is something like 'localhost'.  We should probably replace this with the FQDN

PR #26, opened 2015-11-20
checking if hostname is "localhost" and using value of hostname comma…

ewolinetz:
…nd if it is

Updating the deployer fluentd template to mount the hosts /etc/hostname and updating fluentd to read from it if hostname is localhost in logs

This addresses #25 

@sosiouxme @jcantrill  PTAL

### DIFF
--- a/fluentd/fluent.conf
+++ b/fluentd/fluent.conf
@@ -227,7 +227,7 @@
     <record>
       # if host == 'localhost' do a fqdn lookup
       ## we pull the hostname from the host's /etc/hostname mounted at /etc/docker-hostname to correctly identify the hostname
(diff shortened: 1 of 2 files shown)


fluentd/fluent.conf
```diff
@@ -225,7 +225,9 @@
     type record_transformer
     enable_ruby
     <record>
-      hostname ${host}
+      # if host == 'localhost' do a fqdn lookup
+      ## we pull the hostname from the host's /etc/hostname mounted at /etc/docker-hostname to correctly identify the hostname
+      hostname ${host.eql?('localhost') ? (File.open('/etc/docker-hostname') { |f| f.readline }.rstrip || host) : host}
```
ewolinetz: the null check condition i'm using here doesn't actually do what i was hoping it would do, which is to use host if we cant get a value from /etc/docker-hostname for some reason.

sosiouxme:
LGTM

Issue #24, opened 2015-11-18
Setting ENABLE_OPS_CLUSTER does not set ES_HOST correctly for kibana-ops DC


ewolinetz:
Within the deployer script when set ENABLE_OPS_CLUSTER to true, we are not setting the value ES_HOST to point to the operations ElasticsSearch service.

sosiouxme:
I think from fluentd https://github.com/openshift/origin-aggregated-logging/blob/master/deployment/run.sh#L187 handles it. Are you saying kibana-ops ES_HOST isn't pointed at the ops ES? Maybe https://github.com/openshift/origin-aggregated-logging/blob/master/deployment/run.sh#L188 needs adjustment.

ewolinetz:
Correct, the value of ES_HOST for kibana-ops is incorrect.  That's the line I was looking at as well :+1:

PR #27, opened 2015-11-20
updating to ensure we set the ES_HOST of the kibana-ops to be logging…

ewolinetz:
…-es-ops

This addresses #24 

@sosiouxme PTAL


sosiouxme:
LGTM except this really should have been caught sooner...

PR #29, opened 2015-12-03
Add logout redirect when logging out of kibana

jcantrill:
cc @sosiouxme @ewolinetz  please review

### DIFF
--- a/deployment/templates/kibana.yaml
+++ b/deployment/templates/kibana.yaml
@@ -109,6 +109,9 @@ objects:
                   name: "OAP_PUBLIC_MASTER_URL"
                   value: ${OAP_PUBLIC_MASTER_URL}
                 -
+                  name: "OAP_LOGOUT_REDIRECT"
+                  value: ${OAP_PUBLIC_MASTER_URL}/console/logout
+                -
                   name: "OAP_MASTER_CA_FILE"
                   value: "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
                 -


sosiouxme:
@jcantrill so you *do* want to log the user out from console if they click the kibana logout link? I guess that's one way, it's just a little inconsistent as it won't happen in the other direction - you'll still be logged in at kibana with a separate token after directly logging out at the console.

jcantrill:
@sosiouxme I had a conversation with @liggitt about this because of another issue I was investigating.  He probably did bring this up, but im not certain how to handle the case you describe.  I am almost of the opinion that you should only get to kibana by way of the web console.  Something along the lines of when you visit kibana directly that it redirects you to the web console if the original link does not have container info.  This would resolve, I think, the scenario you describe and also the case where, you see the name of the container you last visited if you first come from the console, and then go to kibana directly

jcantrill:
@sosiouxme Also, I find 'logout' of kibana only a little odd user experience as well; seems like it should be log out completely or not at all.

sosiouxme:
@jcantrill i agree the experience suffers.

I can easily add the redirect to console logout  upon kibana logout. The token for kibana will be different from the one the console has, unless they got to it via console link. You're suggesting making the console link the only way to get a kibana session; that's doable, but it is kind of a lousy experience when the user's token expires and they have to go back through the console to find where they were. It feels like no matter what we do, there will be some flaws in the UX. The only complete fix I can think of is to have stateful SSO so that the user is always using the same token wherever they're logged in.

liggitt:
> The only complete fix I can think of is to have stateful SSO so that the user is always using the same token wherever they're logged in.

Or have kibana always destroy the token it is using, and redirect to /console/logout (which will no-op if the token is already destroyed)

sosiouxme:
On Mon, Dec 7, 2015 at 7:47 AM, Jordan Liggitt <notifications@github.com>
wrote:

> Or have kibana always destroy the token it is using, and redirect to
> /console/logout (which will no-op if the token is already destroyed)
>
>
> That part works fine, but if they're on the console and log out there, it
won't destroy kibana's token (unless it's the same token for both).

liggitt:
True. We're not likely to link all the OAuth tokens together, either

jcantrill:
Maybe the link should say nothing about logout and just be a link to th we
console. And returning or navigating to the console by default destroys
only your Kibana session

On Monday, December 7, 2015, Jordan Liggitt <notifications@github.com>
wrote:

> True. We're not likely to link all the OAuth tokens together, either
>
> —
> <https://github.com/openshift/origin-aggregated-logging/pull/29#issuecomment-162520426>
> .
>


-- 
--
Jeff Cantrill
Red Hat, Inc.
jcantril@redhat.com
http://www.redhat.com

PR #30, opened 2015-12-09
Created initial E2E script…

ewolinetz:
... for verifying everything is configured and running before validating data

### DIFF
--- a/hack/testing/check-EFK-running.sh
+++ b/hack/testing/check-EFK-running.sh
@@ -1,0 +1,200 @@
+#!/bin/bash
+
+if [[ $VERBOSE ]]; then
+  set -ex
+else
+  set -e
+fi
+
+ADDITIONAL_PODS=$((KIBANA_CLUSTER_SIZE + ES_CLUSTER_SIZE - 2))
+
+EXIT_CODE=0
+
+if [[ $# -ne 1 ]]; then
+  # assuming not using OPS cluster
+  CLUSTER="false"
+else
+  CLUSTER="$1"
+  ADDITIONAL_PODS=$((ADDITIONAL_PODS + KIBANA_OPS_CLUSTER_SIZE + ES_OPS_CLUSTER_SIZE - 2))
+fi
+
+if [[ "$CLUSTER" == "true" ]]; then
+  NEEDED_COMPONENTS=("logging-es-[a-ZA-Z0-9]+?0" "logging-kibana0" "logging-fluentd0" "logging-es-ops-[a-zA-Z0-9]+?0" "logging-kibana-ops0")
+else
+  NEEDED_COMPONENTS=("logging-es-[a-zA-Z0-9]+?0" "logging-kibana0" "logging-fluentd0")
+fi
+
+TEST_DIVIDER="-------------------------------------------------------"
+NEEDED_STREAMS=("logging-fluentd" "logging-elasticsearch" "logging-auth-proxy" "logging-kibana")
+COMPONENTS_COUNT=${#NEEDED_COMPONENTS[@]}
+STREAMS_COUNT=${#NEEDED_STREAMS[@]}
+
+echo "Checking component installation and if pods are running:"
+# Check that we have all IS
+echo $TEST_DIVIDER
+
+FOUND_STREAMS=(`oc get is | grep 'logging-' | cut -d" " -f 1`)
+IS_COUNT=${#FOUND_STREAMS[@]}
+IS_MESSAGE="[$IS_COUNT/$STREAMS_COUNT] image streams found."
+
+if [[ $IS_COUNT -ne $STREAMS_COUNT ]]; then
+  echo "Error - $IS_MESSAGE"
+  EXIT_CODE=1
+
+  # check for elasticsearch, kibana, auth-proxy, and kibana images
+  for stream in "${NEEDED_STREAMS[@]}"; do
+    if [[ ! ( ${FOUND_STREAMS[@]} =~ $stream ) ]]; then
+      echo " ! image stream $stream is missing..."
+    fi
+  done
+
+  echo "* Please rerun \`oc process logging-support-template | oc create -f -\` to generate missing image streams."
+else
+  echo "Success - $IS_MESSAGE"
+fi
+
+echo $TEST_DIVIDER
+# Check that we have DC
+
+FOUND_DC=(`oc get dc | grep 'logging-' | awk '{print $1 "0"}'`)
+DC_COUNT=${#FOUND_DC[@]}
+DC_MESSAGE="[$DC_COUNT/$COMPONENTS_COUNT] deployment configs found."
+
+if [[ $DC_COUNT -ne $COMPONENTS_COUNT ]]; then
+  echo "Error - $DC_MESSAGE"
+  EXIT_CODE=1
+
+  # check which DC are missing
+  for dc in "${NEEDED_COMPONENTS[@]}"; do
+    if [[ ! ( ${FOUND_DC[@]} =~ $dc ) ]]; then
+
+      PRINTED_DC=`echo $dc | cut -d"[" -f 1 | rev | cut -c 2- | rev`
+      echo " ! deployment config for $PRINTED_DC is missing..."
+    fi
+  done
+
+  echo "* Please rerun the deployer to generate missing deployment configs."
+else
+  echo "Success - $DC_MESSAGE"
+fi
+
+echo $TEST_DIVIDER
+# Check that we have RC
+
+FOUND_RC=(`oc get rc | grep 'component' | grep 'logging-' | cut -d" " -f 1 | rev | cut -d"-" -f 2- | rev | sort | uniq | awk '{print $1 "0"}'`)
+RC_COUNT=${#FOUND_RC[@]}
+RC_MESSAGE="[$RC_COUNT/$COMPONENTS_COUNT] unique replication controllers found."
+
+if [[ $RC_COUNT -ne $COMPONENTS_COUNT ]]; then
+  echo "Error - $RC_MESSAGE"
+  EXIT_CODE=1
+
+  # check which RC are missing
+  for rc in "${NEEDED_COMPONENTS[@]}"; do
+    if [[ ! ( ${FOUND_RC[@]} =~ $rc ) ]]; then
+      PRINTED_RC=`echo $rc | cut -d"[" -f 1 | rev | cut -c 2- | rev`
+      echo " ! unique replication controller for $PRINTED_RC is missing..."
+    fi
+  done
+
+  #TODO: there is another way to generate the RC from a DC... update message to use that *'if able, otherwise'
+  echo "* Please rerun the deployer or redeploy the appropriate DC to generate missing replication controllers."
+else
+  echo "Success - $RC_MESSAGE"
+fi
+
+echo $TEST_DIVIDER
+# Check that we have Routes
+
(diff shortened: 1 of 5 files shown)


hack/testing/check-EFK-running.sh
```diff
@@ -0,0 +1,171 @@
...
+  echo "Success - $RC_MESSAGE"
+fi
+
+echo $TEST_DIVIDER
+# Check that we have Routes
+# TODO: add logic for if we have an Ops cluster deployment
+
+# we add a '0' to deal with false positives of 'kibana' matching 'kibana' and 'kibana-ops' when checking what is found
+NEEDED_ROUTES=("kibana0" "kibana-ops0")
+FOUND_ROUTES=(`oc get routes | grep 'kibana' | cut -d" " -f 1 | awk '{print $1 "0"}'`)
```
ewolinetz: cut and awk are redundant here, remove cut

hack/testing/check-EFK-running.sh
```diff
@@ -0,0 +1,171 @@
...
+  echo "Success - $ROUTE_MESSAGE"
+fi
+
+echo $TEST_DIVIDER
+# Check that we have Services
+# TODO: add logic for if we have an Ops cluster deployment?
+
+# we add a '0' to deal with false positives of when checking what is found, similar to what we do for routes
+NEEDED_SERVICES=("logging-es0" "logging-es-cluster0" "logging-ops0" "logging-ops-cluster0" "logging-kibana0" "logging-kibana-ops0")
+FOUND_SERVICES=(`oc get svc | grep 'logging-' | cut -d" " -f 1 | awk '{print $1 "0"}'`)
```
ewolinetz: cut and awk are redundant here, remove cut

hack/testing/check-EFK-running.sh
```diff
@@ -0,0 +1,171 @@
...
+
+FOUND_STREAMS=(`oc get is | grep 'logging-' | cut -d" " -f 1`)
+IS_COUNT=${#FOUND_STREAMS[@]}
+IS_MESSAGE="[$IS_COUNT/$STREAMS_COUNT] image streams found."
+
+if [[ $IS_COUNT -ne $STREAMS_COUNT ]]; then
+  echo "Error - $IS_MESSAGE"
+
+  # check for elasticsearch, kibana, auth-proxy, and kibana images
+  for stream in "${NEEDED_STREAMS[@]}"; do
```
jcantrill: Maybe early optimization, but you have some repeated logic here it looks like that is used below and could maybe be moved to a function.  Might also help clarify what is going on in the test.  Also, depending what @mwringe is doing for metrics e2e, maybe an opportunity for reuse.

jcantrill:
This is mostly deployment testing, but I wonder if there is a way to validate logs are being collected and viewable.  Might be an enhancement to work in a headless browser or something to see if we could retrieve logs via kibana (if that is even possible)

ewolinetz:
@jcantrill that is the plan; we would most likely have multiple scripts, this first one would be to verify things are installed and running correctly, and the second one would try to validate the actual log entries in ES and Kibana (possibly leverage Selenium to do our browser-based checks?)

ewolinetz:
Using grep vs checking for message in the log files using go is orders of magnitude faster for large numbers of messages.  So opting to use that strategy.

@sosiouxme @jcantrill PTAL

ewolinetz:
Some perf numbers:
```
QUERY_SIZE=500 ./e2e-testing.sh true

Non-grep:
Failure - [451/500] log entries found in /var/log/containers/*_logging_*.log
Comparision of messages took 1.267653347s

Success - [500/500] log entries found in /var/log/messages*
Comparision of messages took 5m46.053883781s


Grep:
Success - [500/500] log entries found in /var/log/containers/*_logging_*.log
Comparision of messages took 1.891256816s

Success - [500/500] log entries found in /var/log/messages*
Comparision of messages took 18.781591857s



QUERY_SIZE=5000 ./e2e-testing.sh true

Non-grep:
Failure - [685/986] log entries found in /var/log/containers/*_logging_*.log
Comparision of messages took 4.855004116s

Success - [5000/5000] log entries found in /var/log/messages*
Comparision of messages took 1h1m45.869733704s


Grep:
Success - [986/986] log entries found in /var/log/containers/*_logging_*.log
Comparision of messages took 4.107102423s

Success - [5000/5000] log entries found in /var/log/messages*
Comparision of messages took 3m11.343944438s
```

hack/testing/check-EFK-running.sh
```diff
@@ -0,0 +1,193 @@
+#!/bin/bash
+
```
sosiouxme: It's good form in bash scripts to output the commands as they're run and to fail if any commands fail (avoid silent bugs).

    set -ex

http://www.peterbe.com/plog/set-ex

Requires a little more rigor but can save a lot of headaches.

hack/testing/check-EFK-running.sh
```diff
@@ -0,0 +1,193 @@
...
+ADDITIONAL_PODS=$((((KIBANA_CLUSTER_SIZE + ES_CLUSTER_SIZE)) - 2))
```
sosiouxme: `$((KIBANA_CLUSTER_SIZE + ES_CLUSTER_SIZE - 2))` ought to be fine...

hack/testing/check-EFK-running.sh
mwringe: I am not a huge fan of -x, I find it makes the output a bit too verbose by default. But I do find it can be useful to enable it with a verbose flag.

You may also want to set -u (variables have to be declared) and '-o pipefail' as extra precautions

sosiouxme: Guess for me it depends if you're ever going to look at the output. For tests I would expect you don't, unless it breaks.

ewolinetz: I'll probably go with the approach of a way to make it verbose with a flag/env var.  I wouldn't expect someone to need to see the output by default (makes it slightly difficult to see the test results)

sosiouxme:
Couple things I ran into:

```
-------------------------------------------------------
Checking for log entry matches between ES and their sources...
   found 1 index(es) [.operations]
------------------------------------------
panic: exit status 1

goroutine 1 [running]:
main.main()
        /data/src/github.com/openshift/origin-aggregated-logging/hack/testing/check-logs.go:35 +0x478
exit status 2
------------------------------------------
```
Not sure what that was about.

Then I redeployed Kibana and -

```
NAME                          READY     STATUS      RESTARTS   AGE
logging-kibana-2-4dx4q        2/2       Running     0          1m
-------------------------------------------------------
Error - [4/3] replication controllers found.
* Please rerun the deployer to generate missing replication controllers.
-------------------------------------------------------
```
Name is too specific there. Redeploying DCs is definitely a use case.

hack/testing/check-EFK-running.sh
```diff
@@ -0,0 +1,193 @@
...
+  echo "* Please rerun \`oc process logging-support-template | oc create -f -\` to generate missing routes."
+else
+  echo "Success - $ROUTE_MESSAGE"
+fi
+
+echo $TEST_DIVIDER
+# Check that we have Services
+
+# we add a '0' to deal with false positives of when checking what is found, similar to what we do for routes
+NEEDED_SERVICE=("logging-es0" "logging-es-cluster0" "logging-ops0" "logging-ops-cluster0" "logging-kibana0" "logging-kibana-ops0")
```
sosiouxme: Re `logging-ops` ITYM `logging-es-ops`

ewolinetz: definitely meant `logging-es-ops` ...

ewolinetz:
@sosiouxme the RC error reported is the script being confused by the previous RC sticking around.  What does your `oc get rc` look like?  I probably can do a `sort | uniq` after dropping the iteration number

sosiouxme:
@ewolinetz 
```
CONTROLLER              CONTAINER(S)            IMAGE(S)
logging-es-ku7p73na-1   elasticsearch           docker.io/openshift/origin-logging-elasticsearch@sha256:574b02307f2dd2f9f5cf53bff56d2721acb0f98ff218c1a41bcf884d500c446a
logging-fluentd-1       fluentd-elasticsearch   docker.io/openshift/origin-logging-fluentd@sha256:cf63d464418bef8f25749088a96409965442d9cd5d4ac3873449b43f2e6a2568
logging-kibana-1        kibana                  docker.io/openshift/origin-logging-kibana@sha256:3b7dec271a0d55b0a6f0c446c48b2e1b421bdbf75919677e6de382871e39e9ac 
                        kibana-proxy            docker.io/openshift/origin-logging-auth-proxy@sha256:88cc9c702e68f35d7fb523bea72278d1994096e5aa0fafb49dff81eea9e5c08d                                                                                                              
logging-kibana-2        kibana                  docker.io/openshift/origin-logging-kibana@sha256:3b7dec271a0d55b0a6f0c446c48b2e1b421bdbf75919677e6de382871e39e9ac
                        kibana-proxy            docker.io/openshift/origin-logging-auth-proxy@sha256:88cc9c702e68f35d7fb523bea72278d1994096e5aa0fafb49dff81eea9e5c08d  
```

ewolinetz:
@sosiouxme I think the reason that you received that panic on line 35 was because the curl to ES failed for some reason... maybe it wasn't ready to accept connections at the time?  But the exec command failed.  I'll investigate how to make that smarter (e.g. wait until we see that the kibana instance we are trying to exec from has actually connected to ES first)

jcantrill:
@ewolinetz  Make sure to look at the origin util script that is used.  It
has several functions that are used through out the e2e test.

On Tue, Dec 22, 2015 at 3:02 PM, Eric Wolinetz <notifications@github.com>
wrote:

> @sosiouxme <https://github.com/sosiouxme> I think the reason that you
> received that panic on line 35 was because the curl to ES failed for some
> reason... maybe it wasn't ready to accept connections at the time? But the
> exec command failed. I'll investigate how to make that smarter (e.g. wait
> until we see that the kibana instance we are trying to exec from has
> actually connected to ES first)
>
> —
> <https://github.com/openshift/origin-aggregated-logging/pull/30#issuecomment-166716669>
> .
>



-- 
--
Jeff Cantrill
Red Hat, Inc.
jcantril@redhat.com
http://www.redhat.com

ewolinetz:
@jcantrill unless I'm misunderstanding how the cert/key are created for the pods, since we are querying ES while it is doing SSL validation; we need to query from one of the other EFK components.  Making it difficult to reuse the 'wait_for_url' function.  So I am using a similar logic, but checking the logs (oc logs) vs (oc exec) multiple times, up to $TIMES times.

@jcantrill  @sosiouxme PTAL

jcantrill:
Generally LGTM other then I would have thought there was a way to not have to shell out to grep from go; this is counter-intuitive to me and seems like it would be resource expensive.  I defer to the tests you have run.  The deployment check seems like it might be valuable as a diagnostic of some kind? @sosiouxme thoughts?

richm:
This will query ES, then check to see if the record found in ES is in the log file on disk.  What about the converse?  That is, going through every record on disk to see if it is in ES?

ewolinetz:
@richm we could do that as well, I would be concerned about the performance impact of doing that -- we would need to go through each file, reach each line, and see if we have a match in the list of record responses we received.  

We would also need to ensure that we pulled back a large enough sample from ES for the index we are checking.  And we would have to figure out how long would we wait to know that all messages from the logs have been read in by Fluentd and are available to query...

PR #32, opened 2015-12-10
Update kibana Dockerfile to use kibana plugin release file

jcantrill:
This PR:
* Allows using the kibana plugin from the release downloads instead of trying to build everything during the docker build

### DIFF
--- a/kibana/Dockerfile
+++ b/kibana/Dockerfile
@@ -6,7 +6,9 @@ EXPOSE 5601
 
 ENV KIBANA_VER=4.1.1 \ 
     ES_HOST=localhost \ 
-    ES_PORT=9200
+    ES_PORT=9200 \
+    AOP_KIBANA_PLUGIN_VER=v0.5.0 \
+    AOP_KIBANA_PLUGIN_REPO=https://github.com/openshift/origin-kibana
 
 LABEL io.k8s.description="Kibana container for querying Elasticsearch for aggregated logs" \
   io.k8s.display-name="Kibana" \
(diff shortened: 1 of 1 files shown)


jcantrill:
@ewolinetz @sosiouxme  thoughts?

kibana/Dockerfile
```diff
@@ -17,16 +19,12 @@ RUN   wget -q https://download.elastic.co/kibana/kibana/kibana-4.1.1-linux-x64.t
       tar -xzf kibana-4.1.1-linux-x64.tar.gz && \
       mv kibana-4.1.1-linux-x64/* /opt/app-root/src/ && \
       rm -rf kibana-4.1.1-linux-x64* && \
-      mkdir -m 755 /opt/origin-kibana && \
-      git clone --branch master --depth 1 https://github.com/openshift/origin-kibana.git /opt/origin-kibana && \
+      mkdir -p -m 755 /opt/origin-kibana/lib && \
+      cd /opt/origin-kibana/lib && \
+      wget -q ${AOP_KIBANA_PLUGIN_REPO}/releases/download/${AOP_KIBANA_PLUGIN_VER}/origin-kibana-${AOP_KIBANA_PLUGIN_VER}.tgz && \
+      tar -xvzf origin-kibana-${AOP_KIBANA_PLUGIN_VER}.tgz && \
```
ewolinetz: since we're suppressing wget's output, do we want to do the same for tar?  I feel like they should be consistent (one way or another)

ewolinetz:
The approach seems right to me, just one question regarding verbosity, otherwise LGTM.

PR #40, opened 2016-01-09
Need to use double quotes for PUSH variable test

richm:
This always returns true:
  if [ -n ${PUSH:-$3} ] ; ...
Have to use double quotes around shell expansion
  if [ -n "${PUSH:-$3}" ] ; ...


jcantrill:
LGTM

PR #45, opened 2016-01-14
make scripts executable - minor test script issues

richm:
scripts should be executable
check-EFK-running should ignore the logging-deployment imagestream
check-logs should ignore the kibana build pod


ewolinetz:
LGTM

PR #47, opened 2016-01-15
use wildcard to look up hostname field

richm:
I'm not sure what is the purpose of::

    hostname = strings.Split(hostname, ".")[0]

but on my system, the ES logs have the FQDN, so doing the search for hostname fails.  If I add a '*' wildcard to the end, it works.


richm:
@ewolinetz PTAL

ewolinetz:
@richm I didn't have a FQDN in my ES logs, just the hostname.  That was the intent of the Split().

LGTM

PR #48, opened 2016-01-15
exclude -deploy pods

richm:
@ewolinetz PTAL


ewolinetz:
@richm wasn't this included in #45 ?

richm:
https://github.com/openshift/origin-aggregated-logging/pull/45 excluded "-build" pods, but I later found that the e2e scripts were being confused by the "-deploy" pods as well.

ewolinetz:
ah, that is correct, I didn't read closely enough.  LGTM.

Issue #46, opened 2016-01-14
Incorrect year used for syslog based message operations index


ewolinetz:
When pulling in log files from the previous year in the current year; the index the logs are first created for are the current year.

E.g. logs from 12/27/2015 will be created in the index ".operations.2016.12.27" if it is read in in 2016 initially

ewolinetz:
Should we do something like this in the record_transformer filter?
```
time ${Time.parse(Time.parse(time).to_s.gsub(Time.parse(time).year.to_s, (Time.parse(time).year - years_ago).to_s )) if Time.parse(time) > Time.now}
```
How can we figure out what `years_ago` should be?
  - parse from file name somehow?

Can using rsyslog resolve this for older, existing log files?

portante:
@ewolinetz, what logs is fluentd reading?  Just `/var/log/messages`, or are you trying to ingest all log files in `/var/log/messages*.gz`?

ewolinetz:
@portante `/var/log/messages*` so that would include any .gz I suppose.  At least, it would attempt to `tail` them.

https://github.com/ewolinetz/origin-aggregated-logging/blob/fix_syslog_date/fluentd/fluent.conf#L131

portante:
@ewolinetz, do we have control over how rsyslog is configured to write to /var/log/messages?  Or is this something that we have no control over?

ewolinetz:
@portante at this time, we should assume we don't have control over it.  When we change to a solution where rsyslog is the primary means of collecting and processing logs, then we would have control over how rsyslog is configured.  We have a card on our trello board to track that future effort (https://trello.com/c/oUvJg3JR/188-r-d-use-rsyslog-instead-of-fluentd-logging).

portante:
It is also important to get the timezone right, normalized to UTC and then to choose the index.  Otherwise, clients might not be looking in the right index entirely.

PR #49, opened 2016-01-19
Updating fluentd.conf to properly set date for syslog lines and 

ewolinetz:
use corrected time for index name

### DIFF
--- a/fluentd/fluent.conf
+++ b/fluentd/fluent.conf
@@ -1,111 +1,104 @@
-# This configuration file for Fluentd / td-agent is used                                                                                   
-# to watch changes to Docker log files. The kubelet creates symlinks that                                                                                                 
-# capture the pod name, namespace, container name & Docker container ID                                                                              
-# to the docker logs for pods in /var/log/containers directory on the host.                                                                
-# If running this fluentd configuration in a Docker container, the /var/log                                                                                               
-# directory should be mounted in the container.                                                                                                      
-#                                                                                                                                          
-# These logs are then submitted to # Elasticsearch which assumes the                                                                                                      
-# installation of the fluent-plugin-elasticsearch & the                                                                                              
-# fluent-plugin-kubernetes_metadata_filter plugins.                                                                                        
-# See https://github.com/uken/fluent-plugin-elasticsearch &                                                                                                               
-# https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter for                                                                          
-# more information about the plugins.                                                                                                      
-# Maintainer: Jimmi Dyson <jimmidyson@gmail.com>                                                                                                                          
-#                                                                                                                                                    
-# Example                                                                                                                                  
-# =======                                                                                                                                                                 
-# A line in the Docker log file might like like this JSON:                                                                                           
-#                                                                                                                                          
-# {"log":"2014/09/25 21:15:03 Got request with path wombat\n",                                                                                                            
-#  "stream":"stderr",                                                                                                                                
-#   "time":"2014-09-25T21:15:03.499185026Z"}                                                                                               
-#                                                                                                                                                                         
-# The time_format specification below makes sure we properly                                                                                         
-# parse the time format produced by Docker. This will be                                                                                   
-# submitted to Elasticsearch and should appear like:                                                                                                                      
-# $ curl 'http://elasticsearch-logging.default:9200/_search?pretty'                                                                                  
-# ...                                                                                                                                      
-# {                                                                                                                                                                       
-#      "_index" : "logstash-2014.09.25",                                                                                                             
-#      "_type" : "fluentd",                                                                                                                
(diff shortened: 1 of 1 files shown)


ewolinetz:
Addresses #46 

@sosiouxme @jcantrill @richm PTAL

ewolinetz:
I think there is an issue that I didn't encounter yesterday but am today; please do not merge yet

ewolinetz:
Needed to update gsub function to use 
```
(tag_parts[3].nil? ? Time.at(time).year.to_s : tag_parts[3][9,4]) )
```
instead of `||` logic

portante:
@ewolinetz, so this file description talks about reading logs from docker logs, but I see two sources, one for /var/log/messages* and one for container logs.  Seems like we should address that with these comment changes, since we appear to be rewriting large sections of the comments.

It also seems like we are doing a lot of work in fluentd to handle rsyslog data when I believe rsyslog can do that much easier, like dealing with proper timestamps and index names, etc.  Then you have a fluentd for processing container logs, and an rsyslog config for system logs (which it is already handling, not extra data flow).

richm:
@portante I am trying to figure out how we can use rsyslog instead of fluentd for this use case, so that we do not have to have fluentd or fluentd+rsyslog running on each node.  We can just have rsyslog, and it can either send the logs directly to elasticsearch, to a message queue, or to some other forwarder/aggregator.

ewolinetz:
@portante the trello card regarding using rsyslog instead is this one: https://trello.com/c/oUvJg3JR/188-r-d-use-rsyslog-instead-of-fluentd-logging

portante:
I would suggest that the we just make a simple change (when we are ready) to the `rsyslog.conf` to get `/var/log/messages` to use a file format with high resolution time stamps (and timezone information): `RSYSLOG_FileFormat`, see http://www.rsyslog.com/doc/v7-stable/configuration/templates.html (near the bottom).

This is a simple change.

fluentd/fluent.conf
```diff
@@ -1,111 +1,111 @@
...
+# installation of the fluent-plugin-elasticsearch & the
+# fluent-plugin-kubernetes_metadata_filter plugins.
+# See https://github.com/uken/fluent-plugin-elasticsearch &
+# https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter for
+# more information about the plugins.
+# Maintainer: Jimmi Dyson <jimmidyson@gmail.com>
+#
+# Example
+# =======
+# A line in the Docker log file might like like this JSON:
```
sosiouxme: "look like"

fluentd/fluent.conf
```diff
@@ -1,111 +1,111 @@
...
+# Kubernetes pod has been restarted (resulting in a several Docker container IDs).
+#
+# TODO: Propagate the labels associated with a container along with its logs
+# so users can query logs using labels as well as or instead of the pod name
+# and container name. This is simply done via configuration of the Kubernetes
 # fluentd plugin but requires secrets to be enabled in the fluent pod. This is a
-# problem yet to be solved as secrets are not usable in static pods which the fluentd         
-# pod must be until a per-node controller is available in Kubernetes.                                                                                                     
+# problem yet to be solved as secrets are not usable in static pods which the fluentd
+# pod must be until a per-node controller is available in Kubernetes.
```
sosiouxme: Above paragraph no longer relevant?

fluentd/fluent.conf
```diff
@@ -228,7 +228,9 @@
       # if host == 'localhost' do a fqdn lookup
       ## we pull the hostname from the host's /etc/hostname mounted at /etc/docker-hostname to correctly identify the hostname
       hostname ${host.eql?('localhost') ? (begin; File.open('/etc/docker-hostname') { |f| f.readline }.rstrip; rescue; host; end) : host}
-      time ${Time.at(time).to_datetime.to_s}
+
+      # tag_parts = system.var.log.messages-20150710
+      time ${ Time.at(time) > Time.now ? (temp_time = Time.parse(Time.at(time).to_s.gsub(Time.at(time).year.to_s, (tag_parts[3].nil? ? Time.at(time).year.to_s : tag_parts[3][9,4]) )).to_datetime.to_s; Time.parse(temp_time) > Time.now ? Time.parse(temp_time.gsub(Time.parse(temp_time).year.to_s, (Time.parse(temp_time).year - 1).to_s )).to_datetime.to_s : Time.parse(temp_time).to_datetime.to_s ) : Time.at(time).to_datetime.to_s }
```
sosiouxme: Clarify this if possible; could also use a comment as to why this is needed

fluentd/fluent.conf
ewolinetz: I'll have to check again on whether or not we are sending labels, I feel like we are but I'd like to be sure before we scrap that paragraph entirely

ewolinetz:
@sosiouxme addressed comments

richm:
lgtm

PR #51, opened 2016-01-21
Adding MASTER_URL variable when processing fluentd template.

danbev:
Motivation:
When run.sh processes templates/fluentd.yaml it does not take a value
for the MASTER_URL which is used for the env var K8S_HOST_URL.

Modifications:
Adding the MASTER_URL to the variables.

Result:
The K8S_HOST_URL environment variable can be overridden.


ewolinetz:
LGTM, thanks!

PR #52, opened 2016-01-25
Fluentd log throttling

ewolinetz:
Reorganized fluentd to use includes and maintain functionality with allowing to throttle

### DIFF
--- a/deployment/templates/fluentd.yaml
+++ b/deployment/templates/fluentd.yaml
@@ -107,6 +107,8 @@ objects:
               value: ${OPS_CLIENT_KEY}
             - name: "OPS_CA"
               value: ${OPS_CA}
+            - name: "THROTTLE_SETTINGS"
+              value: ${THROTTLE_SETTINGS}
           volumes:
           - name: varlog
             hostPath:
@@ -165,3 +167,7 @@ parameters:
   description: "Location of CA cert for validating connectiong to ElasticSearch to write cluster logs"
   name: OPS_CA
   value: "/etc/fluent/keys/ca"
+-
+  description: "JSON format configuration for log throttle management"
+  name: THROTTLE_SETTINGS
+  value: "{}"
--- a/fluentd/generate_throttle_configs.rb
+++ b/fluentd/generate_throttle_configs.rb
@@ -1,0 +1,108 @@
+require 'json'
+require 'date'
+require 'logger'
+
+# or should this be stderr?  -- we only want to print out warns?
+@log = Logger.new(STDOUT)
+@log.level = Logger::WARN
+
+@Valid_Settings = {"buffer_chunk_limit" => "size", "buffer_queue_limit" => "number", "flush_interval" => "time", "flush_at_shutdown" => "boolean", "max_retry_wait" => "time", "disable_retry_limit" => "boolean", "retry_limit" => "number", "retry_wait" => "time", "num_threads" => "number"}
+
+def get_file_name(name)
+  ## file_name follows pattern: gen-#{name}-YYYYMMDD.conf ##
+
+  file_name = "/etc/fluent/configs.d/output/gen-"
+  file_name << name
+  file_name << "-"
+  file_name << Date.today.strftime("%Y%m%d")
+  file_name << ".conf"
+
+  return file_name
+end
+
+def seed_file(file_name, project)
+
+  File.open(file_name, 'w') { |file|
+    file.write("<match **_#{project}_**>\n@type forward\nsend_timeout 60s\nrecover_wait 10s\nheartbeat_interval 1s\nphi_threshold 16\nhard_timeout 60s\n##\n")
+  }
+
+end
+
+def write_to_file(project, key, value)
+
+  file_name = get_file_name(project)
+
+  # check if the file already exists, if not create it
+  seed_file(file_name, project) if !File.exist?(file_name)
+
+  File.open(file_name, 'a') { |file|
+    file.write("#{key} #{value}\n")
+  }
+
+end
+
+def close_file(project)
+  file_name = get_file_name(project)
+
+  File.open(file_name, 'a') { |file|
+    file.write("##\n<server>\nname '127.0.0.1:24224'\nhost 127.0.0.1\nport 24224\n</server>\n</match>")
+  } if File.exist?(file_name)
+end
+
+def validate(key, value)
+
+  #@log.info "Checking #{key} and #{value}"
+
+  # if the key is in valid settings, validate the value by required type
+  if @Valid_Settings.keys.include?(key)
+    case @Valid_Settings[key]
+    when "time"
+      valid = value.to_s.match('^\d+[sSmMhH]$') { |m| if m.nil?; false; else m.to_s.match('\d+').to_s.to_i > 0 end }
+    when "size"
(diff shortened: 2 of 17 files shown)


ewolinetz:
@sosiouxme @richm @jcantrill PTAL

Please note that this method will most likely change when we transition to use rsyslog and messaging

richm:
fluentd forwards logs to itself?

ewolinetz:
@richm its pretty weird looking and I'm open for changing it.   

What I was looking to accomplish was give a buffer to write into so we could rate limit logs going to ES without opening a connection for each project with a rate restriction.  I was also thinking that avoiding installing a message broker with fluentd would help with performance overhead on the pod and reduce complexity.

So fluentd restricts the rate at which logs are looped back to itself, and then from that it sends it off to ES at the current rate that we have implemented (5s).

richm:
We do not want to require a message queue at this point, and certainly not to solve this particular problem.  I'm a bit surprised that this actually works, and I'm curious to see how it works at scale.

lgtm

sosiouxme:
@ewolinetz what happens when the rate limit is exceeded? logs dropped or just delayed?

It would be nice of course if the limit were per-project so one project couldn't DoS the rest. But I suppose that'll be pretty difficult without either having ES or some intermediary (message bus) centrally keep track of rates per index.

ewolinetz:
@sosiouxme so there isn't quite a way to say 'we only allow a throughput of x logs per y' instead we are updating how long we will buffer for a configured project.  (We can also configure how large of a buffer we use, and the size of blocks).  This may be the wrong approach that we want to take.. 

But we can specify that a certain project can only send logs every 1 minute, while the default for others would still be every 5s.  But the caveat of how this is set up is we buffer those logs for that long  (we also process them first.. but that may be able to be moved to the egress label? -- I'd need to test that)

What we end up with is a different config file to be included here:
https://github.com/openshift/origin-aggregated-logging/pull/52/files#diff-3ba06ea3ce4d4e1a023d40b8f6d98776R115

And each config will have different buffer setting values.

richm:
Have you investigated the read_lines_limit?
What happens if you have a fluentd config like this::

    <source>
      @type tail
      @label @KUBERNETES
      path /var/log/containers/badproject.log
      ...
    </source>
    <source>
      @type tail
      @label @KUBERNETES
      path /var/log/containers/*.log
      ...
    </source>

Will badproject.log also be processed by the second source `@type tail`?

ewolinetz:
@richm I had not looked into the read_lines_limit flag, that might be a better approach for rate limiting so long as we can ensure one `<source>...</source>` doesn't block another.

I think that if they are using the same `pos_file` it might prevent any duplicates. Since it contains the log file name that was read, and the current read position. This would have to be tested though

ewolinetz:
@richm there is also an `exclude_path` option for the `type tail` source.  It takes an array of paths, so it might make sense to treat this how we handled the default curator call.

deployment/templates/fluentd.yaml
```diff
@@ -165,3 +167,7 @@ parameters:
   description: "Location of CA cert for validating connectiong to ElasticSearch to write cluster logs"
   name: OPS_CA
   value: "/etc/fluent/keys/ca"
+-
+  description: "JSON format configuration for log throttle management"
+  name: THROTTLE_SETTINGS
```
sosiouxme: What would be setting this - someone manually editing the fluentd DC/daemonset?

I don't really like to put JSON inside a string or in an env var... too many ways to muff that and it's unreadable. Is there any way to just have this specify a filename to read, and have that default to a file in the fluentd secret, and have the generator just use the default if that file's not there?

fluentd/generate_throttle_configs.rb
```diff
@@ -0,0 +1,108 @@
+require 'json'
+require 'date'
+require 'logger'
+
+# or should this be stderr?  -- we only want to print out warns?
```
sosiouxme: Does anything in the fluentd container *have* any output that isn't a log message? I'm inclined to leave it stdout unless there's output it would be useful to distinguish from log messages.

fluentd/generate_throttle_configs.rb
```diff
@@ -0,0 +1,108 @@
...
+  file_name << Date.today.strftime("%Y%m%d")
+  file_name << ".conf"
+
+  return file_name
+end
+
+def seed_file(file_name, project)
+
+  File.open(file_name, 'w') { |file|
+    file.write("<match **_#{project}_**>\n@type forward\nsend_timeout 60s\nrecover_wait 10s\nheartbeat_interval 1s\nphi_threshold 16\nhard_timeout 60s\n##\n")
```
sosiouxme: please use a heredoc so this is legible...

```
    file.write(<<-CONF)
      <match **_#{project}_**>
      @type forward
      send_timeout 60s
      recover_wait 10s
      heartbeat_interval 1s
      phi_threshold 16
      hard_timeout 60s
      ##
    CONF
```

fluentd/generate_throttle_configs.rb
```diff
@@ -0,0 +1,108 @@
...
+    file.write("#{key} #{value}\n")
+  }
+
+end
+
+def close_file(project)
+  file_name = get_file_name(project)
+
+  File.open(file_name, 'a') { |file|
+    file.write("##\n<server>\nname '127.0.0.1:24224'\nhost 127.0.0.1\nport 24224\n</server>\n</match>")
```
sosiouxme: heredoc here too

sosiouxme:
On the whole I like having the monolithic config broken out into smaller files.

sosiouxme:
So really, it sounds like it's a bit of a misnomer to call this "throttling" since it doesn't normally limit the actual number of log messages. Can we describe this as "batching" and/or "buffering"? I think anything with "throttle" in the name is going to lead people into invalid assumptions.

deployment/templates/fluentd.yaml
ewolinetz: Ideally we would most likely use the ConfigMap api when that's available... we could probably use a secret. My initial thought was that using a structured format would make it a) easier to parse and b) give an 'easier' way to specify x* options for y* projects

ewolinetz:
@sosiouxme I think the fact that it is a misnomer means that this isn't doing what it should be doing. The purpose of this was to throttle the rate of logs;  I'm currently looking into the `read_lines_limit` option that Rich had mentioned since that sounds like it would actually change the number of log lines we read in at a time.  We could also remove the self-looping logic that is currently being used here.  (Unless we want to still give options to buffer for projects?)  (cc @richm)

ewolinetz:
@sosiouxme @richm expanding on previous discussions... I'll have to verify that throttling is occurring but does this approach seem more straight forward and actually seem to throttle rather than just buffer?
https://github.com/ewolinetz/origin-aggregated-logging/commit/1e3d7325e8a609be3e5470e00a2011120c343ed9

richm:
To me, the intention of throttling is to prevent fluentd from being overloaded, which means slowing down the rate at which fluentd reads from a particular file.  I like this approach better.

If we need buffering, I think that should be a separate story/card/pr

ewolinetz:
Closing this PR and will open it with the other intended means of throttling (limiting rate at which files are read).  I'll maintain some files pertaining to buffering in the case that we want to also provide a means to adjust for performance sending to ES.

PR #53, opened 2016-01-25
do not process the kibana template twice - add master_url to kibana-ops

richm:
@ewolinetz @sosiouxme  PTAL

I'm not sure how the previous commit ever worked.  The kibana.yaml template was being processed twice.  The second one causes an error. The script has set -e so it bails out.

The fix is to process kibana.yaml once, with both the master and public master url. I also added the master url to the kibana-ops processing.


ewolinetz:
LGTM

PR #56, opened 2016-02-02
addressing k8s_host missing

ewolinetz:
I noticed warn statements that variables or methods could not be found for 'kubernetes_host', same for 'err'.

@sosiouxme @richm PTAL

### DIFF
--- a/fluentd/fluent.conf
+++ b/fluentd/fluent.conf
@@ -157,7 +157,7 @@
     type record_transformer
     enable_ruby
     <record>
-      log ${(err || msg) || log}
+      log ${(err rescue nil) || (msg rescue nil) || log}
     </record>
     remove_keys req,res,msg,name,level,v,pid,err
   </filter>
@@ -169,7 +169,7 @@
     type record_transformer
     enable_ruby
     <record>
-      hostname ${kubernetes_host}
+      hostname ${(kubernetes_host rescue nil) || File.open('/etc/docker-hostname') { |f| f.readline }.rstrip}
       message ${log}
       version 1.0.6
     </record>


fluentd/fluent.conf
```diff
@@ -157,7 +157,7 @@
     type record_transformer
     enable_ruby
     <record>
-      log ${(err || msg) || log}
+      log ${( (err unless (defined?(err)).nil?) || (msg unless (defined?(msg)).nil?) ) || log}
```
sosiouxme: Little convoluted. This would be clearer:

    (err if defined? err) || (msg if defined? msg) ) || log

Even shorter and less redundant, if more opaque:

    (err rescue nil) || (msg rescue nil) ) || log

fluentd/fluent.conf
```diff
@@ -169,7 +169,7 @@
     type record_transformer
     enable_ruby
     <record>
-      hostname ${kubernetes_host}
+      hostname ${(kubernetes_host unless (defined?(kubernetes_host)).nil?) || (begin; File.open('/etc/docker-hostname') { |f| f.readline }.rstrip; end)}
```
sosiouxme: (kubernetes_host rescue nil) || File.open('/etc/docker-hostname') { |f| f.readline }.rstrip

fluentd/fluent.conf
ewolinetz: never thought to use rescue... I like it

ewolinetz:
addressed comments

fluentd/fluent.conf
```diff
@@ -157,7 +157,7 @@
     type record_transformer
     enable_ruby
     <record>
-      log ${(err || msg) || log}
+      log ${( (err rescue nil) || (msg rescue nil) ) || log}
```
sosiouxme: for that matter, can get rid of the outer set of parens :)

then :+1:

ewolinetz:
nits addressed

ewolinetz:
@richm it should, assuming the host has a /etc/hostname file  https://github.com/openshift/origin-aggregated-logging/blob/master/deployment/templates/fluentd.yaml#L120-L122

We could do a `rescue nil` in the case where it doesn't.  We'd be in no worse shape than the scenario we're trying to solve if it doesn't exist.

richm:
ack - lgtm

PR #57, opened 2016-02-05
Add support for curator for Elasticsearch index management

richm:
Curator allows the admin to remove old data from Elasticsearch
on a per-project basis.  It runs as a separate pod within the
logging project.  See the README.md for more details about how
it is configured and run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,76 @@ proxy is required that runs in front of Kibana.
 The deployer enables the user to generate all of the necessary
 key/certs/secrets and deploy all of the components in concert.
 
-## Defining local builds
+#### Curator
+
+Curator allows the admin to remove old indices from Elasticsearch on a per-project
+basis.  The pod will read its configuration from a mounted yaml file that
+is structured like this:
+
+    $PROJECT_NAME:
+      $ACTION:
+        $UNIT: $VALUE
+
+    $PROJECT_NAME:
+      $ACTION:
+        $UNIT: $VALUE
+     ...
+
+* $PROJECT_NAME - the actual name of a project - "myapp-devel"
+** For operations logs, use the name `.operations` as the project name
+* $ACTION - the action to take - currently only "delete"
+* $UNIT - one of "days", "weeks", or "months"
+* $VALUE - an integer for the number of units
+
+For example, using::
+
+    myapp-dev:
+     delete:
+       days: 1
+
+    myapp-qe:
+      delete:
+        weeks: 1
+
+    .operations:
+      delete:
+        weeks: 8
+     ...
+
+Every day, curator will run, and will delete indices in the myapp-dev
+project older than 1 day, and indices in the myapp-qe project older than 1
+week.
+
+*WARNING*: Using `months` as the unit
+
+When you use month based trimming, curator starts counting at the _first_ day of
+the current month, not the _current_ day of the current month.  For example, if
+today is April 15, and you want to delete indices that are 2 months older than
+today (`delete: months: 2`), curator doesn't delete indices that are dated
+older than February 15, it deletes indices older than _February 1_.  That is,
+it goes back to the first day of the current month, _then_ goes back two whole
+months from that date.
+If you want to be exact with curator, it is best to use `days` e.g. `delete: days: 30`
+[Curator issue](https://github.com/elastic/curator/issues/569)
+
+To create the curator configuration, do the following:
+
+Create a yaml file with your configuration settings using your favorite editor.
+Next create a secret from your created yaml file:
+`oc secrets new index_management settings=</path/to/your/yaml/file>`
+
+Then mount your created secret as a volume in your Curator DC:
+`oc volumes dc/logging-curator --add --type=secret --secret-name=index_management --mount-path=/etc/curator --name=index_management --overwrite`
+
+The mount-path value e.g. `/etc/curator` must match the `CURATOR_CONF_LOCATION`
+in the environment.
+
+By default curator will run at midnight and deletes logs older than 30 days.
+The environment variables `CURATOR_CRON_HOUR` AND `CURATOR_CRON_MINUTE` can be
+used to define another hour and minute while `DEFAULT_DAYS` will change the
+default number of days logs will be kept for.
+
+# Defining local builds
 
 Choose the project you want to hold your logging infrastructure. It can be
 any project.
--- a/curator/install.sh
+++ b/curator/install.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+set -ex
+
+rpm -q epel-release || yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+yum install -y --setopt=tsflags=nodocs \
+  python-pip \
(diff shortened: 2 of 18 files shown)


ewolinetz:
LGTM, but I helped work on it.  Second opinion @sosiouxme @jcantrill  ?

README.md
```diff
@@ -39,7 +39,54 @@ proxy is required that runs in front of Kibana.
 The deployer enables the user to generate all of the necessary
 key/certs/secrets and deploy all of the components in concert.
 
-## Defining local builds
+#### Curator
```
jcantrill: It's not obvious to me what happens in the default scenario.  The case where there is an empty hash for INDEX_MGMT.  The way it reads is that I will need to add an entry for EVERY project which is not what we want.  I would add  a section here that makes it more obvious that doing nothing still gets you deleted indexes.  Also, are project names evaluated based on a wild card since the actual indexes are a combination of a date and project?

ewolinetz:
Should we have this avoid our .searchguard index?  It already avoids the .kibana indices

jcantrill:
probably, though i dont recall what is stashed in that index.

On Tue, Feb 9, 2016 at 2:19 PM, Eric Wolinetz <notifications@github.com>
wrote:

> Should we have this avoid our .searchguard index? It already avoids the
> .kibana indices
>
> —
> <https://github.com/openshift/origin-aggregated-logging/pull/57#issuecomment-182016018>
> .
>



-- 
--
Jeff Cantrill
Red Hat, Inc.
jcantril@redhat.com
http://www.redhat.com

ewolinetz:
@jcantrill I think that is where the ACL info is stored

README.md
ewolinetz: correct, we evaluate project names with ".*" appended to account for the date format at the end, and the inability for project names to contain a period (when created through OpenShift)

richm:
The curator configuration needs to change.

curator delete indices --time default_time --unit default_unit --timestring x.y.z --all-indices --except '^.kibana*' --except '^.searchguard*' --except '^project1.*' --except '^project2.*' ...

curator delete indices --time time1 --unit unit1 --timestring x.y.z --prefix 'project1.' --prefix 'project2.'

curator delete indices --time time2 --unit unit2 --timestring x.y.z --prefix 'project3.' --prefix 'project3.'

The default should work on all indices except kibana, searchguard, and any projects for which we have specific trimming times/units.  Each group of projects that share a common time/unit will have their own curator run listing each project in the prefix.

elasticsearch/install.sh
```diff
@@ -15,7 +16,7 @@ export CLUSTER_NAME=placeholder
 mkdir -p ${HOME}
 ln -s /usr/share/elasticsearch /usr/share/java/elasticsearch
 /usr/share/elasticsearch/bin/plugin -i com.floragunn/search-guard/0.5
-/usr/share/elasticsearch/bin/plugin -i io.fabric8.elasticsearch/openshift-elasticsearch-plugin/0.7
+/usr/share/elasticsearch/bin/plugin -i io.fabric8.elasticsearch/openshift-elasticsearch-plugin/0.8
```
ewolinetz: since we just released version 0.9 of this plugin could you update to use that version instead?

richm: @ewolinetz That should be a separate commit, and I should rebase my commit on that commit

ewolinetz: @richm in that case I'd say we can do that as a follow up PR after this is merged in. I don't see a need to wait on another commit to rebase on top of.

ewolinetz:
@richm Do we still need to update the curator configuration as you indicated in your above comment?

richm:
@ewolinetz I think so - I'm testing it now

ewolinetz:
other than my question LGTM

curator/install.sh
```diff
@@ -0,0 +1,28 @@
...
+yum install -y --setopt=tsflags=nodocs \
+  python-pip \
+  PyYAML
+pip install elasticsearch-curator python-crontab
+yum clean all
+
+# HACK HACK HACK - remove when fixed upstream
+yum install -y --setopt=tsflags=nodocs \
+    git
+git clone -b issue-344 https://github.com/richm/elasticsearch-py.git
```
sosiouxme: good gorm that's wretched... i hope we have a more sanctioned way to apply this patch for enterprise?

README.md
```diff
@@ -39,7 +39,71 @@ proxy is required that runs in front of Kibana.
...
+       days: 1
+
+    myapp-qe:
+      delete:
+        weeks: 1
+     ...
+
+Every day, curator will run, and will delete indices in the myapp-dev
+project older than 1 day, and indices in the myapp-qe project older than 1
+week.
```
sosiouxme: Can we call out how to configure curating ops logs? I assume you just use project name `.operations` in the yaml but I think we need to say that.

deployment/README.md
```diff
@@ -31,6 +31,11 @@ The deployer enables the system administrator to generate all of the
 necessary key/certs/secrets and deploy all of the logging components
 in concert.
 
+## Curator
+
+Curator allows the admin to remove old data from Elasticsearch on a per-project
```
sosiouxme: could we say "and is configurable on a per-project basis"?

elasticsearch/install.sh
```diff
@@ -2,6 +2,7 @@
 
 set -ex
 
+yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
```
sosiouxme: what did we need to add epel for?

curator/install.sh
richm: Yes.  For enterprise we will have to go through the usual approved patch process.  It's going to be a pain anyway, since we're using epel and/or pip packages.

deployment/README.md
richm: done

elasticsearch/install.sh
richm: I think that is a holdover for when we had curator as part of the elasticsearch image.  I will remove that.

deployment/templates/curator.yaml
```diff
@@ -0,0 +1,129 @@
...
+      triggers:
+      - type: ConfigChange
+      - type: ImageChange
+        imageChangeParams:
+          automatic: true
+          containerNames:
+          - curator
+          from:
+            kind: ImageStreamTag
+            name: logging-curator:latest
```
sosiouxme: This is only going to exist if you've created the ImageStream, which looks to be missing - should be added to deployment/templates/support.yaml I think for deploying from official registry.

richm: Ok - check the latest commit - does that look correct?

deployment/README.md
sosiouxme: :+1:

deployment/templates/curator.yaml
sosiouxme: :+1:

sosiouxme:
minor comments, LGTM

ewolinetz:
comments look to be addressed, merging

PR #58, opened 2016-02-09
Log throttle

ewolinetz:
Adding ability to throttle rate at which log files are read by fluentd

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ you can get from:
     logging-deployment      172.30.90.128:5000/logs/logging-deployment
 
 In order to run a deployment with these images, you would process the
-[deployer template](deployment/deployer.yaml) with the 
+[deployer template](deployment/deployer.yaml) with the
 `IMAGE_PREFIX=172.30.90.128:5000/logs/` parameter. Proceed to the
 [deployer instructions](./deployment) to run a deployment.
 
@@ -102,3 +102,36 @@ order to create everything with the right parameters. E.g.:
 There are a number of env vars this script looks at which are useful
 when running directly; check the script headers for details.
 
+## Throttling logs in Fluentd
+
+For projects that are especially verbose, an administrator can throttle
+down the rate at which the logs are read in by Fluentd at a time before being
+processed. Note: this means that aggregated logs for the configured projects
+could fall behind and even be deleted if the pod were deleted before Fluentd
+caught up.  
+
+To tell Fluentd which projects it should be restricting you will
+need to do the following:
+
+Create a yaml file that contains project names and the desired rate at which
+logs are read in. (Default is 1000)
+```
+logging:
+  read_lines_limit: 500
+
+test-project:
+  read_lines_limit: 10
+
+.operations:
+  read_lines_limit: 100
+```
+
+Create a secret providing this file as the source
+```
+oc secrets new fluentd-throttle settings=</path/to/your/yaml>
+```
+
+Mount the created secret to your Fluentd container
+```
+oc volumes dc/logging-fluentd --add --type=secret --secret-name=fluentd-throttle --mount-path=/etc/throttle-settings --name=throttle-settings --overwrite
+```
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -1,0 +1,12 @@
+if [[ $VERBOSE ]]; then
+  set -ex
+else
+  set -e
+fi
+
+mkdir -p /etc/fluent/configs.d/input/docker
+mkdir -p /etc/fluent/configs.d/input/syslog
+
+ruby generate_throttle_configs.rb
+
+fluentd
--- a/fluentd/Dockerfile
+++ b/fluentd/Dockerfile
@@ -6,7 +6,8 @@ ENV HOME=/opt/app-root/src \
   PATH=/opt/app-root/src/bin:/opt/app-root/bin:$PATH \
   RUBY_VERSION=2.0 \
   FLUENTD_VERSION=0.12.6 \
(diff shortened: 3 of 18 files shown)


ewolinetz:
@richm @sosiouxme PTAL - will now use a yaml file mounted as a secret for the throttle configurations such as:
```
logging:
  read_lines_limit: 1000

.operations:
  read_lines_limit: 100
```

Not sure if we want to maintain the extra lines in the Dockerfile where we mkdir -p and touch the throttle config, was thinking it was good security in the case where a user deletes the volume mount so that fluentd can still start up...

ewolinetz:
@richm I'll have to update the documentation to explain the process but basically it goes like this:

1. Admin creates yaml file with settings for throttling
2. Admin creates a secret using that created file as a source
```
oc secrets new fluentd-throttle settings=path/to/yaml/file
```
3. Admin mounts secret as a volume in DC (https://docs.openshift.org/latest/dev_guide/secrets.html#examples)

richm:
Ok.  And this method is only temporary until ConfigMap support is available?

ewolinetz:
I'm not sure we actually want to go the route of ConfigMap since it currently looks like it can be exposed to the pod via ENV vars.  IMO it seems a little restrictive vs what we can do with files and secrets

richm:
Ok.  Then lgtm, and curator and es-copy PRs need to change based on the way the config is passed in, and how the fluentd config is refactored.  I would like to see this PR merged, then those other two rebased on top of this new code.

ewolinetz:
@richm +1

fluentd/run.sh
```diff
@@ -0,0 +1,12 @@
+if [[ $VERBOSE ]]; then
+  set -ex
+else
+  set -e
+fi
+
+mkdir -p /etc/fluent/configs.d/input/docker
+mkdir -p /etc/fluent/configs.d/input/syslog
+
+ruby generate_throttle_configs.rb "$(cat ${THROTTLE_CONF_LOCATION}/settings)"
```
sosiouxme: I don't like reading in the config as a parameter like this. I think ruby should just load yaml from `#{ENV['THROTTLE_CONF_LOCATION']}`

deployment/templates/fluentd.yaml
```diff
@@ -84,6 +84,9 @@ objects:
             - name: dockerhostname
               mountPath: /etc/docker-hostname
               readOnly: true
+            - name: throttle-settings
```
sosiouxme: what's the migration plan for existing fluentd without this secret? are we basically going to tell them to reinstall from scratch...

ewolinetz: they shouldn't need to... they'll be pulling in the latest image, so that should cover the logic behind it and then it would be adding the volume mount to the DC (working on docs right now on how to do that).  Unless I'm missing something, it shouldn't be much more complicated than that

ewolinetz:
Updated README to describe how to configure throttling (should also cover upgrade path)

deployment/templates/fluentd.yaml
sosiouxme: for some reason i had the misconception this was a new secret...

sosiouxme: this is a host mount looking for a directory on the local host... i think it needs to be a secret? or does this DC declaration get overwritten by the secret volume later in deployment?

ewolinetz: Hmm, this is probably redundant since the readme gives the command to mount the secret where this volume mount currently is.  I'll update the yaml to remove this mount and the emptyDir volume definition.  The Dockerfile creates the structure and empty file to be overwritten, so this shouldn't be needed.

ewolinetz:
Removed unnecessary template change due to documentation covering mounting a secret as a volume

README.md
```diff
@@ -102,3 +102,31 @@ order to create everything with the right parameters. E.g.:
...
+```
+
+Create a secret providing this file as the source
+```
+oc secrets new fluentd-throttle settings=</path/to/your/yaml>
+```
+
+Mount the created secret to your Fluentd container
+```
+oc volumes dc/logging-fluentd --add --type=secret --secret-name=fluentd-throttle --mount-path=/etc/throttle-settings --name=throttle-settings --overwrite
```
sosiouxme: I'm a little surprised... doesn't the secret need to be added to the fluentd service account too?

README.md
```diff
@@ -102,3 +102,31 @@ order to create everything with the right parameters. E.g.:
 There are a number of env vars this script looks at which are useful
 when running directly; check the script headers for details.
 
+## Throttling logs in Fluentd
+
+For projects that are especially verbose, an administrator can now throttle
+down the rate at which the logs are read in.  To tell Fluentd which projects
```
sosiouxme: I assume this is the basis for pertinent docs changes...

`s/now//` as all docs are in the present :)

Also can you clarify where/how exactly the rate is being affected? I.e. I think you mean logs in excess of the limit per project per node are delayed at fluentd (not entered into ES) until a later round of log gathering.

sosiouxme:
Other than nits, LGTM

README.md
ewolinetz: That's what I had previously seen, but with the version I had tested this with it wasn't necessary... I'll try on a new system to confirm

ewolinetz: Confirmed, I didn't need to add the secret to the service account
```
oc v1.1.2-268-ga722cb8
kubernetes v1.2.0-origin
```

ewolinetz:
Addressed nits and squashed.

ewolinetz:
[merge]

PR #60, opened 2016-02-11
Allow fluentd to send copies of logs to another Elasticsearch

richm:
If the environment variable ES_COPY is set, fluentd will send a copy
of the logs to another Elasticsearch.  The settings for the copy are
just like the current ES_HOST, etc. and OPS_HOST, etc. settings, except
that they add _COPY: ES_COPY_HOST, OPS_COPY_HOST, etc.
There are some additional parameters added:
ES_COPY_SCHEME, OPS_COPY_SCHEME - can use either http or https - defaults to https
ES_COPY_USERNAME, OPS_COPY_USERNAME - user name to use to authenticate
    to elasticsearch using username/password auth
ES_COPY_PASSWORD, OPS_COPY_PASSWORD - password to use to authenticate
    to elasticsearch using username/password auth

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,42 @@ Mount the created secret to your Fluentd container
 oc volumes dc/logging-fluentd --add --type=secret --secret-name=fluentd-throttle --mount-path=/etc/throttle-settings --name=throttle-settings --overwrite
 ```
 
+## Have Fluentd send logs to another Elasticsearch
+
+You can configure Fluentd to send a copy of each log message to both the
+Elasticsearch instance included with OpenShift aggregated logging, _and_ to an
+external Elasticsearch instance.  For example, if you already have an
+Elasticsearch instance set up for auditing purposes, or data warehousing, you
+can send a copy of each log message to that Elasticsearch, in addition to the
+the Elasticsearch hosted with OpenShift aggregated logging.
+
+If the environment variable `ES_COPY` is `"true"`, Fluentd will send a copy of
+the logs to another Elasticsearch. The settings for the copy are just like the
+current `ES_HOST`, etc. and `OPS_HOST`, etc. settings, except that they add
+`_COPY`: `ES_COPY_HOST`, `OPS_COPY_HOST`, etc.  There are some additional
+parameters added:
+* `ES_COPY_SCHEME`, `OPS_COPY_SCHEME` - can use either http or https - defaults
+  to https
+* `ES_COPY_USERNAME`, `OPS_COPY_USERNAME` - user name to use to authenticate to
+  elasticsearch using username/password auth
+* `ES_COPY_PASSWORD`, `OPS_COPY_PASSWORD` - password to use to authenticate to
+  elasticsearch using username/password auth
+
+To set the parameters::
+
+    oc edit -n logging template logging-fluentd-template
+    # add/edit ES_COPY to have the value "true" - with the quotes
+    # add or edit the COPY parameters listed above
+    # automated:
+    #   oc get -n logging template logging-fluentd-template -o yaml > file
+    #   edit the file with sed/perl/whatever
+    #   oc replace -n logging -f file
+    oc delete daemonset logging-fluentd
+    # wait for fluentd to stop
+    oc process -n logging logging-fluentd-template | \
+      oc create -n logging -f -
+    # this creates the daemonset and starts fluentd with the new params
+
 ## Upgrading your EFK stack
 
 If you need to upgrade your EFK stack with new images, you'll need to take the
--- a/deployment/templates/fluentd.yaml
+++ b/deployment/templates/fluentd.yaml
@@ -86,6 +86,40 @@ objects:
               value: ${OPS_CLIENT_KEY}
             - name: "OPS_CA"
               value: ${OPS_CA}
+            - name: "ES_COPY"
+              value: ${ES_COPY}
+            - name: "ES_COPY_HOST"
+              value: ${ES_COPY_HOST}
+            - name: "ES_COPY_PORT"
+              value: ${ES_COPY_PORT}
+            - name: "ES_COPY_SCHEME"
+              value: ${ES_COPY_SCHEME}
+            - name: "ES_COPY_CLIENT_CERT"
+              value: ${ES_COPY_CLIENT_CERT}
+            - name: "ES_COPY_CLIENT_KEY"
+              value: ${ES_COPY_CLIENT_KEY}
+            - name: "ES_COPY_CA"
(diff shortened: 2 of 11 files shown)


deployment/templates/fluentd.yaml
```diff
@@ -107,6 +107,40 @@ objects:
...
+              value: ${ES_COPY_CLIENT_CERT}
+            - name: "ES_COPY_CLIENT_KEY"
+              value: ${ES_COPY_CLIENT_KEY}
+            - name: "ES_COPY_CA"
+              value: ${ES_COPY_CA}
+            - name: "ES_COPY_USERNAME"
+              value: ${ES_COPY_USERNAME}
+            - name: "ES_COPY_PASSWORD"
+              value: ${ES_COPY_PASSWORD}
+            - name: "OPS_COPY_HOST"
```
ewolinetz: I feel like we shouldn't care about a separate ops cluster for sending a copy, we should just send everything to a second location. @jcantrill thoughts?

jcantrill: @ewolinetz It would simplify the deployment but what could be the usecases for needing to separate the two?  Might there be a scenario where a customer's logs are sensitive to a cluster administrator. Contrived case, but what about departments in the same organization which need some sort of authorization to work with one another for legal reasons. You could make the argument it would be OK for an admin to understand what is going on with the node, but its not acceptable for them to read app logs without approval.

ewolinetz: That is true, I guess it's just that the template looks very noisy.  I'm not sure if there's another way we would accomplish this though.

ewolinetz:
We should have some further documentation around what it means to use this feature and describe the different options that can be provided

richm:
@ewolinetz I updated the README

richm:
check-es-copy.sh test passes

richm:
[test]

richm:
[test]

richm:
[test]

richm:
Once issue #89 is merged I will submit another PR for this fix without that commit in it.

richm:
Is this PR ok?  Is it ready to merge?

README.md
```diff
@@ -223,6 +223,41 @@ Mount the created secret to your Fluentd container
 oc volumes dc/logging-fluentd --add --type=secret --secret-name=fluentd-throttle --mount-path=/etc/throttle-settings --name=throttle-settings --overwrite
 ```
 
+## Have Fluentd send logs to another Elasticsearch
+
+You can configure Fluentd to send a copy of each log message to another
+Elasticsearch instance, _not_ the instance included with the log aggregation
+deployment.  Perhaps you already have an Elasticsearch instance that you are
```
ewolinetz: I don't think this is the correct use case for this feature.  To me, this is for when there are two different Elasticsearch instances that should be receiving the same logs, such as for auditing.  To simply send logs from the OpenShift fluentd to an external ES, we could just modify the current env vars.

I would make mention of sending copies of the logs to both the OpenShift hosted and an external.

fluentd/Dockerfile
```diff
@@ -14,21 +14,14 @@ LABEL io.k8s.description="Fluentd container for collecting of docker container l
...
-  mkdir -p ${HOME} && \
-  gem install --no-ri fluentd fluent-plugin-kubernetes_metadata_filter fluent-plugin-elasticsearch fluent-plugin-flatten-hash && \
-  mkdir -p ${THROTTLE_CONF_LOCATION} && \
-  touch ${THROTTLE_CONF_LOCATION}/settings
-
 ADD fluent.conf /etc/fluent/fluent.conf
 ADD configs.d/ /etc/fluent/configs.d/
-ADD run.sh generate_throttle_configs.rb ${HOME}/
+ADD install.sh run.sh generate_throttle_configs.rb \
+    fluentd_es_config.conf fluentd_es_ops_config.conf \
```
ewolinetz: we generate these configs now

ewolinetz: why wouldn't we already have these two under configs.d/output/ ?

fluentd/install.sh
```diff
@@ -0,0 +1,17 @@
...
+  ruby \
+  ruby-devel \
+  libcurl-devel \
+  make && \
+  gem install --no-ri fluentd fluent-plugin-kubernetes_metadata_filter fluent-plugin-elasticsearch fluent-plugin-flatten-hash && \
+  mkdir -p ${THROTTLE_CONF_LOCATION} && \
+  touch ${THROTTLE_CONF_LOCATION}/settings
+
+
+cp $HOME/fluentd_es_config.conf $HOME/fluentd_es_ops_config.conf /etc/fluent/configs.d/output
```
ewolinetz: same question as in Dockerfile

fluentd/Dockerfile
richm: They are.  I just thought it would be more consistent to have all of the fluentd_es_*_config.conf files under $HOME, then copy from there to /etc/fluent/configs.d/output as needed.

fluentd/install.sh
richm: I suppose I could just get rid of `$HOME/fluentd_es_config.conf` and  `$HOME/fluentd_es_ops_config.conf` and just copy them to `/etc/fluent/configs.d/output` in Dockerfile.

fluentd/configs.d/output/fluentd_es_ops_config.conf
```diff
@@ -0,0 +1,17 @@
+    <store>
+      @type elasticsearch_dynamic
+      host "#{ENV['OPS_HOST']}"
+      port "#{ENV['OPS_PORT']}"
+      scheme https
+      index_name .operations.${Time.at(time).getutc.strftime(@logstash_dateformat)}
```
ewolinetz: Can you match the index_name pattern from the original .../output/operations.conf ?

fluentd/fluentd_es_ops_copy_config.conf
```diff
@@ -0,0 +1,17 @@
+    <store>
+      @type elasticsearch_dynamic
+      host "#{ENV['OPS_COPY_HOST']}"
+      port "#{ENV['OPS_COPY_PORT']}"
+      scheme "#{ENV['OPS_COPY_SCHEME']}"
+      index_name .operations.${Time.at(time).getutc.strftime(@logstash_dateformat)}
```
ewolinetz: Same here

ewolinetz:
other than nits LGTM

ewolinetz:
[test]

ewolinetz:
pending test LGTM

PR #63, opened 2016-02-16
Daemon controller changes

ewolinetz:


deployment/templates/fluentd.yaml
```diff
@@ -58,20 +44,15 @@ objects:
             component: "fluentd"
         spec:
           serviceAccountName: aggregated-logging-fluentd
+          nodeSelector:
+            logging-infra: fluentd
           containers:
           - name: fluentd-elasticsearch
-            image: logging-fluentd
+            image: ${IMAGE_PREFIX}logging-fluentd
```
jcantrill: Is this necessary since we should be consuming an image from an imagestream which would abstract the location?

hack/templates/dev-builds.yaml
```diff
@@ -110,7 +110,7 @@ objects:
       dockerStrategy:
         from:
           kind: ImageStreamTag
-          name: centos:centos7
+          name: centos:7
```
jcantrill: Does this change need to be propagated to the other templates?

deployment/README.md
```diff
@@ -286,12 +286,11 @@ should be sufficient.
...
-actual number of nodes in your cluster.
+running as desired, label the nodes to deploy fluentd to in order to feed logs
+into ES. The example below would be to label a node named
+'ip-172-18-2-170.ec2.internal'.
 
-    $ oc scale dc/logging-fluentd --replicas=3
-    $ oc scale rc/logging-fluentd-1 --replicas=3
+    oc label node/ip-172-18-2-170.ec2.internal logging-infra=fluentd
 
 Kubernetes is developing a method of scheduling a pod to dynamically run
```
jcantrill: We need to evaluate this paragraph as it looks like this change counters this statement

deployment/templates/fluentd.yaml
ewolinetz: I need to verify further, but I did notice that I had issues with the daemonset finding the image when I was using a dev-build unless I specified an Image Prefix.

hack/templates/dev-builds.yaml
ewolinetz: Probably, it was an annoyance for me when I was testing this which is why it ended up committed.  I think if we're going to update all the sources for dev-builds it should be a separate PR.  Until I updated these `centos7` -> `7` they wouldn't even build.

deployment/templates/fluentd.yaml
ewolinetz: if we leave off ${IMAGE_PREFIX} it will try to pull in the logging-fluentd image from `library/logging-fluentd`.  Unless I'm mistaken that seems to be different than our other containers trying to pull from `openshift/` if they aren't dev builds.

hack/templates/dev-builds.yaml
```diff
@@ -17,7 +17,7 @@ objects:
       build: logging-elasticsearch
     name: centos
   spec:
-    dockerImageRepository: library/centos:centos7
+    dockerImageRepository: library/centos:7
```
richm: the "centos:centos7" to "centos:7" changes should be in a separate pr

ewolinetz: Thanks @richm it makes it easier for when I rebase

deployment/README.md
```diff
@@ -112,10 +112,10 @@ create secrets, templates, and deployments in the project. By default
 service accounts are not allowed to do this.
 
 The fluentd deployment also requires a service account which the deployer
-will create that must be given special privileges. Edit the privileged
+will create that must be given special privileges. Edit the hostmount-anyuid
 security configuration with the following:
 
-    $ oc edit scc/privileged
+    $ oc edit scc/hostmount-anyuid
```
sosiouxme: Use "oadm policy add-scc-to-user" command.

deployment/README.md
```diff
@@ -398,9 +388,9 @@ between the instances as they are restarted, we advise halting traffic to
 ElasticSearch and then taking down the entire cluster for maintenance. No
 logs will be lost; Fluentd simply blocks until the cluster returns.
 
-Halting traffic to ElasticSearch requires scaling down Fluentd and Kibana:
+Halting traffic to ElasticSearch requires scaling down Kibana and removing node labels for Fluentd:
 
-    $ oc scale rc/logging-fluentd-1 --replicas=0
+    $ oc label node <nodes> logging-infra-
```
sosiouxme: oc label node --all logging-infra-

deployment/README.md
```diff
@@ -286,22 +286,11 @@ should be sufficient.
...
-down as nodes come and go. The pod reserves a host port (1095 - which it
-does not actually listen on) in order to prevent multiple instances from
-running on the same node. However any pods that cannot be scheduled are
-still defined in the list of pods and may be perceived as clutter. This
-is only a workaround until the proper method of deployment is available.
+running as desired, label the nodes to deploy fluentd to in order to feed logs
+into ES. The example below would be to label a node named
+'ip-172-18-2-170.ec2.internal'.
+
+    oc label node/ip-172-18-2-170.ec2.internal logging-infra=fluentd
```
sosiouxme: Assuming they want all:

    oc label node --all logging-infra=fluentd

deployment/run.sh
```diff
@@ -230,7 +230,7 @@ Create the supporting definitions:
     oc process logging-support-template | oc create -f -
 
 Enable fluentd service account - edit SCC with the following
-    oc edit scc/privileged
+    oc edit scc/hostmount-anyuid
```
sosiouxme: `oadm policy` here too

ewolinetz:
@sosiouxme updated documentation per comments

ewolinetz:
as discussed with sosiouxme, want to update so the nodeSelector can be configured and will be `logging-infra-fluentd=true` by default

deployment/templates/fluentd.yaml
sosiouxme: In a DC the imagechangetrigger substitutes the real image name. I guess we don't have that in the case of a DaemonSet, it being a kube object. We would need the image suffix too...

ewolinetz:
Added nodeSelector configuration to be provided by deployer, added IMAGE_VERSION to image definition within daemonset

sosiouxme:
LGTM, squash and merge...

PR #67, opened 2016-02-18
Adding oc patch option for adding node selector for Elasticsearch

ewolinetz:
@sosiouxme PTAL


ewolinetz:
cc @jcantrill @richm

jcantrill:
LGTM

ewolinetz:
@richm the only gotcha I can think of is if you incorrectly add something or try to add something where it doesn't belong it's silently ignored.  So there's no direct feedback if you use this incorrectly.

PR #70, opened 2016-02-23
Allow ability to specify node selectors for ES, Kibana and Curator with deployer

ewolinetz:


ewolinetz:
Proposed changes for #62

ewolinetz:
@sosiouxme @richm PTAL

deployment/deployer.yaml
```diff
@@ -148,4 +172,51 @@ parameters:
   description: "Timeout for *expected* ops nodes to be present when cluster is recovering from a full restart."
   name: ES_OPS_RECOVER_AFTER_TIME
   value: "5m"
-
+-
+  description: "Node selector key for Elasticsearch cluster (key: value)."
+  name: ES_NS_KEY
```
sosiouxme: why not just use ES_NODESELECTOR (BTW NS = namespace) and not have separate KEY/VALUE?

    ES_NODESELECTOR=type=infra

Either way you can't specify multiple labels, so we don't get away from documenting the oc edit approach.

ewolinetz: That was my original plan, but i didn't try something like `...=key=value'.  It may be that since I was trying to assign with 'key: value' I ended up writing it this way to match the yaml structure.

I'll try to assign the value that way... and also not confuse nomenclature :)

ewolinetz: with using the sed approach, if I have the user pass in a single variable as `type=infra`, the result after the `oc process` is `"nodeSelector": "type=infra"` instead of the expected:
```
nodeSelector:
    type: infra
```

Or am I incorrect in thinking those two are different?

ewolinetz:
@sosiouxme addressed nomenclature and condensed number of parameters to provide

deployment/run.sh
```diff
@@ -24,6 +24,19 @@ es_ops_node_quorum=${ES_OPS_NODE_QUORUM:-$((es_ops_cluster_size/2+1))}
 es_ops_recover_after_nodes=${ES_OPS_RECOVER_AFTER_NODES:-$((es_ops_cluster_size-1))}
 es_ops_recover_expected_nodes=${ES_OPS_RECOVER_EXPECTED_NODES:-$es_ops_cluster_size}
 es_ops_recover_after_time=${ES_OPS_RECOVER_AFTER_TIME:-5m}
+# node selectors
+es_ns_key=$(echo $ES_NODESELECTOR | cut -f1 -d=)
+es_ns_value=$(echo $ES_NODESELECTOR | cut -f2 -d=)
```
sosiouxme: I think we should have a function here for turning the env var specification into a nodeselector. This could provide some validation and string safety as well as enable multi-label selectors (even though there's no way to enter one in the template, that could get fixed, or the env var could be directly edited instead of filled by parameter). Something like:

```
function join { local IFS="$1"; shift; echo "$*"; }
# untested, probably needs some fixes:
function extract_nodeselector() {
  local inputstring="${1//"/}"  # remove any errant double quotes in the inputs
  local -a selectors=()
  for keyvalstr in ${inputstring//\,/ }; do
    keyval=( "${keyvalstr//=/ }" )
    if [[ -n "${keyval[0]}" && -n "${keyval[1]}" ]]; then
      selectors=( "${selectors[@]}" "\"${keyval[0]}\": \"${keyval[1]}\"")
    else
      echo "Could not make a node selector label from '$keyval'"
      exit 255
    fi
    if [[ "${#selectors[*]}" > 0 ]]; then
      echo "nodeSelector: { $(join , ${selectors[*]}) }"
    fi
}
```

And then in your code:

```
es_ns=$(extract_nodeselector $ES_NODESELECTOR)
if [[ -n "${ES_NODESELECTOR}" ]]; then
	sed "/serviceAccountName/ i\
          $es_ns" templates/es.yaml | oc process ...
```

ewolinetz: #63 contains the function above, after that merges I'll reuse it for this branch

ewolinetz:
rebased on top of daemon controller changes and reused extract_nodeselector function

richm:
code lgtm, but I don't know anything about node selectors

ewolinetz:
[test]

deployment/run.sh
```diff
@@ -210,16 +217,55 @@ es_ops_params=$(join , \
...
+fi
+
 es_host=logging-es
 es_ops_host=${es_host}
-oc process -f templates/kibana.yaml -v "OAP_PUBLIC_MASTER_URL=${public_master_url},OAP_MASTER_URL=${master_url}" | oc create -f -
-oc process -f templates/curator.yaml -v "ES_HOST=${es_host},MASTER_URL=${master_url},CURATOR_DEPLOY_NAME=curator"| oc create -f -
+
+if [[ -n "${KIBANA_NODESELECTOR}" ]]; then
+	sed "/serviceAccountName/ i\
+\          ${kibana_nodeselecto}" templates/kibana.yaml | oc process -v "OAP_PUBLIC_MASTER_URL=${public_master_url},OAP_MASTER_URL=${master_url}" | oc create -f -
```
sosiouxme: typo here on nodeselector

richm: suggest using `set -o nounset` to avoid these sorts of errors
it will be painful at first . . .

ewolinetz:
This is currently failing due to openshift-elasticsearch-plugin/0.10 not available on maven.. working to resolve now

ewolinetz:
[test]

Issue #54, opened 2016-01-26
logging-fluentd connection refused to ElasticSearch


danbev:
We are running into an issue following the [documentation](https://access.redhat.com/documentation/en/openshift-enterprise/3.1/installation-and-configuration/chapter-17-aggregating-container-logs). We have basically followed the docs apart from calling the our project ```mbaas-logging``` instead of the default. 

The issue we are seeing is with the fluentd pod:
```shell
$ oc logs -f logging-fluentd-1-3925v
2016-01-25 08:02:35 -0500 [info]: reading config file path="/etc/fluent/fluent.conf"
2016-01-25 08:04:06 -0500 [warn]: temporarily failed to flush the buffer. next_retry=2016-01-25 08:02:48 -0500 error_class="Fluent::ElasticsearchOutput::ConnectionFailure" error="Can not reach Elasticsearch cluster ({:host=>\"logging-es.mbaas-logging.svc.cluster.local\", :port=>9200, :scheme=>\"https\", :user=>\"fluentd\", :password=>\"obfuscated\"})! Connection refused - connect(2) (Errno::ECONNREFUSED)" plugin_id="object:1421250"
  2016-01-25 08:04:06 -0500 [warn]: /usr/share/gems/gems/fluent-plugin-elasticsearch-1.0.0/lib/fluent/plugin/out_elasticsearch_dynamic.rb:61:in `rescue in client'
  2016-01-25 08:04:06 -0500 [warn]: /usr/share/gems/gems/fluent-plugin-elasticsearch-1.0.0/lib/fluent/plugin/out_elasticsearch_dynamic.rb:58:in `client'
  2016-01-25 08:04:06 -0500 [warn]: /usr/share/gems/gems/fluent-plugin-elasticsearch-1.0.0/lib/fluent/plugin/out_elasticsearch_dynamic.rb:184:in `rescue in send'
  2016-01-25 08:04:06 -0500 [warn]: /usr/share/gems/gems/fluent-plugin-elasticsearch-1.0.0/lib/fluent/plugin/out_elasticsearch_dynamic.rb:182:in `send'
  2016-01-25 08:04:06 -0500 [warn]: /usr/share/gems/gems/fluent-plugin-elasticsearch-1.0.0/lib/fluent/plugin/out_elasticsearch_dynamic.rb:174:in `block in write'
[...]

PR #71, opened 2016-02-24
Updating fluentd ES_HOST and OPS_HOST to just be service name and mat…

ewolinetz:
…ch pattern for kibana

Addresses #54


ewolinetz:
@richm or @sosiouxme ptal

PR #77, opened 2016-02-29
Remove wildcard from /var/log/messages

wshearn:
Having that wildcard causes fluentd to try and watch compressed messages files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,3 +210,50 @@ Mount the created secret to your Fluentd container
 ```
 oc volumes dc/logging-fluentd --add --type=secret --secret-name=fluentd-throttle --mount-path=/etc/throttle-settings --name=throttle-settings --overwrite
 ```
+
+## Upgrading your EFK stack
+
+If you need to upgrade your EFK stack with new images, you'll need to take the
+following steps to safely upgrade with minimal disruption to your log data.
+
+Scale down your Fluentd instances to 0.
+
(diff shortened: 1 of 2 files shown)


fluentd/configs.d/input/old-default-syslog.conf
```diff
@@ -23,7 +23,7 @@
 <source>
   @type tail
   @label @INGRESS
-  path /var/log/messages*
+  path /var/log/messages
```
ewolinetz: This will make it so we are only reading in the current syslog messages.  I'd recommend that we instead add an `exclude_path`

```
exclude_path ["/path/to/*.gz", "/path/to/*.zip"]
```

ewolinetz:
@wshearn we also need to make the same change in  close_file(...)

wshearn:
I am trying to think of how to handle the close, may have to handle it in the open section as it loops through file names there.

ewolinetz:
What about something like https://gist.github.com/ewolinetz/13655948a718ef4f4584

wshearn:
But you can see the issue here https://github.com/wshearn/origin-aggregated-logging/blob/wshearn-patch-1/fluentd/generate_throttle_configs.rb#L23 doing it that way.

ewolinetz:
@wshearn not following on how that'd be an issue.. sorry

PR #78, opened 2016-02-29
Exclude compressed logs

wshearn:
Re-open of PR #77 because I fail at git.


ewolinetz:
LGTM, thanks @wshearn !

PR #80, opened 2016-03-03
refine curator secret name according to secret naming rule

xiazhao2015:
Renamed curator secret name from index_management to index-management in the example doc  according to secret naming rule(could not contain underscore "_")


ewolinetz:
LGTM, thanks @xiazhao2015 !

PR #81, opened 2016-03-03
Add support for ops cluster, .defaults, runhour and runminute

richm:
Add support for testing curator with the ops cluster.

You can now use the settings .yaml file to set the default parameters
for the index deletion age, and the run hour and minute, instead of
using the env. vars./deploy parameters.

    .defaults:
      delete:
        days: 30
      runhour: 0
      runminute: 0

Changed CURATOR_CRON_HOUR and MINUTE to CURATOR_RUN_HOUR and
CURATOR_RUN_MINUTE.  Can use the settings yaml .default instead
to set runhour and runminute

Renamed DEFAULT_DAYS to CURATOR_DEFAULT_DAYS.

Added check-curator.sh test for runhour and runminute

### DIFF
--- a/curator/run_cron.py
+++ b/curator/run_cron.py
@@ -27,16 +27,28 @@ curator_settings = {'delete': {}}
 
 filename = os.getenv('CURATOR_CONF_LOCATION', '/etc/curator') + '/settings'
 
-decoded = []
+decoded = {}
 with open(filename, 'r') as stream:
-    decoded = yaml.load(stream) or []
+    decoded = yaml.load(stream) or {}
 
 connection_info = '--host ' + os.getenv('ES_HOST') + ' --port ' + os.getenv('ES_PORT') + ' --use_ssl --certificate ' + os.getenv('ES_CA') + ' --client-cert ' + os.getenv('ES_CLIENT_CERT') + ' --client-key ' + os.getenv('ES_CLIENT_KEY')
 
+defaults = {'delete': {'days': int(os.getenv('CURATOR_DEFAULT_DAYS', 30))}}
+deldefaults = defaults['delete']
+
+default_time_unit = decoded.get('.defaults', defaults).get('delete', deldefaults).keys()[0]
+if not default_time_unit in allowed_units:
+    logger.error('an unknown time unit of ' + default_time_unit + ' was provided... using days')
+    default_time_unit = 'days'
+
+default_value = int(decoded.get('.defaults', defaults).get('delete', deldefaults)[default_time_unit])
+
 base_default_cmd = '/usr/bin/curator --loglevel ERROR ' + connection_info + ' delete indices --timestring %Y.%m.%d'
-default_command = base_default_cmd + ' --older-than ' + os.getenv('DEFAULT_DAYS') + ' --time-unit days' + ' --exclude .searchguard*' + ' --exclude .kibana*'
+default_command = base_default_cmd + ' --older-than ' + str(default_value) + ' --time-unit ' + default_time_unit + ' --exclude .searchguard*' + ' --exclude .kibana*'
 
 for project in decoded:
+    if project == '.defaults':
+        continue
     for operation in decoded[project]:
         if operation in allowed_operations:
             for unit in decoded[project][operation]:
@@ -90,13 +102,30 @@ def run_all_jobs(joblist):
 # run jobs now
 run_all_jobs(my_cron)
 
-thehour = int(os.environ.get('CURATOR_CRON_HOUR', 0))
-theminute = int(os.environ.get('CURATOR_CRON_MINUTE', 0))
+thehour = decoded.get('.defaults', {}).get('runhour', None)
+if not thehour:
+    thehour = os.getenv('CURATOR_RUN_HOUR', None)
+if not thehour:
+    thehour = defaults.get('runhour', 0)
 
+theminute = decoded.get('.defaults', {}).get('runminute', None)
+if not theminute:
+    theminute = os.getenv('CURATOR_RUN_MINUTE', None)
+if not theminute:
+    theminute = defaults.get('runminute', 0)
+
+thehour = int(thehour)
+theminute = int(theminute)
 while True:
     # get time when next run should happen
     nextruntime = time.mktime(datetime.now().replace(hour=thehour, minute=theminute,
-                                                     second=0, microsecond=0).timetuple())+86400
+                                                     second=0, microsecond=0).timetuple())
+    timenow = time.time()
+    if nextruntime < timenow:
+        # the next runtime is less than now, so run a day from now
+        nextruntime = nextruntime + 86400
+        # else run later today
+    logger.debug("logging-curator hour [%d] minute [%d] nextruntime [%d] now [%d]" % (thehour, theminute, nextruntime, time.time()))
     # sleep until then
     time.sleep(nextruntime - time.time())
     run_all_jobs(my_cron)
--- a/hack/testing/check-curator.sh
+++ b/hack/testing/check-curator.sh
@@ -11,16 +11,116 @@ if [[ $# -ne 1 || "$1" = "false" ]]; then
   CLUSTER="false"
 else
   CLUSTER="$1"
-  ops="-ops"
+  OPS="-ops"
 fi
 
-TEST_DIVIDER="------------------------------------------"
+add_message_to_index() {
+    # index is $1
+    # message is $2
+    # ops is $3
+    message=${2:-"curatortest message"}
+    if [ -z "$3" ] ; then
+        host=logging-es
+    else
+        host=logging-es-ops
+    fi
+    ca=/etc/fluent/keys/ca
+    cert=/etc/fluent/keys/cert
+    key=/etc/fluent/keys/key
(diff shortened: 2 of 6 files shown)


richm:
@ewolinetz @sosiouxme ptal

curator/run_cron.py
```diff
@@ -27,16 +27,23 @@
...
 connection_info = '--host ' + os.getenv('ES_HOST') + ' --port ' + os.getenv('ES_PORT') + ' --use_ssl --certificate ' + os.getenv('ES_CA') + ' --client-cert ' + os.getenv('ES_CLIENT_CERT') + ' --client-key ' + os.getenv('ES_CLIENT_KEY')
 
+defaults = {'.defaults': {'delete': {'days': int(os.getenv('CURATOR_DEFAULT_DAYS', 30))}}}
+
+default_time_unit = decoded.get('.defaults', defaults).get('delete', defaults['.defaults']['delete']).keys()[0]
+default_value = int(decoded.get('.defaults', defaults).get('delete', defaults['.defaults']['delete'])[default_time_unit])
+
 base_default_cmd = '/usr/bin/curator --loglevel ERROR ' + connection_info + ' delete indices --timestring %Y.%m.%d'
-default_command = base_default_cmd + ' --older-than ' + os.getenv('DEFAULT_DAYS') + ' --time-unit days' + ' --exclude .searchguard*' + ' --exclude .kibana*'
+default_command = base_default_cmd + ' --older-than ' + str(default_value) + ' --time-unit ' + default_time_unit + ' --exclude .searchguard*' + ' --exclude .kibana*'
```
ewolinetz: We should probably validate `default_time_unit` against the allowed units
On that note, should we also validate that the time values for our projects are actually numbers?

curator/run_cron.py
```diff
@@ -27,16 +27,23 @@
...
 with open(filename, 'r') as stream:
-    decoded = yaml.load(stream) or []
+    decoded = yaml.load(stream) or {}
 
 connection_info = '--host ' + os.getenv('ES_HOST') + ' --port ' + os.getenv('ES_PORT') + ' --use_ssl --certificate ' + os.getenv('ES_CA') + ' --client-cert ' + os.getenv('ES_CLIENT_CERT') + ' --client-key ' + os.getenv('ES_CLIENT_KEY')
 
+defaults = {'.defaults': {'delete': {'days': int(os.getenv('CURATOR_DEFAULT_DAYS', 30))}}}
+
+default_time_unit = decoded.get('.defaults', defaults).get('delete', defaults['.defaults']['delete']).keys()[0]
+default_value = int(decoded.get('.defaults', defaults).get('delete', defaults['.defaults']['delete'])[default_time_unit])
```
sosiouxme: Python isn't my first language but... shouldn't `decoded.get('.defaults', defaults)` be `decoded.get('.defaults', defaults['.defaults'])` ?

Also the repetition of `decoded.get('.defaults', defaults).get('delete', defaults['.defaults']['delete']` seems like something that's begging to get put in a variable... DRY

hack/testing/check-curator.sh
```diff
@@ -29,15 +40,20 @@ TEST_DIVIDER="------------------------------------------"
 # key=./key-fluentd
 # oc get secret logging-fluentd --template='{{.data.cert}}' | base64 -d > cert-fluentd
 # cert=./cert-fluentd
-fpod=`oc get pods | awk '/-deploy/ {next}; /-build/ {next}; /^logging-fluentd-.* Running / {print $1}'`
-
-# where "yesterday", "today", etc. are the dates in UTC in YYYY.mm.dd format
+fpod=`oc get pods -l component=fluentd | awk '/fluentd/ {print $1}'`
```
sosiouxme: This is not prettier but it *does* prevent relying on the default output template:

    oc get pods -l provider=openshift,component=fluentd --template '{{with index .items 0}}{{.metadata.name}}{{end}}'

hack/testing/check-curator.sh
```diff
@@ -57,36 +73,36 @@ twomonthsago=`date -u +"$tf" --date="$(date +%Y-%m-1) -2 months -1 day"`
...
-# show current indices
-echo current indices are:
-oc exec $curpod -- curator --host logging-es${ops} --use_ssl --certificate /etc/curator/keys/ca \
-   --client-cert /etc/curator/keys/cert --client-key /etc/curator/keys/key --loglevel ERROR \
-   show indices --all-indices
-# add the curator config yaml settings file
-curtest=`mktemp --suffix=.yaml`
-cat > $curtest <<EOF
+    # get current curator pod
+    curpod=`oc get pods -l component=curator${ops} | awk '$1 ~ curator && $3 == "Running" {print $1}'`
```
sosiouxme: Do you really want to only use running ones, and if so should you have a check below to see if this comes back with anything?

hack/testing/check-curator.sh
```diff
@@ -100,66 +116,108 @@ project-prod:
...
-            echo good - index "$idx" is present
+    oc delete secret curator-config${ops} || echo no such secret curator-config${ops} - ignore
+    oc secrets new curator-config${ops} settings=$curtest
+    oc volumes dc/logging-curator${ops} --add --type=secret --secret-name=curator-config${ops} --mount-path=/etc/curator --name=curator-config --overwrite
+    # scale down dc
+    oc scale --replicas=0 dc logging-curator${ops}
+    # wait for pod to go away
+    ii=120
+    incr=10
+    while [ $ii -gt 0 ] ; do
```
sosiouxme: you have at least two of these... could they be a function? preferably one that results in an error if you actually time out on the while loop?

hack/testing/check-curator.sh
sosiouxme: It *is* possible to specify template with this too, but boy is it ugly:

```
curpod=`oc get pods -l component=curator${ops} --template='{{range .items}}{{.metadata.name}} {{.status.phase}}
{{end}}' | awk '$1 ~ curator && $2 == "Running" {print $1}'`
```

Sadly `oc get` doesn't provide any other filters except labels and namespace.

hack/testing/check-curator.sh
richm: Is there a problem with relying on the default output template?

hack/testing/check-curator.sh
richm: At this point in the code, yes, I really want to only use a running one - can I do an oc exec $pod if $pod is not running?  Most all of the tests in hack/testing make the implicit assumption that the pod under test is running.  This script is intended to be run via the e2e-test script which first runs check-EFK-running.  If using check-curator without that, the user should ensure that there is a curator pod running.

curator/run_cron.py
richm: The time values are "validated" by the `int()` function::

    >python
    >>> int("a")
      File "<stdin>", line 1, in <module>
    ValueError: invalid literal for int() with base 10: 'a'
    >>> 

So there will be a traceback in the curator logs.

richm:
@ewolinetz @sosiouxme Thanks for the reviews.  I think I have addressed all of your concerns.  ptal

curator/run_cron.py
```diff
@@ -90,13 +102,30 @@ def run_all_jobs(joblist):
 # run jobs now
 run_all_jobs(my_cron)
 
-thehour = int(os.environ.get('CURATOR_CRON_HOUR', 0))
-theminute = int(os.environ.get('CURATOR_CRON_MINUTE', 0))
+thehour = decoded.get('.defaults', {}).get('runhour', None)
+if not thehour:
+    thehour = os.getenv('CURATOR_RUN_HOUR', None)
+if not thehour:
```
ewolinetz: I'm not sure if this is possible in python, but couldn't you collapse this into a single statement and have something like `thehour = os.getenv('CURATOR_RUN_HOUR', defaults.get('runhour', 0))` ?

curator/run_cron.py
```diff
@@ -90,13 +102,30 @@ def run_all_jobs(joblist):
...
+    thehour = defaults.get('runhour', 0)
+
+theminute = decoded.get('.defaults', {}).get('runminute', None)
+if not theminute:
+    theminute = os.getenv('CURATOR_RUN_MINUTE', None)
+if not theminute:
```
ewolinetz: same question as above with `thehour`

richm: I did have it as a python one-liner originally, but based on the feedback from @sosiouxme I split it up this way for readability.

ewolinetz: fair enough :+1:

ewolinetz:
LGTM

hack/testing/check-curator.sh
sosiouxme: well... it can change. it has in the past (beta timeframe). it's not technically part of the API, just a client implementation.

now you could also say "well if the template changes our test will break and we'll fix it." and that's true enough.

still, if it's not outlandishly complicated, i'd prefer to have the client give us exactly what we're looking for rather than try to extract things from its output.

hack/testing/check-curator.sh
sosiouxme: come to think of it you *could* implement filtering in the template... but nevermind :)

PR #85, opened 2016-03-08
updating e2e testing to accomodate daemonset changes

ewolinetz:
Addressing - https://gist.github.com/ewolinetz/f403b9f37a0f79e16086

@richm @sosiouxme PTAL

### DIFF
--- a/hack/testing/check-EFK-running.sh
+++ b/hack/testing/check-EFK-running.sh
@@ -6,7 +6,9 @@ else
   set -e
 fi
 
-ADDITIONAL_PODS=$((KIBANA_CLUSTER_SIZE + ES_CLUSTER_SIZE - 2))
+# add one since fluentd will be deployed via a daemonset
+# keeping as -2 + 1 for readibilty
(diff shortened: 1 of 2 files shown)


ewolinetz:
@richm i'll check that out! if it works as expected i'll also try it out with the other elements we check on

ewolinetz:
@richm looks like that is really only feasible when we have one resource we're looking for (unless we iterate, which might be more complicated?) -- possibly in a future iteration?

richm:
jsonpath supports wildcards and iteration - http://kubernetes.io/v1.1/docs/user-guide/jsonpath.html::

    * 	wildcard. Get all objects 	{.items[*].metadata.name} 	[127.0.0.1 127.0.0.2]
    range, end 	iterate list 	{range .items[*]}[{.metadata.name}, {.status.capacity}] {end} 	[127.0.0.1, map[cpu:4]] [127.0.0.2, map[cpu:8]]

ewolinetz:
@richm awesome! let me try to use that

ewolinetz:
@richm @sosiouxme updating to use jsonpath to simplify getting current EFK objects PTAL

richm:
lgtm

PR #86, opened 2016-03-08
convert curator units of weeks to days

richm:
Because our timestring is %Y.%m.%d and does not contain weeks,
curator doesn't like asking for trimming in weeks, so convert
weeks to days.
Sample error message:

    ERROR. Timestring %Y.%m.%d does not contain time unit weeks


ewolinetz:
LGTM

PR #88, opened 2016-03-09
several logging automated testing improvements

richm:
* override LOG_DIR and ARTIFACT_DIR for vagrant
  download-artifacts-origin-aggregated-logging
* have tests output logs and other files to ARTIFACT_DIR so that
  they can be downloaded to the ci system
* use sudo to run check-logs.go because it needs access to the system
  logs and other protected files
* make e2e-test.sh only run the check-* scripts which do not modify
  data and are safe to run in production
* rename the other test scripts to be test-* - these do modify data
  and are not generally safe to run in production
* make the check and test output go directly to the console instead
  of being buffered by the test command
* remove the wildcard at the end of the hostname search in check-logs.go
* if check-logs.go finds no messages in the index, this is an error -
  if there is an index in ES, there must be at least 1 message in it
* fixes to allow working with set -o nounset

### DIFF
--- a/hack/testing/logging.sh
+++ b/hack/testing/logging.sh
@@ -32,6 +32,7 @@ USE_LOGGING_DEPLOYER=
 USE_LOGGING_DEPLOYER_SCRIPT=
 ENABLE_OPS_CLUSTER=${ENABLE_OPS_CLUSTER:-false}
 DEBUG_FAILURES=${DEBUG_FAILURES:-false}
+USE_LOCAL_SOURCE=${USE_LOCAL_SOURCE:-false}
 
 # includes util.sh and text.sh
 source "${OS_ROOT}/hack/cmd_util.sh"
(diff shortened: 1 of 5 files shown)


richm:
@ewolinetz @sosiouxme @jcantrill ptal
[test]

richm:
[test]

richm:
I think I finally have all of the issues resolve.  PTAL and merge asap and this should help with the issues with the other PR tests.

ewolinetz:
LGTM

hack/testing/logging.sh
```diff
@@ -264,7 +293,16 @@ if [ "$ENABLE_OPS_CLUSTER" = "true" ] ; then
 else
     USE_CLUSTER=
 fi
-os::cmd::expect_success "./e2e-test.sh $USE_CLUSTER"
+# e2e-test runs checks which do not modify any data - safe to use
+# in production environments
+./e2e-test.sh $USE_CLUSTER
```
ewolinetz: why do we no longer expect_success here?

richm: os::cmd::expect_success is really only for commands that either run for a very short period of time, or don't have a lot of output.  It saves the stderr and stdout until the command returns, which is not useful when you want to tail the logs in real time (i.e. from jenkins). Our test scripts `e2e-test.sh` and the other `test-*` scripts run for a long period of time and produce a lot of output so it isn't really suitable to use  `expect_success`.  `expect_success` will time the command and tell you how much time it took, and will also annotate the output e.g. red for failure and green for success.

ewolinetz: makes sense, sounds good

PR #102, opened 2016-04-01
deployer: enable using PVCs

sosiouxme:
Creates PVCs if they don't exist, attaches to ES DCs.

### DIFF
--- a/deployment/README.md
+++ b/deployment/README.md
@@ -141,16 +141,18 @@ are available:
 
 * `KIBANA_HOSTNAME` (required): External hostname where web clients will reach Kibana
 * `PUBLIC_MASTER_URL` (required): External URL for the master, for OAuth purposes
-* `IMAGE_PREFIX`: Specify prefix for logging component images; e.g. for "docker.io/openshift/origin-logging-deployer:v1.1", set prefix "docker.io/openshift/origin-"
-* `IMAGE_VERSION`: Specify version for logging component images; e.g. for "docker.io/openshift/origin-logging-deployer:v1.1", set version "v1.1"
-* `ES_INSTANCE_RAM`: Amount of RAM to reserve per ElasticSearch instance (e.g. 1024M, 2G). Defaults to 8GB; must be at least 512M (Ref.: [ElasticSearch documentation](https://www.elastic.co/guide/en/elasticsearch/guide/current/hardware.html#_memory).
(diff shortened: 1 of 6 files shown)


sosiouxme:
[test]

richm:
lgtm - In logging.sh, what do I need to do if I want to deploy with a PVC?  Create a PVC claim for es and es-ops?  Create directories in the filesystem for es and es-ops?

sosiouxme:
@richm I am not sure the best route to test this. If you're going to deploy with a PVC you have to have a PV or ES won't run. I would use host-mounts except those almost certainly have extra complications. Maybe we could set up an NFS server on the host?

richm:
egads - I don't think ES likes to use nfs e.g. google "elasticsearch nfs"
ES recommends the use of the local disk for storage (or something more fault tolerant and performant than nfs).
So the logging test should try to configure a local disk PV (which is what it does now).

richm:
ok - so we should test both (in case there is some weird bug that causes writes inside the pod to fail, but writes to "real" PVs still work).  The test will enable the ops cluster, but only configure the es-ops instance to use the real PV.  So how do I create a PV claim to use /var/lib/es-ops?

sosiouxme:
See hack/templates/pv-hostmount.yaml for an example (needs a tweak or two)

sosiouxme:
[test]

sosiouxme:
[test] again, ya?

sosiouxme:
etcd flake [test]

sosiouxme:
fscking flakes come on!
[test]

sosiouxme:
CI seems hosed today, try again later.

sosiouxme:
Let's try this again. [test]

richm:
The test failure is in the ES OPS instance: https://ci.openshift.redhat.com/jenkins/job/test-origin-aggregated-logging/116/artifact/origin/artifacts/logs/container-logging-es-ops-o5edh1wd-1-19fs8-elasticsearch.log/*view*/::

    [2016-04-15 14:19:17,369][INFO ][plugins                  ] [Shaper of Worlds] loaded [searchguard, openshift-elasticsearch-plugin, cloud-kubernetes], sites []
    {1.5.2}: Initialization Failed ...
    - ElasticsearchIllegalStateException[Failed to created node environment]
	AccessDeniedException[/elasticsearch/persistent/logging-es-ops]

sosiouxme:
Yeah, that's probably a real error, still with the permissions on the
persistent volume.

sosiouxme:
@ewolinetz PTAL

ewolinetz:
squash and LGTM!

PR #110, opened 2016-04-05
fix build-images script for origin release process

jcantrill:
cc @richm @sosiouxme  Updates to allow the origin release build to call our build script with a desired version for tagging.

### DIFF
--- a/hack/build-images.sh
+++ b/hack/build-images.sh
@@ -1,14 +1,35 @@
 #!/bin/bash
 
-set -ex
+set -o errexit
+set -o nounset
+set -o pipefail
 
-prefix=${PREFIX:-${1:-openshift/origin-}}
-version=${VERSION:-${2:-latest}}
-for component in fluentd elasticsearch kibana deployment curator ; do
-    docker build -t "${prefix}logging-${component}:${version}"       ../$component/
+STARTTIME=$(date +%s)
+
+prefix="${PREFIX:-openshift/origin-}"
+version="${OS_TAG:-latest}" 
+
+source_root=$(dirname "${0}")/..
+
+#################################
+for component in deployment fluentd elasticsearch kibana curator ; do
+  BUILD_STARTTIME=$(date +%s)
+  comp_path=$source_root/$component/
+  docker_tag=${prefix}logging-${component}:${version}
+  echo
+  echo
+  echo "--- Building component '$comp_path' with docker tag '$docker_tag' ---"
+  docker build -t $docker_tag       $comp_path
+  BUILD_ENDTIME=$(date +%s); echo "--- $docker_tag took $(($BUILD_ENDTIME - $BUILD_STARTTIME)) seconds ---"
+  echo
+  echo
 done
-if [ -n "${PUSH:-$3}" ]; then
-    for component in fluentd elasticsearch kibana deployment curator ; do
-	    docker push "${prefix}logging-${component}:${version}"
-    done
-fi
+
+echo
+echo
+echo "++ Active images"
+docker images | grep ${prefix}logging | grep ${version} | sort
+echo
+
+
+ret=$?; ENDTIME=$(date +%s); echo "$0 took $(($ENDTIME - $STARTTIME)) seconds"; exit "$ret"
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+
+# This script builds and pushes a release to DockerHub.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+OS_ROOT=$(dirname "${BASH_SOURCE}")/..
+source "${OS_ROOT}/hack/common.sh"
+
+# Go to the top of the tree.
+cd "${OS_ROOT}"
+
+if [[ -z "${OS_TAG}" ]]; then
+  echo "You must specify the OS_TAG variable as the name of the tag to create, e.g. 'v1.0.1'."
+  exit 1
+fi
+tag="${OS_TAG}"
+
+if [[ "$(git name-rev --name-only --tags HEAD)" != "${tag}^0" ]]; then
+  if git rev-parse -q --short "${tag}" &>/dev/null; then
+    echo "Tag ${tag} already exists"
+    exit 1
+  else
+    git tag "${tag}" -a -m "${tag}" HEAD
+  fi
+fi
+
+function removeimage() {
+  for i in $@; do
+    if docker inspect $i &>/dev/null; then
+      docker rmi $i
+    fi
+    if docker inspect docker.io/$i &>/dev/null; then
+      docker rmi docker.io/$i
+    fi
+  done
+}
+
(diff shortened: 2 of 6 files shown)


hack/build-images.sh
```diff
@@ -2,12 +2,18 @@
 
 set -ex
 
-prefix=${PREFIX:-${1:-openshift/origin-}}
-version=${VERSION:-${2:-latest}}
+prefix="openshift/origin-"
+version="latest"
+push=false
```
richm: So how is it possible to pass in the prefix, version, or push flag?

hack/build-images.sh
```diff
@@ -2,12 +2,18 @@
...
+
+source_root=$(dirname "${BASH_SOURCE}")/..
```
richm: how about:

    source_root=$(dirname "${0}")/..

I don't like using bash-isms where a pobs (plain old bourne shell) feature works.

hack/build-images.sh
jcantrill: looks like my changes did not make it in the comment.  fixing

hack/build-images.sh
jcantrill: fixed

jcantrill:
Updated per @richm comments

richm:
lgtm 

What is the usage for doing a release for a tag?  Is the user supposed to do a `git checkout $gittag` before running this script?

jcantrill:
We will tag our repo per the tag specified by @smarterclayton or whomever is doing a release.  The build and release scripts from origin will then call this script to build IS images

richm:
Ok - just as long as @smarterclayton or whoever does the release knows to do a `git clone https://github.com/openshift/origin-aggregated-logging -b $thetag`, or `git clone https://github.com/openshift/origin-aggregated-logging; cd origin-aggregated-logging; git checkout $thetag` before doing the `cd hack; ./build-images.sh ....`

jcantrill:
Per our recent conversation, set of scripts to perform releasing based on similar process for origin:

`OS_TAG=v1.1.7 hack/release.sh` to build and push images

jcantrill:
@richm @smarterclayton @sosiouxme please review

hack/release.sh
```diff
@@ -0,0 +1,53 @@
...
+    if docker inspect $i &>/dev/null; then
+      docker rmi $i
+    fi
+    if docker inspect docker.io/$i &>/dev/null; then
+      docker rmi docker.io/$i
+    fi
+  done
+}
+
+removeimage openshift/origin centos openshift/base-centos7
```
jcantrill: @richm @smarterclayton i have concerns about these following lines.  They are to ensure we have a clean build and the desired image.  This is basically taken from origin but does it apply here?  Can we defer to the build script to get the correct versions? Do we need to pull specific tags?

richm: Good question - I don't know

jcantrill:
fixes https://github.com/openshift/origin/issues/6628

hack/release.sh
smarterclayton: Isn't openshift/origin (the deployer) supposed to move to origin?  Removing openshift/base-centos7 seems appropriate since we'll need to grab the latest.

smarterclayton: I wouldn't expect to see "centos" in this list, but "centos:7" or "centos:centos7".  Our images should not be unqualified tags for core images.

richm: Yes, we talked about moving the deployer to origin, but that is a very big change - will have to be coordinated across a couple of origin+integration team sprints and releases

jcantrill: As long as we can parameterize the Dockerfile to use a base image of our choosing does it need to move to origin?  That is possible right?

jcantrill:
@smarterclayton @sosiouxme @richm Updated so the deployer image can now be based on a specific tagged version of origin.  please review.

hack/build-images.sh
```diff
@@ -1,14 +1,75 @@
...
+prefix="${PREFIX:-openshift/origin-}"
+version="${VERSION:-latest}"
+os_tag="${OS_PUSH_TAG:-latest}"
+
+source_root=$(dirname "${0}")/..
+
+# update the FROM statement in the Dockerfile to be a specific
+# tag.  Assumes there is only 1 FROM but will change them all
+# if there is more than one
+function create_dockerfile_using_base_tag {
```
smarterclayton: Two points:

1. Deployer shouldn't be in this repo - how far away are we from fixing that?
2. This is probably too complicated.  Just use :latest and have the caller be required to set the tag.

hack/release.sh
```diff
@@ -0,0 +1,53 @@
...
+    fi
+    if docker inspect docker.io/$i &>/dev/null; then
+      docker rmi docker.io/$i
+    fi
+  done
+}
+
+removeimage openshift/origin:${tag} centos:centos7 openshift/base-centos7
+
+docker pull openshift/origin:${tag}
```
smarterclayton: I'm going to run this script before I build origin.  So this will fail.

hack/build-images.sh
jcantrill: 1. I have no idea how far away we would be from that because there are all the associated scripts that get built into the deployer image that need to be validated.  Additionally, @richm has been working on CI for logging that executes e2e tests which need to be validated.  We need to think through what has to happen and put it through some testing cycles.  I imagine its at least a sprint worth of effort.
2. This allows us to have a working Dockerfile and additionally be able to modify the base image as needed.  Is that not desirable?

hack/release.sh
jcantrill: Shouldn't the logging release script be run after origin since the deployer depends on it?  I thought the idea was for you to have us tag, and then you would build all the images, with logging and metrics afterwards.

hack/build-images.sh
richm: Because of CI, once a commit is merged, it has been tested with the e2e checks and tests, with the latest origin.  So if we then git tag that commit that was merged, do we have to do another e2e test?

sosiouxme: @smarterclayton what's the logic behind not having the deployer code in this repo? It's the first I've heard of it. Most of this repo is deployer code.

hack/build-images.sh
```diff
@@ -1,14 +1,75 @@
...
+      if [[ $line =~ $regex ]]; then
+         from=${BASH_REMATCH[1]}:$tag
+         echo "Found FROM statement. updating to $from"
+         echo 
+         echo $from >> $file
+      else
+        echo $line >> $file
+      fi
+      
+  done < $source
```
sosiouxme: I think you could replace this with:

    sed -e "s/^\([:blank:]*FROM[^:]*:\).*/\1${tag}/" $source > $file

jcantrill: updated.

jcantrill:
@smarterclayton is this finally in the format desire and adhere to the process you would expect?

smarterclayton:
Yeah, looks fine.

Issue #107, opened 2016-04-05
hardcode for image version for es,curator,kibana components


chunyunchen:
https://github.com/openshift/origin-aggregated-logging/blob/master/deployment/templates/es.yaml#L43
https://github.com/openshift/origin-aggregated-logging/blob/master/deployment/templates/curator.yaml#L45
https://github.com/openshift/origin-aggregated-logging/blob/master/deployment/templates/kibana.yaml#L45

It's better to use IMAGE_VERSION parameter to specify image tag for above template instead of hardcode.

sosiouxme:
I think I tried it and something about it didn't work. But I need to take another look at this. We need to ensure it's not too hard to deploy a specific version as opposed to "latest".

chunyunchen:
For now, the "latest" tag is not point to the latest image version. Like ES image, the "latest" tag always point to tag "3.1.1-x", but not "3.2.0-x", so if we always use "latest", we should make sure it point to latest image.

Sometimes, we need to specify image tag for a special image, not all images use same tag like "latest".

sosiouxme:
As a short-term fix, you can tag anything you want as "latest" in the imagestream and it will be deployed that way, e.g.:

`oc tag logging-kibana:3.1.1 logging-kibana:latest`

This is obviously not ideal and I'm working on fixing the deployer to put the configured version in the deployments.

sosiouxme:
Fix in Origin. I also built OSE 3.2.0-5 and 3.1.1-13 with the fix; I'm not sure they made it to the QE registry but they should be available from pulp-brew.

chunyunchen:
@sosiouxme I tried with latest images from pulp-brew, the issue is still reproduced, only deployer pod can be used specific image with specifying tag, the kibana/es/fluentd pods still using "latest" tag, please refer to [1].

My test commands with specifying IMAGE_VERSION=3.2.0 as below:

oc process openshift//logging-deployer-template -v ENABLE_OPS_CLUSTER=false,IMAGE_PREFIX={pulp-brew-prefix},KIBANA_HOSTNAME={kibana-route},KIBANA_OPS_HOSTNAME={kibana-ops-route},PUBLIC_MASTER_URL=https://{master-domain}:8443,ES_INSTANCE_RAM=1024M,ES_CLUSTER_SIZE=1,IMAGE_VERSION=3.2.0,MASTER_URL=https://{master-doamin}:8443|oc create -f -

[...]

PR #111, opened 2016-04-06
deployer: use IMAGE_VERSION as intended

sosiouxme:
ref https://github.com/openshift/origin-aggregated-logging/issues/107


sosiouxme:
[test]

PR #112, opened 2016-04-06
some improvements for testing the es-copy feature

richm:
* code refactor
* add timeouts for curl so curl won't hang forever if es is down
* use polling instead of sleep to speed up test
* make sure the initial configuration does not use ES_COPY, and test
  it to make sure no copies are found
* try very hard to ensure final configuration does not use ES_COPY
* for the .operations log, search for the unique string in the ident
  field to avoid false positives

The way elasticsearch works, it will return a match with a very low
score, even when doing a search like message=$uuid, because some
components of the uuid may be in other log messages.

### DIFF
--- a/hack/testing/test-es-copy.sh
+++ b/hack/testing/test-es-copy.sh
@@ -23,6 +23,27 @@ if [ ! -d $ARTIFACT_DIR ] ; then
     mkdir -p $ARTIFACT_DIR
 fi
 
+# $1 - shell command or function to call to test if wait is over -
+#      this command/function should return true if the condition
+#      has been met, or false if still waiting for condition to be met
+# $2 - shell command or function to call if we timed out for error handling
+# $3 - timeout in seconds - should be a multiple of $4 (interval)
+# $4 - loop interval in seconds
+wait_until_cmd_or_err() {
+    let ii=$3
+    interval=${4:-1}
+    while [ $ii -gt 0 ] ; do
+        $1 && break
+        sleep $interval
+        let ii=ii-$interval
+    done
+    if [ $ii -le 0 ] ; then
+        $2
+        return 1
+    fi
+    return 0
+}
+
 get_running_pod() {
     # $1 is component for selector
     oc get pods -l component=$1 | awk -v sel=$1 '$1 ~ sel && $3 == "Running" {print $1}'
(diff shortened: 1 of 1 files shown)


richm:
[test]

richm:
@sosiouxme @jcantrill @ewolinetz ptal

hack/testing/test-es-copy.sh
```diff
@@ -23,6 +23,21 @@ if [ ! -d $ARTIFACT_DIR ] ; then
     mkdir -p $ARTIFACT_DIR
 fi
 
+wait_until_cmd_or_err() {
```
sosiouxme: with 4 args could you make a comment on what they are... just for clarity?

hack/testing/test-es-copy.sh
```diff
@@ -23,6 +23,21 @@ if [ ! -d $ARTIFACT_DIR ] ; then
...
+    ii=$3
+    interval=${4:-1}
+    while [ $ii -gt 0 ] ; do
+        $1 && break
+        sleep $interval
+        ii=`expr $ii - $interval`
```
sosiouxme: don't shell out for this...

    let ii=ii-interval

(nit)

richm:
[test]

@sosiouxme concerns addressed

hack/testing/test-es-copy.sh
```diff
@@ -65,21 +86,38 @@ wait_for_pod_ACTION() {
     return 0
 }
 
+test_count_expected() {
+    myfield=${myfield:-message}
+    nrecs=`oc exec $kpod -- curl --connect-timeout 1 -s -k --cert /etc/kibana/keys/cert --key /etc/kibana/keys/key \
+           https://${myhost}:9200/${myproject}*/_count\?q=${myfield}:${mymessage} | \
+           python -c 'import json, sys; print json.loads(sys.stdin.read())["count"]'`
```
sosiouxme: Factor this out? I hate to see something this intricate repeated.

hack/testing/test-es-copy.sh
```diff
@@ -65,21 +86,38 @@ wait_for_pod_ACTION() {
...
+}
+
+test_count_err() {
+    myfield=${myfield:-message}
+    nrecs=`oc exec $kpod -- curl --connect-timeout 1 -s -k --cert /etc/kibana/keys/cert --key /etc/kibana/keys/key \
+           https://${myhost}:9200/${myproject}*/_count\?q=${myfield}:${mymessage} | \
+           python -c 'import json, sys; print json.loads(sys.stdin.read())["count"]'`
+    echo Error: found $nrecs for project $myproject message $mymessage - expected $expected
+    oc exec $kpod -- curl --connect-timeout 1 -s -k --cert /etc/kibana/keys/cert --key /etc/kibana/keys/key \
+       https://${myhost}:9200/${myproject}*/_count\?q=${myfield}:${mymessage} | python -mjson.tool
```
sosiouxme:
```
for endpoint in _count _search; do ...
```

sosiouxme:
nits, otherwise LGTM

richm:
[test]

@sosiouxme concerns addressed

ewolinetz:
lgtm too, thanks @richm !

PR #113, opened 2016-04-08
Add steps to print ES query results more times

xiazhao2015:
This is in order to ease the curator testing observations in verbose log. The pass log of new script: http://pastebin.test.redhat.com/363864


xiazhao2015:
@richm Could you please help me review it? Thanks!

PR #117, opened 2016-04-16
deployer: cleanup; reencrypt routes

sosiouxme:
We can use new-app (-p) for processing templates in one step.
No longer have to string all params into one comma-sep param.

Also I would like service accounts all to be created up front.
This will enable future validation to check that the right roles and
SCCs are in place.

Also, use internal/external kibana certs and reencrypt route.
This will enable using a real cert with Kibana, which could be hard
before since the cert had to support multiple names.


sosiouxme:
[test]

sosiouxme:
flake on es_copy test? not sure. try [test] again.

sosiouxme:
@ewolinetz @richm PTAL

sosiouxme:
Forgot I need to make one tweak, to use the router default cert on the route if a cert isn't supplied.

sosiouxme:
There we go. Now PTAL.

deployment/README.md
```diff
@@ -156,9 +156,11 @@ are available:
 
 You run the deployer by instantiating a template. Here is an example with some parameters:
 
-    $ oc process -n openshift logging-deployer-template \
-               -v KIBANA_HOSTNAME=kibana.example.com,PUBLIC_MASTER_URL=https://localhost:8443 \
-               | oc create -f -
+    $ oc new-app logging-deployer-template \
+               -p KIBANA_HOSTNAME=kibana.example.com \
+               -p ES_CLUSTER_SIZE=1 \
+               -p ES_INSTANCE_RAM=1G \
```
ewolinetz: why are we setting this here? does `new-app -p` not pick up the template defaults?

sosiouxme: Sure it does, but this is a readme, so this is just an example. BTW the default is 8G I think...

ewolinetz: ack

deployment/scripts/install.sh
```diff
@@ -286,22 +286,22 @@ fi
 echo "(Re-)Creating deployed objects"
 if [ "${KEEP_SUPPORT}" != true ]; then
 	oc process logging-support-template | oc delete -f - || :
-	oc delete serviceaccount,service,route --selector logging-infra=support
+	oc delete imagestream,service,route --selector logging-infra=support
```
ewolinetz: do we want to also delete image streams by default? won't that make dev builds more difficult?

deployment/scripts/install.sh
```diff
@@ -286,22 +286,22 @@ fi
...
 	# note: no automatic deletion of persistentvolumeclaim; didn't seem wise
 	oc process logging-support-template | oc create -f -
+	kibana_keys=""; [ -e "$dir/kibana.crt" ] && kibana_keys="--cert='$dir/kibana.crt' --key='$dir/kibana.key'"
 	oc create route reencrypt --service="logging-kibana" \
 	                          --hostname="${hostname}" \
 	                          --{dest-,}ca-cert="$dir/ca.crt" \
-	                          --cert="$dir/kibana.crt" \
-	                          --key="$dir/kibana.key"
+                                  $kibana_keys
+	kibana_keys=""; [ -e "$dir/kibana-ops.crt" ] && kibana_keys="--cert='$dir/kibana-ops.crt' --key='$dir/kibana-ops.key'"
```
ewolinetz: is this a dupe of the line above it?

ewolinetz:
Other than a few questions LGTM

deployment/scripts/install.sh
sosiouxme: It only deletes them if they have the label indicating it also created
them. So it won't impact dev builds (which don't have the label). The only
impact is on testing where they have to do some fiddling to get the right
tags from nonsecure registry into the IS. I don't think they'll be doing
that iteratively (but if they are, they could remove the label to prevent
it being deleted). I think it's worth it because if we don't delete them,
changes to the IMAGE_PREFIX (oops! typo!) don't get picked up in the ISs
from one run to the next.

ewolinetz: okay that makes sense; maybe we can add a quick documentation 'gotcha' ?

richm:
LGTM

deployment/scripts/install.sh
sosiouxme: sure. ah, what's the "gotcha"? :) that ISs are normally recreated? that you
can remove the label to prevent this?

ewolinetz: :) yeah exactly, let users know that image streams will now be recreated and if you don't want this to happen (for some reason) you can remove the label to prevent that.

PR #119, opened 2016-04-18
Upgrade/uninstall/reinstall modes for deployer

ewolinetz:


deployment/scripts/upgrade.sh
```diff
@@ -0,0 +1,129 @@
...
+
+fluentd_nodeselector=${FLUENTD_NODESELECTOR:-logging-infra-fluentd=true}
+TIMES=300
+
+fluentd_nodes=""
+fluentd_nodeselector=""
+
+function getSvcVersion {
+  set +e
+  local version=`oc get svc logging-es -o jsonpath='{.metadata.annotations.LOGGING_VERSION}' 2>1`
```
richm: `2>&1` ?

ewolinetz: oops, thanks for the catch!

and that makes more sense to do; i'll make that change too

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
to make this easier to look at for review time, i'll break this into two commits -- one with the further refactoring and one with the upgrade functionality

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
[test]

richm:
aos-ci-test

ewolinetz:
[test]

ewolinetz:
I need to create automated testing, but doing manual tests worked.
In the interim @richm @sosiouxme PTAL

deployment/scripts/upgrade.sh
```diff
@@ -0,0 +1,370 @@
...
+  # generate the template
+  generate_curator_template
+  generate_curator
+}
+
+function fluentd_daemonset() {
+  echo "Updating Fluentd to use Daemonset..."
+
+  # Delete old deployment and create new one
+  generate_fluentd_template
```
ewolinetz: the nodeselector for fluentd doesn't appear to be on the resulting daemonset

ewolinetz: Seeing this in the logs:
`scripts/util.sh: line 110: selectors[@]: unbound variable`

I'll check if this also appears in the normal install...

deployment/scripts/install.sh
```diff
@@ -59,6 +59,33 @@ function procure_server_cert() {
...
+     --param KIBANA_HOSTNAME=${hostname} \
+     --param KIBANA_OPS_HOSTNAME=${ops_hostname} \
+     --param IMAGE_PREFIX_DEFAULT=${image_prefix}
+
+  oc process logging-support-template | oc delete -f - || :
+  oc delete imagestream,service,route --selector logging-infra=support
+  # note: dev builds aren't labeled and won't be deleted. if you need to preserve imagestreams, you can just remove the label.
+  # note: no automatic deletion of persistentvolumeclaim; didn't seem wise
+  oc new-app logging-support-template
+  kibana_keys=""; [ -e "$dir/kibana.crt" ] && kibana_keys="--cert='$dir/kibana.crt' --key='$dir/kibana.key'"
```
richm: https://github.com/openshift/origin-aggregated-logging/pull/122

deployment/scripts/upgrade.sh
```diff
@@ -0,0 +1,370 @@
...
+  # check for a fluentd dc, if there isn't one it was either deleted or using a daemonset
+  local fluentd_dc=(`oc get dc -l logging-infra=fluentd -o jsonpath='{.items[*].metadata.name}'`)
+  if [[ -z "$fluentd_dc" ]]; then
+    # oc get daemonset -l logging-infra=fluentd -o jsonpath='{.items[*].metadata.name}'
+    local selector=$(oc get daemonset -l logging-infra=fluentd -o jsonpath='{.items[*].spec.template.spec.nodeSelector}')
+    if [[ -n "$selector" ]]; then
+      fluentd_nodeselector=$(echo ${selector:4:-1} | sed 's/:/=/g')
+
+      # get a list of all nodes that we have labeled for fluentd to be deployed to so we can deploy to them again
+      fluentd_nodes=(`oc get node -l ${fluentd_nodeselector} -o jsonpath='{.items[*].metadata.name}'`)
```
ewolinetz: we can't do this with the latest roles on the deployer...

deployment/scripts/upgrade.sh
```diff
@@ -0,0 +1,370 @@
...
+
+      oc label node --all $(echo ${fluentd_nodeselector} | sed 's/=.*/-/g')
```
ewolinetz: we also can't do this with the latest roles on the deployer...

deployment/scripts/install.sh
ewolinetz: Thanks @richm I'll keep an eye on that PR and rebase on top of it

richm:
couple of minor nits, otherwise, lgtm

deployment/scripts/upgrade.sh
ewolinetz: I'm having trouble recreating this with an origin base install.  Not sure if this is due to me testing going from an enterprise install to an origin one... i'll have to investigate this further

deployment/deployer.yaml
```diff
@@ -67,6 +67,16 @@ items:
       - watch
       - delete
       - update
+  - apiVersion: v1
```
ewolinetz: @sosiouxme @liggitt any objections to letting the deployer be able to list and label (patch) nodes in the cluster for scaling down/up the fluentd pods as part of upgrading?

deployment/README.md
```diff
@@ -114,6 +114,8 @@ The deployer must run under a service account defined as follows:
     $ oc policy add-role-to-user daemonset-admin --serviceaccount logging-deployer
     $ oadm policy add-cluster-role-to-user oauth-editor \
               system:serviceaccount:logging:logging-deployer
+    $ oadm policy add-cluster-role-to-user daemonset-scheduler \
```
ewolinetz: I tried to add this first as a normal role but it didn't let me list and patch the nodes in the cluster as it would want to

deployment/deployer.yaml
liggitt: edit permission on nodes is an extremely high permission... how is that needed to scale up/down particular pods?

ewolinetz: We need to be able to label nodes so the fluentd daemonset (un)schedules pods.  Unless there is a verb for just labelling that I'm not aware of?

the deployer is trying to use `oc label nodes --all ${nodeSelectorLabel}-`

deployment/scripts/upgrade.sh
ewolinetz: I'm able to get this when it is an upgrade starting from version 0

deployment/deployer.yaml
liggitt: Again,  adding or removing labels on notes is a very highly privileged operation. It seems a little bit odd to make the nodes adjust to fit a daemonset.  Does this mean we're not setting nodes up correctly?

ewolinetz: The thought was to automate the upgrade process for the EFK stack as much as possible since its fairly involved.

After discussing with @sosiouxme we can get away with deleting the daemonset itself to unschedule the pods since we'll need to do that anyway if we are to make any changes to the daemonset as part of the upgrade.  And we can output instructions for labelling nodes to redeploy Fluentd after the upgrade instead.

sosiouxme: On Fri, Apr 22, 2016 at 6:03 PM, Jordan Liggitt <notifications@github.com>
wrote:
>
> Again, adding or removing labels on notes is a very highly privileged
> operation. It seems a little bit odd to make the nodes adjust to fit a
> daemonset. Does this mean we're not setting nodes up correctly?
>

@liggitt how would you suggest indicating which nodes a daemonset is
supposed to land on? The way we know of is to have a matching label on the
pod spec and on the nodes. So we label all the nodes we want the daemonset
to deploy to. This is easy for a cluster-admin, but makes an automated
upgrade (run by the deployer without such capabilities) impossible. But as
you say, node editing doesn't feel like something that should be handed to
the deployer.

So it's not that nodes are labeled wrong. It's that they require a
particular labeling for use with the fluentd daemonset, and that labeling
has to be done by a cluster-admin at some point. It adds a manual step for
the upgrade from pre-daemonset to post- but I don't really see that as a
problem.

liggitt: I would expect the admin to provide a label selector indicating what nodes they want fluentd to run on. How were the nodes being targeted before?

deployment/scripts/install.sh
sosiouxme: just remove the single quotes

sosiouxme:
What I did with apiman and metrics, and didn't get around to yet with logging, is split the delete+install bits out into modes "install" and "uninstall". Then include a "reinstall" mode that does both. I know it's a break from the current "install" functionality but I think it makes more sense to be explicit about what you want to happen. Is that something we can slip in here? It would improve some of the awkwardness.

deployment/deployer.yaml
sosiouxme: The admin can optionally provide one, or there's a default. Either way they
have to label all the nodes for fluentd...

Before daemonsets the nodes were targeted by the DC being manually scaled
up to the number of nodes.

ewolinetz:
@sosiouxme I'll take a look at those modes in the other deployers and see about getting those changes in here

ewolinetz:
[test]

ewolinetz:
etcd issue that time... [test]

ewolinetz:
etcd again [test]

ewolinetz:
Addressed nits, PTAL

ewolinetz:
aos-ci-test

deployment/scripts/upgrade.sh
```diff
@@ -0,0 +1,375 @@
...
+
+To label all nodes at once:
+    oc label nodes --all ${fluentd_nodeselector}
+
+Note: if your previous deployment used a DaemonSet for Fluentd, there should be
+no additional actions to deploy your pods -- the deployer did not unlabel any nodes.
+EOF
+}
+
+function regenerate_secrets() {
```
sosiouxme: This seems a bit, err... misnamed? given it nukes a lot more than secrets.

deployment/scripts/upgrade.sh
```diff
@@ -0,0 +1,375 @@
...
+  local elasticsearch_dc=(`oc get dc -l logging-infra=elasticsearch -o jsonpath='{.items[*].metadata.name}'`)
+  for dc in "${elasticsearch_dc[@]}"; do
+    oc scale --replicas=0 dc $dc
+  done
+  for pod in $(oc get pods -l component=es -o jsonpath='{.items[*].metadata.name}'); do
+    waitForStop $pod "component=es"
+  done
+}
+
+function scaleUp() {
```
sosiouxme: Can we have this and scaleDown as MODE entries so folks can do this "safely" should they need to do a cluster restart or something?

deployment/scripts/upgrade.sh
```diff
@@ -0,0 +1,375 @@
...
+
+  # scale up ES slowly -- we want to wait for each instance to recover its indices first
```
sosiouxme: Is it actually a problem to start all of them at once? Do they not recover their own indices before they start syncing with the other instances?

The pod will be marked Ready before they cluster - actually, as soon as it's running, I think. I'm not thrilled about starting them serially though because if you have enough members in the cluster and a few of them need to pull the ES image (for instance) you could exceed the cluster start timeout; then when the others join later, they'll be out of sync.

deployment/scripts/upgrade.sh
```diff
@@ -0,0 +1,375 @@
+#! /bin/bash
+
+set -ex
+
+TIMES=300
+
+fluentd_nodes=""
+fluentd_nodeselector="logging-infra-fluentd=true"
+
+# this should be updated to get the lowest found version
```
sosiouxme: ITYM highest found version.

Also could you rename this, something like `getDeploymentVersion` ...? Something that hopefully doesn't confuse with the image/platform version, but indicates the structural version. Or `getIntegrationVersion` or `getStructuralVersion`?

deployment/templates/fluentd.yaml
```diff
@@ -11,7 +11,7 @@ objects:
   metadata:
     name: logging-fluentd-template
     annotations:
-      description: "Template for logging fluentd deployment. Currently uses terrible kludge to restrict to one per node. TODO: adapt for daemon controller"
+      description: "Template for logging fluentd deployment."
```
sosiouxme: :+1: kept forgetting to do this

deployment/scripts/upgrade.sh
ewolinetz: my only concern with having it as a mode is we currently delete the daemonset to scale down Fluentd, not sure if we want to do that for just a cluster restart...

deployment/scripts/upgrade.sh
ewolinetz: yeah, i'll split up the logic within generate_secrets to not also create the support objects...

deployment/scripts/upgrade.sh
ewolinetz: i'll experiment with this in the latest commit so we have starts and stops more as the elastic docs state to do their major version upgrades.

deployment/scripts/upgrade.sh
```diff
@@ -0,0 +1,410 @@
...
+  local migrate=
+
+  # VERSIONS
+  # 0 -- just EFK
+  # 1 -- admin cert
+  # 2 -- curator & daemonset
+
+  # start with installed version and go until we're at $LOGGING_VERSION
+  if [[ $installedVersion -eq $LOGGING_VERSION ]]; then
+    echo "Already at latest version. No upgrade will be done at this time."
```
sosiouxme: Here's a thought; say they're at 3.2.1 and what they really want is to just move to 3.2.2 (no structural changes, let's hope). Could "upgrade" mode do that for them? They've specified IMAGE_VERSION=3.2.2 so scale everything down, go into the DCs and patch them (or just import updated image stream tags, if the version hasn't changed) and recreate the daemonset...? Otherwise they'll have to do all that manually.

ewolinetz: Hmm, yeah that shouldn't be too difficult to implement.  Do we think most users would be using the "latest" tag or an actual version?  I guess we could make the script just evaluate either, if its latest we can pretty much leave it and just import the latest images otherwise we'll also have to patch... hopefully patching the daemonset template isn't tricky :)

ewolinetz:
[test]

ewolinetz:
I temporarily changed our `oc new-app` commands to be `oc process ... | oc create -f -` until https://github.com/openshift/origin/pull/8674 is merged and available so that tests can still run.

ewolinetz:
currently this doesn't patch correctly with origin since the kibana-auth-proxy image tags differ from our other origin images.

@jcantrill can we remedy this?

ewolinetz:
[test]

deployment/scripts/upgrade.sh
ewolinetz: @sosiouxme how do you think we should handle revisions for our images especially since they may fall out of sync?  The deployer may only have 3.2.1-2 as the latest but ES may have 3.2.1-8.

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
Other than cleaning up some unused code (can't decide if its worth keeping the functions around), ready for review

sosiouxme:
Looks relatively clean for all you need to do here. Nice job. I'll try to review more in depth later.

I can't help thinking that all the oc commands should be some kind of safe_oc wrapper that is resilient to transitory issues like connections getting dropped or the master getting bounced, to the extent possible. Ideally this would be pretty robust, people don't like reinstalling from scratch when their upgrade chokes in the middle.

deployment/run.sh
```diff
@@ -30,11 +30,27 @@ case "${mode}" in
...
     install_logging
     ;;
+  uninstall)
+    delete_logging
+    ;;
+  reinstall)
+    delete_logging
+    install_logging
+    ;;
   migrate)
```
sosiouxme: do we still need migrate) now that upgrade) does the migration? could it just be combined into migrate|upgrade) ?

deployment/scripts/install.sh
```diff
@@ -1,9 +1,19 @@
 #!/bin/bash
 set -ex
 
+function delete_logging() {
+  initialize_install_vars
+  echo "Attempting to delete supporting objects (may fail)"
+  # delete oauthclient created in template; we can't search for it by label. the rest is incidental.
+  oc process logging-support-template | oc delete -f - || :
```
sosiouxme: echo "Deleting all other logging objects"

deployment/scripts/install.sh
```diff
@@ -1,9 +1,19 @@
...
+  oc delete all,templates,secrets --selector logging-infra
```
sosiouxme: might as well include `,daemonset` in this list?

deployment/run.sh
ewolinetz: in my mind they were still two different things... but we can probably just combine it.

deployment/scripts/install.sh
ewolinetz: I thought that when I had `,daemonset` in that list it didn't actually remove it.. i'll double check and make the appropriate changes.

deployment/run.sh
ewolinetz: on second thought, given that upgrade) only does the migration _if_ the admin-cert doesn't exist I think we should keep it separate so someone can explicitly make a migrate) call if needed...

ewolinetz:
Added doc updates and addressed previous comments

README.md
```diff
@@ -264,86 +264,35 @@ To set the parameters::
...
+           -p ES_CLUSTER_SIZE=1 \
+           -p PUBLIC_MASTER_URL=https://localhost:8443 \
+           -p MODE=upgrade
+
+Upgrade mode will take care of the following for you:
+  * Scale down your EFK deployment in a manner that will have minimal disruption
+  to your log data.
+  * Pull down the latest EFK image tags and patch your templates/DCs
+  * Perform any infrastructure changes in a non-destructive manner -- if you don't
+  yet have Curator, we won't delete your old ES instances
```
sosiouxme: saying "we" (here and below) feels odd. suggest "the upgrade" instead.

deployment/run.sh
sosiouxme: fair enough.

deployment/scripts/upgrade.sh
```diff
@@ -0,0 +1,470 @@
...
+  local version=0
+
+  # Check for the admin cert
+  if [[ -n "$(oc get secrets -o jsonpath='{.items[?(@.data.admin-cert)].metadata.name}')" ]]; then
+    version=1
+  else
+    echo "$version"
+    return
+  fi
+
```
sosiouxme: Could shorten above to this:

    if [[ -z "$(oc get secrets -o jsonpath='{.items[?(@.data.admin-cert)].metadata.name}')" ]]; then
      echo 0
      return
    fi

and similarly below

deployment/scripts/upgrade.sh
```diff
@@ -0,0 +1,470 @@
...
+  echo "$version"
+}
+
+function waitFor() {
+
```
sosiouxme: could shorten to

```
for (( i=1; i<=$TIMES; i++ )); do
  eval "$statement" && return 0
done
return 1
```

deployment/scripts/upgrade.sh
```diff
@@ -0,0 +1,470 @@
...
+  else
+    return 0
+  fi
+}
+
+# This lets us wait until the pod has been scheduled before we try to grab its name
+function getPodName() {
+
+  local dc=$1
+  local pod=$(waitForValue "oc get pod -l deploymentconfig=$dc -o jsonpath='{.items[?(.metadata.labels.deploymentconfig==\"$dc\")].metadata.name}'")
```
sosiouxme: the `local` here changes $? so it will always be 0 - do it separately.

deployment/scripts/upgrade.sh
```diff
@@ -0,0 +1,470 @@
...
+  else
+    echo "$pod"
+  fi
+}
+
+function checkESStarted() {
+
+  local pod=$1
+
+  local cluster_service=$(waitForValue "oc logs $pod | grep '\[cluster\.service[[:space:]]*\]'")
```
sosiouxme: local changes $?

ewolinetz:
[test]

deployment/scripts/upgrade.sh
```diff
@@ -0,0 +1,452 @@
...
+  if [[ $? -ne 0 ]]; then
+    echo "Kibana was not able to start up within $TIMES seconds"
+    return 1
+  fi
+}
+
+function scaleDown() {
+
+  # check for a fluentd dc, if there isn't one it was either deleted or using a daemonset
+  local fluentd_dc=(`oc get dc -l logging-infra=fluentd -o jsonpath='{.items[*].metadata.name}'`)
```
sosiouxme: do you think there's really any probability there will be multiple fluentd DCs here?

deployment/scripts/upgrade.sh
```diff
@@ -0,0 +1,452 @@
...
+
+  if [[ -z "$fluentd_dc" ]]; then
+    local selector=$(oc get daemonset -l logging-infra=fluentd -o jsonpath='{.items[*].spec.template.spec.nodeSelector}')
```
sosiouxme: ... and if so, what about multiple daemonsets here? :)

deployment/scripts/upgrade.sh
```diff
@@ -0,0 +1,452 @@
...
+    for dc in "${fluentd_dc[@]}"; do
+      oc scale --replicas=0 dc $dc
+    done
+  fi
+
+  for pod in $(oc get pods -l component=fluentd -o jsonpath='{.items[*].metadata.name}'); do
+    waitForStop $pod "component=fluentd"
+  done
+
+  local kibana_dc=(`oc get dc -l logging-infra=kibana -o jsonpath='{.items[*].metadata.name}'`)
```
sosiouxme: likewise... multiple kibana DCs?

deployment/scripts/upgrade.sh
sosiouxme: oh, wait, i see. one or none.

ewolinetz: it also loops through all found fluentd DCs just in case there are multiple

deployment/scripts/upgrade.sh
```diff
@@ -0,0 +1,452 @@
...
+  for dc in "${elasticsearch_dc[@]}"; do
+    pod=$(getPodName $dc)
+    waitForStart $pod
+
+    checkESStarted $pod
+  done
+
+  local kibana_dc=(`oc get dc -l logging-infra=kibana -o jsonpath='{.items[*].metadata.name}'`)
+  for dc in "${kibana_dc[@]}"; do
+    oc scale --replicas=1 dc $dc
```
sosiouxme: minor issue: this doesn't scale kibana back to where the user had it. that may be acceptable but we should be sure and note it somewhere...

ewolinetz: I thought about having the deployer do that, I guess since we also have this as a start/stop mode it should scale back up as best it can -- keeping track of the number of replicas it had before if possible and defaulting to replicas=1 otherwise.

sosiouxme: maybe put an annotation on the DC when shutting it down :)

ewolinetz: Sounds like an nice feature -- scaling back up to the previously specified amount when we scale down to 0 :)

ewolinetz:
Added feature to scale back up to previous number of replicas per DC based on replicas when stopping (Defaults to 1) and will honor replicas=0

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
Refactored how we were scaling up and down pods.  PTAL

ewolinetz:
[test]

richm:
aos-ci-test

deployment/scripts/uuid_migrate.sh
```diff
@@ -50,7 +50,8 @@ function create_context() {
 }
 
 function recreate_admin_certs(){
-
+  # lets not broadcast what our keys are for security reasons...
+  set +x
```
richm: Will this always be run with -x?  If you don't want to make that assumption, you can do this::

    usingx=$(echo $- | grep x)
    if [ -n "$usingx" ] ; then
      set +x
    fi
    ... # no trace here
    if [ -n "$usingx" ] ; then
      set -x
    fi

richm:
one optional nit
LGTM

ewolinetz:
aos-ci-test

ewolinetz:
will merge after tests complete

PR #122, opened 2016-04-21
Route reencrypt fails because of kibana_keys vars

thomaspepio:
The command ```oc create route reencrypt --service="logging-kibana" ...``` leads to :

```
# oc create route reencrypt --service=logging-kibana --hostname=kibana.node2.cnamts.octo.com --dest-ca-cert=/etc/deploy/ca.crt --ca-cert=/etc/deploy/ca.crt '--cert='\''/etc/deploy/kibana.crt'\''' '--key='\''/etc/deploy/kibana.key'\'''
# error: open '/etc/deploy/kibana.crt': no such file or directory
```

The error comes from mis-evaluation/interpretation of : ```kibana_keys="--cert='$dir/kibana.crt' --key='$dir/kibana.key'"```

Hence : splitting ```kibana_keys``` in two vars (```kibana_cert``` and ```kibana_key```), providing each flag with its content only.

### DIFF
--- a/deployment/scripts/install.sh
+++ b/deployment/scripts/install.sh
@@ -261,16 +261,20 @@ function generate_objects() {
     # note: dev builds aren't labeled and won't be deleted. if you need to preserve imagestreams, you can just remove the label.
     # note: no automatic deletion of persistentvolumeclaim; didn't seem wise
     oc process logging-support-template | oc create -f -
-    kibana_keys=""; [ -e "$dir/kibana.crt" ] && kibana_keys="--cert='$dir/kibana.crt' --key='$dir/kibana.key'"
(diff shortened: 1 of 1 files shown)


richm:
[test]

richm:
aos-ci-test

sosiouxme:
I'd rather not split these into two variables; it would allow defining a route with only one of key/cert, which would be invalid. Better to fail to create the route if one is missing (actually better to catch it earlier, but that's another story).

The problem is the superfluous single quotes. I have no idea why/when I put those in. Just take those out and it should work.

thomaspepio:
You're right, I amended the file according to your suggestions.

sosiouxme:
[test]

sosiouxme:
Thanks!

PR #123, opened 2016-04-21
fix prefix and pass through tag

jcantrill:


jcantrill:
updates per request from @smarterclayton

PR #127, opened 2016-04-29
handle authenticated/non-secured registry

sosiouxme:
provide deployer parameters for:
1. "insecure" registry in imagestream
2. applying pull secret if provided

also, DC pod specs now specify the actual image expected, although
normally the trigger will overwrite this. may lead to less confusion
when users try `oc deploy` explicitly due to IS failures.

### DIFF
--- a/deployment/scripts/install.sh
+++ b/deployment/scripts/install.sh
@@ -27,6 +27,7 @@ function install_logging() {
 function initialize_install_vars() {
   image_prefix=${IMAGE_PREFIX:-openshift/origin-}
   image_version=${IMAGE_VERSION:-latest}
+  insecure_registry=${INSECURE_REGISTRY:-false}
   hostname=${KIBANA_HOSTNAME:-kibana.example.com}
   ops_hostname=${KIBANA_OPS_HOSTNAME:-kibana-ops.example.com}
   public_master_url=${PUBLIC_MASTER_URL:-https://kubernetes.default.svc.cluster.local:443}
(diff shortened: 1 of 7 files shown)


sosiouxme:
@ewolinetz something to slip in for 3.2.1 - untested as yet due to the ongoing breakage of new-app.

deployment/scripts/install.sh
```diff
@@ -159,6 +160,12 @@ CONF
                    logging-fluentd
     oc secrets add serviceaccount/aggregated-logging-curator \
                    logging-curator
+    if [ -n "$IMAGE_PULL_SECRET" ]; then
+      for account in default aggregated-logging-{elasticsearch,fluentd,kibana}; do
```
ewolinetz: We should probably also add for curator and the auth-proxy too...

deployment/scripts/install.sh
```diff
@@ -183,27 +190,32 @@ function generate_templates() {
...
+
   create_template_optional_nodeselector "${ES_NODESELECTOR}" es \
     --param ES_CLUSTER_NAME=es \
     --param ES_INSTANCE_RAM=${es_instance_ram} \
     --param ES_NODE_QUORUM=${es_node_quorum} \
     --param ES_RECOVER_AFTER_NODES=${es_recover_after_nodes} \
     --param ES_RECOVER_EXPECTED_NODES=${es_recover_expected_nodes} \
     --param ES_RECOVER_AFTER_TIME=${es_recover_after_time} \
-    --param IMAGE_VERSION_DEFAULT=${image_version}
+    --param "$image_params"
```
ewolinetz: :+1: didn't realize we weren't already passing the prefix for these

ewolinetz:
@sosiouxme how do we want to order this with #119 ?

deployment/scripts/install.sh
sosiouxme: good point, auth-proxy doesn't have a separate SA but I think curator does.

deployment/scripts/install.sh
sosiouxme: well it's not strictly necessary; the trigger for getting the tag from the imagestream is supposed to overwrite whatever we put in the image field. however if that doesn't happen for some reason, and the user manually deploys, it's really confusing if the image is not what we intended.

sosiouxme:
@ewolinetz I imagine there are minor conflicts with #119 - if it's easy to rebase that on this then it would be nice to get it in sooner, but it can wait if that's going to cause pain.

sosiouxme:
#119 merged so rebased and [test]

ewolinetz:
LGTM

PR #131, opened 2016-05-11
updating fluentd filter to not overwrite message field if it already exists

### LINKED ISSUE #130: JSON is overwritten in Fluentd
If a log message is already contained within the 'message' field that Fluentd picks up, it is being overwritten by the contents of 'log'.

https://bugzilla.redhat.com/show_bug.cgi?id=1324996

ewolinetz:
this is to resolve #130 

@sosiouxme @richm

### DIFF
--- a/fluentd/configs.d/filter/k8s_record_transform.conf
+++ b/fluentd/configs.d/filter/k8s_record_transform.conf
@@ -3,8 +3,8 @@
   enable_ruby
   <record>
     hostname ${(kubernetes_host rescue nil) || File.open('/etc/docker-hostname') { |f| f.readline }.rstrip}
-    message ${log}
-    version 1.1.4
+    message ${(message rescue nil) || log}
+    version 1.2.0
   </record>
   remove_keys log,stream
 </filter>
--- a/fluentd/configs.d/filter/syslog_record_transform.conf
+++ b/fluentd/configs.d/filter/syslog_record_transform.conf
@@ -14,7 +14,7 @@
       time ${ Time.at(time) > Time.now ? (temp_time = Time.parse(Time.at(time).to_s.gsub(Time.at(time).year.to_s, (tag_parts[3].nil? ? Time.at(time).year.to_s : tag_parts[3][9,4]) )).to_datetime.to_s; Time.parse(temp_time) > Time.now ? Time.parse(temp_time.gsub(Time.parse(temp_time).year.to_s, (Time.parse(temp_time).year - 1).to_s )).to_datetime.to_s : Time.parse(temp_time).to_datetime.to_s ) : Time.at(time).to_datetime.to_s }
 
       #tag ${tag}_.operations_log
-      version 1.1.4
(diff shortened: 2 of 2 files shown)


richm:
LGTM

ewolinetz:
aos-ci-test

fluentd/configs.d/filter/k8s_record_transform.conf
```diff
@@ -3,7 +3,7 @@
   enable_ruby
   <record>
     hostname ${(kubernetes_host rescue nil) || File.open('/etc/docker-hostname') { |f| f.readline }.rstrip}
-    message ${log}
+    message ${(message rescue nil) || log}
     version 1.1.4
   </record>
   remove_keys log,stream
```
sosiouxme: probably fine, but note - we're deleting "log" whether or not it's been preserved. if the metadata has both "message" and "log" we throw the second away.

i can't think of a better approach though other than maybe concatenating them.

ewolinetz: Should we just preserve `log` then?  I was wanting to avoid having duplicate entries to avoid confusion, but that might be better to the alternative of dropping things...

richm: Under what conditions will a message have both `message` and `log`?  Can you provide an example?
Perhaps compare `message` and `log` - if they are the same, just keep `message`, otherwise keep both.  Also note that this may add to the amount of data per log entry in ES, if it is a frequent occurrence that an entry has both `log` and `message` and they are different.

sosiouxme: I don't think there are any *common* conditions where you'll have both. You
can put any metadata in that you want with JSON logging though, and that's
just the kind of thing that spurred this issue.

I'm not sure this DSL is adequately expressive for the logic you'd want.

On Wed, May 11, 2016 at 4:09 PM, Richard Megginson <notifications@github.com

richm: You could probably do it with some ruby - the hard part would be the remove_keys part - you could set log to '' or nil if you don't need it, and see what happens when the elasticsearch plugin formats it

ewolinetz: I think when ES gets it if its empty it gets entered as nil

ewolinetz: I noticed when building my test case to verify this issue that the contents of `log` include all the metadata as well as `message` when coming from Docker.  So I think we can safely remove it still.

ewolinetz:
We can test this using https://github.com/ewolinetz/log4j2-ex

ewolinetz:
bumped version fields since we're changing log formatting

sosiouxme:
worksforme

ewolinetz:
[test]

ewolinetz:
[test]

Issue #19, opened 2015-11-12
need docs on scaling elasticsearch


thoraxe:
we currently allude to some gotchas with scaling elasticsearch, but we may need to more explicitly document how to do it (simply).

ewolinetz:
Docs were updated per https://github.com/openshift/openshift-docs/pull/2136

ewolinetz:
it may also be good to note that we have https://github.com/openshift/origin-aggregated-logging/issues/151 to track creating some sort of scale up/down mode.  This _should_ all go away with the use of PetSets when available.

PR #136, opened 2016-05-16
Adding documentation on how to increase ES replicas

ewolinetz:
Addresses #19 

@sosiouxme @jcantrill PTAL

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,6 +294,32 @@ Upgrade mode will take care of the following for you:
   same as in previous versions. There should be minimal performance impact to ES
   while running this and it will not perform an install.
 
+## Adding Elasticsearch replicas
+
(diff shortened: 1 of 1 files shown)


sosiouxme:
This seems partially redundant with the [explanation in the deployer docs](https://github.com/ewolinetz/origin-aggregated-logging/tree/42d6f46a26f8baa927a74cfa272aec66520f06c7/deployment#elasticsearch-1). I would suggest merging something into that, and mention the ES parameters that should probably be adjusted for the new size of the cluster.

But I think issue 19 was really wanting something for this in official docs.

ewolinetz:
@sosiouxme i definitely missed that section in the deployer docs... i'll close this out and open a PR against openshift-docs that touches on your suggestion.

PR #139, opened 2016-05-18
Adding allowAnyScope to oauthclient definition

ewolinetz:
This is to resolve https://bugzilla.redhat.com/show_bug.cgi?id=1337083

Necessary since https://github.com/openshift/origin/commit/b59219345a959c9bae36516670cbf80678c87d0d

### DIFF
--- a/fluentd/Dockerfile
+++ b/fluentd/Dockerfile
@@ -5,12 +5,12 @@ MAINTAINER OpenShift Development <dev@lists.openshift.redhat.com>
 ENV HOME=/opt/app-root/src \
   PATH=/opt/app-root/src/bin:/opt/app-root/bin:$PATH \
   RUBY_VERSION=2.0 \
-  FLUENTD_VERSION=0.12.6 \
+  FLUENTD_VERSION=0.12.23 \
   GEM_HOME=/opt/app-root/src \
   THROTTLE_CONF_LOCATION=/etc/throttle-settings
 
 LABEL io.k8s.description="Fluentd container for collecting of docker container logs" \
-  io.k8s.display-name="Fluentd 0.12.6" \
+  io.k8s.display-name="Fluentd ${FLUENTD_VERSION}" \
   io.openshift.expose-services="9200:http, 9300:http" \
   io.openshift.tags="logging,elk,fluentd"
 
@@ -22,7 +22,7 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
   libcurl-devel \
   make && \
   mkdir -p ${HOME} && \
-  gem install --no-ri fluentd fluent-plugin-kubernetes_metadata_filter fluent-plugin-elasticsearch fluent-plugin-flatten-hash && \
+  gem install --no-rdoc --no-ri fluentd:${FLUENTD_VERSION} fluent-plugin-kubernetes_metadata_filter fluent-plugin-elasticsearch fluent-plugin-flatten-hash && \
   mkdir -p ${THROTTLE_CONF_LOCATION} && \
   touch ${THROTTLE_CONF_LOCATION}/settings
 
--- a/deployment/templates/support.yaml
+++ b/deployment/templates/support.yaml
@@ -128,6 +128,11 @@ objects:
     redirectURIs:
     - https://${KIBANA_HOSTNAME}
     - https://${KIBANA_OPS_HOSTNAME}
+    scopeRestrictions:
+    - literals:
+      - user:info
+      - user:check-access
+      - user:list-projects
 - apiVersion: "v1"
   kind: "Template"
   metadata:


ewolinetz:
@deads2k is this the best scope to use since this is just used to redirect requests?

deads2k:
> @deads2k is this the best scope to use since this is just used to redirect requests?

Probably not.  What permissions do you actually need?  If a cluster-admin uses your client to go through an oauth flow, whatever code gets that token will end up having the power to do things like deleting nodes.  You probably want to scope yourself down the powers you actually need.

This could be done by creating a role that enumerates the power you want and then creating a scope restriction saying this client can only request the powers of that role.  See https://github.com/openshift/openshift-docs/pull/2111 for an explanation of the current powers of scopes.

ewolinetz:
per conversation with @deads2k will be looking to change this to use a scoped user token with something like `user:info user:check-access user:list-projects`  currently `user:list-projects` doesn't exist

sosiouxme:
Arguably this only matters marginally as the logging auth proxy will have "powerful" tokens passed in all the time from the console. Unless the console is going to down-scope the tokens it passes on somehow?

Looks like the new scope will be in shortly, but if not we could probably get away with allowAnyScope.

deads2k:
> Arguably this only matters marginally as the logging auth proxy will have "powerful" tokens passed in all the time from the console. Unless the console is going to down-scope the tokens it passes on somehow?

@jwforres you pass the user's full-access token to other locations?  I'm assuming this is a curated list?  If I made us respect then, would you consider including a macaroon implementation into the console to down-scope them?  Cue @smarterclayton's smirk....

smarterclayton:
Smirk.

ewolinetz:
[test]

sosiouxme:
Looks like it ought to be:

```
    scopeRestrictions:
      exactValues:
      - user:info
      - user:check-access
      - user:list-projects
```

@deads2k does that look right?

deads2k:
> @deads2k does that look right?

Yes.  One wrinkle.  I found the need to eliminate scopes from the access check using `scopes = []string{}` when you make the actual SAR.  It doesn't affect the config, it affects the code and the results it will get.

What's neat is that old servers will drop both fields and I think they'll continue to work how you'd expect.

ewolinetz:
I had tried that but got the following:
```
[ec2-user@ip-172-18-2-204 ~]$ oc process logging-support-template | oc apply -f -
only encoded map or array can be decoded into a struct
[ec2-user@ip-172-18-2-204 ~]$ oc version
oc v1.3.0-alpha.0-615-gb1c26c8
kubernetes v1.3.0-alpha.1-331-g0522e63
```

```
- apiVersion: v1
  kind: OAuthClient
  metadata:
    name: kibana-proxy
  redirectURIs:
  - https://kibana.example.com
  - https://kibana-ops.example.com
  scopeRestrictions:
    exactValues:
    - user:info
    - user:check-access
    - user:list-projects
```

deads2k:
`scopeRestrictions` is an array, so you need the dash, right?

sosiouxme:
So

    scopeRestrictions:
      - exactValues:
        - user:info
        - user:check-access
        - user:list-projects

?

ewolinetz:
Ah, missed that it was an array...
```
  scopeRestrictions:
  - literals:
    - user:info
    - user:check-access
    - user:list-projects
```
seemed to be the magic incantation

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
[test]

fluentd/Dockerfile
```diff
@@ -22,7 +22,7 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
   libcurl-devel \
   make && \
   mkdir -p ${HOME} && \
-  gem install --no-ri fluentd fluent-plugin-kubernetes_metadata_filter fluent-plugin-elasticsearch fluent-plugin-flatten-hash && \
+  gem install --no-rdoc --no-ri fluentd:0.12.23 fluent-plugin-kubernetes_metadata_filter fluent-plugin-elasticsearch fluent-plugin-flatten-hash && \
```
richm: Can you update `FLUENTD_VERSION=0.12.23` and use `fluentd:${FLUENTD_VERSION}` instead?

ewolinetz:
fixed tests failing
@sosiouxme @richm ptal

richm:
lgtm

sosiouxme:
[merge]

richm:
[merge]

ewolinetz:
[merge]

richm:
@danmcp any idea why the `test_pull_requests --merge_pull_request 139 ..` is failing?::

    + test_pull_requests --merge_pull_request 139 --repo origin-aggregated-logging --config /var/lib/jenkins/.test_pull_requests_logging.json
      Merging pull request #139 for repo 'origin-aggregated-logging'
      Test merging pull request #139 for repo 'origin-aggregated-logging'
      Updating status of '7103c2aeb27ed2578ab48c1d83e9372a27b89887' with state: success
      Retrying...  attempt: 2
      Retrying...  attempt: 3
      Retrying...  attempt: 4
      Retrying...  attempt: 5
    /usr/share/ruby/net/http/response.rb:119:in `error!': 405 "Method Not Allowed (Net::HTTPServerException)
    	from /bin/test_pull_requests:487:in `merge_pull_request'
	from /bin/test_pull_requests:1454:in `<main>'
    Build step 'Execute shell' marked build as failure

richm:
Since the actual test is passing, just not the merge part, can we manually merge, until we figure out what's wrong with test_pull_requests --merge_pull_request?

danmcp:
@richm @ewolinetz I changed the perms on the repo (after I added merge) but forgot the bot needed special permissions.  Should be fixed.

PR #141, opened 2016-05-19
Updating upgrade to check before patching; removed dc change triggers

ewolinetz:
…after patch

### DIFF
--- a/deployer/scripts/install.sh
+++ b/deployer/scripts/install.sh
@@ -205,7 +205,7 @@ function generate_secrets() {
                    logging-curator
     if [ -n "$IMAGE_PULL_SECRET" ]; then
       for account in default aggregated-logging-{elasticsearch,fluentd,kibana,curator}; do
-        oc secrets add --for=pull "serviceaccount/$account" "secret/$IMAGE_PULL_SECRET" 
+        oc secrets add --for=pull "serviceaccount/$account" "secret/$IMAGE_PULL_SECRET"
       done
     fi
 
--- a/deployer/scripts/upgrade.sh
+++ b/deployer/scripts/upgrade.sh
@@ -5,6 +5,8 @@ set -ex
 TIMES=300
 fluentd_nodeselector="logging-infra-fluentd=true"
 
+patchPIDs=()
+
 function getDeploymentVersion() {
   #base this on what isn't installed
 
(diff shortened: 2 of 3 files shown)


ewolinetz:
Dependent on https://github.com/openshift/origin-aggregated-logging/pull/139

richm:
I don't see any obvious problems, other than we are really exercising bash more than I would like to . . .

ewolinetz:
updated import image per conversation around https://github.com/openshift/origin-aggregated-logging/pull/146

ewolinetz:
@richm @sosiouxme PTAL

ewolinetz:
currently checking if we actually need to `oc deploy` all the dc after an installation or not...

ewolinetz:
[test]

ewolinetz:
[test]

deployer/scripts/install.sh
```diff
@@ -373,6 +373,10 @@ function generate_objects() {
   generate_kibana
   generate_curator
   generate_fluentd
+
+  for dc in $(oc get dc -l logging-infra -o jsonpath='{.items[*].metadata.name}'); do
```
sosiouxme: If you just need the name, then `-o name` will do the job and is clearer.

Note output includes the `dc/`

deployer/scripts/upgrade.sh
```diff
@@ -278,20 +286,103 @@ function getArrayIndex() {
   fi
 }
 
+# This function will wait until the latestVersion changes
+function waitForChange() {
+
+  local currentVersion=$1
+  local dc=$2
```
sosiouxme: doesn't have to be a DC, could be any single object... maybe just change name of var to `object`?

ewolinetz: From what I've seen only DC have a `status.latestVersion`

deployer/scripts/install.sh
ewolinetz: ah cool, yeah that is clearer.  it would just be shortened to `oc deploy $dc --latest`

deployer/scripts/upgrade.sh
```diff
@@ -278,20 +286,103 @@ function getArrayIndex() {
...
+  return 1
+}
+
+# This function will tell us if the proposed patches are different than the current values
+function isValidChange() {
+
+  local object=$1
+  local patches=$2
+
+  for patch in $patches; do
```
sosiouxme: I don't like using a single parameter for multiple things like this, and I think it would be just as clear and more correct (if you ever had a patch with a space) as:

```
function isValidChange() {

  local object=$1; shift

  for patch in "$@"; do
  ...
}

isValidChange "$object" "${patches[@]}"

File: deployer/scripts/upgrade.sh [hunk 4]
```diff
@@ -278,20 +286,103 @@ function getArrayIndex() {
...
+  done
+
+  echo "Same values as intended patches for $object"
+  return 1
+}
+
+function patchIfValid() {
+
+  local object=$1
+  local patches=$2
```
sosiouxme: Likewise here. Of course you could be explicit:

    local patches=( "$@" )

deployer/scripts/upgrade.sh
sosiouxme: you're right, sorry, was thinking of metadata latest version which would be
different

ewolinetz:
there currently seems to be issues with es-ops starting up in the tests; investigating

deployer/scripts/upgrade.sh
```diff
@@ -278,20 +286,101 @@ function getArrayIndex() {
...
+
+# This function will tell us if the proposed patches are different than the current values
+function isValidChange() {
+
+  local object=$1; shift
+
+  for patch in "$@"; do
+    path="$(echo $patch | cut -d"=" -f 1)"
+    value="$(echo $patch | cut -d"=" -f 2)"
+
```
sosiouxme: result=$(oc get $object -o jsonpath="$path")

and use $result below

deployer/scripts/upgrade.sh
```diff
@@ -278,20 +286,102 @@ function getArrayIndex() {
...
+    currentVersion=$(oc get $object -o jsonpath='{.status.latestVersion}')
+  fi
+
+  if ! isValidChange "$object" "$@"; then
+    return 0
+  fi
+
+  for patch in "$@"; do
+    # delimeter is '='
+    path=$(echo $patch | cut -d"=" -f 1 | sed 's/[\.\[]/\//g ; s/\]//g ; s/[\}\{]//g')
```
sosiouxme: could this have a comment? it looks like you're patching the patch, but I don't see why.

Also looks like the last two could be compressed to one?

     sed 's/[\.\[]/\//g ; s/[\]\}\{]//g'

sosiouxme: e.g. "make 'foo' look like 'bar' when we're done"

ewolinetz: I'm not sure why it doesn't escape the `]` when doing the above, but it complains about an unmatched `\{`.  I'll play around with it a bit more to see if we can make it only two statements

ewolinetz:
[test]

deployer/scripts/upgrade.sh
```diff
@@ -278,20 +286,102 @@ function getArrayIndex() {
...
+    value=$(echo $patch | cut -d"=" -f 2)
+
+    actualPatch+=( "{\"op\":\"replace\",\"path\":\"$path\",\"value\":\"$value\"}")
```
sosiouxme: IMHO this looks better as

    actualPatch+=( '{"op":"replace","path":"'"$path"'","value":"'"$value"'"}')

deployer/scripts/upgrade.sh
```diff
@@ -278,20 +286,102 @@ function getArrayIndex() {
...
+  done
+
+  if oc patch $object --type=json -p="[$(join , ${actualPatch[@]})]"; then
```
sosiouxme: ${actualPatch[@]}

Then you could put actual spaces in the patch...

deployer/scripts/upgrade.sh
sosiouxme: you just about need a unit test for this.

FWIW you can do it easily with bash substitution but that doesn't look much prettier.

deployer/scripts/upgrade.sh
ewolinetz: is there a reason why we'd put quotes around `$path` and `$value`?  couldn't we just have 
```
actualPatch+=( '{"op":"replace","path":"'$path'","value":"'$value'"}' )
```

sosiouxme: That works fine as long as there are no spaces in $path or $value, which as
you can tell i hate to rely on...

sosiouxme:
:shipit:

deployer/scripts/upgrade.sh
ewolinetz: I don't think I'm following.. even if there's still spaces it should expand all the same.  

```
[ec2-user@ip-172-18-2-165 ~]$ value="logging-es something else"
[ec2-user@ip-172-18-2-165 ~]$ echo '{"op":"replace","path":"'$path'","value":"'$value'"}'
{"op":"replace","path":"/spec/containers/0/spec/image","value":"logging-es something else"}
```

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
to address https://bugzilla.redhat.com/show_bug.cgi?id=1339513 i'm going to split the dc trigger removal into a separate PR and change this PR to just address the upgrader timing/patch changes.

ewolinetz:
[test]

richm:
LGTM

ewolinetz:
[test]

ewolinetz:
[merge]

PR #142, opened 2016-05-19
Allow reading system logs from the journal instead of /var/log/messages

richm:
This introduces a new environment variable: USE_JOURNAL.  If this is
set to "true", fluentd will use the systemd-journal input plugin to
read log messages from the journal instead of /var/log/messages*.
The immediate benefit is performance - reading log entries directly
from the journal rather than having them go from the journal to
rsyslog to /var/log/messages.  Another benefit is the microsecond
precision timestamps instead of seconds used by /var/log/messages.
There is a lot of metadata available from the journal which we can
now use.
This is not yet a complete replacement for the file based input.

* Need to handle log throttling.  Might have to be done at the
  journal level - there aren't any sort of throttling configuration
  options in https://github.com/reevoo/fluent-plugin-systemd
  journald (man journald.conf) has RateLimitInterval=, RateLimitBurst=
  options.
* Need to handle the case if user has configured persistent journal
  logs (man journald.conf - Storage=) so may need to read from
  /var/log/journal in addition to/instead of /run/log/journal
* Need to handle historical logs.  The default configuration for
  journald.conf on el7 is `Storage=auto` which means /run/log/journal
  will be used for `volatile` storage unless the user has created
  /var/log/journal.  `volatile` means not persistent, so the persistent
  logs by default will be /var/log/messages*.  We need some way to grab
  those logs, and de-dup them before storing in elasticsearch.


richm:
[test]

richm:
This requires https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/24 and docker version 1.10 or later and docker daemon configured to use `--log-driver=journald` in order to actual use the fluentd journal reader.

richm:
@ewolinetz @sosiouxme PTAL

fluentd/configs.d/filter/retag_journal.conf
```diff
@@ -0,0 +1,66 @@
...
+  rewriterule1 CONTAINER_NAME ^k8s_.*kibana          kubernetes.journal.container.kibana
+  # mark default container logs as system logs
+  rewriterule2 CONTAINER_NAME ^k8s_.*default         kubernetes.journal.container._default_
+  # mark openshift-infra container logs as system logs
+  rewriterule3 CONTAINER_NAME ^k8s_.*openshift-infra kubernetes.journal.container._openshift-infra_
+  # mark openshift container logs as system logs
+  rewriterule4 CONTAINER_NAME ^k8s_.*openshift       kubernetes.journal.container._openshift_
+  # not kibana - see if this is a kubernetes container log
+  rewriterule5 CONTAINER_NAME ^k8s_.*fluentd         kubernetes.journal.container.fluentd
+  rewriterule6 CONTAINER_NAME ^k8s_                  kubernetes.journal.container
```
ewolinetz: are we losing the project that the logs came from in this case, or do we recover that with the metadata later on?

richm: The full metadata is still encoded in the `CONTAINER_NAME` field - this re-tagging is just for routing purposes.  Every message tagged with `kubernetes.*` still goes through the k8s metadata plugin filter.

ewolinetz:
lgtm

richm:
[test]

richm:
@ewolinetz @sosiouxme @timothysc @detiber PTAL - also requires https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/25

richm:
fabric8io/fluent-plugin-kubernetes_metadata_filter#25 was merged - waiting for build and release of 0.23.0

fluentd/configs.d/filter/cloud_init.conf
```diff
@@ -0,0 +1,5 @@
+<filter journal>
+  @type grep
+  # drop messages from cloud-init - they are not logged to /var/log/messages by default
+  exclude1 _COMM ^cloud-init$
```
t0ffel: is there really a need to remove this? it still goes to the logs which admin can see.

fluentd/configs.d/filter/retag_journal.conf
```diff
@@ -0,0 +1,74 @@
...
+
+<match journal>
+  @type rewrite_tag_filter
+  # see if this is a kibana container for special log handling
+  # looks like this:
+  # k8s_kibana.a67f366_logging-kibana-1-d90e3_logging_26c51a61-2835-11e6-ad29-fa163e4944d5_f0db49a2
+  # we filter these logs through the kibana_transform.conf filter
+  rewriterule1 CONTAINER_NAME ^k8s_kibana\. kubernetes.journal.container.kibana
+  # mark default container logs as system logs
+  rewriterule2 CONTAINER_NAME ^k8s_.*_default_ kubernetes.journal.container._default_
```
t0ffel: can user's pod named xxx_default_xx match this?

fluentd/configs.d/filter/k8s_record_transform.conf
```diff
@@ -1,4 +1,24 @@
-<filter kubernetes.**>
+<filter kubernetes.journal.container**>
+  type record_transformer
+  enable_ruby
+  <record>
+    hostname ${(kubernetes_host rescue nil) || File.open('/etc/docker-hostname') { |f| f.readline }.rstrip}
+    # log is set if MESSAGE is JSON valued and has err or msg field
+    message ${record['MESSAGE'] || log}
+    version 1.2.0
+    time ${Time.at((record["_SOURCE_REALTIME_TIMESTAMP"] || record["__REALTIME_TIMESTAMP"]).to_f / 1000000.0).to_datetime.rfc3339(6)}
```
t0ffel: is this different from syslog_record_transform.conf line 32 because docker always fills one of these fields?

fluentd/configs.d/filter/cloud_init.conf
richm: In general, what I'm trying to do with this patch is complete feature parity with the current version of OpenShift with aggregated logging - no more, no less - so we can easily verify with current testing, CI, QE, and documentation that using fluentd with journal works exactly like using fluentd with in_tail files.

So, this is needed for the use case in the current version of OpenShift.  I don't know what customers will do if they suddenly begin seeing cloud-init logs in their Elasticsearch - will it flood ES?  Will it cause strange behavior with existing kibana queries e.g. queries return a lot more hits than expected and other false positives?

fluentd/configs.d/filter/k8s_record_transform.conf
richm: It is different because log records that come from a kubernetes container will have the kubernetes_host field added by the k8s metadata plugin.  For syslog records, there is only the _HOSTNAME field.

fluentd/configs.d/filter/retag_journal.conf
richm: Yes, but that's the way it works now, when using in_tail to read from the /var/log/containers/*.log files.

richm:
[test]

fluentd/configs.d/filter/k8s_record_transform.conf
```diff
@@ -1,4 +1,24 @@
-<filter kubernetes.**>
+<filter kubernetes.journal.container**>
+  type record_transformer
+  enable_ruby
+  <record>
+    hostname ${(kubernetes_host rescue nil) || File.open('/etc/docker-hostname') { |f| f.readline }.rstrip}
+    # log is set if MESSAGE is JSON valued and has err or msg field
+    message ${record['MESSAGE'] || log}
+    version 1.2.0
```
ewolinetz: we should probably bump all `version` fields up to `1.3.0` since we're introducing some formatting changes (with the introduction of journal)

fluentd/generate_throttle_configs.rb
```diff
@@ -102,8 +102,9 @@ def create_default_docker(excluded)
 
   file_name = "/etc/fluent/configs.d/input/docker/default-docker.conf"
 
-  File.open(file_name, 'w') { |file|
-    file.write(<<-CONF)
+  unless ENV['USE_JOURNAL'] == "true"
```
ewolinetz: so using Journal replaces the use of the default Docker config?  what bout if the user is trying to also throttle, would they end up with a combination of journal and in_tail?

fluentd/configs.d/filter/k8s_record_transform.conf
richm: @ewolinetz This commit does not introduce any formatting changes AFAIK.  I wanted to make the output, when using journal, identical to the output when reading from files, for ease of migration.  The only change is that the timestamp now has microsecond resolution when reading from journal and writing to the `.operations.*` index.

hack/testing/logging.sh
```diff
@@ -438,6 +454,10 @@ else
     done
 fi
 
+if [ "$TEST_MIGRATE_UPGRADE" = "false" ] ; then
```
ewolinetz: this overlaps and is probably of no use with https://github.com/openshift/origin-aggregated-logging/pull/162

fluentd/generate_throttle_configs.rb
richm: Yeah, that's a big problem.  I'm not sure what to do about that since we can't throttle from journal (in fluentd, anyway).

fluentd/configs.d/filter/k8s_record_transform.conf
ewolinetz: @richm I guess that is more of what I meant, its different than what we are publishing with our current configuration.  `version` is more just for a quick sanity check and would make it easier for verifying that logs are the same -- we could look at logs with version=1.2.0 and version=1.3.0 and ensure they are similar enough.

fluentd/generate_throttle_configs.rb
ewolinetz: Can we just not create any throttle configs in the case where USE_JOURNAL == true?  e.g. move this check to here? https://github.com/richm/origin-aggregated-logging/blob/11e11e93caec61914ab853db912ebcc928ad318c/fluentd/generate_throttle_configs.rb#L241

ewolinetz: Sorry, I meant here https://github.com/richm/origin-aggregated-logging/blob/11e11e93caec61914ab853db912ebcc928ad318c/fluentd/generate_throttle_configs.rb#L201

richm:
[test]

richm:
[test]

fluentd/configs.d/filter/retag_journal.conf
sosiouxme: pod names can't have underscores so it's nota problem

sosiouxme:
Late to the party...

- Need to handle the case if user has configured persistent journal logs (man journald.conf - Storage=) so may need to read from /var/log/journal in addition to/instead of /run/log/journal

I think we do need to do this. Requiring ephemeral journal seems like something customers would balk at. It looks to me like the journal will write to either ephemeral or persistent, but not both at the same time (the complication is at boot time, before /var/log is RW, it writes to /run/log, then switches over). We're already mounting /var/log so ... why not just read the journal in both locations?

- Need to handle historical logs. The default configuration for journald.conf on el7 is Storage=auto which means /run/log/journal will be used for volatile storage unless the user has created /var/log/journal. volatile means not persistent, so the persistent logs by default will be /var/log/messages*. We need some way to grab those logs, and de-dup them before storing in elasticsearch.

Actually I would count this as a feature. I don't think people expect us to index the syslogs going back to the beginning of time if they've configured this feature. In fact they might be surprised if their voluminous old logs are there taking up space - has been a problem in the past.

fluentd/configs.d/filter/retag_journal.conf
richm: hmm - the current code doesn't have this problem::

    var\.log\.containers\.(?<pod_name>[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*)_(?<namespace>[^_]+)_(?<container_name>.+)-(?<docker_id>[a-z0-9]{64})\.log$

The pod_name always follows `var.log.containers.` so there is no ambiguity between a pod name of `default` vs. a namespace name of `default`.

richm:
@sosiouxme "I don't think people expect us to index the syslogs going back to the beginning of time if they've configured this feature."  Well, this is how it works now, when reading from `/var/log/messages*` https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/generate_throttle_configs.rb#L110 and https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/generate_throttle_configs.rb#L132

In both of these cases, `read_from_head` is `true` - which AFAICT means when fluentd is installed for the first time on the host/node, it will read everything from every log going back to be beginning of time.  My initial journal commit was just to keep feature parity with the existing way of doing things, which also used `read_from_head true`

However, when upgrading/changing an existing deployment to use journal, you _do not_ want `read_from_head true` because you have no way of de-duping all of those records.

So I think the best way to do it is to make the default `read_from_head true` as it is in the fluentd case, and use `read_from_head false` when upgrading.

sosiouxme:
I know the current implementation reads /var/log/messages since the
beginning of time, but I don't think users actually want that. I wasn't
aware there was an option to read from tail. I think we do not want to read
from head in either case when first installed to the system, however I
don't know how we would make the distinction as we almost certainly want to
read from head the rest of the time. If it is simpler to always read from
head, we should do that.

Mainly, if users choose to read from the journal, I don't think they care
if fluentd goes back and reads up on /var/log/messages since the beginning
of time and tries to dedup that. I think it's an unintended consequence
that the current implementation does that.

On Fri, Jun 17, 2016 at 2:28 PM, Richard Megginson <notifications@github.com

ewolinetz:
The more I think about it, the more I feel like we shouldn't be doing this with the current implementation either especially with us making Curator available in 3.3 (enterprise).
Most of the old logs that Fluentd would read in would then end up being deleted by Curator.

I second journal not reading back into old logs; I'm not sure this is something that customers aren't even really aware happens (until their ES volumes get filled up very quickly after first installing).

richm:
@sosiouxme @ewolinetz re: read_from_head - I agree - we shouldn't read from tail unless explicitly told to do so.  For new deployments, users won't know the difference, and can look up in the docs to see what journal-read-from-head does and if they want to deploy with it.  For existing deployments that upgrade to journal, the logic in fluentd/run.sh will figure out that they want `read_from_head false`.
The next commit will default to `false`, and I think I should get rid of the proposed `auto` option.

richm:
@ewolinetz @sosiouxme @t0ffel Thanks for the reviews.  I think I have addressed all of your concerns, and rebased to use the configmap code.  PTAL.

fluentd/configs.d/filter/syslog_record_transform.conf
```diff
@@ -1,22 +1,38 @@
...
+    ## we attempt to use the date format used for rolled over logs 'YYYYMMDD' if possible to get the correct* year, otherwise we subtract the year by 1
+    ## *we cannot know if the year is always correct e.g. logs from 20121230 may be in the file 20130101 which would resolve to still be in the past -- 20131230
+    time ${ ((Time.at(time) > Time.now) && tag_parts[3] && tag_parts[3][9,4]) ? (temp_time = Time.parse(Time.at(time).to_s.gsub(Time.at(time).year.to_s, tag_parts[3][9,4] )).to_datetime.to_s; Time.parse(temp_time) > Time.now ? Time.parse(temp_time.gsub(Time.parse(temp_time).year.to_s, (Time.parse(temp_time).year - 1).to_s )).to_datetime.to_s : Time.parse(temp_time).to_datetime.to_s ) : Time.at(time).to_datetime.to_s }
 
-      #tag ${tag}_.operations_log
-      version 1.2.0
-    </record>
-    remove_keys host
+    #tag ${tag}_.operations_log
+    version 1.2.0
```
ewolinetz: 1.3.0

hack/testing/test-upgrade.sh
```diff
@@ -159,7 +159,10 @@ function upgrade() {
                         -p ES_CLUSTER_SIZE=1 \
                         -p PUBLIC_MASTER_URL=https://localhost:8443${masterurlhack} \
                         -p MODE=upgrade \
-                        -p IMAGE_VERSION=$version
+                        -p IMAGE_VERSION=$version \
+                        -p USE_JOURNAL=${USE_JOURNAL:-false} \
```
ewolinetz: Nit picking here, but do we need to set these as parameters since we're creating the logging-deployer configmap with these settings?

ewolinetz:
Nit otherwise LGTM.  Would want other "LGTM"s since this is a big PR

richm:
@sosiouxme @ewolinetz @jcantrill @t0ffel any further objections?

richm:
@sosiouxme @ewolinetz @jcantrill @t0ffel any further objections?

richm:
WIP - actual enabling the journal doesn't work . . .

richm:
[test]

richm:
[test]
etcd :-(

richm:
[test]
wth, etcd????

richm:
[test]
whack-a-mole :p

richm:
now that tests are passing - @sosiouxme @ewolinetz @jcantrill @t0ffel any further objections?

richm:
now that tests are passing - @sosiouxme @ewolinetz @jcantrill @t0ffel any further objections?

deployer/deployer.yaml
```diff
@@ -323,4 +329,16 @@ items:
     description: "(Deprecated) Node selector operations Curator (label=value)."
     name: CURATOR_OPS_NODESELECTOR
     value: ""
+  -
+    description: 'Use the systemd journal for the fluentd log source'
+    name: USE_JOURNAL
+    value: "false"
```
jcantrill: Don't we want journal to be our default behaviour?

jcantrill:
@richm aside from my one comment LGTM from what I know on the matter

deployer/deployer.yaml
richm: The default should be - whatever docker is set up to use.  If we are deploying logging (fluentd) on a system which has docker `--log-driver=journald` we should configure fluentd to read from journald.  So I guess I need to add logic to the deployer and logging.sh to do that.

deployer/deployer.yaml
```diff
@@ -323,4 +329,16 @@ items:
...
+  -
+    description: 'The journal source to use when USE_JOURNAL=true'
+    name: JOURNAL_SOURCE
+    value: ""
+  -
+    description: 'Read the journal starting from the very beginning (false|true).'
+    name: JOURNAL_READ_FROM_HEAD
+    value: "false"
```
sosiouxme: The non-deprecated parameters and env vars in the deployer template are really just to prevent breaking ansible and the instructions (while I work to revamp them). We want to push toward using the deployer configmap and secret for configuration. If new parameters are to be added, they should be marked "(deprecated)" however I think it is best just not to add them at all.

deployer/scripts/install.sh
```diff
@@ -44,7 +44,8 @@ function initialize_install_vars() {
   # if legacy variables set, use them to fill unset inputs
   for var in KIBANA_HOSTNAME KIBANA_OPS_HOSTNAME PUBLIC_MASTER_URL ENABLE_OPS_CLUSTER IMAGE_PULL_SECRET \
              ES{_OPS,}_{INSTANCE_RAM,PVC_SIZE,PVC_PREFIX,PVC_DYNAMIC,CLUSTER_SIZE,NODE_QUORUM,RECOVER_AFTER_NODES,RECOVER_EXPECTED_NODES,RECOVER_AFTER_TIME} \
-             {ES,ES_OPS,KIBANA,KIBANA_OPS,CURATOR,CURATOR_OPS,FLUENTD}_NODESELECTOR
+             {ES,ES_OPS,KIBANA,KIBANA_OPS,CURATOR,CURATOR_OPS,FLUENTD}_NODESELECTOR USE_JOURNAL JOURNAL_SOURCE \
+             JOURNAL_READ_FROM_HEAD
```
sosiouxme: per comments above, these would not be needed

deployer/README.md
```diff
@@ -462,6 +484,32 @@ To set the parameters:
       oc create -n logging -f -
     # this creates the daemonset and starts fluentd with the new params
 
+#### Have Fluentd use the systemd journal as the log source
+
+By default, fluentd will read from `/var/log/messages*` and
+`/var/log/containers/*.log` for system logs and container logs, respectively.
+You can use the systemd journal instead as the log source.  There are three new
```
sosiouxme: the word "new" has a way of sticking around long after it's ceased to be new. How about "related"

deployer/README.md
```diff
@@ -385,6 +404,9 @@ processed. Note: this means that aggregated logs for the configured
 projects could fall behind and even be lost if the pod were deleted
 before Fluentd caught up.
 
+*NOTE* Throttling does not work when using the systemd journal as the
```
sosiouxme: Would prefer a little more detail here... e.g. "Throttling relies on the use of Docker json-log files and will not be active when using the systemd journal as the log source."

fluentd/configs.d/filter/cloud_init.conf
sosiouxme: So you're saying cloud-init logs are left out of syslog, but go into the journal? I looked at to config to see why that might be true, guessing it's some kind of boot-time weirdness. In any case, it seems odd in my opinion to explicitly exclude something that goes into the journal unless there's a specific problem with it. Cloud-init runs once at boot and I can't see how including its output would cause any problems. Meanwhile, sysadmins might be mystified why it doesn't show up in Kibana and start to distrust the solution.

So it seems to me we should leave cloud-init entries alone.

fluentd/configs.d/filter/max_priority.conf
```diff
@@ -0,0 +1,4 @@
+<filter journal>
```
sosiouxme: could the file be "max-priority.conf" for consistency?

Would actually be clearer to me if it were named "exclude-journal-debug.conf"

fluentd/configs.d/openshift/filter-retag-journal.conf
```diff
@@ -0,0 +1,74 @@
...
+#    container name in pod: bob
+#    pod name: bob-iq0d4
+#    namespace: default
+#    pod uid: 2d67916a-1eac-11e6-94ba-001c42e13e5d
+
+# in addition, there are a few other differences about /var/log/messages and journal messages
+# * the default rsyslog.conf on el7 will only log messages to /var/log/messages at INFO level
+#   or higher
+# * the default rsyslog.d includes 21-cloudinit.conf which redirects all cloud-init messages
+#   to /var/log/cloud-init.log
```
sosiouxme: Ah, I see. That doesn't come with rsyslog.d though, that belongs to cloud-init itself. Looking at the file it seems like they just didn't want messages duplicated by default.

richm: Ok.  I can remove the filtering out of cloud-init messages.

richm:
[test]

richm:
[test]
```
8, 6, 5, 4, 1, 2, 3
How I loathe thee, et-see-dee
```

richm:
[test]
can't blame this one on etcd . . .

richm:
[test]
or this one . . .

richm:
[test]
etcd, etc.

richm:
[test]
hmm - deployer couldn't start kibana when re-deploying

richm:
@sosiouxme I think I have addressed your concerns - PTAL

richm:
This is ready to go.  Once it gets final approval, I will squash the commits and resubmit prior to merging.

sosiouxme:
@richm LGTM :shipit: 

Only one question, is the idea to have Jenkins run it in both modes - journal and syslog? It seems like we ought to exercise both paths at least once at some point.

richm:
@sosiouxme re: testing both modes - definitely - Jenkins has something called "parameterized builds" or perhaps some way to specify a "test matrix" - I will look into that

richm:
[merge]

richm:
[merge]
This is the sporadic 
```
SUCCESS after 0.192s: /data/src/github.com/openshift/origin-aggregated-logging/hack/testing/logging.sh:501: executing 'oc patch deploymentconfig/logging-es-ops-777byytx    -p '{"spec":{"template":{"spec":{"containers":[{"name":"elasticsearch","securityContext":{"privileged": true}}]}}}}'' expecting success
"logging-es-ops-777byytx" patched

There was no error output from the command.
error: #1 is already in progress (Running).
Optionally, you can cancel this deployment using the --cancel option.
```
Error

richm:
[merge]

richm:
[merge]
```error: #1 is already in progress (Running).``` again

**Comment by richm** [comment]
[merge]
```
++ oc start-build --from-dir /data/src/github.com/openshift/origin-aggregated-logging logging-auth-proxy
Uploading directory "/data/src/github.com/openshift/origin-aggregated-logging" as binary input for the build ...
Error from server: cannot upload file to build logging-auth-proxy-2 with status Failed
```
I'm guessing etcd?

richm:
[merge]
back to ```error: #1 is already in progress (Running).```

richm:
[merge]

richm:
[merge]

richm:
[merge]

PR #146, opened 2016-05-23
deployer: ensure desired tag is imported

sosiouxme:
technical fix for https://github.com/openshift/origin/issues/8948
In order to ensure that the triggers on the DCs fire, seed the tags on
each ImageStream with the desired version. Otherwise, if there are
more than 5 tags available, the IS may not have the desired tag
imported automatically (or may have it imported by another name).

### DIFF
--- a/deployer/scripts/install.sh
+++ b/deployer/scripts/install.sh
@@ -82,6 +82,7 @@ function generate_support_objects() {
      --param KIBANA_HOSTNAME=${hostname} \
      --param KIBANA_OPS_HOSTNAME=${ops_hostname} \
      --param IMAGE_PREFIX_DEFAULT=${image_prefix} \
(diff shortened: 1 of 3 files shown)


sosiouxme:
[test]

sosiouxme:
@ewolinetz PTAL

ewolinetz:
@sosiouxme does this mean we shouldn't be doing import-image as part of the upgrade path?  it doesn't seem like this fixes that unless we're recreating the IS there too...

sosiouxme:
@ewolinetz This only operates on the initial deployment (or I suppose if the ISes get deleted somehow, when they're generated on an upgrade). As we discussed, the upgrade needs to do a specific import to ensure that new tags are imported.

ewolinetz:
LGTM

sosiouxme:
[merge]

PR #154, opened 2016-05-26
Adding tests to verify that patching templates and dc work

ewolinetz:


ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
re[test]

ewolinetz:
re[test]

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
[test]

hack/testing/logging.sh
```diff
@@ -441,9 +470,30 @@ function removeAdminCert() {
   os::cmd::expect_success "oc patch secret logging-elasticsearch -p '{\"data\":{\"admin-cert\": null}}'"
 }
 
+function rebuildVersion() {
```
ewolinetz: for some reason the source ref is changing when i'm rebuilding the deployer for the patching test.  that is why the test is failing -- it is trying to run the deployer out of master.

ewolinetz:
[test]

ewolinetz:
[test]

hack/testing/logging.sh
richm: Right.  Look at https://github.com/openshift/origin-aggregated-logging/blob/master/hack/testing/logging.sh#L124 - we have to build from local source because that's what the testing CI does - it checks out the master source locally on the jenkins machine, merges it with the code from the PR, and copies it to the test machine.  So there isn't a github or source repo with the actual code being tested - it only really exists on the local machine - so we have to do `oc start-build --from-dir=/path/to/origin-aggregated-logging`

There should probably be a shell function to do builds which should check for `$USE_LOCAL_SOURCE` - if true, do `oc start-build --from-dir=/path/to/origin-aggregated-logging` - and then have no direct calls to start-build in logging.sh.

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
@sosiouxme @richm PTAL

ewolinetz:
[test]

deployer/scripts/upgrade.sh
```diff
@@ -306,7 +312,7 @@ function isValidChange() {
     # the IS is between a "/" and ":" in value
     if [[ "$path" =~ "containers" ]]; then
       #we want to check if the tag or the sha256 match
-      image=$(echo $value | sed 's/^.*\///')
+      image=$(echo $value | sed 's,'"$IMAGE_PREFIX"',,')
```
sosiouxme: looks like this could be simplified to

    sed "s,$IMAGE_PREFIX,,"

hack/testing/logging.sh
```diff
@@ -76,6 +76,58 @@ function cleanup()
...
+
+  interval=30
+  waittime=120
+
+  local bc=$1
+  local lastVersion=$(oc get bc $bc -o jsonpath='{.status.lastVersion}')
+  local status
+
+  for (( i = 1; i <= $waittime; i++ )); do
+    status=$(oc get build -o jsonpath='{.items[?(@.metadata.name=="'$bc'-'$lastVersion'")].status.phase}')
```
sosiouxme: I think this would be clearer/simpler as:

    status=$(oc get build/$bc-$lastVersion -o jsonpath='{.status.phase}')

hack/testing/logging.sh
sosiouxme: is there actually a use case for building from non-local source?

hack/testing/logging.sh
ewolinetz: yeah, no idea why i did it this way

ewolinetz:
[test]

hack/testing/logging.sh
ewolinetz: only if we wanted to build what was in master as a comparison/regression maybe.

sosiouxme:
lgtm :shipit:

ewolinetz:
[merge]

PR #157, opened 2016-05-29
Removing DC triggers and updating upgrade to account for this.  Bumpi…

ewolinetz:
…ng LOGGING_VERSION to 3

@sosiouxme

### DIFF
--- a/deployer/scripts/install.sh
+++ b/deployer/scripts/install.sh
@@ -367,12 +367,15 @@ function generate_es() {
 #
 function generate_objects() {
   echo "(Re-)Creating deployed objects"
-  oc new-app logging-imagestream-template || : # these may fail if created independently; that's ok
 
   generate_es
   generate_kibana
   generate_curator
   generate_fluentd
+
+  for dc in $(oc get dc -l logging-infra -o name); do
+    oc deploy $dc --latest
+  done
 } #generate_objects()
 
 ######################################
--- a/deployer/scripts/upgrade.sh
+++ b/deployer/scripts/upgrade.sh
@@ -22,6 +22,12 @@ function getDeploymentVersion() {
     return
   fi
 
+  # Check for DC triggers
+  if [[ -n "$(oc get dc -l logging-infra -o jsonpath='{.items[?(@.spec.triggers[*].type)].metadata.name}')" ]]; then
+    echo 2
+    return
+  fi
+
   echo "$LOGGING_VERSION"
 }
 
@@ -363,7 +369,7 @@ function patchIfValid() {
 
   if oc patch $object --type=json -p="[$(join , "${actualPatch[@]}")]"; then
     if [[ $isDC = true ]]; then
-      #oc deploy $object --latest
+      oc deploy $object --latest
       waitForChange $currentVersion $object &
       patchPIDs+=( $!)
     fi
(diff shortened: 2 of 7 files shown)


richm:
lgtm

richm:
[test]

ewolinetz:
re[test]

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
This should go in after #154

deployer/scripts/upgrade.sh
```diff
@@ -513,6 +514,14 @@ function regenerate_secrets_and_support_objects() {
   generate_support_objects
 }
 
+function remove_triggers() {
+
+  for dc in $(oc get dc -l logging-infra -o jsonpath='{.items[?(@.spec.triggers[*].type)].metadata.name}'); do
```
sosiouxme: having a hard time seeing what this jsonpath is for. can this be simplified to:

```
for dc in $(oc get dc -l logging-infra -o name; do
    oc patch $dc -p='{"spec": { "triggers" : [] } }'
    oc deploy $dc --latest
done
```

deployer/scripts/upgrade.sh
```diff
@@ -513,6 +514,14 @@ function regenerate_secrets_and_support_objects() {
...
+    oc patch dc/$dc -p='{"spec": { "triggers" : [] } }'
+    oc deploy dc/$dc --latest
```
sosiouxme: come to think of it, is there any reason to redeploy? changing the triggers has no implications for the existing pods.

deployer/scripts/install.sh
```diff
@@ -373,6 +373,10 @@ function generate_objects() {
   generate_kibana
   generate_curator
   generate_fluentd
+
```
sosiouxme: can we get rid of `oc new-app logging-imagestream-template` above as well? the imagestreams are no longer necessary.

deployer/scripts/upgrade.sh
ewolinetz: doing the `?(@.spec.triggers[*].type)` check in the jsonpath makes it so it only returns dc that still have a trigger.  it's so that we aren't unnecessarily patching DC.

deployer/scripts/upgrade.sh
ewolinetz: this is true, i think i was doing this because i wasn't sure if just removing it would unregister the dc triggers; should be fine though, will remove

deployer/scripts/upgrade.sh
ewolinetz: I can add some more comments around what this is doing exactly

sosiouxme: ok i see, but i'm not sure that we care about unnecessary patching when it's idempotent and doesn't really have any immediate effect.

deployer/scripts/install.sh
ewolinetz: should that also be removed from https://github.com/ewolinetz/origin-aggregated-logging/blob/644af15b3d1722be7df576136917963616194373/deployer/scripts/upgrade.sh#L431 ?

hack/testing/logging.sh
```diff
@@ -261,8 +261,11 @@ if [ "$ENABLE_OPS_CLUSTER" = "true" ] ; then
          system:serviceaccount:logging:aggregated-logging-elasticsearch"
     # update the ES_OPS DC to be in the privileged context
     # TODO: should not have to do this - should work the same as regular hostmount
-    os::cmd::expect_success "oc patch $(oc get dc -o name -l component=es-ops) \
+    ops_dc=$(oc get dc -o name -l component=es-ops)
+    os::cmd::expect_success "oc patch $ops_dc \
        -p '{\"spec\":{\"template\":{\"spec\":{\"containers\":[{\"name\":\"elasticsearch\",\"securityContext\":{\"privileged\": true}}]}}}}'"
+
+    os::cmd::try_until_text "oc deploy $ops_dc --latest" "Started" "$(( 3 * TIME_MIN ))"
```
sosiouxme: `oc deploy` only supports one deployment config name as argument. so:

```
for dc in $ops_dc; do
  os::cmd::try_until_text "oc deploy $dc --latest" "Started" "$(( 3 * TIME_MIN ))"
done

File: deployer/scripts/install.sh (hunk 3 above)
Comment (sosiouxme) [inline deployer/scripts/install.sh]: yes, and the `oc import-image` lines too. this complicates the concept of
an upgrade but i think in the end, if everything is scaled down and then we
get everything up to the latest before scaling up... it won't matter that
we're skipping updating imagestreams that don't end up being relevant. we
should probably even delete the imagestreams (if they have the template
label) in the upgrade to reduce confusion.

File: hack/testing/logging.sh (hunk 4 above)
Comment (sosiouxme) [inline hack/testing/logging.sh]: I guess since this is logging.sh there's only gonna be one, so it doesn't matter.

File: hack/testing/logging.sh [hunk 5]
```diff
@@ -261,8 +261,11 @@ if [ "$ENABLE_OPS_CLUSTER" = "true" ] ; then
          system:serviceaccount:logging:aggregated-logging-elasticsearch"
     # update the ES_OPS DC to be in the privileged context
     # TODO: should not have to do this - should work the same as regular hostmount
-    os::cmd::expect_success "oc patch $(oc get dc -o name -l component=es-ops) \
+    ops_dc=$(oc get dc -o name -l component=es-ops)
+    os::cmd::expect_success "oc patch $ops_dc \
        -p '{\"spec\":{\"template\":{\"spec\":{\"containers\":[{\"name\":\"elasticsearch\",\"securityContext\":{\"privileged\": true}}]}}}}'"
+
+    os::cmd::try_until_text "oc deploy $dc --latest" "Started" "$(( 3 * TIME_MIN ))"
```
sosiouxme: ITYM `$ops_dc` ?

ewolinetz: woof... yeah :)

sosiouxme:
LGTM :shipit:

ewolinetz:
forgot to re-enable deploying after we patch [test]

ewolinetz:
[merge]

PR #158, opened 2016-06-01
deployer improvements

sosiouxme:
A few changes intended to make this easier to run.

Where possible, the roles required for the service accounts are added in
the account template.

All parameters to the deployer template are now optional, as is the
logging-deployer secret. Parameters and secret contents can now be
accepted from three sources (which are checked in this order):
1. A configmap named logging-deployer
2. A secret named logging-deployer
3. Environment variables
Most deployer parameters are deprecated, but they will still work.
However the intent is that most parameters will come from the configmap.
Due to fallback behavior, using the secret as before also works.

Tests were updated to reflect these changes.

### DIFF
--- a/deployer/README.md
+++ b/deployer/README.md
@@ -42,11 +42,8 @@ for details.
 The deployer pod can enable deploying the full stack of the aggregated
 logging solution with just a few prerequisites:
 
-1. A "logging-deployer" Secret
-2. A set of ServiceAccounts with the right privileges
-3. Optionally, a key and certificate for the Kibana server to be deployed.
-4. Sufficient volumes defined for ElasticSearch cluster storage.
-5. A router deployment for serving cluster-defined routes for Kibana.
+1. Sufficient volumes defined for ElasticSearch cluster storage.
+2. A router deployment for serving cluster-defined routes for Kibana.
 
 The deployer generates all the necessary certs/keys/etc for cluster
 communication and defines secrets and templates for all of the necessary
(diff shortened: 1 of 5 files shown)


sosiouxme:
[test]

sosiouxme:
@ewolinetz FYI (don't know if tests will pass yet, probably not on first try)

deployer/deployer.yaml
```diff
@@ -173,130 +191,127 @@ items:
...
-          secretName: logging-deployer
   parameters:
   -
-    description: "If true, set up to use a second ES cluster for ops logs."
+    description: "The mode that the deployer runs in."
+    name: MODE
+    value: "install"
+  -
+    description: 'Specify prefix for logging components; e.g. for "openshift/origin-logging-deployer:v1.1", set prefix "openshift/origin-"'
+    name: IMAGE_PREFIX
```
ewolinetz: Why aren't we also deprecating the prefix and version?

sosiouxme: Because they are needed by the template to even determine what version of
the deployer to run. The template can't read a configmap... well, not yet.
Also, version is one of the things that's likely to vary per invocation.

sosiouxme:
Obviously this will need some README changes... although things will work OK with present directions.

ewolinetz:
LGTM

deployer/deployer.yaml
```diff
@@ -67,13 +66,35 @@ items:
...
       - update
+  -
+    apiVersion: v1
+    kind: RoleBinding
+    metadata:
+      name: logging-deployer-edit-role
+    roleRef:
+      kind: ClusterRole
+      name: edit
+    subjects:
```
ewolinetz: so this lets us avoid needing to manually grant the deployer serviceaccount these roles?

sosiouxme: yes! unfortunately we can't do the same thing yet for ClusterRoleBindings
or SCC membership. But we can at least save the step for roles local to the
project.

sosiouxme:
@ewolinetz any further thoughts/concerns? otherwise I think it's ready.

deployer/README.md
```diff
@@ -101,91 +167,51 @@ The following files may be supplied in the deployer secret:
...
+### Choose Template Parameters
 
-Also give the account access to read labels from all pods (again with the correct project):
+When running the deployer in the next step, there are a few parameters
+that are specified directly if needed:
 
-    $ oadm policy add-cluster-role-to-user cluster-reader system:serviceaccount:logging:aggregated-logging-fluentd
+* `IMAGE_PREFIX`: Specify the prefix for logging component images; e.g. for "docker.io/openshift/origin-logging-deployer:v1.2.0", set prefix "docker.io/openshift/origin-"
+* `IMAGE_VERSION`: Specify version for logging component images; e.g. for "docker.io/openshift/origin-logging-deployer:v1.2.0", set version "v1.2.0"
+* `MODE`: Mode to run the deployer in; one of `install`, `uninstall`, `reinstall`, `upgrade`, `migrate`, `start`, `stop`. `migrate` refers to the ES UUID data migration that is required for upgrading from version 1.1. `stop` and `start` can be used to safely pause the cluster for maintenance.
```
ewolinetz: +1

deployer/README.md
```diff
@@ -101,91 +167,51 @@ The following files may be supplied in the deployer secret:
...
-
 You run the deployer by instantiating a template. Here is an example with some parameters:
 
     $ oc new-app logging-deployer-template \
-               -p KIBANA_HOSTNAME=kibana.example.com \
-               -p ES_CLUSTER_SIZE=1 \
-               -p ES_INSTANCE_RAM=1G \
-               -p PUBLIC_MASTER_URL=https://localhost:8443
+               -p IMAGE_VERSION=v1.2.0 \
+               -p MODE=install
```
ewolinetz: Should we mention that this is the default mode or do we want to encourage users to be explicit?

sosiouxme: I think that's worth mentioning.

ewolinetz:
One nit, otherwise LGTM

sosiouxme:
[merge]

PR #162, opened 2016-06-07
Adding a+x perms to test-upgrade.sh

ewolinetz:


ewolinetz:
[test]

ewolinetz:
[test]

hack/testing/test-upgrade.sh
```diff
@@ -185,7 +185,11 @@ function verifyUpgrade() {
   os::cmd::try_until_text "oc get pods -l component=curator" "Running" "$(( 3 * TIME_MIN ))"
 }
 
-source "./logging.sh"
+#source "./logging.sh"
+source "${OS_ROOT}/hack/cmd_util.sh"
```
ewolinetz: this may not even be necessary anymore...

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
failed for reason other than ruby image

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
ran successfully without using the ruby s2i image just now. will revert so that tests should pass again once that issue is resolved.

ewolinetz:
@richm @sosiouxme PTAL

ewolinetz:
[test]

ewolinetz:
[test]

hack/testing/test-upgrade.sh
```diff
@@ -196,7 +222,7 @@ addTriggers && \
 rebuildVersion "upgraded" && \
 
 upgrade "upgraded" && \
-verifyUpgrade true && \
+verifyUpgrade "upgraded" true && \
 
 ./e2e-test.sh $USE_CLUSTER && \
```
sosiouxme: 1. don't need backslashes after `&&`
2. could get rid of all of these `&&` and the `|| return 1` mess above - the `&&` is what is making `-e` ineffective and we don't even need to do it. the point of `-e` is exactly what we're doing, where we want to quit if anything fails.

ewolinetz: My thought was we want each step to pass, otherwise we would kick out to:
```
oc get events -o yaml > $ARTIFACT_DIR/all-events.yaml 2>&1
exit 1
```
I can probably push the tests into their own functions to mirror what I'm trying to do, but I thought `-e` will exit the script if a step fails not just the function it's in.

sosiouxme: I missed that there was something useful happening after failure. Carry on :)

ewolinetz: I'll play around with using a trap to clean up the script a little

ewolinetz:
looks like the trap works... but now to make the tests not fail again :)

ewolinetz:
[test]

ewolinetz:
image pull backoff [test]

hack/testing/logging.sh
```diff
@@ -421,7 +421,7 @@ if [ "$TEST_PERF" = "true" ] ; then
     echo "Running performance tests"
     for test in perf-*.sh ; do
         if [ -x ./$test ] ; then
-            ./$test $USE_CLUSTER
+            (. ./$test $USE_CLUSTER)
```
richm: If you are sourcing the file it doesn't have to be executable, but I guess the `-x` test is useful to know which files are meant to be run.

richm:
LGTM

ewolinetz:
need to figure out what was going on with the migrate script.  the deployer wasn't able to see all projects using the `oc config set-context` trick, I faked it by giving the deployer cluster-reader which we shouldn't keep if we can help it

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
Since the migration isn't currently working using `oc config set-context` we are trying to use `--as` but it isn't yet available in origin:v1.2.0 and the origin:latest was built with Docker 1.10 which fails to build currently on our test machine.

May be blocked until this is merged in https://trello.com/c/CJiUnVUm/136-3-docker-1-10-push-force-schema1-manifest-via-daemon-flag unless the image is rebuilt.

ewolinetz:
To be clear, this works if we grant the deployer `cluster-reader` privs but we would like to avoid doing that as a temporary fix.

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
not sure if it was trying to build with origin:latest or not... re[test]

ewolinetz:
[test]

ewolinetz:
error during upgrade for config maps logic

ewolinetz:
I think this happens because Curator is created earlier in the migration step and already is set up with a config map

ewolinetz:
[test]

ewolinetz:
[test] for real

ewolinetz:
[test]

hack/testing/check-EFK-running.sh
```diff
@@ -18,6 +18,58 @@ function waitFor() {
   return 1
 }
 
+function checkESStarted() {
+
+  local pod=$1
+  local cluster_service
+
+  if ! cluster_service=$(waitForValue "oc logs $pod | grep '\[cluster\.service[[:space:]]*\]'"); then
```
richm: This is causing the test failures::

    ./check-EFK-running.sh: line 26: waitForValue: command not found

Where is `waitForValue()` defined?

ewolinetz: I forgot to copy it over to that script :-/

ewolinetz:
... trying to do 1.3.0 things with a 1.2.0 client ...

ewolinetz:
etcd error... re[test]

ewolinetz:
[test]

ewolinetz:
etcd re[test]

ewolinetz:
[test]

ewolinetz:
so close... failed after reinstall..

ewolinetz:
looks like it is a permissions issue on recovering a node when starting ES back up... lets re[test] ?

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
in the mean time @sosiouxme @richm PTAL

deployer/scripts/upgrade.sh
```diff
@@ -489,22 +489,37 @@ function add_fluentd_daemonset() {
...
   for dc in $(get_curator_dcs); do
-    oc patch $dc --type=json --patch "[$patch]"
+    # we want to ignore a failure here for the scenario where we are installing Curator as part of this upgrade
+    # since the curator will already contain the changes oc patch provides and will fail due to duplicate volume and mountPath
+    currentVersion=$(oc get $dc -o jsonpath='{.status.latestVersion}')
+    if oc patch $dc --type=json --patch "[$patch]"; then
+      oc deploy $dc --latest
+      waitForChange $currentVersion $dc &
+      patchPIDs+=( $!)
+    fi
```
richm: else error?  what happens if `oc patch` fails?

deployer/scripts/upgrade.sh
```diff
@@ -489,22 +489,37 @@ function add_fluentd_daemonset() {
...
   done
+
+  wait ${patchPIDs[@]}
```
richm: If any of the `oc deploy` fail, how do we find out why?  Where are the logs?

deployer/scripts/upgrade.sh
ewolinetz: not in this case, if it fails it is most likely due to the fact that it already contains the volume and mounts that the patch is providing

richm: so in that case it is ok to skip the deploy - ok

deployer/scripts/upgrade.sh
ewolinetz: i believe `oc deploy` will fail in the case where there is currently a deployment scheduled for `--latest`.  In that case the output _should_ be captured in the deployer log file.  Maybe we should have `oc deploy $dc --latest && waitForChange ...` ?

richm: I don't know.  I just want to know where to look if `oc deploy` blows up.  If that's the deployer pod log, that's fine.

ewolinetz:
[test]

ewolinetz:
One thing I noticed with this is that when doing a `MODE=reinstall` when reusing the PV, ES was lacking the permissions to open up its node information.  I'm not sure if this is a valid use-case since if you are doing a reinstall of aggregated logging one would _assume_ that they do not want to maintain their previous logs.

sosiouxme:
Actually I wouldn't assume that. So I'm guessing that on a reinstall, the
pod user gets randomly assigned a new UID or something like that, and it
doesn't happen if you just nuke the pod and get a new one? I think that's
important to look into, the PVCs should be completely reusable. That
doesn't need to block this PR though.

ewolinetz:
[merge]

ewolinetz:
[merge]

PR #156, opened 2016-05-28
generic with configmaps

sosiouxme:
use the generic ES and curator from common and apply configuration as a configmap.

PR #172, opened 2016-06-16
os::log::install_errexit bash function not found

richm:
origin hack/cmd_util.sh changed - it does not include util.sh and
text.sh any more.  Change the test to see if os::log::install_errexit
and if not, source util.sh and text.sh

### DIFF
--- a/hack/testing/logging.sh
+++ b/hack/testing/logging.sh
@@ -42,6 +42,12 @@ source "${OS_ROOT}/hack/cmd_util.sh"
 source "${OS_ROOT}/hack/lib/test/junit.sh"
 source "${OS_ROOT}/hack/common.sh"
 source "${OS_ROOT}/hack/lib/log.sh"
+# in case cmd_util.sh does not include util.sh and text.sh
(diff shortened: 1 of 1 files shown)


richm:
[test]

richm:
@ewolinetz @sosiouxme @jcantrill PTAL - we need this due to a recent change in origin/hack/cmd_util.sh

ewolinetz:
lgtm

richm:
[merge]

sosiouxme:
I hate that we're even following the origin libs at all.

FWIW I fixed it for my testing a little more generally like this: https://github.com/openshift/origin-aggregated-logging/pull/156/commits/1668c938256798e946a5779f835fec2aea452bf2

May take care of some future case where something is moved into a lib.

richm:
@sosiouxme yeah, not so good to be dependent on origin for this when arbitrary changes like this wipe us out.  My change is good enough to get us unstuck in the very short term.  Your change is probably better long term.

PR #185, opened 2016-06-24
build/release: transition "deployment" to "deployer"

sosiouxme:
Reduce confusion as much as possible by calling the image "logging-deployer". Keeping "logging-deployment" around for one release cycle to prevent a period where the deployer template refers to the wrong one.


sosiouxme:
[merge]

PR #187, opened 2016-06-29
curator date/times are with respect to the curator pod timezone

richm:
@sosiouxme @ewolinetz bump submodule commit for testing curator


richm:
[test]

PR #191, opened 2016-07-01
Fix recent build test issues

richm:
fix several recent issues with building images and running tests

### DIFF
--- a/hack/testing/logging.sh
+++ b/hack/testing/logging.sh
@@ -36,6 +36,31 @@ USE_LOCAL_SOURCE=${USE_LOCAL_SOURCE:-false}
 TEST_PERF=${TEST_PERF:-false}
 ES_VOLUME=${ES_VOLUME:-/var/lib/es}
 ES_OPS_VOLUME=${ES_OPS_VOLUME:-/var/lib/es-ops}
+# default CI image uses journal by default - allow to control if logging uses it or not
+USE_JOURNAL=${USE_JOURNAL:-false}
+
+if [ $USE_JOURNAL = false ] ; then
+    # see if docker is using the journal log driver - if so, change it to json-file
+    if grep -q -- '--log-driver=journald' /etc/sysconfig/docker ; then
+        sudo sed -i.bak 's/--log-driver=journald/--log-driver=json-file/' /etc/sysconfig/docker
+        sudo systemctl restart docker
+    fi
+else
+    # see if docker is explicitly configured to use the json-file log driver
+    if grep -q -- '--log-driver=json-file' /etc/sysconfig/docker ; then
+        sudo sed -i.bak 's/--log-driver=json-file/--log-driver=journald/' /etc/sysconfig/docker
+        sudo systemctl restart docker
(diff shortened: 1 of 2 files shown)


richm:
[test]

richm:
@ewolinetz @sosiouxme @jcantrill PTAL

hack/testing/logging.sh
```diff
@@ -36,6 +36,31 @@ USE_LOCAL_SOURCE=${USE_LOCAL_SOURCE:-false}
 TEST_PERF=${TEST_PERF:-false}
 ES_VOLUME=${ES_VOLUME:-/var/lib/es}
 ES_OPS_VOLUME=${ES_OPS_VOLUME:-/var/lib/es-ops}
+# default CI image uses journal by default - allow to control if logging uses it or not
+USE_JOURNAL=${USE_JOURNAL:-false}
+
+if [ $USE_JOURNAL = false ] ; then
+    # see if docker is using the journal log driver - if so, change it to json-file
+    if grep -- '--log-driver=journald' /etc/sysconfig/docker > /dev/null 2>&1 ; then
```
sosiouxme: I suggest using `grep -q` instead of blackholing output - then you shouldn't have any output unless something is really wrong.

hack/testing/logging.sh
```diff
@@ -36,6 +36,31 @@ USE_LOCAL_SOURCE=${USE_LOCAL_SOURCE:-false}
...
+        sudo sed -i.bak 's/--log-driver=json-file/--log-driver=journald/' /etc/sysconfig/docker
+        sudo systemctl restart docker
+    elif grep '^OPTIONS=' /etc/sysconfig/docker > /dev/null 2>&1 ; then
+        # using default log driver - make it explicit to use journald
+        sudo sed -i.bak "/^OPTIONS=/ s/'$/ --log-driver=journald'/" /etc/sysconfig/docker
+        sudo systemctl restart docker
+    else
+        # using default log driver - make it explicit to use journald
+        sudo cp /etc/sysconfig/docker /etc/sysconfig/docker.bak
+        echo "OPTIONS=--log-driver=journald" | sudo tee -a /etc/sysconfig/docker
```
sosiouxme: Just curious, why `| sudo tee -a` instead of `>>`?

sosiouxme:
Nits, otherwise, brilliant!

ewolinetz:
LGTM

richm:
@sosiouxme `$ sudo echo "foo" >> /root/bar` doesn't work - something about the `>>` is evaulated before the sudo or outside of the sudo, so you have to use `| sudo tee -a` - see http://stackoverflow.com/questions/82256/how-do-i-use-sudo-to-redirect-output-to-a-location-i-dont-have-permission-to-wr

sosiouxme:
oh right, the `>>` operates in the context of the shell outside of sudo.
makes sense then.

On Fri, Jul 1, 2016 at 10:06 AM, Richard Megginson <notifications@github.com

richm:
[merge]

PR #193, opened 2016-07-05
extract components from common

sosiouxme:
Since testing elasticsearch and curator PRs is awkward and they probably
won't really be used by another project, this pulls them back in
directly.

There's no real reason to keep kibana-proxy as a sub-sub-module either,
so now it is just a top-level submodule.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "deployer/common"]
 	path = deployer/common
 	url = https://github.com/openshift/origin-integration-common
+[submodule "kibana-proxy"]
+	path = kibana-proxy
+	url = https://github.com/fabric8io/openshift-auth-proxy.git
--- a/curator/Dockerfile
+++ b/curator/Dockerfile
@@ -1,0 +1,29 @@
+FROM centos:centos7
+
+MAINTAINER OpenShift Development <dev@lists.openshift.redhat.com>
+
+# by default, run the curator cron job at midnight every night, and delete
+# indices older than 30 days old
+ENV HOME=/opt/app-root/src \
+    ES_HOST=localhost \
+    ES_PORT=9200 \
+    CURATOR_CRON_MINUTE=0 \
+    CURATOR_CRON_HOUR=0 \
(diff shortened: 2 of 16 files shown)


richm:
We're still keeping origin-integration-common, so we still need some way to test PRs in that module.

What about the git history for curator and es?  Will we lose that?

richm:
Do we use the `bash/` subdirectory?  If not, then LGTM

richm:
[test]

richm:
hmm - not quite sure how to make the tests pass
```
+ git submodule update --recursive
fatal: reference is not a tree: cfa3fa43fe9ca55298631a54d26bcc5bc0b037e7
Unable to checkout 'cfa3fa43fe9ca55298631a54d26bcc5bc0b037e7' in submodule path 'deployer/common'
~/jobs/test-origin-aggregated-logging/workspace/origin-aggregated-logging ~/jobs/test-origin-aggregated-logging/workspace
```

Where is commit cfa3fa43fe9ca55298631a54d26bcc5bc0b037e7?  Did you push that change to https://github.com/sosiouxme/origin-integration-common?

sosiouxme:
I haven't pushed that commit to common yet. Since you're OK with it I'll push it and do the merge tests.

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "deployer/common"]
 	path = deployer/common
 	url = https://github.com/openshift/origin-integration-common
+[submodule "kibana-proxy"]
```
sosiouxme: For the purposes of the release scripts we can call it anything. That being the case I thought it might be nice to have it named according to the function it plays in this system. I don't feel strongly about it, though, if you think `openshift-auth-proxy` is better to minimize confusion I'll do that.

jcantrill: I dont feel strongly about it either.  As long as it works, fine with me.  Less work.

sosiouxme:
Note, we still refer to `logging-deployment` for the deployer image because we haven't tagged/built/pushed one called `logging-deployer` yet. Once we do that we can change the remaining "deployment" references.

[merge]

PR #195, opened 2016-07-07
fluentd configmap

sosiouxme:
Supply a throttle config in a configmap on the fluentd daemonset so that users can easily edit or replace it.

Also move fluent.conf into the same configmap to allow users to edit fluentd config easily.


sosiouxme:
[test]

sosiouxme:
test https://ci.openshift.redhat.com/jenkins/job/test-origin-aggregated-logging/605/ seems to have lost its brain seeing a merge conflict. not sure what happened there, the PR is based on master.

Try [test] again.

sosiouxme:
@ewolinetz @richm PTAL

sosiouxme:
[test] and don't give me no lip about it

ewolinetz:
so the manner of which we'll allow others to update the fluentd config is with the `pre-*` and `post-*` hooks?

sosiouxme:
No, they can update the fluent.conf, so effectively anything. The hooks are
there to enable us to add stuff in `configs.d/openshift` later on without
having to update fluent.conf. It's not perfectly future-proofed but I'm
hoping this will make most upgrades possible.

fluentd/Dockerfile
```diff
@@ -16,23 +15,28 @@ LABEL io.k8s.description="Fluentd container for collecting of docker container l
...
-  mkdir -p ${THROTTLE_CONF_LOCATION} && \
-  touch ${THROTTLE_CONF_LOCATION}/settings
+    yum install -y --setopt=tsflags=nodocs \
+      gcc-c++ \
+      ruby \
+      ruby-devel \
+      libcurl-devel \
+      make && \
+    yum clean all
+RUN mkdir -p ${HOME} && \
```
ewolinetz: should probably remove this RUN

deployer/conf/fluent.conf
```diff
@@ -0,0 +1,33 @@
...
+  @include configs.d/openshift/filter-k8s-record-transform.conf
+  @include configs.d/openshift/filter-syslog-record-transform.conf
+  @include configs.d/openshift/filter-post-*.conf
+##
+
+## matches
+  @include configs.d/openshift/output-pre-*.conf
+  @include configs.d/openshift/output-operations.conf
+  @include configs.d/openshift/output-applications.conf
+  @include configs.d/openshift/output-post-*.conf
```
ewolinetz: important to note that since `output-applications.conf` does a match on ** we won't ever get to this `output-post-*`

fluentd/configs.d/user/fluent.conf
```diff
@@ -0,0 +1,35 @@
...
+  @include configs.d/openshift/filter-k8s-record-transform.conf
+  @include configs.d/openshift/filter-syslog-record-transform.conf
+  @include configs.d/openshift/filter-post-*.conf
+##
+
+## matches
+  @include configs.d/openshift/output-pre-*.conf
+  @include configs.d/openshift/output-operations.conf
+  @include configs.d/openshift/output-applications.conf
+  @include configs.d/openshift/output-post-*.conf
```
ewolinetz: same as other fluent conf

fluentd/Dockerfile
sosiouxme: Any particular reason? I wanted to put the yum stuff in a separate layer so
it would cache even if the stuff after it changed.

deployer/conf/fluent.conf
sosiouxme: Hadn't thought of that. Seems like we might as well just remove it then?

fluentd/Dockerfile
ewolinetz: I was thinking to reduce the number of layers, but I've been swayed :)

deployer/conf/fluent.conf
ewolinetz: sounds good to me

fluentd/configs.d/openshift/output-applications.conf
```diff
@@ -0,0 +1,5 @@
+<match **>
+   @type copy
+   @include output-es-config.conf
+   @include ../dynamic/es-copy-config.conf
```
richm: I suggest to have an `@include output-extra-*.conf` here - then the user could add a config map for `output-extra-splunk.conf` or `output-extra-fluentd-secure-forward.conf` or whatever

fluentd/configs.d/openshift/output-operations.conf
```diff
@@ -1,5 +1,5 @@
 <match system.var.log** **_default_** **_openshift_** **_openshift-infra_**>
   @type copy
-  @include fluentd_es_ops_config.conf
-  @include fluentd_es_ops_copy_config.conf
+  @include output-es-ops-config.conf
+  @include ../dynamic/es-ops-copy-config.conf
```
richm: I suggest to have an `@include output-extra-ops-*.conf` here - then the user could add a config map for `output-extra-ops-splunk.conf` or `output-extra-ops-fluentd-secure-forward.conf` or whatever

fluentd/configs.d/openshift/output-applications.conf
ewolinetz: The only catch with a user wanting to configure fluentd to send to anything
extra is that we'd need to make sure our image provides that plugin...

On Tue, Jul 12, 2016 at 4:21 PM, Richard Megginson <notifications@github.com
> wrote:

> In fluentd/configs.d/openshift/output-applications.conf
> <https://github.com/openshift/origin-aggregated-logging/pull/195#discussion_r70524938>
> :
>
> > @@ -0,0 +1,5 @@
> > +<match **>
> > +   @type copy
> > +   @include output-es-config.conf
> > +   @include ../dynamic/es-copy-config.conf
>
> I suggest to have an @include output-extra-*.conf here - then the user
> could add a config map for output-extra-splunk.conf or
> output-extra-fluentd-secure-forward.conf or whatever
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/openshift/origin-aggregated-logging/pull/195/files/c79a4f8ff336fad4187dcb823b73638724df6f84#r70524938>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ABXIEf0T3MbS75dk4DtWD-Cb9MZ_l0oNks5qVAVcgaJpZM4JHWV8>
> .
>



-- 
Eric Wolinetz

Red Hat Engineering | OpenShift
ewolinet@redhat.com
IRC: ewolinetz
C: (847) 641-0143

richm: True, but

- the same thing applies to any user provided configuration, including the support this commit is providing - the user can't use any plugin they want in their custom input-post-foo.conf, they have to use the ones we provide
- there are some things they can do with the plugins built-in to fluentd: out_stream, out_roundrobin, out_file, out_exec, out_forward, in addition to es.

If we give users the ability to provide their own config, we should also give them a way to provide their own plugins . . . http://docs.fluentd.org/articles/plugin-management

ewolinetz: I think I'd argue against letting them install their own plugins, however I feel like we should provide them with the secure forward plugin.  That way they can forward their logs to any other fluentd instance they may want to use (to forward to any other log store)

deployer/scripts/install.sh
```diff
@@ -255,7 +255,13 @@ function generate_configmaps() {
     # generate curator configmap
     oc create configmap logging-curator \
       --from-file=config.yaml=conf/curator.yml
-    oc label configmap/logging-curator logging-infra=support # make easier to delete later
+    oc label configmap/logging-curator logging-infra=support
+
+    # generate fluentd configmap
+    oc create configmap logging-fluentd \
+      --from-file=fluent.conf=conf/fluent.conf \
+      --from-file=throttle-config.yaml=conf/fluentd-throttle-config.yaml || return
```
t0ffel: Why not move hardcoded fluentd config files and related env variables to configmaps as well?
Is there a plan to change dynamically generated throttled configs to configmaps?

fluentd/configs.d/openshift/output-applications.conf
richm: @ewolinetz ok - then I think this commit should include the ability to provide `output-extra-*` and `output-extra-ops-*`, so that customers can use the plugins we already include, and we should implement support for out_secure_forward in a separate commit

deployer/scripts/install.sh
sosiouxme: In this PR, the throttle configs are dynamically generated from a configmap.

There is a delicate balance here in how much we allow users to edit; after
all, we're responsible for providing sane configuration. Users want to edit
everything, but we can't really support them editing *anything* if we want
to hold open the option of updates in the future.

The middle ground I've proposed here is to provide most of the config on
the image (the configs.d/openshift directory which we'll own updates for)
but provide a configmap for editing the fluent.conf that brings it all
together with @include (which we hope not to change substantially for the
future). You can change the provided files by adding your own version into
the configmap and editing fluent.conf to point at it. Doing so will provide
a clear separation; if we update our image and change something about the
file they "edited" then it will be easy to figure out what happened and fix
it.

I'm honestly a bit nervous about even making fluent.conf configurable, but
it has already proved really useful in my testing, and I know customers
will want *some* power to modify config without having to incorporate a
custom image.

fluentd/configs.d/openshift/output-applications.conf
sosiouxme: That seems reasonable.

deployer/scripts/install.sh
richm: @t0ffel - are you suggesting to be able to completely replace fluent.conf and the other hardcoded config files?  or just to be able to provide the configuration we now have to provide with environment variables via a configmap instead?

t0ffel: My primary concern is that right now there are 4 different places where fluentd configuration is defined:
1) hardcoded config files
2) dynamically generated config files
3) config files passed via configmaps
4) env variables

I think this is not an easy setup.
User needs to have knowledge of fluentd, then user also needs to understand 1)-4) to know where to add his modifications(because w/o knowing the full config and where it is coming from it is very easy to screw up).
On update there are 4 different items to track.

To me personally it is much easier to see the entire fluentd configuration accurately(i.e. dynamically generated configs output to configmap, rather than to .conf file) taken from configmaps.
I'm not arguing against good directory structure which will advise how/where users need to add their modifications.
If everything(most stuff) is in configmaps on upgrade you can still check what was edited and act accordingly.

t0ffel: @richm I think (eventually) changing all the fluentd config files and the related env variables to configmaps will be awesome.

sosiouxme: I agree it's not an easy setup to explain or understand.

Having all the files in the configmap was my first design. But when we
started talking about upgrades and users modifying things I realized I
wanted as little in their hands as I could get away with.

It does make some sense to have the dynamically generated stuff all in the
configmap, but then it would have to be generated in the deployer, not in
fluentd, and the throttle conf generator is in ruby to match fluentd... so
that seemed ugly.

sosiouxme:
try that [test] again

sosiouxme:
https://ci.openshift.redhat.com/jenkins/job/test-origin-aggregated-logging/634 might be a flake... failed by deploying `deploymentconfig/logging-es-ops-xg5m1514` when it was in the middle of being deployed already.

sosiouxme:
[test] again for fun

sosiouxme:
We're getting flakes then :(

Anyway this seems ready. Any further comments?

richm:
LGTM

ewolinetz:
:shipit:

sosiouxme:
[merge]

richm:
[merge]

PR #184, opened 2016-06-24
Allow setting curator runhour/runminute timezone - default is UTC

richm:
https://github.com/openshift/origin-aggregated-logging/issues/182
Clarify in the documentation that the curator date/time settings
are with respect to the timezone of the curator pod, and change
the test to use that timezone.

PR #197, opened 2016-07-08
Curator ignores more than one project with same time and time unit

richm:
[test]

### DIFF
--- a/curator/run_cron.py
+++ b/curator/run_cron.py
@@ -9,6 +9,16 @@ import logging
 from crontab import CronTab
 from datetime import datetime, timedelta
 from pytz import timezone, UnknownTimeZoneError
+try:
+    from shlex import quote as shellquote
+except:
+    from pipes import quote as shellquote
+import re
+
+# metadata.name: Invalid value: "F00": must be a DNS label (at most 63
+# characters, matching regex [a-z0-9]([-a-z0-9]*[a-z0-9])?): e.g. "my-name"
+projectre = re.compile(r'^[a-z0-9]([-a-z0-9]*[a-z0-9])?$')
+projectmaxlen = 63
 
 logger = logging.getLogger()
 # log at INFO by default
@@ -41,6 +51,15 @@ filename = os.getenv('CURATOR_CONF_LOCATION', '/etc/curator/settings/config.yaml
 decoded = {}
 with open(filename, 'r') as stream:
     decoded = yaml.load(stream) or {}
+for project in decoded:
+    if project == '.defaults' or project == '.operations':
+        continue
+    if len(project) > projectmaxlen:
+        logger.error('The project name length must be less than or equal to %d characters.  This is too long: [%s]' % (projectmaxlen, project))
+        sys.exit(1)
+    if not projectre.match(project):
+        logger.error('The project name must match this regex: [%s] This does not match: [%s]' % (projectre.pattern, project))
+        sys.exit(1)
 
 tzstr = decoded.get('.defaults', {}).get('timezone', os.getenv('CURATOR_RUN_TIMEZONE', 'UTC'))
 tz = None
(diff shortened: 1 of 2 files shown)


richm:
[test]

richm:
[test]

richm:
@ewolinetz @sosiouxme PTAL

ewolinetz:
LGTM, do you want to wait until after the default UTC TZ changes are merged in?

richm:
@ewolinetz yes, this commit depends on https://github.com/openshift/origin-aggregated-logging/pull/197

curator/run_cron.py
```diff
@@ -81,7 +86,7 @@
                 value = int(decoded[project][operation][unit])
 
                 if unit in allowed_units:
-                    default_command = default_command + " --exclude " + project + '.*'
+                    default_command = default_command + " --exclude " + shellquote(re.escape(project + '.')) + '*'
```
sosiouxme: shouldn't the '*' be inside the shellquote, but outside the re.escape?

sosiouxme:
So... since project names can't properly include anything but alphanum and hyphen anyway, wouldn't it be better to complain and omit any that do? If someone thinks they can use "foo-*" to represent a bunch of projects, this'll just escape that and match nothing, which isn't what they wanted; while if we give them an error in the logs, they have a better chance of figuring out they did something wrong.

richm:
@sosiouxme Is there a canonical regex that must match project names e.g. `^[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9]$`
that is - name must begin and end with alphanum, and may contain hyphens?  If so, I can add a regex to curator to reject names in the config not matching.

sosiouxme:
If you want to be quite correct it needs to be a domain label, lowercase,
no unicode.

```
$ oadm new-project F00
The Project "F00" is invalid.
metadata.name: Invalid value: "F00": must be a DNS label (at most 63
characters, matching regex [a-z0-9]([-a-z0-9]*[a-z0-9])?): e.g. "my-name"
```

However is that the goal here? If we're just trying to filter out project
names that couldn't be correct and could cause parsing problems, surely
`^[-a-z0-9]+$` would do the trick. With `.operations` being an exception
obviously.

On Tue, Jul 12, 2016 at 3:31 PM, Richard Megginson <notifications@github.com

richm:
[test]

richm:
[test]

richm:
@ewolinetz @sosiouxme PTAL

ewolinetz:
LGTM

richm:
depends on https://github.com/openshift/origin-aggregated-logging/pull/184

curator/run_cron.py
```diff
@@ -107,9 +126,15 @@
...
+            tab_command = tab_command + ' --regex '
+            regex = '('
+            sep = ''
             for project in curator_settings[operation][unit][value]:
-                tab_command = tab_command + ' --prefix ' + project + '.'
-
+                regex = regex + sep + '^' + re.escape(project + '.')
+                sep = '|'
+            regex = regex + ')'
+            tab_command = tab_command + shellquote(regex)
```
sosiouxme: this works ok but it feels like it ought to be reduced to a one-liner, something like (not a pythonista):

    tab_command = tab_command + ' --regex ' + shellquote('(' + '|'.join(map(lambda project:'^'+re.escape(project+'.'),curator_settings[operation][unit][value] )) + ')')

OK, maybe two lines, or to taste. I leave it up to you though.

sosiouxme:
nit, otherwise LGTM

richm:
[test]
```error: #1 is already in progress (Running).``` again

richm:
[merge]

richm:
[merge]

PR #203, opened 2016-07-25
Add Common Data Model index template

lukas-vlcek:
We make sure Common Data Model index template is available on the Elasticsearch cluster side.

### DIFF
--- a/elasticsearch/Dockerfile
+++ b/elasticsearch/Dockerfile
@@ -32,6 +32,7 @@ RUN rpm --import https://packages.elastic.co/GPG-KEY-elasticsearch && \
 # initial boring elasticsearch.yml is just to enable plugin installation.
 # put the real config in place in a deployment with a ConfigMap.
 ADD elasticsearch.yml logging.yml /usr/share/elasticsearch/config/
+ADD index_templates/ /usr/share/elasticsearch/index_templates/
 ADD install.sh ${HOME}/
 RUN ${HOME}/install.sh
 ADD run.sh ${HOME}/
--- a/elasticsearch/run.sh
+++ b/elasticsearch/run.sh
@@ -2,6 +2,15 @@
 
 set -euo pipefail
 
+ES_REST_BASEURL=https://localhost:9200
+LOG_FILE=elasticsearch_connect_log.txt
+RETRY_COUNT=30		# how many times
+RETRY_INTERVAL=1	# how often (in sec)
+
+retry=${RETRY_COUNT}
+max_time=$(( RETRY_COUNT * RETRY_INTERVAL ))	# should be integer
+timeouted=false
+
 mkdir -p /elasticsearch/$CLUSTER_NAME
 secret_dir=/etc/elasticsearch/secret/
 [ -f $secret_dir/searchguard-node-key ] && ln -s $secret_dir/searchguard-node-key /elasticsearch/$CLUSTER_NAME/searchguard_node_key.key
@@ -59,5 +68,73 @@ else
 	echo "INSTANCE_RAM env var is invalid: ${INSTANCE_RAM:-}"
 	exit 1
 fi
+
+
+# Wait for Elasticsearch port to be opened. Fail on timeout or if response from Elasticsearch is unexpected.
+wait_for_port_open() {
+	rm -f ${LOG_FILE}
+	echo -n "Checking if Elasticsearch is ready on ${ES_REST_BASEURL} "
+	while ! curl -s -k \
+			--cert ${secret_dir}admin-cert \
+			--key  ${secret_dir}admin-key \
+			--max-time ${max_time} \
+			-o ${LOG_FILE} \
+			${ES_REST_BASEURL} && [ ${timeouted} == false ]
+	do
+		echo -n "."
+		sleep ${RETRY_INTERVAL}
+		(( retry -= 1 ))
+		if (( retry == 0 )) ; then
+			timeouted=true
+		fi
+	done
+
+	# Test for response code 200 in Elasticsearch output. This can be sensitive to Elasticsearch version.
+	if [ -f ${LOG_FILE} ] && grep -q "200" ${LOG_FILE} ; then
+		echo "- connection successful"
+	else
+		if [ ${timeouted} == true ] ; then
+			echo -n "[timeout] "
+		fi
+		echo "failed"
+		cat ${LOG_FILE}
+		exit 1
+	fi
+}
+
+verify_or_add_index_templates() {
+	wait_for_port_open
+	# Try to wait for cluster become more stable before index template being pushed in.
+	# Give up on timeout and continue...
+	curl -v -s -k -X GET \
+		--cert ${secret_dir}admin-cert \
+		--key  ${secret_dir}admin-key \
+		"${ES_REST_BASEURL}/_cluster/health?wait_for_status=yellow&timeout=${max_time}s"
+
+	shopt -s failglob
+	for template_file in /usr/share/elasticsearch/index_templates/*.json
+	do
+		template=`basename $template_file`
+		# Check if index template already exists
+		response_code=$(curl -s -k -X HEAD \
+			--cert ${secret_dir}admin-cert \
+			--key  ${secret_dir}admin-key \
+			-w '%{response_code}' \
+			${ES_REST_BASEURL}/_template/$template)
+		if [ $response_code == "200" ]; then
+			echo "Index template '$template' already present in ES cluster"
+		else
+			echo "Create index template '$template'"
+			curl -v -s -k -X PUT \
+				--cert ${secret_dir}admin-cert \
+				--key  ${secret_dir}admin-key \
+				-d@$template_file \
+				${ES_REST_BASEURL}/_template/$template
+		fi
+	done
+	shopt -u failglob
+}
+
+verify_or_add_index_templates &
 
 exec /usr/share/elasticsearch/bin/elasticsearch
--- a/hack/testing/check-EFK-running.sh
+++ b/hack/testing/check-EFK-running.sh
@@ -82,6 +82,27 @@ function checkKibanaStarted() {
   fi
 }
 
+function checkESContainsIndexTemplates() {
+
+  local pod=$1
+  local template_files
+  local secret_dir=/etc/elasticsearch/secret/
+ 
+  if ! template_files=$(waitForValue "oc exec $pod -- ls -1 /usr/share/elasticsearch/index_templates"); then
+    echo "No index template files found"
+    return 1
+  fi
+
+  echo "Checking presence of index templates: ${template_files}"
+  for template in $template_files; do
+    echo "  - verify ${template}"
+    if ! waitForValue "oc exec $pod -- curl -s -k -X HEAD -x '%{response_code}' --cert ${secret_dir}admin-cert --key ${secret_dir}admin-key https://localhost:9200/_template/$template"; then
+      return 1
+    fi
+  done
+
+}
+
 # add one since fluentd will be deployed via a daemonset
 # keeping as -2 + 1 for readibilty
 ADDITIONAL_PODS=$((KIBANA_CLUSTER_SIZE + ES_CLUSTER_SIZE - 2 + 1))
@@ -290,6 +311,17 @@ done
 for pod in $(oc get pods -l component=kibana-ops -o name); do
   checkKibanaStarted "$pod" || EXIT_CODE=1
 done
+
+echo $TEST_DIVIDER
+echo "Checking if ES contains common data model index templates"
+for pod in $(oc get pods -l component=es -o name | sed 's,pod/,,'); do
+  checkESContainsIndexTemplates "$pod" || EXIT_CODE=1
+done
+if [[ "$CLUSTER" == "true" ]]; then
+  for pod in $(oc get pods -l component=es-ops -o name | sed 's,pod/,,'); do
+    checkESContainsIndexTemplates "$pod" || EXIT_CODE=1
+  done
+fi
 
 echo $TEST_DIVIDER
 exit $EXIT_CODE
--- a/hack/testing/check-EFK-running.sh
+++ b/hack/testing/check-EFK-running.sh
@@ -96,7 +96,8 @@ function checkESContainsIndexTemplates() {
   echo "Checking presence of index templates: ${template_files}"
   for template in $template_files; do
     echo "  - verify ${template}"
-    if ! waitForValue "oc exec $pod -- curl -s -k -X HEAD -x '%{response_code}' --cert ${secret_dir}admin-cert --key ${secret_dir}admin-key https://localhost:9200/_template/$template"; then
+    if ! response_code=$(waitForValue "oc exec $pod -- curl -s -k -X HEAD -w '%{response_code}' --cert ${secret_dir}admin-cert --key ${secret_dir}admin-key https://localhost:9200/_template/$template") || test "$response_code" != "200" ; then
+      echo "Could not find index template https://localhost:9200/_template/$template - $response_code"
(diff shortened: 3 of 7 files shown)


lukas-vlcek:
First kick-off, do not merge now. @jcantrill Do you think we can have a label for "do-not-merge"? I assume we will be working on Common Data Model for some time in various PRs and branches for some time before anything gets merged.

richm:
@lukas-vlcek Just add [do-not-merge] to the PR description

jcantrill:
@lukas-vlcek @richm  for your labeling pleasure.... added 'do not merge'... or the title is probably more obvious when looking through the list

lukas-vlcek:
The best DO-NOT-MERGE label ever! ++1 @jcantrill 
We can add it to title too @richm, we will figure out later what works best for us.

elasticsearch/run.sh
```diff
@@ -28,4 +37,64 @@ else
...
+	done
+
+	# Test for response code 200 in Elasticsearch output. This can be sensitive to Elasticsearch version.
+	if [ -f ${LOG_FILE} ] && grep -q "200" ${LOG_FILE} ; then
+		echo "- connection successful"
+	else
+		if [ ${timeouted} == true ] ; then
+			echo -n "[timeout] "
+		fi
+		echo "failed"
```
richm: add `cat ${LOG_FILE}`

lukas-vlcek:
[test]

elasticsearch/run.sh
lukas-vlcek: Added.

lukas-vlcek:
@richm I was thinking about adding a test to verify the index template is there, but I am not sure I see a good script to hijack to. Simple solution would be to add this to `/hack/testing/check-logs.go` but that feels hack-inch (and non systematic). My be the best and most understandable solution is creating a whole new script for this?

All we need to do for now is just call `https://${HOST}:${PORT}/_template/_template_name_` and test the index template is there. WDYT?

richm:
@lukas-vlcek It's just a couple of lines of bash right?  Could go directly into logging.sh.

OTOH, I don't think a test like this is necessary, as long as the ES pod fails if the index is not added successfully.  Once we have the integration with fluentd, we can test that the data is flowing correctly from fluentd into ES and have some tests with real data in the index.

lukas-vlcek:
@richm ok, implementing some test later make sense.

ewolinetz:
@lukas-vlcek I think that since this is something that needs to be there, we should make it part of the e2e-test.  I think it would make most sense for it to be a part of the `check-EFK-running.sh` script, to check if the template is there or not since we are already using that script to verify that things we expect to be there, are.

lukas-vlcek:
[test]

lukas-vlcek:
[test]

richm:
[test]

lukas-vlcek:
@ewolinetz I was looking at `check-EFK-running.sh`. May be you can recommend me better idea about implementing basic test (we want to test that ES contains specific index templates upon startup).

Here is my idea:
- I will get list of expected index templates from ES pod (my PR install all cda index templates to `/usr/share/elasticsearch/config/index_templates` in `Dockerfile`). So I can use something like `oc exec $pod ls [...]`
- then I can verify presence of each index template using `oc exec $pod curl -X HEAD --key --cert localhost:9200/_template/_expected_index_template_`

Before I do any of these steps I will want to verify the ES cluster health status is yellow or green (this is [simple curl statement](https://github.com/openshift/origin-aggregated-logging/pull/203/files#diff-05261c9e7776c31e9b2fed5a68db6a3aR77)).

All this can be done in `check-EFK-running.sh` script though I am not sure if similar functionality is already avail in other scripts? So may be I can/should `source` some other scripts first?

WDYT? //cc @richm

ewolinetz:
@lukas-vlcek I'm not sure our other scripts are similar enough... the closest would be `check-logs.go`, that is in go but its what we're using to query against ES.  We don't check the health of ES first (this is something that should probably be added to `check-logs.sh`...).  So where you think it makes the most sense to insert it would probably be good for me.  Hopefully that answers your question

lukas-vlcek:
Pushed a bugfix.
[test]

elasticsearch/Dockerfile
```diff
@@ -31,7 +31,8 @@ RUN rpm --import https://packages.elastic.co/GPG-KEY-elasticsearch && \
     yum clean all
 # initial boring elasticsearch.yml is just to enable plugin installation.
 # put the real config in place in a deployment with a ConfigMap.
-ADD elasticsearch.yml logging.yml /usr/share/elasticsearch/config/
```
richm: you are getting rid of logging.yml?

elasticsearch/Dockerfile
```diff
@@ -31,7 +31,8 @@ RUN rpm --import https://packages.elastic.co/GPG-KEY-elasticsearch && \
...
+ADD elasticsearch.yml /usr/share/elasticsearch/config/
+ADD index_templates /usr/share/elasticsearch/config/index_templates
```
richm: This is one of the problems.  I think this should be 
```
ADD index_templates/ /usr/share/elasticsearch/config/index_templates/
```
that is, they should both end in `/`, so that `docker build` (or the openshift image builder) knows that both the source and destination are directories. For example, from the fluentd pod:
```
```
The pod log shows this:
```
Create index template '*.json'
...
> PUT /_template/*.json HTTP/1.1
...
< HTTP/1.1 400 Bad Request
```

This means it is trying to use a file literally named `*.json`.  This for loop:
```
	for template_file in /usr/share/elasticsearch/config/index_templates/*.json
```
One of the annoying quirks of bash is that if there are no files which match `/usr/share/elasticsearch/config/index_templates/*.json`, then `template_file` will get set to the literal value `/usr/share/elasticsearch/config/index_templates/*.json`.  I would suggest using `shopt -o failglob` before the for loop.  This will cause the script to fail if there are no matching files.  For more information, see `man bash` - look for shopt and failglob.

elasticsearch/run.sh
```diff
@@ -60,4 +69,70 @@ else
...
+verify_or_add_index_templates() {
+	wait_for_port_open
+	# Try to wait for cluster become more stable before index template being pushed in.
+	# Give up on timeout and continue...
+	curl -v -s -k -X GET \
+		--cert ${secret_dir}admin-cert \
+		--key  ${secret_dir}admin-key \
+		"${ES_REST_BASEURL}/_cluster/health?wait_for_status=yellow&timeout=${max_time}s"
+
+	for template_file in /usr/share/elasticsearch/config/index_templates/*.json
```
richm: See above for more information about why this is problematic.

hack/testing/check-EFK-running.sh
```diff
@@ -82,6 +82,26 @@ function checkKibanaStarted() {
   fi
 }
 
+function checkESContainsIndexTemplates() {
+
+  local pod=$1
+  local template_files
+  local secret_dir=/etc/elasticsearch/secret/
+ 
+  if ! template_files=$(waitForValue "oc exec -it $pod ls /usr/share/elasticsearch/config/index_templates"); then
```
richm: `oc exec $pod -- ls -1 /usr/share/elasticsearch/config/index_templates`

hack/testing/check-EFK-running.sh
```diff
@@ -292,4 +312,13 @@ for pod in $(oc get pods -l component=kibana-ops -o name); do
 done
 
 echo $TEST_DIVIDER
+echo "Checking ES contains common data model index templates"
+for pod in $(oc get pods -l component=es -o name | sed 's,pod/,,'); do
+  checkESContainsIndexTemplates "$pod" || EXIT_CODE=1
+done
+for pod in $(oc get pods -l component=es-ops -o name | sed 's,pod/,,'); do
```
richm: Better to explicitly check if using the ops cluster and skip this check, to avoid confusion in the future.
```
if [[ "$CLUSTER" == "true" ]]; then
    for pod in $(oc get pods -l component=es-ops -o name | sed 's,pod/,,'); do
...
```

hack/testing/check-EFK-running.sh
```diff
@@ -82,6 +82,26 @@ function checkKibanaStarted() {
...
+  local secret_dir=/etc/elasticsearch/secret/
+ 
+  if ! template_files=$(waitForValue "oc exec -it $pod ls /usr/share/elasticsearch/config/index_templates"); then
+    return 1
+  fi
+
+  echo "Checking presence of index templates: ${template_files}"
+  for template in $template_files; do
+    echo "  - verify ${template}"
+    if ! waitForValue "oc exec -it $pod curl -s -k -X HEAD -x '%{response_code}' --cert ${secret_dir}admin-cert --key ${secret_dir}admin-key https://localhost:9200/_template/$template"; then
```
richm:
```
oc exec $pod -- curl -s -k ....
```

elasticsearch/Dockerfile
lukas-vlcek: Ahh... good catch! How I managed to drop it?

elasticsearch/Dockerfile
lukas-vlcek: Good to know. Interesting that this works fine for [ViaQ](https://github.com/ViaQ/docker-elasticsearch/pull/12) (/me confused 😖 ). I will fix it.

hack/testing/check-EFK-running.sh
lukas-vlcek: Yea, I screwed up here! Fix is on the way...

lukas-vlcek:
@richm Thanks for looking at this. Somehow I missed your comments. On it now.

elasticsearch/Dockerfile
lukas-vlcek: @richm Wouldn't it be better to do `shopt -o failglob` in the beginning of the script instead before the loop? (May this can clash with some assumptions and break existing logic in the original script? Or is there any reason why it is good to do it just before the loop?)

richm: If you add `shopt -o failglob` at the beginning it will affect everything.  If you are absolutely sure that's safe, that's fine.  If not, then add `shopt -o failglob` before the loop, and add `shopt -o nofailglob` after the loop.

elasticsearch/Dockerfile
```diff
@@ -32,6 +32,7 @@ RUN rpm --import https://packages.elastic.co/GPG-KEY-elasticsearch && \
 # initial boring elasticsearch.yml is just to enable plugin installation.
 # put the real config in place in a deployment with a ConfigMap.
 ADD elasticsearch.yml logging.yml /usr/share/elasticsearch/config/
+ADD index_templates/ /usr/share/elasticsearch/config/index_templates
```
richm: `/usr/share/elasticsearch/config/index_templates/` - if the destination is a directory, add the trailing `/` - that's the way it is used elsewhere in the file.

elasticsearch/run.sh
```diff
@@ -60,4 +69,71 @@ else
...
+verify_or_add_index_templates() {
+	wait_for_port_open
+	# Try to wait for cluster become more stable before index template being pushed in.
+	# Give up on timeout and continue...
+	curl -v -s -k -X GET \
+		--cert ${secret_dir}admin-cert \
+		--key  ${secret_dir}admin-key \
+		"${ES_REST_BASEURL}/_cluster/health?wait_for_status=yellow&timeout=${max_time}s"
+
+	shopt -o failglob
```
richm: This should be `shopt -s -o failglob`

elasticsearch/run.sh
```diff
@@ -60,4 +69,71 @@ else
...
+			echo "Index template '$template' already present in ES cluster"
+		else
+			echo "Create index template '$template'"
+			curl -v -s -k -X PUT \
+				--cert ${secret_dir}admin-cert \
+				--key  ${secret_dir}admin-key \
+				-d@$template_file \
+				${ES_REST_BASEURL}/_template/$template
+		fi
+	done
```
richm: `shopt -u -o failglob` - turn off failglob after the loop

richm:
some minor nits, otherwise, LGTM

lukas-vlcek:
@richm the PR is now failing due to independent Ruby issue in fluentd component, see the [Travis CI log](https://travis-ci.org/openshift/origin-aggregated-logging/builds/156537610). Specifically:

````
...
Successfully installed fluent-plugin-kubernetes_metadata_filter-0.24.0
Building native extensions.  This could take a while...
ERROR:  Error installing fluent-plugin-kubernetes_metadata_filter:
	serverengine requires Ruby version >= 2.1.0.
ERROR:  Error installing fluent-plugin-elasticsearch:
	serverengine requires Ruby version >= 2.1.0.
ERROR:  Error installing fluent-plugin-flatten-hash:
	serverengine requires Ruby version >= 2.1.0.
...
Successfully installed fluent-plugin-flatten-hash-0.4.0
Building native extensions.  This could take a while...
ERROR:  Error installing fluent-plugin-systemd:
	serverengine requires Ruby version >= 2.1.0.
ERROR:  Error installing fluent-plugin-rewrite-tag-filter:
	serverengine requires Ruby version >= 2.1.0.
````

May be `fluent-plugin-kubernetes_metadata_filter` and `fluent-plugin-flatten-hash` plugins have been released in the mean time and they depend on newer Ruby? We might require specific versions of those plugins?

File: elasticsearch/run.sh (hunk 10 above)
Comment (lukas-vlcek) [inline elasticsearch/run.sh]: Done.

**Comment by richm** [comment]
@lukas-vlcek I think the problem is that gem by default tries to update dependencies of dependencies to the latest version, then that triggers other updates.  We should disable this behavior.  Try adding this to the gem install line `--conservative --minimal-deps` - see `man gem` or `gem install --help` for more information.
```
        --conservative               Don't attempt to upgrade gems already
                                               meeting version requirement
        --minimal-deps               Don't upgrade any dependencies that already
                                               meet version requirements
```

**Comment by ewolinetz** [comment]
@lukas-vlcek would it makes sense/is there also space to update the templates to restrict the number of indices to start with 1 primary shard and 0 replicas?  not sure if this would make more sense to be a separate template... this way we can make recommendations to increase the number of replicas as the ES cluster size increases.  otherwise we'll still be doing the default of 5/1

If it doesn't make sense to add that in here we can always add it as a config in ES

**Comment by lukas-vlcek** [comment]
@richm yep, `gem install [gems] --conservative --minimal-deps` did the trick!

**Comment by lukas-vlcek** [comment]
Getting: `/opt/app-root/src/run.sh: line 114: shopt: failglob: invalid option name`
Looking at it...

**Comment by lukas-vlcek** [comment]
@ewolinetz I am not sure what will 0 replicas buy us. Is there any specific use case this will help with?

File: elasticsearch/run.sh (hunk 10 above)
Comment (richm) [inline elasticsearch/run.sh]: sorry - you are correct - `-o` is only needed for `set` options, not `shopt`

**Comment by richm** [comment]
The error is
```
/opt/app-root/src/run.sh: line 105: no match: /usr/share/elasticsearch/config/index_templates/*.json
```

lukas-vlcek:
@richm right, that is the latest failure, not sure why ... currently trying to get more info about this by logging more info from this script.

ewolinetz:
@lukas-vlcek its more for the fact that most of our openshift users may not have more than one instance of ES running at first. It'll prevent their instance from needing to hold on to information about the replica shards that it can never allocate

lukas-vlcek:
[test]

lukas-vlcek:
[test]

richm:
LGTM

lukas-vlcek:
Closing this PR and opening a new PR against `es2.x` branch instead.

richm:
@lukas-vlcek can we close this PR?

richm:
@lukas-vlcek never mind - already closed

PR #208, opened 2016-08-10
[DO_NOT_MERGE] Logging upgrade

ewolinetz:
@sosiouxme @jcantrill @richm PTAL - this isn't to be merged in, but will be used for QE to test from.
I plan to update the Readme as well to cover topics such as de-dotting and to update the health-check information.  Also it having a PR will allow it to go through the CI testing process.

Other things to note: 
SG currently doesn't let us configure its index. [1]
And if I remember correctly we are impacted by this when trying to connect to another ES node for clustering (will retest) [2]

[1] https://github.com/floragunncom/search-guard/issues/176
[2] https://github.com/floragunncom/search-guard/issues/153


ewolinetz:
There seems to be some strangeness with SG initializing itself when there are multiple nodes in the cluster. I'm going to investigate seeding SG sooner in the openshift-elasticsearch-plugin...

deployer/conf/elasticsearch.yml
```diff
@@ -1,97 +1,60 @@
...
-      system.logging.kibana.*.indices: [".kibana.*"]
-      system.logging.curator:
-        execute: ["actionrequestfilter.curator"]
-        actionrequestfilter.curator.comment: "Curator can list all indices and delete them"
-      system.admin:
-        bypass: ["*"]
-      system.admin.*.comment: "Admin user can do anything"
+      enabled: true
+      enforce_hostname_verification: false
+      enabled_protocols: ["TLSv1", "TLSv1.1", "TLSv1.2"]
```
sosiouxme: Only internals should be contacting ES. Shouldn't we drop TLSv1 here?

ewolinetz: I think I was playing around with the TLS versions ES supports when Curator wasn't able to connect.  I'll retest without TLSv1 now that Curator is working

deployer/scripts/install.sh
```diff
@@ -185,11 +185,17 @@ function generate_config() {
...
+    generate_JKS_client_cert "$admin_user"
 
-    # generate java store/trust for the ES SearchGuard plugin
-    generate_JKS_chain logging-es "$(join , logging-es{,-ops}{,-cluster}{,.${project}.svc.cluster.local})"
     # generate common node key for the SearchGuard plugin
-    openssl rand 16 | openssl enc -aes-128-cbc -nosalt -out $dir/searchguard_node_key.key -pass pass:pass
+    # we use a JKS chain now instead for inter-node communication
+    #openssl rand 16 | openssl enc -aes-128-cbc -nosalt -out $dir/searchguard_node_key.key -pass pass:pass
+    generate_JKS_chain true elasticsearch "$(join , logging-es{,-ops})"
+    mv $dir/keystore.jks $dir/searchguard_node_key.key
```
sosiouxme: i would lose the suffix or make it .jks

deployer/templates/es.yaml
```diff
@@ -92,7 +96,7 @@ objects:
                   mountPath: /etc/elasticsearch/secret
                   readOnly: true
                 - name: elasticsearch-config
-                  mountPath: /usr/share/elasticsearch/config
+                  mountPath: /etc/elasticsearch/config
```
sosiouxme: i like the consistency but everything is still looking in `/usr/share/elasticsearch/config` so isn't this making the configmap useless?

elasticsearch/logging.yml
```diff
@@ -24,7 +24,7 @@ logger:
 
   # search-guard
   # Due to search-guard issue #33 this should be set to ERROR (too verbose at WARN)
```
sosiouxme: nuke the comment then...

elasticsearch/Dockerfile
```diff
@@ -8,17 +8,19 @@ USER 0
 
 ENV HOME=/opt/app-root/src \
   JAVA_VER=1.8.0 \
-  ES_VER=1.5.2 \
-  ES_CONF=/usr/share/elasticsearch/config/elasticsearch.yml \
+  ES_VER=2.3.3 \
+  ES_CONF=/usr/share/elasticsearch/config \
```
sosiouxme: couldn't we just make this /etc/elasticsearch/config ?

elasticsearch/run.sh
```diff
@@ -2,12 +2,23 @@
...
-mkdir -p /elasticsearch/$CLUSTER_NAME
-secret_dir=/etc/elasticsearch/secret/
-[ -f $secret_dir/searchguard-node-key ] && ln -s $secret_dir/searchguard-node-key /elasticsearch/$CLUSTER_NAME/searchguard_node_key.key
-[ -f $secret_dir/searchguard.key ] && ln -s $secret_dir/searchguard.key /elasticsearch/$CLUSTER_NAME/searchguard_node_key.key
-[ -f $secret_dir/keystore.password ] && export KEYSTORE_PASSWORD=$(cat $secret_dir/keystore.password)
-[ -f $secret_dir/truststore.password ] && export TRUSTSTORE_PASSWORD=$(cat $secret_dir/truststore.password)
+secret_dir=/etc/elasticsearch/secret
+config_dir=/etc/elasticsearch/config
+
+# ES seems to be picky about its config/ dir and permissions within it -- ln -s would yield incorrect permissions...
```
sosiouxme: hm, OK... well, it seems like there oughta be a way to use the directories directly rather than copying everything around...

elasticsearch/run.sh
```diff
@@ -28,4 +39,24 @@ else
...
 
-exec /usr/share/elasticsearch/bin/elasticsearch
+ln -s /usr/share/elasticsearch/config/${CLUSTER_NAME}.truststore /usr/share/elasticsearch/config/${CLUSTER_NAME}.truststore.jks
+cp /etc/elasticsearch/secret/admin-jks /usr/share/elasticsearch/config/admin.jks
+
+TIMES=60
+function waitForES() {
+  for (( i=1; i<=$TIMES; i++ )); do
+    # test for ES to be up first
+		# don't provide a client cert since ES doesn't validate us correctly
```
sosiouxme: what's the story here... SG hasn't built its ACL yet?

elasticsearch/sgconfig/sg_rolesmapping.yml
```diff
@@ -0,0 +1,15 @@
+sg_role_kibana:
+  users:
+    - 'CN=system.logging.kibana,OU=OpenShift,O=Logging,L=Test,C=DE'
```
sosiouxme: so we have to match the entire subject now, not just the CN? that's... odd.

kibana/kibana.yml
```diff
@@ -1,67 +1,79 @@
...
-# used by the Kibana server to perform maintence on the kibana_index at statup. Your Kibana
-# users will still need to authenticate with Elasticsearch (which is proxied thorugh
+# If your Elasticsearch is protected with basic auth, these are the user credentials
+# used by the Kibana server to perform maintenance on the kibana_index at startup. Your Kibana
+# users will still need to authenticate with Elasticsearch (which is proxied through
 # the Kibana server)
-kibana_elasticsearch_username: kibana
-kibana_elasticsearch_password: changeme
+elasticsearch.username: "kibana"
+elasticsearch.password: "changeme"
```
sosiouxme: it's a tiny thing but could we change this :)

something like "dummy" or "foo" or... could this bit just be left out?

sosiouxme:
We wanted a custom searchguard index because we don't want multiple instances overwriting each others' ACLs (instead give each its own ACL). Is that maybe related to what's happening?

sosiouxme:
This doesn't include the changes to our ES plugin. Question regarding authnz. Would it be correct to say that under this config you can't access ES/SG at all without a client cert? In the old configuration, if you had no client cert but did have a token bearer, that would still give you access as your token - I was hoping to use that from an upgraded `oc logs` command in the future. Is this ability still there, or should we also define some kind of "dummy" client cert that's essentially like the kibana cert (lets you connect but your real auth is your token)?

Also, I wouldn't rely on the x-proxy-remote-user header unless it was coming from kibana (not clear if we are, just noting this).

jcantrill:
@sosiouxme are you referring to @ewolinetz comment regarding strangeness with multiple ES nodes?  Now that you mention it, I recall what you are saying being a specific design point we discussed.  Each ES node instance with SG installed would have its own index specifically for auth purposes.  We wanted each node to have to pull ACL's and expire them independently so we didn't need to worry about syncing across nodes.

sosiouxme:
Yes, that's what I was referring to.

deployer/scripts/install.sh
ewolinetz: yeah that's weird... why'd I do that?  I'll drop it to match the truststore

deployer/templates/es.yaml
ewolinetz: I ended up doing a bunch of odd things to appease the java security manager, I believe that ES picks up changes from the configmap currently... but now I'm not sure how :-/ I'll probably need to rip and replace some things around the configs, this included.

elasticsearch/run.sh
ewolinetz: it seems that we were able to do things like that before in 1.5 since the java security manager wasn't enabled.  any objections to running with it disabled in 2.x?

elasticsearch/run.sh
ewolinetz: SG doesn't poll its index anymore as it did in 1.5, it listens for a change on specific ES types.  We can seed correctly when we receive a REST request without this function; but in the scenario where we just start up an ES cluster with say 2 nodes, one may not initialize its ACL and then SG starts to complain that it isn't initialized...

ewolinetz: This also goes along with what is done in #203

elasticsearch/sgconfig/sg_rolesmapping.yml
ewolinetz: yup, for the client cert auth, it pulls the whole subject to be the user name

ewolinetz:
@sosiouxme this does contain the changes to our ES plugin... the plugin now reads from the `sgconfig` directory (which can be configured, but defaults to where it currently is).  

Currently we can either connect to ES giving either a client cert, or using a proxy header that we can specify, so I don't think we can just provide a token unless we patch it into SG somehow...

deployer/templates/es.yaml
ewolinetz: the changes in the 'testing' commit seem to work around all that wackiness i was doing before...

elasticsearch/run.sh
sosiouxme: You know, if the security manager ever is responsible for neutralizing some CVE, this minor hack is probably worth it. So I guess I'd say leave it on, if it doesn't become a blocker.

ewolinetz: IIRC it actually caused some minor issues with the K8S client in our plugin... but that may just require a plugin whitelist.  I'll check it out...

elasticsearch/run.sh
lukas-vlcek: @ewolinetz checking if I understand correctly: Are you saying that when we try to push index templates to ES in `run.sh` we can run into issues because SG did initialize this node correctly?

ewolinetz: @lukas-vlcek no, so this is specific to the work for es 2.3 due to how SG now initializes itself. It basically blocks traffic to ES unless it is configured which is what this loop is ensuring (we set up the openshift-elasticsearch-plugin to check if its initialized when a rest request comes in)

ewolinetz:
moment of truth... [test]

ewolinetz:
for some reason curator isn't able to pull the list of indices currently during the curator tests... need to investigate further

kibana/Dockerfile
```diff
@@ -4,31 +4,22 @@ MAINTAINER OpenShift Development <dev@lists.openshift.redhat.com>
 
 EXPOSE 5601
 
-ENV KIBANA_VER=4.1.2 \
+ENV KIBANA_VER=4.5.1 \
```
jcantrill: @ewolinetz  Need to bump to 4.5.4 to fix ESA-2016-04 and ESA-2016-03 identified here: https://www.elastic.co/community/security

ewolinetz:
the kibana version needs to be bumped up to at least 4.5.4 due to:

    bug 1364389 XSS vulnerability - ESA-2016-03
    bug 1364394 Session hijacking - ESA-2016-04

ewolinetz:
[test]

ewolinetz:
there was an issue with the upgrade path where the admin cert isn't letting us do a uuid migration

ewolinetz:
Also need to bump ES to at least 2.3.4 per https://www.elastic.co/community/security

ewolinetz:
I'm seeing something odd in kibana when viewing logs "/elasticsearch/.operations.*/_field_stats?level=indices"  currently trying to resolve.. otherwise stack seems to be good :)

ewolinetz:
Currently getting
`
[2016-08-29 21:43:46,093][WARN ][rest.suppressed          ] /.operations.*/_field_stats Params: {level=indices, index=.operations.*}
java.lang.UnsupportedOperationException: custom format isn't supported
`
only for operations logs from Kibana
while using journald driver

richm:
@ewolinetz 
> while using journald driver
Does that mean you see the problem when docker is using log-driver=journald, but not when docker is using log-driver=json-file?

ewolinetz:
@richm I've only tested this using the journald log driver and I've only seen this happen with the newer kibana/ES, if i rebuild and run on the 1.5 stack I see no issue.

I haven't yet tested if there is an issue as well when using json-file.  I will verify that next

ewolinetz:
@richm this doesn't appear when using the json-file log driver for docker...

lukas-vlcek:
Just for the record this issue seem similar: https://discuss.elastic.co/t/custom-format-isnt-supported/41083

ewolinetz:
@lukas-vlcek what is odd though, is if we use the `json-file` log driver instead of the `journald` log driver for docker I don't see this issue come about.  And if we are using `jourald` this error only comes up for the operations index, not the normal application indices.

We could switch to using the non-pre-flight request mode, but it seems like there might be something in our formatting of the syslog records in our fluentd journald logic that is slightly "odd" for Kibana...

Thanks for taking a look!

ewolinetz:
@richm @lukas-vlcek I think this stems from the rfc3339(6) format of the time stamp for operations journal entries... I'll see if there's a way to allow that as a format, if not i'll switch it to a non-preflight request as described above

ewolinetz:
@richm @lukas-vlcek the commit to add format to the default journald `time` value resolved it...

richm:
lgtm

richm:
find the troublesome record:
```
     time ${(record["_SOURCE_REALTIME_TIMESTAMP"] || record["__REALTIME_TIMESTAMP"]) ? Time.at((record["_SOURCE_REALTIME_TIMESTAMP"] || record["__REALTIME_TIMESTAMP"]).to_f / 1000000.0).to_datetime.rfc3339(6) : puts "bad record #{record}"}

ewolinetz:
after #215 and #217 merge in, i'll rebase and have this start being tested

ewolinetz:
closing in favor of #219

PR #213, opened 2016-08-29
fixes security issues identified in bug 1370892

jcantrill:
- bug 1364389 XSS vulnerability - ESA-2016-03
- bug 1364394 Session hijacking - ESA-2016-04

[test] cc @ewolinetz @sosiouxme @richm

### DIFF
--- a/kibana/Dockerfile
+++ b/kibana/Dockerfile
@@ -4,7 +4,7 @@ MAINTAINER OpenShift Development <dev@lists.openshift.redhat.com>
 
 EXPOSE 5601
 
-ENV KIBANA_VER=4.1.2 \
+ENV KIBANA_VER=4.1.11 \
     ES_HOST=localhost \
     ES_PORT=9200 \
     AOP_KIBANA_PLUGIN_VER=v0.6.0 \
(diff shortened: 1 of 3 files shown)


jcantrill:
@ewolinetz  does this also require an update to elasticsearch?  wasn't there a value in the config that referenced the kibana version?

kibana/Dockerfile
```diff
@@ -4,7 +4,7 @@ MAINTAINER OpenShift Development <dev@lists.openshift.redhat.com>
 
 EXPOSE 5601
 
-ENV KIBANA_VER=4.1.2 \
+ENV KIBANA_VER=4.1.11 \
```
ewolinetz: we also need to update the ES config setting that specifies the kibana version, otherwise we won't be able to accurately set the default index
[1] https://github.com/fabric8io/openshift-elasticsearch-plugin/blob/ES1.5.x/src/main/java/io/fabric8/elasticsearch/plugin/ConfigurationSettings.java#L29

ewolinetz:
@jcantrill yeah, there's a value in the ES config we need to also update.  The setting is `kibana.version`

jcantrill:
@ewolinetz updated deployer/conf and README to reflect bump of 4.1.11

ewolinetz:
LGTM

jcantrill:
[merge]

PR #215, opened 2016-08-30
origin fix for bug 1369646

sosiouxme:
Make the upgrade set the correct secret volume on the logging-kibana-ops
DC; in earlier versions, it got a separate-but-equal secret, but in the
present versions both kibana DCs should use the same
logging-kibana-proxy secret.

### DIFF
--- a/deployer/scripts/upgrade.sh
+++ b/deployer/scripts/upgrade.sh
@@ -591,6 +591,17 @@ function remove_triggers_and_IS() {
   oc delete is -l logging-infra=support
 }
 
+# at some point we dropped the separate-but-equal kibana secrets
+function update_kibana_ops_proxy_secret() {
+  local dc
+  for dc in $(oc get dc -o name -l component=kibana-ops,logging-infra=kibana); do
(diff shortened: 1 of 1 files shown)


sosiouxme:
@ewolinetz FYI also I wasn't sure about testing for this in the upgrade... would have to set the kibana-proxy DC volume back to the earlier secret, and actually copy that secret, and then validate that the upgrade puts it back how it's supposed to be... what do you think?

ewolinetz:
@sosiouxme yeah, i think we take that for granted. 
I think it'd be kind of tricky to put it back in a bad state unless we just use _any_ cert.. not necessarily the one it had before. Then we could verify if kibana-ops and kibana have the same certificate?

sosiouxme:
[merge] now, add tests later

ewolinetz:
[merge]

sosiouxme:
builds will fail until https://github.com/openshift/origin-aggregated-logging/pull/217 goes in

sosiouxme:
alright now [merge]

ewolinetz:
i dont think #217 went in yet...

PR #219, opened 2016-09-02
Logging upgrade

ewolinetz:


elasticsearch/Dockerfile
```diff
@@ -8,17 +8,19 @@ USER 0
 
 ENV HOME=/opt/app-root/src \
   JAVA_VER=1.8.0 \
-  ES_VER=1.5.2 \
-  ES_CONF=/usr/share/elasticsearch/config/elasticsearch.yml \
+  ES_VER=2.3.3 \
```
richm: 2.3.5 is the latest 2.3 branch release

elasticsearch/Dockerfile
```diff
@@ -8,17 +8,19 @@ USER 0
...
+  ES_CONF=/usr/share/elasticsearch/config \
```
richm: Why is some config in /etc/elasticsearch, and some in /usr/share/elasticsearch?

elasticsearch/sgconfig/sg_rolesmapping.yml
```diff
@@ -0,0 +1,15 @@
+sg_role_kibana:
+  users:
+    - 'CN=system.logging.kibana,OU=OpenShift,O=Logging,L=Test,C=DE'
```
richm: I think we need a better subjectname . . . or at least, no Test and c=DE . . .

fluentd/Dockerfile
```diff
@@ -5,7 +5,7 @@ MAINTAINER OpenShift Development <dev@lists.openshift.redhat.com>
 ENV HOME=/opt/app-root/src \
   PATH=/opt/app-root/src/bin:/opt/app-root/bin:$PATH \
   RUBY_VERSION=2.0 \
-  FLUENTD_VERSION=0.12.23 \
+  FLUENTD_VERSION=0.12.27 \
```
richm: 12.28

kibana/kibana.repo
```diff
@@ -0,0 +1,6 @@
+[kibana-4.5]
```
richm: This .repo file isn't used anywhere?

richm:
Are you planning to squash these?

kibana/kibana.repo
ewolinetz: good catch, i had meant to update the kibana install to use the rpm again but lost track of that

ewolinetz:
@richm yeah, i had squashed and then through testing let the commits expand again so i could remove any unnecessary ones

elasticsearch/Dockerfile
ewolinetz: ack, i'll have to check how SG is doing since we now have to specify an ES version in our plugins I believe

elasticsearch/Dockerfile
ewolinetz: it should only be `/usr/share/elasticsearch/config`
in fact the /etc/elasticsearch line in the dockerfile is commented out... i'll clean it up

ewolinetz:
[test]

ewolinetz:
ES plugins are the worst... they're required to be built against specific ES versions now. rebuilding plugins for 2.3.5

ewolinetz:
couldn't connect to frontend for generating logs... re[test]

elasticsearch/sgconfig/sg_rolesmapping.yml
ewolinetz: @richm  should we just leave it at 'CN=system.logging.kibana,OU=OpenShift,O=Logging' ? Any preferences on L and C if we should have them...

richm: `CN=system.logging.kibana,OU=OpenShift,O=Logging` works for me - afaik, no need for L or C.  If for some reason it is an SG requirement, then I suggest L=Raleigh,c=US

ewolinetz:
fluentd build took too long [test]

ewolinetz:
@jcantrill @richm @sosiouxme PTAL

curator/run.sh
```diff
@@ -1,4 +1,19 @@
 #!/bin/bash
 
+TIMES=60
+
+function waitForES() {
+  for ((i=1; i<=$TIMES; i++ )); do
+    # test for ES to be up first
+    result=$(curl --cacert $ES_CA --key $ES_CLIENT_KEY --cert $ES_CLIENT_CERT -s -w "%{http_code}" -XGET "https://$ES_HOST:$ES_PORT/" -o /dev/null)
```
jcantrill: Minor tweak, but thinking of the kibana plugin issue we discussed where users MIGHT want to run un-secure.  Should we consider allowing the protocol to be added in via an undocumented ENV_VAR?

deployer/conf/elasticsearch-logging.yml
```diff
@@ -23,7 +23,6 @@ logger:
   index.indexing.slowlog: TRACE, index_indexing_slow_log_file
 
   # search-guard
-  # Due to search-guard issue #33 this should be set to ERROR (too verbose at WARN)
   com.floragunn.searchguard: ERROR
```
jcantrill: Can we drop to WARN now?

deployer/scripts/util.sh
```diff
@@ -83,7 +88,47 @@ function generate_PEM_cert() {
...
+
+function generate_JKS_client_cert() {
+  NODE_NAME="$1"
+  ks_pass=${KS_PASS:-kspass}
+  ts_pass=${TS_PASS:-tspass}
+  dir=${SCRATCH_DIR:-_output}  # for writing files to bundle into secrets
+
+  echo Generating keystore and certificate for node ${NODE_NAME}
+
+  "$JAVA_HOME/bin/keytool" -genkey \
```
jcantrill: worth refactoring into an env var like '$KEY_TOOL' ???

elasticsearch/install.sh
```diff
@@ -5,11 +5,15 @@ set -ex
 mkdir -p ${HOME}
 ln -s /usr/share/elasticsearch /usr/share/java/elasticsearch
 
-/usr/share/elasticsearch/bin/plugin -i com.floragunn/search-guard/0.5.1 -url https://github.com/lukas-vlcek/origin-aggregated-logging/releases/download/v0.1/search-guard-0.5.1.zip
+/usr/share/elasticsearch/bin/plugin install -b com.floragunn/search-guard-ssl/2.3.5.15
```
jcantrill: Does it make sense to simplify future changes by moving the versions to env vars?

elasticsearch/run.sh
```diff
@@ -2,12 +2,23 @@
...
 set -euo pipefail
 
-mkdir -p /elasticsearch/$CLUSTER_NAME
-secret_dir=/etc/elasticsearch/secret/
-[ -f $secret_dir/searchguard-node-key ] && ln -s $secret_dir/searchguard-node-key /elasticsearch/$CLUSTER_NAME/searchguard_node_key.key
-[ -f $secret_dir/searchguard.key ] && ln -s $secret_dir/searchguard.key /elasticsearch/$CLUSTER_NAME/searchguard_node_key.key
-[ -f $secret_dir/keystore.password ] && export KEYSTORE_PASSWORD=$(cat $secret_dir/keystore.password)
-[ -f $secret_dir/truststore.password ] && export TRUSTSTORE_PASSWORD=$(cat $secret_dir/truststore.password)
+secret_dir=/etc/elasticsearch/secret
+#config_dir=/etc/elasticsearch/config
```
jcantrill: Reason to keep these commented lines?

elasticsearch/run.sh
```diff
@@ -60,4 +71,24 @@ else
 	exit 1
 fi
 
-exec /usr/share/elasticsearch/bin/elasticsearch
+#ln -s /usr/share/elasticsearch/config/${CLUSTER_NAME}.truststore /usr/share/elasticsearch/config/${CLUSTER_NAME}.truststore.jks
+#cp /etc/elasticsearch/secret/admin-jks /usr/share/elasticsearch/config/admin.jks
+
+TIMES=60
+function waitForES() {
```
jcantrill: Duplicate of whats used by curator?  Seems like we have a case for some common scripts

kibana/install.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# Kibana starts up slowly because it tries to optimize and cache bundles
+# so we start it up as part of install and then stop it
+mv ${KIBANA_HOME}/config/kibana.yml ${KIBANA_HOME}/config/hidden_kibana.yml
+touch ${KIBANA_HOME}/config/kibana.yml
+
+${KIBANA_HOME}/bin/kibana > ${KIBANA_HOME}/kibana.out &
+pid=$!
+
+until [ -n "$(grep 'Optimization of bundles for kibana and statusPage complete' ${KIBANA_HOME}/kibana.out)" ]; do
```
jcantrill: Timeout?  Any chance of infinitely waiting?

kibana/kibana.yml
```diff
@@ -1,67 +1,79 @@
 # Kibana is served by a back end server. This controls which port to use.
-port: 5601
+# server.port: 5601
```
jcantrill: I noticed a few other places where we either grep/look for 5601.  If we allow it to change, that seems to be a problem with 'the check'

kibana/kibana.yml
```diff
@@ -1,67 +1,79 @@
...
- - plugins/metric_vis/index
- - plugins/origin-kibana/index
- - plugins/settings/index
- - plugins/table_vis/index
- - plugins/vis_types/index
- - plugins/visualize/index
+# pid.file: /var/run/kibana.pid
+
+# If you would like to send the log output to a file you can set the path below.
+# logging.dest: stdout
```
jcantrill: seems like this would be useful

elasticsearch/run.sh
ewolinetz: Right now it is a bit of a duplicate, but i'm not sure we'd be able to have common scripts since they're in two separate images.  This will also change a little when we integrate the template seeding from #222

kibana/install.sh
ewolinetz: It shouldn't wait forever...., the other side of this is if we don't do it here Kibana would do it upon normal start up and then it if there is a chance this would happen it would wait forever when trying to start up for a customer.

kibana/kibana.yml
ewolinetz: Kibana doesn't use a configmap, so it shouldn't be able to be changed from 5601by a user...

kibana/kibana.yml
ewolinetz: the default is stdout which is what Docker is expecting.  Its commented out and kept around just as a 'fyi' that it can be changed (if need be) in the future

elasticsearch/run.sh
ewolinetz: I forgot I have a little bit more investigation to do as a follow up on this, otherwise no reason to keep them

ewolinetz:
[test]

curator/run.sh
richm: I think we should address that in a different commit/PR - The change to allow `http` instead of only `https` will hit a lot of code.

ewolinetz: @richm agreed -- i think we discussed this past Friday that we don't plan to support the switch to allow both because it would impact so many components.

elasticsearch/run.sh
```diff
@@ -60,4 +57,20 @@ else
 	exit 1
 fi
 
-exec /usr/share/elasticsearch/bin/elasticsearch
+TIMES=60
+function waitForES() {
+  for (( i=1; i<=$TIMES; i++ )); do
+    # test for ES to be up first
+		# don't provide a client cert since ES doesn't validate us correctly
```
richm: Is this comment still correct?  the command is supplying `--cert` and `--key`

fluentd/configs.d/openshift/filter-syslog-record-transform.conf
```diff
@@ -32,6 +32,6 @@
     # if field is optional you have to use record[name] or (name rescue nil) or something like that
     ident ${record["SYSLOG_IDENTIFIER"] || record["_COMM"]}
     message ${record["MESSAGE"]}
-    version 1.3.0
+    version 1.3.1
```
richm: Why is the data version different?

richm:
Just a couple of minor issues.  LGTM

fluentd/configs.d/openshift/filter-syslog-record-transform.conf
ewolinetz: This originally also had the changes to the default time being shipped to ES, that was why the version was bumped.  Since that is part of master now i'll drop the version back down (I was anticipating that change going out first here)

elasticsearch/install.sh
```diff
@@ -5,11 +5,16 @@ set -ex
...
 
-/usr/share/elasticsearch/bin/plugin -i com.floragunn/search-guard/0.5.1 -url https://github.com/lukas-vlcek/origin-aggregated-logging/releases/download/v0.1/search-guard-0.5.1.zip
+/usr/share/elasticsearch/bin/plugin install -b com.floragunn/search-guard-ssl/${SG_SSL_VER}
+#/usr/share/elasticsearch/bin/plugin install -b com.floragunn/search-guard-2/2.3.3.3
 
-/usr/share/elasticsearch/bin/plugin -i io.fabric8.elasticsearch/openshift-elasticsearch-plugin/0.15
-
-/usr/share/elasticsearch/bin/plugin -i io.fabric8/elasticsearch-cloud-kubernetes/1.3.0
+# We are currently carrying a patch for sg-2; cannot use normal install
+/usr/share/elasticsearch/bin/plugin install -b file://${HOME}/search-guard-2-2.3.5.6.zip
```
ewolinetz: This change was recently merged into their codebase; will change this to not be a fork once they release

ewolinetz: This can be carried forward as tech debt so long as there are no other comments on this PR

jcantrill:
No objections here.  LGTM

deployer/README.md
```diff
@@ -901,47 +901,50 @@ are using the systemd journal as the log source, look in `/var/log/journal.pos`.
...
+[2016-09-07 19:39:16,895][INFO ][env                      ] [Kofi Whitemane] using [1] data paths, mounts [[/elasticsearch/persistent (/dev/xvda2)]], net usable_space [15.8gb], net total_space [24.9gb], spins? [possibly], types [xfs]
+[2016-09-07 19:39:16,895][INFO ][env                      ] [Kofi Whitemane] heap size [3.9gb], compressed ordinary object pointers [true]
+[2016-09-07 19:39:17,502][INFO ][http                     ] [Kofi Whitemane] Using [org.elasticsearch.http.netty.NettyHttpServerTransport] as http transport, overridden by [search-guard2]
+[2016-09-07 19:39:17,647][INFO ][transport                ] [Kofi Whitemane] Using [com.floragunn.searchguard.transport.SearchGuardTransportService] as transport service, overridden by [search-guard2]
+[2016-09-07 19:39:17,647][INFO ][transport                ] [Kofi Whitemane] Using [com.floragunn.searchguard.ssl.transport.SearchGuardSSLNettyTransport] as transport, overridden by [search-guard-ssl]
+[2016-09-07 19:39:19,019][INFO ][plugins                  ] [Johnny Blaze] modules [], plugins [search-guard-ssl, search-guard2], sites []
```
sosiouxme: minor but shouldn't this still be Kofi Whitemane?

deployer/deployer.yaml
```diff
@@ -88,6 +88,17 @@ items:
...
+    apiVersion: v1
+    kind: RoleBinding
+    metadata:
+      name: logging-elasticsearch-view-role
+    roleRef:
+      kind: ClusterRole
+      name: view
+    subjects:
+    - kind: ServiceAccount
+      name: aggregated-logging-elasticsearch
```
sosiouxme: Why does elasticsearch need view on the logging project? I'm not seeing anything that needs it...

deployer/scripts/install.sh
```diff
@@ -184,12 +184,19 @@ function generate_config() {
...
+
+    generate_JKS_client_cert "$admin_user"
     generate_PEM_cert "$admin_user"
 
-    # generate java store/trust for the ES SearchGuard plugin
-    generate_JKS_chain logging-es "$(join , logging-es{,-ops}{,-cluster}{,.${project}.svc.cluster.local})"
     # generate common node key for the SearchGuard plugin
-    openssl rand 16 | openssl enc -aes-128-cbc -nosalt -out $dir/searchguard_node_key.key -pass pass:pass
+    # we use a JKS chain now instead for inter-node communication
+    #openssl rand 16 | openssl enc -aes-128-cbc -nosalt -out $dir/searchguard_node_key.key -pass pass:pass
```
sosiouxme: may as well nuke this cruft

deployer/README.md
ewolinetz: This is actually due to the openshift-elasticsearch-plugin connecting to the cluster as a client.  So this output is coming from that client as it's starting up.

deployer/deployer.yaml
ewolinetz: hmm, this may be vestigial from when I was trying to get the openshift-client to work within the openshift-elasticsearch-plugin.  let me verify whether or not this is needed still.

elasticsearch/run.sh
```diff
@@ -60,4 +57,18 @@ else
 	exit 1
 fi
 
-exec /usr/share/elasticsearch/bin/elasticsearch
+TIMES=60
```
sosiouxme: so... if recovery takes longer than 60 seconds, we bail out? or am I misreading what's below?

ewolinetz: basically -- we try to connect to ES for 60 seconds until we see that the searchguard index has been created. ES will start allowing connections through once it finishes loading in indices, so if it doesn't start up within 60 seconds then we bail.

Think this should be bumped up from 60?

sosiouxme:
one concern, otherwise LGTM

elasticsearch/run.sh
ewolinetz: @sosiouxme i removed the `exit 255` since it probably isn't doing what i think/want it to do.
I also bumped us up to 300 from 60; I considered a do-while but I didn't want this out these spinning forever if someone ES gets stuck...

sosiouxme: I was thinking the background process would kill ES by exiting 255, but
that was just my poor understanding of bash. Five minutes of trying should
be fine, LGTM :)

ewolinetz:
lets do this... [merge]!

ewolinetz:
the test failed with this in the deployer
`Error from server: Operation cannot be fulfilled on deploymentconfigs "logging-es-ops-gw6323nm": the object has been modified; please apply your changes to the latest version and try again`

jcantrill:
@danmcp how were we able to merge but fail a test?

danmcp:
@jcantrill Merge and test are separate.  You can see the merge run just above that was successful.

jcantrill:
@danmcp I did notice that but is merge not gated on test?

danmcp:
@jcantrill Test runs the tests and reports back the results.  Merge runs the tests (completely separate) and merges on success.  You don't need to run test and merge.  Test is just there to work out the kinks before you are ready to merge.

PR #222, opened 2016-09-07
Add Common Data Model index templates

lukas-vlcek:
Opening instead of #203. We are going to work against `es2.x` branch until it is merged into master.

### DIFF
--- a/elasticsearch/index_templates/com.redhat.viaq-openshift-operations.template.json
+++ b/elasticsearch/index_templates/com.redhat.viaq-openshift-operations.template.json
@@ -1,0 +1,569 @@
+{
+  "aliases": {},
+  "mappings": {
+    "_default_": {
+      "_meta": {
+        "version": "2016.10.12.0"
+      },
+      "date_detection": false,
+      "dynamic_templates": [
+        {
+          "message_field": {
+            "mapping": {
+              "index": "analyzed",
+              "omit_norms": true,
+              "type": "string"
+            },
+            "match": "message",
+            "match_mapping_type": "string"
+          }
+        },
+        {
+          "string_fields": {
+            "mapping": {
+              "fields": {
+                "raw": {
+                  "ignore_above": 256,
+                  "index": "not_analyzed",
+                  "type": "string"
+                }
+              },
+              "index": "analyzed",
+              "omit_norms": true,
+              "type": "string"
+            },
+            "match": "*",
+            "match_mapping_type": "string"
+          }
+        }
+      ],
+      "properties": {
+        "@timestamp": {
+          "doc_values": true,
+          "fields": {
+            "raw": {
+              "doc_values": true,
+              "ignore_above": 256,
+              "index": "not_analyzed",
+              "type": "string"
+            }
+          },
+          "format": "yyyy-MM-dd'T'HH:mm:ss.SSSSSSZ||yyyy-MM-dd'T'HH:mm:ssZ||dateOptionalTime",
+          "index": "not_analyzed",
+          "type": "date"
+        },
+        "docker": {
+          "properties": {
+            "container_id": {
+              "doc_values": true,
+              "index": "not_analyzed",
+              "type": "string"
+            }
+          }
+        },
+        "file": {
+          "doc_values": false,
+          "fields": {
+            "raw": {
+              "doc_values": true,
+              "ignore_above": 256,
+              "index": "not_analyzed",
+              "type": "string"
+            }
+          },
+          "index": "analyzed",
+          "norms": {
+            "enabled": true
+          },
+          "type": "string"
+        },
+        "geoip": {
+          "dynamic": true,
+          "properties": {
+            "location": {
+              "type": "geo_point"
+            }
+          },
+          "type": "object"
+        },
+        "hostname": {
+          "doc_values": true,
+          "index": "not_analyzed",
+          "type": "string"
+        },
+        "ipaddr4": {
+          "doc_values": true,
+          "fields": {
+            "raw": {
+              "doc_values": true,
+              "ignore_above": 256,
+              "index": "not_analyzed",
+              "type": "string"
+            }
+          },
+          "index": "not_analyzed",
+          "norms": {
+            "enabled": false
+          },
+          "type": "ip"
+        },
+        "ipaddr6": {
+          "doc_values": true,
+          "index": "not_analyzed",
+          "type": "string"
+        },
+        "kubernetes": {
+          "properties": {
+            "host": {
+              "doc_values": true,
+              "index": "not_analyzed",
+              "norms": {
+                "enabled": true
+              },
+              "type": "string"
+            },
+            "namespace_id": {
+              "doc_values": true,
+              "index": "not_analyzed",
+              "norms": {
+                "enabled": true
+              },
+              "type": "string"
+            },
+            "namespace_name": {
+              "doc_values": true,
+              "index": "not_analyzed",
+              "norms": {
+                "enabled": true
+              },
+              "type": "string"
+            },
+            "pod_id": {
+              "doc_values": true,
+              "index": "not_analyzed",
+              "norms": {
+                "enabled": true
+              },
+              "type": "string"
+            },
+            "pod_name": {
+              "doc_values": true,
+              "index": "not_analyzed",
+              "norms": {
+                "enabled": true
+              },
+              "type": "string"
+            }
+          }
+        },
+        "level": {
+          "doc_values": true,
+          "index": "not_analyzed",
+          "type": "string"
+        },
+        "message": {
+          "doc_values": false,
+          "index": "analyzed",
+          "norms": {
+            "enabled": false
+          },
+          "type": "string"
+        },
+        "offset": {
+          "doc_values": true,
+          "index": "not_analyzed",
+          "type": "long"
+        },
+        "pid": {
+          "doc_values": true,
+          "index": "not_analyzed",
+          "type": "string"
+        },
+        "pipeline_metadata": {
+          "properties": {
+            "@version": {
+              "doc_values": true,
+              "index": "not_analyzed",
+              "type": "string"
+            },
+            "collector": {
+              "properties": {
+                "hostname": {
+                  "doc_values": true,
+                  "index": "not_analyzed",
+                  "type": "string"
+                },
+                "inputname": {
+                  "doc_values": true,
+                  "index": "not_analyzed",
+                  "type": "string"
+                },
+                "ipaddr4": {
+                  "doc_values": true,
+                  "fields": {
+                    "raw": {
+                      "doc_values": true,
+                      "ignore_above": 256,
+                      "index": "not_analyzed",
+                      "type": "string"
+                    }
+                  },
+                  "index": "not_analyzed",
+                  "norms": {
+                    "enabled": false
+                  },
+                  "type": "ip"
+                },
+                "ipaddr6": {
+                  "doc_values": true,
+                  "index": "not_analyzed",
+                  "type": "string"
+                },
+                "name": {
+                  "doc_values": true,
+                  "index": "not_analyzed",
+                  "type": "string"
+                },
+                "original_raw_message": {
+                  "doc_values": false,
+                  "fields": {
+                    "raw": {
+                      "doc_values": true,
+                      "ignore_above": 256,
+                      "index": "not_analyzed",
+                      "type": "string"
+                    }
+                  },
+                  "index": "analyzed",
+                  "type": "string"
+                },
+                "received_at": {
+                  "doc_values": true,
+                  "format": "yyyy-MM-dd'T'HH:mm:ss.SSSSSSZ||yyyy-MM-dd'T'HH:mm:ssZ||dateOptionalTime",
+                  "index": "not_analyzed",
+                  "type": "date"
+                },
+                "version": {
+                  "doc_values": true,
+                  "index": "not_analyzed",
+                  "type": "string"
+                }
+              }
+            },
+            "normalizer": {
+              "properties": {
+                "hostname": {
+                  "doc_values": true,
+                  "index": "not_analyzed",
+                  "type": "string"
+                },
+                "inputname": {
+                  "doc_values": true,
+                  "index": "not_analyzed",
+                  "type": "string"
+                },
+                "ipaddr4": {
+                  "doc_values": true,
+                  "fields": {
+                    "raw": {
+                      "doc_values": true,
+                      "ignore_above": 256,
+                      "index": "not_analyzed",
+                      "type": "string"
+                    }
+                  },
+                  "index": "not_analyzed",
+                  "norms": {
+                    "enabled": false
+                  },
+                  "type": "ip"
+                },
+                "ipaddr6": {
+                  "doc_values": true,
+                  "index": "not_analyzed",
+                  "type": "string"
+                },
+                "name": {
+                  "doc_values": true,
+                  "index": "not_analyzed",
+                  "type": "string"
+                },
+                "original_raw_message": {
+                  "doc_values": false,
+                  "fields": {
+                    "raw": {
+                      "doc_values": true,
+                      "ignore_above": 256,
+                      "index": "not_analyzed",
+                      "type": "string"
+                    }
+                  },
+                  "index": "analyzed",
+                  "type": "string"
+                },
+                "received_at": {
+                  "doc_values": true,
+                  "format": "yyyy-MM-dd'T'HH:mm:ss.SSSSSSZ||yyyy-MM-dd'T'HH:mm:ssZ||dateOptionalTime",
+                  "index": "not_analyzed",
+                  "type": "date"
+                },
+                "version": {
+                  "doc_values": true,
+                  "index": "not_analyzed",
+                  "type": "string"
+                }
+              }
+            },
+            "trace": {
+              "analyzer": "whitespace",
+              "doc_values": false,
+              "index": "analyzed",
+              "type": "string"
+            }
+          }
+        },
+        "rsyslog": {
+          "properties": {
+            "appname": {
+              "doc_values": true,
+              "index": "not_analyzed",
+              "type": "string"
+            },
+            "facility": {
+              "doc_values": true,
+              "index": "not_analyzed",
+              "type": "string"
+            },
+            "msgid": {
+              "doc_values": true,
+              "index": "not_analyzed",
+              "type": "long"
+            },
+            "protocol-version": {
+              "doc_values": true,
+              "index": "not_analyzed",
+              "type": "string"
+            },
+            "structured-data": {
+              "doc_values": true,
+              "index": "not_analyzed",
+              "norms": {
+                "enabled": false
+              },
+              "type": "string"
+            }
+          }
+        },
+        "service": {
+          "doc_values": true,
+          "index": "not_analyzed",
+          "type": "string"
+        },
+        "systemd": {
+          "properties": {
+            "k": {
+              "properties": {
+                "KERNEL_DEVICE": {
+                  "doc_values": true,
+                  "index": "not_analyzed",
+                  "type": "string"
+                },
+                "KERNEL_SUBSYSTEM": {
+                  "doc_values": true,
+                  "index": "not_analyzed",
+                  "type": "string"
+                },
+                "UDEV_DEVLINK": {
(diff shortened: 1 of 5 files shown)


lukas-vlcek:
[test]

elasticsearch/index_templates/com.redhat.viaq-openshift-project.template.json
```diff
@@ -0,0 +1,332 @@
...
+          }
+        }
+      }
+    }
+  },
+  "order": 11,
+  "settings": {
+    "index.refresh_interval": "5s"
+  },
+  "template": "project.*"
```
ewolinetz: does this need to be the actual value of the project name not just "project"?

elasticsearch/index_templates/com.redhat.viaq-openshift-project.template.json
```diff
@@ -0,0 +1,332 @@
...
+                "enabled": true
+              },
+              "type": "string"
+            }
+          }
+        }
+      }
+    }
+  },
+  "order": 11,
```
ewolinetz: higher order means it'll be executed later, correct?

t0ffel: higher order override lower order: https://www.elastic.co/guide/en/elasticsearch/reference/2.4/indices-templates.html

elasticsearch/index_templates/com.redhat.viaq-openshift-project.template.json
t0ffel: the index name supposed to be `project.<projectname>.<projectid>.YYYY.MM.DD` for tenant's index.

ewolinetz: ack, we probably need to make a similar change for the `.operations.*` template to be `.operations.YYYY.MM.DD` 

Do we also need to make fluentd changes then so we can match the template? or is that in another PR

richm: > Do we also need to make fluentd changes then so we can match the template? or is that in another PR

That's another PR (coming soon)

elasticsearch/run.sh
```diff
@@ -60,4 +69,72 @@ else
...
+}
+
+verify_or_add_index_templates() {
+	wait_for_port_open
+	# Try to wait for cluster become more stable before index template being pushed in.
+	# Give up on timeout and continue...
+	curl -v -s -k -X GET \
+		--cert ${secret_dir}admin-cert \
+		--key  ${secret_dir}admin-key \
+		"${ES_REST_BASEURL}/_cluster/health?wait_for_status=yellow&timeout=${max_time}s"
```
ewolinetz: what if the cluster is in 'red' because it cant recover for an index? this will time out even though the cluster is up and accepting connections

lukas-vlcek: I was trying to get more information about this situation here https://discuss.elastic.co/t/creating-index-template-anytime-regardless-cluster-health-status/56194 (it is bit confusing).
But let's think about this... if the cluster is red due to index data (recovering indices or the like) then there is a chance the index template is already there (may be upgrade scenario is exception). If the cluster is red and there is no data/indices in it then we are probably in trouble anyway.

This process is still far from ideal. I am for example not sure if we can currently avoid data being indexed into ES cluster before the index template is in place. In which case the index template will apply only to indices created bit later (next day?).

I think the issue here is that we try to handle initialization at individual node level, however, we in fact should init cluster only after it has been fully formed. It would be great if we had mechanism of blocking indexing into cluster until we can make sure things are correctly initialized. Not sure if this is possible now.

ewolinetz: I ran into a similar [inner] debate for the openshift-elasticsearch-plugin which does the ACL seeding for searchguard. Based on scenarios seen for a customer (where their cluster was taking a long time to recover and some indices had even failed to recover). I would still want the ACLs in place so they could go in as an admin and clean up/recover manually if need be.

In that case I changed my health check from a waitforyellow to just getting a response back from the cluster health api.

I think we would want to consider something similar here. If we haven't yet created the index templates, we would want to make sure we seed them as soon as the cluster is responsive in case a customer begins to send logs to it (via Fluentd) because they see that ES has completed loading its indices in. The cluster may be in a 'red' state at this time because it hadn't finished recovering but the customer may not be aware of this.  And we would want to make sure that the index templates were there before any new indices are created.

I think regarding the individual vs cluster claim isn't an issue if we're waiting for the cluster/node to be responsive. A non-master node would be checking to see if its going to be master before it starts to do recovery I thought. If so, I think it would be connected to the cluster before it was responsive to health checks.

elasticsearch/run.sh
```diff
@@ -60,4 +69,72 @@ else
...
+
+	shopt -s failglob
+	for template_file in /usr/share/elasticsearch/index_templates/*.json
+	do
+		template=`basename $template_file`
```
ewolinetz: can we also confirm that the versions match? we may want to push an updated template out in the future and we would want to make sure an updated version overlaps an older index template that we provided

lukas-vlcek: Index templates versioning and dealing with this is whole a different debate. We definitely want to address this in the future but I am not sure we have a clear strategy what we want to do about this. Things can be more complicated because clients may want to introduce their custom index templates or modify our templates...

@t0ffel @portante feel free to jump in if we already know, otherwise we can probably leave it as it is now and get back to this later.

elasticsearch/run.sh
lukas-vlcek: @ewolinetz I commented the `waitForYellow` check out. It will try to push index templates as soon as possible.

lukas-vlcek:
[test]

elasticsearch/run.sh
```diff
@@ -2,9 +2,22 @@
...
+retry=$RETRY_COUNT
+max_time=$(( RETRY_COUNT * RETRY_INTERVAL ))	# should be integer
+timeouted=false
+
+mkdir -p /elasticsearch/$CLUSTER_NAME
+secret_dir=/etc/elasticsearch/secret
+[ -f $secret_dir/searchguard-node-key ] && ln -s $secret_dir/searchguard-node-key /elasticsearch/$CLUSTER_NAME/searchguard_node_key.key
+[ -f $secret_dir/searchguard.key ] && ln -s $secret_dir/searchguard.key /elasticsearch/$CLUSTER_NAME/searchguard_node_key.key
+[ -f $secret_dir/keystore.password ] && export KEYSTORE_PASSWORD=$(cat $secret_dir/keystore.password)
+[ -f $secret_dir/truststore.password ] && export TRUSTSTORE_PASSWORD=$(cat $secret_dir/truststore.password)
```
richm: @ewolinetz should this be removed?

deployer/common
```diff
@@ -1 +1 @@
-Subproject commit 45bf993212cdcbab5cbce3b3fab74a72b851402e
+Subproject commit 472dac81ec41046f7245ce7a2cec0af584bdafc2
```
ewolinetz: I think this is going back to a previous commit... 45bf993 is the most up to date commit in origin-integration-common

elasticsearch/run.sh
ewolinetz: +1 this may be leftover from a previous rebase

deployer/common
richm: To get rid of this, do
```
rm -rf deployer/common
git checkout deployer/common
```

elasticsearch/run.sh
lukas-vlcek: Removed.

deployer/common
lukas-vlcek: Removed.

richm:
```
The ImageStream "centos" is invalid: spec.dockerImageRepository: Invalid value: "library/centos:7": the repository name may not contain a tag
[ERROR] PID 2171: hack/lib/cmd.sh:229: `return "${return_code}"` exited with status 1.
```
[test] again?

ewolinetz:
@richm I think this gets us past that (as my PR test is currently at the check log portion)
https://github.com/openshift/origin-aggregated-logging/pull/241/commits/807c033cea5709b509a8f2939a20f0de2fe94d6c

lukas-vlcek:
Error:
The ImageStream "centos" is invalid: spec.dockerImageRepository: Invalid value: "library/centos:7": the repository name may not contain a tag.
😕

lukas-vlcek:
[test]

richm:
@lukas-vlcek looks like something has changed in openshift that is causing the image with tag to fail.  This is fixed in https://github.com/openshift/origin-aggregated-logging/pull/241/commits/807c033cea5709b509a8f2939a20f0de2fe94d6c
once that merges, this PR should also pass tests

lukas-vlcek:
@richm so we are now waiting for #241 to be merged?

richm:
@lukas-vlcek yes

richm:
[test]

ewolinetz:
so that it can eventually be removed from the `openshift-elasticsearch-plugin` can we also add `.all` to the aliases for both of these templates?  Currently we have the logic to update this alias in a plugin and it shouldn't live there anymore.

Since we will be adding an alias for the projects index template also... I figured it would be appropriate to capture that as well.

lukas-vlcek:
@ewolinetz @richm Rebased and updated.

I added the following aliases:

- `.all` alias to both `operations` and `project` index templates
- `project.{index}` alias to `project` index template only

--

The following is not important for `openshift/origin-aggregated-logging`, it is rather headsup note to /me and @richm: Unfortunately, I was not able to re-generate ViaQ templates directly due to some Python issue (see https://github.com/ViaQ/elasticsearch-templates/issues/15) so I modified existing Openshift templates manually not to block this PR. Going forward we __need to fix this__!

lukas-vlcek:
[test]

richm:
This doesn't seem to help with the problem for projects, to allow you to search for projects without having to specify `project.`  For example:
```
$ oc exec logging-kibana- ... 'https://logging-es:9200/test.*/_search?size=10000'|python -mjson.tool
{
    "hits": {
        "hits": [],
        "max_score": 0.0,
        "total": 0
    },
}
```
returns nothing - but if I specify `project.test.*` I get all of the records:
```
$ oc exec logging-kibana- ... 'https://logging-es:9200/project.test.*/_search?size=10000'|python -mjson.tool
{
    "hits": {
        "hits": [
            {
                "_id": "AVd3FloVSD7WiOPKimBp",
                "_index": "project.test.a5a373d0-866c-11e6-8b01-fa163e2719e9.2016.09.29",
                "_score": 1.0,
...
        ],
        "max_score": 1.0,
        "total": 211
    },
}
```
So I guess I misunderstand what aliases give us and how they work.  At least, it doesn't seem as though they will help us with the openshift console ui, or upgrade backward compatability, or curator, or ...

Do I have this right?

lukas-vlcek:
I thought the main motivation was the Kibana. As it just needs to be pointed to all `project.*` indices and thanks to SG users will be allowed to only relevant project indices. @ewolinetz am I right?

lukas-vlcek:
Wait a sec. All (new) indices (i.e. those created according to Common Data Model) will already start with `project.` prefix, right? So they will now get alias in form of `project.project.<projectname>.<projectid>.YYYY.MM.DD` which is what we wanted?
If we wanted to create index aliases for old (pre-Common Data model) indices to make them available via `project.*` then we can not use aliases in index templates.

richm:
> All (new) indices (i.e. those created according to Common Data Model) will already start with project. prefix, right?

Right.  So `project.test.a5a373d0-866c-11e6-8b01-fa163e2719e9.2016.09.29` is the actual index name created by fluentd.

> So they will now get alias in form of project.project.<projectname>.<projectid>.YYYY.MM.DD which is what we wanted?

I don't think `project.project.$name` is useful.

> If we wanted to create index aliases for old (pre-Common Data model) indices to make them available via project.* then we can not use aliases in index templates.

There isn't a way to make `test.$uid.$date` an alias for `project.test.$uid.$date`?  Is there a way to make an alias like this: If I ask for an index that begins with `project.` or `.operations.` return the index.  Otherwise, return `project.$indexthatIaskedfor`.

lukas-vlcek:
So it seems we can drop the `project.{index}` alias from index templates. How about the `.all` alias, will this one still needed?

> There isn't a way to make `test.$uid.$date` an alias for `project.test.$uid.$date`? Is there a way to make an alias like this: If I ask for an index that begins with `project.` or `.operations.` return the index. Otherwise, return `project.$indexthatIaskedfor`.

Not within the `aliases` section in index templates.

Seems there has been already some discussions about adding functionality like this, see https://github.com/elastic/elasticsearch/pull/5180 or https://github.com/elastic/elasticsearch/issues/5359 but now users seem to workaround this by implementing simple cron jobs to periodically check for indices and add/modify index aliases as needed.

richm:
Ok.  We should go ahead and use the `project.` prefix everywhere - openshift-elasticsearch-plugin, openshift UI, and curator.  I guess you can remove the alias from the index template, unless it is good for something else.
@nhosoi ^^^

lukas-vlcek:
@richm @nhosoi Removed the `project.{index}` alias. Kept the `.all` alias in both index templates for now //cc @ewolinetz

ewolinetz:
@richm @lukas-vlcek if we're going to update the aliases to be '.all' and 'project.{index}' then we'd also need to update Fluentd to not prepend 'project.' to its indices. I thought that was the plan... this way we'd have an index `logging.myuuid.YYYY.MM.DD` and it would have an alias `project.logging.myuuid.YYYY.MM.DD` so that querying with either one would fetch the data...

richm:
@t0ffel Would that work?

t0ffel:
@richm @lukas-vlcek @ewolinetz the only concern is to distinguish user's vs op's indices in "template" field of the respective index template. If they can be distinguished clearly - works for me.

also in this case "project.$projectname..." alias is not needed at all.

richm:
Just to close the discussion above - we are going with `project.` as the prefix for container/k8s logs.  This will require changes to the openshift-elasticsearch-plugin as well as the openshift ui (View Archive link).  These PRs have already been submitted.

lukas-vlcek:
[test]

richm:
[test] - not sure why the tests are aborting

fluentd/Dockerfile
```diff
@@ -34,6 +34,7 @@ RUN mkdir -p ${HOME} && \
       systemd-journal \
       fluent-plugin-rewrite-tag-filter \
       fluent-plugin-secure-forward
+      --conservative --minimal-deps
```
richm: remove this line - I think this is causing the test failures

richm:
LGTM
@ewolinetz @jcantrill PTAL

jcantrill:
LGTM

elasticsearch/index_templates/com.redhat.viaq-openshift.template.json
```diff
@@ -0,0 +1,569 @@
...
+          "type": "string"
+        }
+      }
+    }
+  },
+  "order": 10,
+  "settings": {
+    "index.refresh_interval": "5s"
+  },
+  "template": "*"
```
ewolinetz: i incorrectly asked about this on #234 but what does this template give us over the ones that match on `.operations.\*` and `project.\*` ?

lukas-vlcek:
[test]

elasticsearch/run.sh
```diff
@@ -57,17 +66,76 @@ else
...
+		(( retry -= 1 ))
+		if (( retry == 0 )) ; then
+			timeouted=true
+            break
+		fi
+	done
+
+	if [ $timeouted = true ] ; then
+		echo -n "[timeout] "
+    else
```
portante: Looks like a mix of tabs and spaces, makes it a tad difficult to line up in review UIs.

elasticsearch/run.sh
```diff
@@ -57,17 +66,76 @@ else
...
+		response_code=$(curl -s -X HEAD \
+			--cacert $secret_dir/admin-ca \
+			--cert $secret_dir/admin-cert \
+			--key  $secret_dir/admin-key \
+			-w '%{response_code}' \
+			$ES_REST_BASEURL/_template/$template)
+		if [ $response_code == "200" ]; then
+			echo "Index template '$template' already present in ES cluster"
+		else
+			echo "Create index template '$template'"
```
portante: If the `response_code` is a `404` then perhaps we should try to create the index template, but if it is a `5xx` code, then we might want to report the error.

elasticsearch/run.sh
```diff
@@ -57,17 +66,76 @@ else
...
+			curl -v -s -X PUT \
```
portante: Shouldn't we also check the `response_code` from this `curl` command to make sure it was applied successfully?

elasticsearch/run.sh
```diff
@@ -57,17 +66,76 @@ else
...
+# Wait for Elasticsearch port to be opened. Fail on timeout or if response from Elasticsearch is unexpected.
+wait_for_port_open() {
+	rm -f $LOG_FILE
     # test for ES to be up first and that our SG index has been created
-		result=$(curl --cacert $secret_dir/admin-ca --cert $secret_dir/admin-cert --key $secret_dir/admin-key -s -w "%{http_code}" -XGET "https://localhost:9200/.searchguard.${HOSTNAME}" -o /dev/null) ||:
-    [[ $result -eq 200 ]] && return 0
-    sleep 1
-  done
+	echo -n "Checking if Elasticsearch is ready on $ES_REST_BASEURL "
+	while ! response_code=$(curl -s \
```
richm: Not sure, but I'm assuming if the port is not opened, there will be no response_code, because there won't be an actual response, because the request didn't go through, so curl will exit with a non-zero code, which will cause the first part of the `||` condition to fail.  So it will sleep and retry.

elasticsearch/run.sh
richm: I suppose, but if the code is not 200 (404 or 5xx or whatever) it will attempt to add the template, which will then fail, assuming the previous condition was not 404, or perhaps it will fail anyway if there is some other problem, and that will be reported.

elasticsearch/run.sh
richm: If the curl fails for any reason, the error will be reported by curl, and the script will exit with an error condition (because the script does `set -e`)

elasticsearch/index_templates/com.redhat.viaq-openshift-operations.template.json
```diff
@@ -0,0 +1,569 @@
...
+          "type": "string"
+        }
+      }
+    }
+  },
+  "order": 10,
+  "settings": {
+    "index.refresh_interval": "5s"
+  },
+  "template": ".operations.*"
```
portante: It appears that the only difference between these two templates is that one has `.operations.*` as the pattern and the other has `project.*`.  Seems like that duplication is going to bite us.

Couldn't we have one base template that applies to all indices, and then add ones specific to a particular index?

richm: How?  Note that, due to the fact that we have to deal with both kibana with `.kibana*` indices, and searchguard with `.searchguard*` indices, we cannot have a `"template":"*"`.  I tried it, it caused many strange problems.

portante: Then we might want to consider having a `mapping.json` file, and then assembling the template by writing out that specific template with the mapping pulled from the single file.  This is fairly easy to do since it is all JSON.  A simple python program would have the dictionary containing the skeleton of the template, reads in the `mapping.json` file, and then writes out via json.dumpf the full template.

elasticsearch/run.sh
richm: no. qed
```
$ curl -s -w '%{response_code}' -X PUT -d junk http://www.google.com
... bunch of error messages ...
405
$ echo $?
0
```
so curl returned a 405 response_code, but the curl command returned 0 to the shell.

elasticsearch/index_templates/com.redhat.viaq-openshift-operations.template.json
richm: ok - https://github.com/openshift/origin-aggregated-logging/issues/259

richm:
@portante Have your concerns been addressed?  Can we merge this PR?  We will fix #258 and #259 before Nov. 11.

richm:
[merge]

richm:
[merge]

PR #223, opened 2016-09-07
[WIP] Enable optimizing and closing indices for curator

ewolinetz:
`close` has the same parameters [1] as `delete` and `optimize` has a few others [2], but the default seems to be okay for now. 

[1] https://www.elastic.co/guide/en/elasticsearch/client/curator/3.5/close.html
[2] https://www.elastic.co/guide/en/elasticsearch/client/curator/3.5/optimize.html

tests will be needed

### DIFF
--- a/curator/run_cron.py
+++ b/curator/run_cron.py
@@ -42,9 +42,9 @@ if not curlvl in logging._levelNames:
 #allowed_units = {'hours': 'hours', 'days': 'days', 'weeks': 'weeks', 'months': 'months'}
 allowed_units = {'days': 'days', 'weeks': 'weeks', 'months': 'months'}
 
-# allowed operations, currently we'll just allow delete
-allowed_operations = {'delete': 'delete'}
-curator_settings = {'delete': {}}
+# allowed operations, currently we'll just allow delete, close, and optimize
+allowed_operations = {'delete': 'delete', 'close': 'close', 'optimize': 'optimize'}
+curator_settings = {'delete': {}, 'close': {}, 'optimize': {}}
 
 filename = os.getenv('CURATOR_CONF_LOCATION', '/etc/curator/settings/config.yaml')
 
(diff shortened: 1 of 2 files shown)


curator/run_cron.py
```diff
@@ -42,9 +42,9 @@
 #allowed_units = {'hours': 'hours', 'days': 'days', 'weeks': 'weeks', 'months': 'months'}
 allowed_units = {'days': 'days', 'weeks': 'weeks', 'months': 'months'}
 
-# allowed operations, currently we'll just allow delete
-allowed_operations = {'delete': 'delete'}
-curator_settings = {'delete': {}}
+# allowed operations, currently we'll just allow delete, close, and optimize
+allowed_operations = {'delete': 'delete', 'close': 'close', 'optimize': 'optimize'}
```
portante: Why does this need to be a dictionary?

richm: Because that's the simplest way to create a lookup table.  Do you have a better suggestion?

deployer/templates/curator.yaml
```diff
@@ -134,9 +138,17 @@ parameters:
   value: "/etc/curator/keys/ca"
 -
   description: "(Deprecated) The default number of days that logs will be retained for a project if an index management configuration for it is not provided"
-  name: CURATOR_DEFAULT_DAYS
+  name: CURATOR_DEFAULT_DELETE_DAYS
   value: "30"
 -
+  description: "(Deprecated) The default number of days that logs will be left open for a project if an index management configuration for it is not provided"
```
jcantrill: Are these deprecated because we prefer to use configmaps over env vars?

curator/run_cron.py
portante: So we are looking up `delete` to know it is `delete`?  I am not following why this is needed?

richm: We are looking up `delete` to see if it is in the table.   The value doesn't have to be the same as the key.  I guess it would be more clear if the value for every key was `True`.

deployer/templates/curator.yaml
ewolinetz: this was more following what we already had, we can probably remove this one and the default for delete TBH

jcantrill:
@ewolinetz any reason not to kickstart this PR and merge?  looks like it would require some changes to also be made to our ansible work.

ewolinetz:
@jcantrill it was put on hold and doesn't currently work

jcantrill:
[(3) Have curator optimize yesterday's indices [logging][epic-ois-agl-perf]](https://trello.com/c/UlmNQ0Av/381-3-have-curator-optimize-yesterdays-indices-loggingepic-ois-agl-perf)

jcantrill:
Closing as bit rot since curator has changed in the most recent releases.

Issue #226, opened 2016-09-08
Increase grace termination period for ES and Fluentd


ewolinetz:
ES and Fluentd do a bit of tasks when closing up shop, and they may end up in a bad state/result in loss of data if they are killed before they are able to complete.

E.g.

* ES removes its lock files and writes its current trans logs out [1]
* Fluentd finishes processing its currently read in records and pushes them to ES

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1371220

ewolinetz:
resolved with #227

PR #227, opened 2016-09-08
Increasing terminationGracePeriod for ES and Fluentd to allow them to…

ewolinetz:
… do their clean up before being killed

To address #226

### DIFF
--- a/deployer/templates/es.yaml
+++ b/deployer/templates/es.yaml
@@ -41,6 +41,7 @@ objects:
             component: ${ES_CLUSTER_NAME}
             deployment: ${ES_DEPLOYMENT_NAME}
         spec:
+          terminationGracePeriodSeconds: 600
           serviceAccountName: aggregated-logging-elasticsearch
           securityContext:
             supplementalGroups:
--- a/deployer/templates/fluentd.yaml
+++ b/deployer/templates/fluentd.yaml
@@ -52,6 +52,7 @@ objects:
             provider: openshift
             component: "fluentd"
         spec:
+          terminationGracePeriodSeconds: 300
           serviceAccountName: aggregated-logging-fluentd
           containers:
           - name: fluentd-elasticsearch


ewolinetz:
@sosiouxme @jcantrill PTAL

deployer/templates/es.yaml
```diff
@@ -41,6 +41,7 @@ objects:
             component: ${ES_CLUSTER_NAME}
             deployment: ${ES_DEPLOYMENT_NAME}
         spec:
+          terminationGracePeriodSeconds: 1800
```
jcantrill: Should we be pulling these in via parameter and/or configmap? to allow them to be tweaked in production.  I realize you can edit but then the settings would disappear on re-install

ewolinetz: We could make this tunable with parameters for the deployer; it wouldn't work for a configmap since this is in the pod spec itself. 
We are just looking to give ES and Fluentd more time to complete their shutdown operations so this is to increase that window of time -- so I'd be cautious of letting users just change this on the fly since that could end up putting ES/Fluentd in a bad state (if they set this to `0` then there is no grace period -- it will SIGKILL right away instead of SIGTERM)

sosiouxme: I think half an hour may be a little over the top... If it's taking that kind of time it's probably hung anyway. What do you think about 10 minutes?

ewolinetz: We can drop it down to 10 minutes. I was planning for a worst case where they have a ton of indices that the metadata needed to be written out for and they could always oc delete the pod if its taking too long. Since we don't have actual performance number on this it was kind of a guess

ewolinetz:
dropped down to 10 minutes for ES

richm:
LGTM

deployer/templates/es.yaml
sosiouxme: oc delete pod goes through the same termination grace period, there's no
"force-delete" other than kill -9 on the PID. 10 is good, LGTM

ewolinetz:
[merge]

PR #230, opened 2016-09-12
Creating initial docs/proposals structure and first draft fluentd-sec…

ewolinetz:
…ure-forward proposal

@jcantrill @sosiouxme @richm @lukas-vlcek

### DIFF
--- a/docs/proposals/fluentd-secure-forward.md
+++ b/docs/proposals/fluentd-secure-forward.md
@@ -1,0 +1,115 @@
+# Allow Fluentd to secure forward to other Fluentd/Logstash
+
+## Why?
+
+* Currently we allow forwarding just to another unsecure Elasticsearch
+* No easy way to ship off Fluentd collected logs for OSE to other log aggregators
+* More simple to support single plugin to do this than adding a number of Fluentd output plugins
+
+## How?
+
+Ship Fluentd image with [fluent-plugin-secure-forward](https://github.com/tagomoris/fluent-plugin-secure-forward)
+and provide a means to configure it with configmaps/environment variables.
+
+Dockerfile:
+```
+RUN mkdir -p ${HOME} && \
+    gem install --no-rdoc --no-ri \
+      --conservative --minimal-deps \
+      fluentd:${FLUENTD_VERSION} \
+      'activesupport:<5' \
+      fluent-plugin-kubernetes_metadata_filter \
+      fluent-plugin-elasticsearch \
+      fluent-plugin-flatten-hash \
+      fluent-plugin-systemd \
+      systemd-journal \
+-     fluent-plugin-rewrite-tag-filter
++     fluent-plugin-rewrite-tag-filter \
++     fluent-plugin-secure-forward
+```
+
+configs.d/openshift/output-applications.conf:
+```
+<match **>
+   @type copy
+   @include output-es-config.conf
+   @include ../user/output-extra-*.conf
+   @include ../dynamic/es-copy-config.conf
++  @include ../user/secure-forward-config.conf
+</match>
+```
+
+configs.d/openshift/output-operations.conf:
+```
+<match journal.system** system.var.log** **_default_** **_openshift_** **_openshift-infra_**>
+  @type copy
+  @include output-es-ops-config.conf
+  @include ../user/output-ops-extra-*.conf
+  @include ../dynamic/es-ops-copy-config.conf
++ @include ../user/secure-forward-config.conf
+</match>
+```
+
+example secure-forward-config.conf:
+```
+@type secure_forward
+
+  self_hostname ${HOSTNAME}
+  shared_key    secret_string
+
+  secure yes
+  # enable_strict_verification yes
+
+  ca_cert_path /path/for/certificate/ca_cert.pem
+  ca_private_key_path /path/for/certificate/ca_key.pem
+  ca_private_key_passphrase passphrase # for private CA secret key
+
+  <server>
+    host server.fqdn.example.com  # or IP
+    # port 24284
+  </server>
+  <server>
+    host 203.0.113.8 # ip address to connect
+    hostlabel server.fqdn.example.com # specify hostlabel for FQDN verification if ipaddress is used for host
+  </server>
+```
+
+## Proposal
+
+Begin to deprecate `ES_COPY` which allows us to simply forward a copy of our logs
+to another, unsecured, Elasticsearch cluster. We will provide a way for customers
+to configure and use the `Secure Forward` Fluentd plugin to enable them to integrate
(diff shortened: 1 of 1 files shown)


docs/proposals/fluentd-secure-forward.md
```diff
@@ -0,0 +1,35 @@
...
+# Limitations
+
+The customer would need to have something that can talk with the Fluentd plugin.
+
+* For Fluentd that means using the Secure Forward input plugin.
+* For Logstash, it looks like it has a Fluent codec [1] that may be able to be
+used in conjunction with secure TCP configurations [2].
+* With the removing of the `ES_COPY` features; we would need to provide a simple path
+'upgrade' to using either secure-forward if they were simply using an external
+ES cluster as a backup (e.g. no other logs collected there)
```
richm: I don't think we can just remove ES_COPY as there may be customers using it.  We should support secure_forward _and_ ES_COPY, or provide some sort of deprecation period.

docs/proposals/fluentd-secure-forward.md
```diff
@@ -0,0 +1,35 @@
...
+to configure and use the `Secure Forward` Fluentd plugin to enable them to integrate
+logs from OSE into their current log aggregation solution.
+
+# Limitations
+
+The customer would need to have something that can talk with the Fluentd plugin.
+
+* For Fluentd that means using the Secure Forward input plugin.
+* For Logstash, it looks like it has a Fluent codec [1] that may be able to be
+used in conjunction with secure TCP configurations [2].
```
richm: What about Splunk?

docs/proposals/fluentd-secure-forward.md
jcantrill: Is it possible for these to co-exist?  Could we provide a switch to enable one or the other and deprecate ES Copy?

docs/proposals/fluentd-secure-forward.md
ewolinetz: It looks like Fluentd has two splunk output plugins [1] [2]  I'd imagine a customer would set up a Fluentd instance that connected to their Splunk instance and would set up the OSE Fluentd to forward to that Fluentd instance.

[1] https://github.com/parolkar/fluent-plugin-splunk
[2] https://github.com/gtrevg/fluent-plugin-splunk-ex

docs/proposals/fluentd-secure-forward.md
ewolinetz: What if we use a configmap to configure either setting, and if both are enabled then we would use both.  We should call out that we will be depricating `ES_COPY` in the docs.

The forwarder config could then be defined/called in the same location as the `ES_COPY` configs

richm: Ok.

docs/proposals/fluentd-secure-forward.md
```diff
@@ -0,0 +1,35 @@
+# Allow Fluentd to secure forward to other Fluentd/Logstash (over ES-Copy)
```
jcantrill: We should mention in this document something about changes to the deployer for cases where only fluentd will be deployed to support an external aggregation (e.g. Splunk).  Are we going to have single flag?  Maybe there is a 'history' captured in the configmap which might allow something like diagnostics to only evaluate the relevant components?

ewolinetz: How automated do we need to make this do you think?  I think if we were going to have a flag it would have fluentd not send to an internal ES at all and only would forward / ES_COPY. And we could look at the configurations to know which one we were going to actually use?

Does this make sense to be in a configmap since Fluentd already uses a configmap for its normal configuration and we could just as easily describe how to update that?

jcantrill: Ultimately it depends on what our fluentd needs to be functional.  At a minimum, it needs permissions to read from the host right?  Isn't that at least a serviceaccount and maybe some policy?  I can envision a customer wanting a simplified experience  for deploying just fluentd with the forwarding option and nothing else.  I would imagine they would prefer to run single command and/or template instead of having to execute various instructions.  Couldn't this be captured as a mode?  There is the caveat also of "I want everything else as well fluentd configured in THIS way..." which means that 'autodeploy' requires several switches.   Maybe its a matter of 'listing' the name of the plugins to add using a well defined name (e.g fluentd_options=es_copy,secure_forward)

ewolinetz: I don't think we would change anything with how fluentd is installed; and I'm not sure if we'd want to have a mode that would just install fluentd for scenarios like this... but it sounds like that's what you're thinking here @jcantrill ?

How would we do a normal upgrade for that mode?

jcantrill: @ewolinetz The only upgrade it seems you could do is to update the base component and its associated plugins for version.  Upgrade with have to be smart enough to determine start and end for fluents only which is kind of what is being requested of the admin command @sosiouxme is designing.

docs/proposals/fluentd-secure-forward.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Why?
+
+* Currently we allow forwarding just to another unsecure Elasticsearch
+* No easy way to ship off Fluentd collected logs for OSE to other log aggregators
+* More simple to support single plugin to do this than adding a number of Fluentd output plugins
+
+## How?
+
+Ship Fluentd image with (https://github.com/tagomoris/fluent-plugin-secure-forward)[fluent-plugin-secure-forward]
```
jcantrill: formatting of link is off here.

ewolinetz: :+1: couldn't remember if its hard brackets first or not

docs/proposals/fluentd-secure-forward.md
ewolinetz: I think i meant more, we would need to split the `upgrade` mode up based on whether or not this was a deployment where only fluentd is used. And was wondering if you'd imagine us still using a single `upgrade` mode, or also have one specific for a solo fluentd deployment?

We currently try to detect the current stack version based on what features are installed. If its a solo fluentd we wouldn't have any other components and the deployer would think its at a version before curator was introduced since it would be absent... 

I guess we could add a label/annotation on the fluentd daemonset to indicate to ourselves that it is solo and if it is absent then we know its a prior version..

ewolinetz:
any further comments on this? if not id like to merge this and begin implementation...

richm:
LGTM

docs/proposals/fluentd-secure-forward.md
ewolinetz: Capturing this in the Additional Considerations section

ewolinetz:
[merge]

PR #234, opened 2016-09-14
Nhosoi fluentd common data model

nhosoi:


richm:
[test]

richm:
@ewolinetz @sosiouxme @lukas-vlcek ptal

deployer/common
```diff
@@ -1 +1 @@
-Subproject commit 45bf993212cdcbab5cbce3b3fab74a72b851402e
+Subproject commit 472dac81ec41046f7245ce7a2cec0af584bdafc2
```
ewolinetz: I think this is going back to using an older commit

deployer/scripts/uuid_migrate.sh
```diff
@@ -88,7 +91,7 @@ function run_uuid_migration() {
     [[ "${OPS_PROJECTS[@]}" =~ "$index" ]] && continue
 
     uid=$(oc get project "$index" -o jsonpath='{.metadata.uid}')
-    create_alias $index $uid
+    create_alias ${PROJ_PREFIX}$index $uid
```
ewolinetz: to start a discussion around this... should we be updating the uuid_migrate script? we were only running it to get users back up to the state where their indices have a UUID in them.  I think we should discuss if we can use an index template alias instead for all other indices as we move forward

hack/testing/test-upgrade.sh
```diff
@@ -227,10 +227,11 @@ function verifyUpgrade() {
   [[ -z "$(oc get secret/logging-elasticsearch -o jsonpath='{.data.admin-cert}')" ]] && return 1
 
 ### check that migration was successful
+  PROJ_PREFIX=project.
   if [ $checkMigrate = true ]; then
     for project in $(oc get projects -o 'jsonpath={.items[*].metadata.name}'); do
       [[ "${OPS_PROJECTS[@]}" =~ $project ]] && continue
-      [[ -z "$(oc logs $UPGRADE_POD | grep 'Migration for project '$project': {"acknowledged":true}')" ]] && return 1
+      [[ -z "$(oc logs $UPGRADE_POD | grep 'Migration for project '${PROJ_PREFIX}$project': {"acknowledged":true}')" ]] && return 1
```
ewolinetz: This may not change pending discussion

fluentd/configs.d/openshift/filter-syslog-record-transform.conf
```diff
@@ -19,19 +19,31 @@
...
     hostname ${_HOSTNAME.eql?('localhost') ? (begin; File.open('/etc/docker-hostname') { |f| f.readline }.rstrip; rescue; _HOSTNAME; end) : _HOSTNAME}
+    ipaddr4 "#{ENV['IPADDR4']}"
+    ipaddr6 "#{ENV['IPADDR6']}"
     # if the field name begins with an uppercase letter, or might not exist, you have to use the record["FIELDNAME"] style
     time ${(record["_SOURCE_REALTIME_TIMESTAMP"] || record["__REALTIME_TIMESTAMP"]) ? Time.at((record["_SOURCE_REALTIME_TIMESTAMP"] || record["__REALTIME_TIMESTAMP"]).to_f / 1000000.0).to_datetime.rfc3339(6) : time.to_datetime.rfc3339(6)}
     # if field is optional you have to use record[name] or (name rescue nil) or something like that
     ident ${record["SYSLOG_IDENTIFIER"] || record["_COMM"]}
     message ${record["MESSAGE"]}
-    version 1.3.0
+    version 2.0.0
```
ewolinetz: The version we were using was tracking to the version of origin this would match up with; can we make this 1.4.0 if this is planned for 3.4.0 enterprise?

fluentd/configs.d/openshift/filter-k8s-record-transform.conf
```diff
@@ -1,28 +1,33 @@
...
+    # Needs to set a field inside a hash.
+    # Plus we'd better avoid the file IO.
+    hostname ${(record['kubernetes']['host'] rescue nil) || File.open('/etc/docker-hostname') { |f| f.readline }.rstrip}
     # log is set if MESSAGE is JSON valued and has err or msg field
     message ${record['MESSAGE'] || log}
-    version 1.3.0
+    level ${record['PRIORITY']}
+    ipaddr4 "#{ENV['IPADDR4']}"
+    ipaddr6 "#{ENV['IPADDR6']}"
+    version 2.0.0
```
ewolinetz: 1.4.0

fluentd/configs.d/openshift/filter-k8s-record-transform.conf
```diff
@@ -1,28 +1,33 @@
...
-
 <filter kubernetes.var.log.containers**>
-  type record_transformer
+  @type record_transformer
   enable_ruby
   <record>
-    hostname ${(kubernetes_host rescue nil) || File.open('/etc/docker-hostname') { |f| f.readline }.rstrip}
+    hostname ${(record['kubernetes']['host'] rescue nil) || File.open('/etc/docker-hostname') { |f| f.readline }.rstrip}
     message ${(message rescue nil) || log}
     version 1.3.0
```
ewolinetz: 1.4.0 simply for the fact that it seems we aren't going to be flattening records anymore.

fluentd/Dockerfile
```diff
@@ -14,25 +14,27 @@ LABEL io.k8s.description="Fluentd container for collecting of docker container l
...
     yum clean all
 RUN mkdir -p ${HOME} && \
-    gem install --no-rdoc --no-ri \
-      --conservative --minimal-deps \
+    gem install -N --conservative --minimal-deps \
       fluentd:${FLUENTD_VERSION} \
       'activesupport:<5' \
       fluent-plugin-kubernetes_metadata_filter \
       fluent-plugin-elasticsearch \
       fluent-plugin-flatten-hash \
```
ewolinetz: if we aren't flattening the hash anymore can this be removed?

elasticsearch/run.sh
```diff
@@ -2,9 +2,22 @@
...
+RETRY_COUNT=300		# how many times
+RETRY_INTERVAL=1	# how often (in sec)
+
+retry=$RETRY_COUNT
+max_time=$(( RETRY_COUNT * RETRY_INTERVAL ))	# should be integer
+timeouted=false
+
+mkdir -p /elasticsearch/$CLUSTER_NAME
+secret_dir=/etc/elasticsearch/secret
+[ -f $secret_dir/searchguard-node-key ] && ln -s $secret_dir/searchguard-node-key /elasticsearch/$CLUSTER_NAME/searchguard_node_key.key
```
ewolinetz: do we need these sym links since we aren't using them in the es config?

richm:
[test]
etcd

elasticsearch/run.sh
```diff
@@ -57,17 +70,76 @@ else
...
+
+verify_or_add_index_templates() {
+	wait_for_port_open
+	# Try to wait for cluster become more stable before index template being pushed in.
+	# Give up on timeout and continue...
+	curl -v -s -X GET \
+        --cacert $secret_dir/admin-ca \
+		--cert $secret_dir/admin-cert \
+		--key  $secret_dir/admin-key \
+		"$ES_REST_BASEURL/_cluster/health?wait_for_status=yellow&timeout=${max_time}s"
```
ewolinetz: what if the cluster is in 'red' because it cant recover for an index? this will time out even though the cluster is up and accepting connections

elasticsearch/run.sh
richm: This is really for https://github.com/openshift/origin-aggregated-logging/pull/222

elasticsearch/run.sh
richm: This is for https://github.com/openshift/origin-aggregated-logging/pull/222

deployer/scripts/uuid_migrate.sh
nhosoi: Hi, to follow your comments, I undo'ed this change as well as the one in hack/testing/test-upgrade.sh.  Then, I reran the test using Rich's script oshift-aws-dev.sh.  But it fails in hack/testing/test-upgrade.sh.  To make this change, could there be any prerequisites?  Am I supposed to pick up some patch?  Thanks!

ewolinetz: What was the failure? Those should be the only changes for undoing this scenario

nhosoi: Thank you for asking!   

This is the end of the output which shows the failure.
http://pastebin.test.redhat.com/412553

It seems it fails at the line 234 in "hack/testing/test-upgrade.sh".
234       [[ -z "$(oc logs $UPGRADE_POD | grep 'Migration for project '$project': {"acknowledged    ":true}')" ]] && return 1

ewolinetz: Yeah, it looks like it didn't find that record...  `++ [[ -z '' ]]`
I'd be curious what the logs looked like in the deployer and what log indices were in that ES instance..

It could be that the uuid migration didn't run successfully...

deployer/common
richm: you can
```
rm -rf deployer/common
git checkout deployer/common
```
to get rid of that

deployer/scripts/uuid_migrate.sh
richm: right - because the index name is now `project.$projectname.$uuid`

ewolinetz: I'm not sure if this is the result of the github layout but it looks like these changes are still in for the uuid_migrate.sh.
`create_alias` takes in the old format project name and its uuid, if we are sending it "project.$projectname" $uid it won't be able to find that project to create an alias for.

Unless we are planning on reusing the uuid_migrate script to migrate again to a new format, I dont think the changes to this file should be kept.

https://github.com/nhosoi/origin-aggregated-logging/blob/cb32d37d2f10d2ccd3f8cdacbacbabf4de924957/deployer/scripts/uuid_migrate.sh#L23

fluentd/configs.d/openshift/filter-syslog-record-transform.conf
nhosoi: All right.  Changing it to 1.4.0.
BTW, I searched "1.2.0" and found these.   Do we want to change them, as well?
find . -type f | xargs egrep "1\.2\.0" | egrep -v artifacts
./deployer/README.md:* `IMAGE_PREFIX`: Specify the prefix for logging component images; e.g. for "docker.io/openshift/origin-logging-deployer:v1.2.0", set prefix "docker.io/openshift/origin-"
./deployer/README.md:* `IMAGE_VERSION`: Specify version for logging component images; e.g. for "docker.io/openshift/origin-logging-deployer:v1.2.0", set version "v1.2.0"
./deployer/README.md:               -p IMAGE_VERSION=v1.2.0 \
./deployer/Dockerfile:FROM openshift/origin:v1.2.0
./hack/templates/dev-builds.yaml:    - name: v1.2.0
./hack/templates/dev-builds.yaml:        name: openshift/origin:v1.2.0
./hack/templates/dev-builds.yaml:          name: origin:v1.2.0

deployer/scripts/uuid_migrate.sh
richm:
```
function create_alias() {
    local output=""
    output=$(curl -s --cacert $CA --key $KEY --cert $CERT -XPOST "https://$es_host:$es_port/_aliases" -d "{ \"actions\": [ { \"add\": { \"index\": \"${1}.*\", \"alias\": \"${1}.${2}.reference\"}} ] }")

    echo Migration for project $1: $output
}
...
    uid=$(oc get project "$index" -o jsonpath='{.metadata.uid}')
    create_alias project.$index $uid
```
There are two cases here:
- upgrade from 1.old to 1.4 - this means they will need to move index `$projname.YYYY.MM.DD` to `project.$projname.$projuuid.YYYY.MM.DD`
- upgrade from 1.newer to 1.4 - this means they will need to move index `$projname.$projuuid.YYYY.MM.DD` to `project.$projname.$projuuid.YYYY.MM.DD`

Do we care about either of these cases?  If so, do we need to handle them with uuid_migrate.sh, or with index template alias definitions, or ???

fluentd/Dockerfile
nhosoi: This means "fluent-plugin-flatten-hash" or this diff?

richm: flatten-hash - we don't need it any more - just get rid of it

hack/testing/test-upgrade.sh
ewolinetz: Per our discussion yesterday, we won't be needing to make this change

deployer/scripts/uuid_migrate.sh
ewolinetz: Per our discussion yesterday, we won't be needing to make any changes for UUID migration -- we will have aliases that match the current index format added to our index template to map the new index formats to.

fluentd/configs.d/openshift/filter-syslog-record-transform.conf
ewolinetz: no, there isn't a need to change those here. those are defaults that I believe we are overwriting... if we aren't they should be taken care of independent of this PR

fluentd/Dockerfile
ewolinetz: how do we query in kibana with hierarchical data?

richm: instead of `"kubernetes_pod_name":"logging-es-1"` you use `"kubernetes.pod_name":"logging-es-1"`

fluentd/configs.d/openshift/filter-k8s-record-transform.conf
```diff
@@ -1,30 +1,35 @@
...
+    # The index mapping has a field record['kubernetes']['hostname'];
+    # it comes from record['kubernetes']['host'] or File.open('/etc/docker-hostname')
+    # Needs to set a field inside a hash.
+    # Plus we'd better avoid the file IO.
+    hostname ${(record['kubernetes']['host'] rescue nil) || File.open('/etc/docker-hostname') { |f| f.readline }.rstrip}
     # log is set if MESSAGE is JSON valued and has err or msg field
     message ${record['MESSAGE'] || log}
-    version 1.3.0
+    level ${record['PRIORITY']}
+    ipaddr4 "#{ENV['IPADDR4']}"
```
t0ffel: Where does the IPDADDR4 come from?
When I don't have this var set - the filter just stops working silently.

fluentd/configs.d/openshift/filter-k8s-record-transform.conf
```diff
@@ -1,30 +1,35 @@
...
+    ipaddr6 "#{ENV['IPADDR6']}"
```
t0ffel: Where does IPADDR6 come from?

fluentd/configs.d/openshift/filter-k8s-record-transform.conf
```diff
@@ -1,30 +1,35 @@
...
+    version 1.4.0
```
t0ffel: What does the 'version' field mean?
if that's the version of the collector it should go under pipeline_metadata.collector, right now it is on the top level

fluentd/configs.d/openshift/filter-syslog-record-transform.conf
```diff
@@ -13,25 +13,37 @@
...
+    # it comes from record['kubernetes']['host'] or File.open('/etc/docker-hostname')
+    # Needs to set a field inside a hash.
+    # Plus we'd better avoid the file IO.
     hostname ${_HOSTNAME.eql?('localhost') ? (begin; File.open('/etc/docker-hostname') { |f| f.readline }.rstrip; rescue; _HOSTNAME; end) : _HOSTNAME}
+    ipaddr4 "#{ENV['IPADDR4']}"
+    ipaddr6 "#{ENV['IPADDR6']}"
     # if the field name begins with an uppercase letter, or might not exist, you have to use the record["FIELDNAME"] style
     time ${(record["_SOURCE_REALTIME_TIMESTAMP"] || record["__REALTIME_TIMESTAMP"]) ? Time.at((record["_SOURCE_REALTIME_TIMESTAMP"] || record["__REALTIME_TIMESTAMP"]).to_f / 1000000.0).to_datetime.rfc3339(6) : time.to_datetime.rfc3339(6)}
     # if field is optional you have to use record[name] or (name rescue nil) or something like that
     ident ${record["SYSLOG_IDENTIFIER"] || record["_COMM"]}
```
t0ffel: what's `ident` field?

fluentd/configs.d/openshift/filter-k8s-record-transform.conf
```diff
@@ -1,30 +1,35 @@
...
     time ${Time.at((record["_SOURCE_REALTIME_TIMESTAMP"] || record["__REALTIME_TIMESTAMP"]).to_f / 1000000.0).to_datetime.rfc3339(6)}
```
t0ffel: should it be `@timestamp`

fluentd/configs.d/openshift/filter-syslog-record-transform.conf
richm: Is `ident` field in common data model?  If not, let's get rid of this line.

fluentd/configs.d/openshift/es-copy-config.conf
```diff
@@ -3,14 +3,17 @@
...
-      index_name ${record['kubernetes_namespace_name']}.${record['kubernetes_namespace_id']}.${Time.at(time).getutc.strftime(@logstash_dateformat)}
+      index_name project.${record['kubernetes']['namespace_name']}.${record['kubernetes']['namespace_id']}.${Time.at(time).getutc.strftime(@logstash_dateformat)}
       user "#{ENV['ES_COPY_USERNAME']}"
       password "#{ENV['ES_COPY_PASSWORD']}"
 
       client_key "#{ENV['ES_COPY_CLIENT_KEY']}"
       client_cert "#{ENV['ES_COPY_CLIENT_CERT']}"
       ca_file "#{ENV['ES_COPY_CA']}"
 
+      type_name ${(record && record.delete('es_type')) || 'fluentd'}
```
richm: Actually `type_name` is set where `es_type` is set here https://github.com/openshift/origin-aggregated-logging/pull/234/files#diff-5871732c8b4b021e8bebf8e5352d6deaR29

fluentd/configs.d/openshift/es-ops-copy-config.conf
```diff
@@ -11,6 +11,9 @@
       client_cert "#{ENV['OPS_COPY_CLIENT_CERT']}"
       ca_file "#{ENV['OPS_COPY_CA']}"
 
+      type_name ${(record && record.delete('es_type')) || 'fluentd'}
```
richm: type is set by `es_type` setting in the filter

fluentd/configs.d/openshift/filter-k8s-record-transform.conf
richm: IPDADDR4 this is set in the run.sh script.  If we can't assume the runner of fluentd will set IPADDR4, how should we calculate the value?

fluentd/configs.d/openshift/filter-k8s-record-transform.conf
richm: IPDADDR6 this is set in the run.sh script. If we can't assume the runner of fluentd will set IPADDR6, how should we calculate the value?

fluentd/configs.d/openshift/filter-k8s-record-transform.conf
richm: This `version` field is special to openshift - it denotes the version of openshift.

fluentd/configs.d/openshift/filter-k8s-record-transform.conf
richm: except that you can't have
```
    @timestamp something
```
in a record_transform filter.   We had to add a custom filter for this.  @nhosoi it looks like this commit still has the old hack for erasing empty fields and setting `@timestamp`.  Where is the code with the new `filter-common-data-model.rb` ?

fluentd/configs.d/openshift/es-ops-copy-config.conf
t0ffel: I think we can move to a static type in the `match`, the additional benefit of that is  - we won't have the extra field `es_type` in ES.

fluentd/configs.d/openshift/filter-k8s-record-transform.conf
t0ffel: ok.

fluentd/configs.d/openshift/filter-k8s-record-transform.conf
t0ffel: ok

fluentd/configs.d/openshift/filter-k8s-record-transform.conf
t0ffel: my concern is that I see that in the ElasticSearch document... not sure that it can be fixed, but it would be nice to get rid of that field and keep it only in `@timestamp`

fluentd/configs.d/openshift/filter-syslog-record-transform.conf
t0ffel: it's not. what is it supposed to mean here?

nhosoi: > Is ident field in common data model?
means if ident is in the templates or not?  If so, no, it is not.

fluentd/configs.d/openshift/filter-k8s-record-transform.conf
richm: @nhosoi Just this one extra commit: https://github.com/richm/origin-aggregated-logging/commit/e1f4aa7d6ccdc3a4d56ee825fc14f91031313563
```
 add common-data-model filter and tests
```
Add a fluentd filter plugin for the common data model processing.
It is too much code for a one-line ruby hack in the es plugin.
```
The other commits were just for me to test the cdm changes in the openshift elasticsearch plugin and openshift ui

File: fluentd/configs.d/openshift/es-ops-copy-config.conf (hunk 16 above)
Comment (richm) [inline fluentd/configs.d/openshift/es-ops-copy-config.conf]: Ok. Now that we have only 1 type, this makes sense.

File: fluentd/configs.d/openshift/filter-syslog-record-transform.conf (hunk 13 above)
Comment (richm) [inline fluentd/configs.d/openshift/filter-syslog-record-transform.conf]: Both `SYSLOG_IDENTIFIER` and `_COMM` are already in the common data model, so we can get rid of `ident`.  We'll have to change the tests that use `ident` to see if the record that was written to the logs has been picked up by fluentd and written to elasticsearch.

Comment (richm) [inline fluentd/configs.d/openshift/filter-syslog-record-transform.conf]: `systemd.t.COMM` and `systemd.u.SYSLOG_IDENTIFIER`

File: fluentd/configs.d/openshift/filter-k8s-record-transform.conf (hunk 14 above)
Comment (richm) [inline fluentd/configs.d/openshift/filter-k8s-record-transform.conf]: The common-data-model plugin adds `@timestamp` and gets rid of `time`.

Comment (nhosoi) [inline fluentd/configs.d/openshift/filter-k8s-record-transform.conf]: @richm Pushed your "add common-data-model filter and tests" patch.

File: fluentd/run.sh [hunk 17]
```diff
@@ -31,6 +31,10 @@ if [ "${USE_JOURNAL:-}" != false ] ; then
     fi
 fi
 
+IPADDR4=`/usr/sbin/ip -4 addr show dev eth0 | grep inet | sed -e "s/[ \t]*inet \([0-9.]*\).*/\1/"`
```
t0ffel: I'm actually getting SDN's pod IP for IPADDR4(same for IPADDR6). My expectation would be that I get node's IP, however I don't see node's interface from inside the pod.
That means `ipaddr4`/`ipaddr6` fields should be moved under `pipeline_metadata.collector`

richm: So `ipaddr4` and `ipaddr6` are for the node?  Do we need that?  If so, we will have to figure out a way to get that from the node, and store the container ips in the pipeline_metadata.collector (or remove them).

nhosoi: I see ipaddr4 but not ipaddr6 in pipeline_metadata.collector in com.redhat.viaq-openshift-project.template.json and com.redhat.viaq-openshift-operations.template.json.  I'm wondering if it is to be added to them or no need?

fluentd/configs.d/openshift/filter-k8s-record-transform.conf
ewolinetz: @richm its not that we would directly query on `version` however it may be useful to customers if they wish to compare different versions of data for consistency.

An added benefit is it helps us get a sense of which version of Fluentd is being used

fluentd/run.sh
richm: @t0ffel is going to add this field to pipeline_metadata.collector.

fluentd/configs.d/openshift/filter-k8s-record-transform.conf
richm: @ewolinetz ok - @t0ffel is going to add a field in the pipeline_metadata.collector namespace for that version.

richm:
@ewolinetz @jcantrill @t0ffel @lukas-vlcek and anyone else - PTAL
[test]

richm:
[test] again to see if this is a transient

deployer/scripts/uuid_migrate.sh
ewolinetz: I hate to do this, but I think we may need to reopen this conversation... Do we want to have another migration type script to make sure that data that matches the previous index pattern matches the new one so that users will be able to see their data in Kibana still?  We would update this to grab anything that doesn't start with the prefix and create an alias for it that matches `$prefix.$project.$uuid.$somesuffix` or something.

elasticsearch/index_templates/com.redhat.viaq-openshift.template.json
```diff
@@ -0,0 +1,569 @@
...
+          "type": "string"
+        }
+      }
+    }
+  },
+  "order": 10,
+  "settings": {
+    "index.refresh_interval": "5s"
+  },
+  "template": "*"
```
ewolinetz: given that this will match any index, what is the intent with this template? How much of an overlap does it have with the '.operations.*' and 'project.*' templates?

fluentd/configs.d/openshift/output-es-config.conf
```diff
@@ -3,14 +3,16 @@
...
-      index_name ${record['kubernetes_namespace_name']}.${record['kubernetes_namespace_id']}.${Time.at(time).getutc.strftime(@logstash_dateformat)}
+      index_name project.${record['kubernetes']['namespace_name']}.${record['kubernetes']['namespace_id']}.${Time.parse(record['@timestamp']).getutc.strftime(@logstash_dateformat)}
       user fluentd
       password changeme
 
       client_key "#{ENV['ES_CLIENT_KEY']}"
       client_cert "#{ENV['ES_CLIENT_CERT']}"
       ca_file "#{ENV['ES_CA']}"
 
+      type_name com.redhat.viaq.common
```
ewolinetz: this will change the ES type this is pushed to from `fluentd` to `com.redhat.viaq.common` correct?

ewolinetz:
We should make sure we are also updating the elasticsearch install script to pull in the latest release of the openshift-elasticsearch-plugin that @richm updated for using the project prefix

fluentd/configs.d/openshift/output-es-config.conf
richm: Which CI tests?
Yes, we will definitely need to update this and several other things in the docs.

richm:
This PR does update the openshift-elasticsearch-plugin version, and we will need an update to that plugin because this PR changes the mapping.

ewolinetz:
@richm ah, i was looking for the `elasticsearch/install.sh` file, I forgot that we use a var in the dockerfile instead to make it easier :)

fluentd/configs.d/openshift/output-es-config.conf
nhosoi: Like this?
-       queryCommand := `oc exec ` + kibana_pod + ` -- curl [...] "https://` + es_svc + `/` + index + `.*/fluentd/_search?q=hostname:` + hostname + `&fields=message&size=` + querySize + `"`
+       queryCommand := `oc exec ` + kibana_pod + ` -- curl [...] "https://` + es_svc + `/` + index + `.*/com.redhat.viaq.common/_search?q=hostname:` + hostname + `&fields=message&size=` + querySize + `"`

ewolinetz:
That should be it, yes

richm:
One more fix - this is currently broken with kibana - the `"template":"*"` match in the index template does indeed make kibana unhappy - the fix is to have two different index templates 
https://github.com/richm/origin-aggregated-logging/commit/cec55612003753fbc53fb7603164b2b520434001

the files are identical - the only difference is that the project file has `"template":"project.*"` and the operations file has `"template":".operations.*"`  I don't know if this is the best way to do it but it makes kibana happy.  @lukas-vlcek you will need to make this change to your PR #222

lukas-vlcek:
@richm ok, I will switch to two index templates in #222 and let you and @nhosoi know again.

fluentd/configs.d/openshift/output-es-config.conf
nhosoi: I fixed it as well as check-logs.sh to adjust the doc type.  Plus added the nil check for record.Fields.Message to check-logs.go to avoid the panic by "index out of range".

hack/testing/check-logs.go
```diff
@@ -91,6 +91,9 @@ func main() {
 
 	for _, record := range dat.Hits.Hits {
 		// for each message, we need to check the logs
+		if record.Fields.Message == nil {
```
richm: Is this necessary?  Are you seeing records without a Message field in them?  If so, that's bad, and we should investigate further.

richm:
[test]
lgtm

richm:
[test] see if transient

richm:
@ewolinetz @jcantrill @sosiouxme @lukas-vlcek PTAL

ewolinetz:
The uuid_migrate is creating an alias of `$prefix.$project.$uuid.reference` for `$prefix.$project.*`

I think we should be creating an alias of `$prefix.$project.$uuid.reference` for both `$project.$uuid.*' and `$project.*`.

https://github.com/openshift/origin-aggregated-logging/pull/234/files#diff-b72ed29e44a16b5f13edfe125361f4acR23

fluentd/Dockerfile
```diff
@@ -14,29 +14,30 @@ LABEL io.k8s.description="Fluentd container for collecting of docker container l
...
       fluent-plugin-elasticsearch \
-      fluent-plugin-flatten-hash \
       fluent-plugin-systemd \
       systemd-journal \
       fluent-plugin-rewrite-tag-filter \
       fluent-plugin-secure-forward
 
 ADD configs.d/ /etc/fluent/configs.d/
 ADD run.sh generate_throttle_configs.rb ${HOME}/
+ADD filter-common-data-model.rb /etc/fluent/plugin/
```
ewolinetz: is there a plan to host this in its own repository?

richm: Yes.  For OCP 3.5, this will be in viaq upstream, and rhlog downstream.

ewolinetz:
If we are going to make sure old data is still available to users with the index formatting change (since Kibana will be updated to reflect it) we also should be adding a check for when we upgrade to see if we need to run the UUID migration to create an alias for the new pattern for the previous indices.  Currently we run a UUID migration when we install the `admin-cert` for ES.

https://github.com/openshift/origin-aggregated-logging/blob/master/deployer/scripts/upgrade.sh#L710

deployer/scripts/uuid_migrate.sh
richm: @ewolinetz https://github.com/richm/origin-aggregated-logging/commit/a8c5361894790aa902214038354aa0502127b147 If this looks good, @nhosoi will push this to the PR

richm: basically, uuid_migrate will only care about the case when project indices have no uuid.  The only change is to create_alias, so that it will skip creating the uuid alias if the project already has an index like `project.name.uuid.*`

hack/testing/test-upgrade.sh
richm: @ewolinetz https://github.com/richm/origin-aggregated-logging/commit/a8c5361894790aa902214038354aa0502127b147 should handle this case too

richm:
@ewolinetz https://github.com/richm/origin-aggregated-logging/commit/a8c5361894790aa902214038354aa0502127b147 covers the case where we need to create `project.name.uuid.*` aliases for old `name.uuid.*` indices

ewolinetz:
@richm added comments on that commit

richm:
@ewolinetz I think we got everything.  If you approve, @nhosoi will squash the commits into a single commit for merging

deployer/scripts/upgrade.sh
```diff
@@ -46,6 +46,12 @@ function getDeploymentVersion() {
     return
   fi
 
+  # check for common data model
+  if [[ -n "$(oc get configmap/logging-elasticsearch -o yaml | grep 'use_common_data_model: true')" ]]; then
```
richm: If someone decides not to use the common data model, then `$(oc get configmap/logging-elasticsearch -o yaml | grep 'use_common_data_model: true')` will return nothing (the empty string), which means the `if [[ -n "" ]]` will return false, which means the `echo 6` will be skipped, which means the script will fall through to `echo "$LOGGING_VERSION"`

portante: Sorry I am so confused by the logic here.  It would seem on first read that for all these conditions, the first one that is hit returns at that point.  E.g., line 38, return 4 if the config map for logging-fluentd does not exist.  So if common data model is the default, then we'll return 6 all the time here, and never return the actual logging version.  Right?

portante:
It looks like this commit includes commit #222, is that right?

richm:
> It looks like this commit includes commit #222, is that right?

Yes and no.  Yes, it is in the "commit chain" of this PR, because github doesn't have a graceful way of having a commit in a PR refer to another commit in a different PR (unlike, say, gerrit).

Once #222 is merged, this PR will be rebased on that "real" commit.

ewolinetz:
@richm i think that was all that was left for me. lets go ahead and squash

-- sorry, i forgot to press send on this comment before

deployer/scripts/upgrade.sh
richm: @portante you are right - version 6 means "has es 2.x but does not have common data model"

richm:
[test]

richm:
[merge]

PR #235, opened 2016-09-14
set up to use kibana - router sa, router, kibtest user

richm:
Set up the router sa, the router, and the kibtest user to allow the
use of kibana
add env var SETUP_ONLY - if "true", just exit after setting
up the logging components

NOTE: Even with this, there is still a bit of setup - you have to make sure the machine allows port 443 and 8443 - with aws/openstack, this means having the correct security group and specifying it when creating the machine - see https://github.com/openshift/origin/pull/10916

You also need to edit your local /etc/hosts to add the public IP of the machine with the aliases for the kibana and the kibana-ops hosts.

### DIFF
--- a/hack/testing/logging.sh
+++ b/hack/testing/logging.sh
@@ -270,11 +270,42 @@ os::cmd::expect_success "oc project logging"
 os::cmd::expect_success "oc create -f $OS_O_A_L_DIR/deployer/deployer.yaml"
 os::cmd::expect_success "oc new-app logging-deployer-account-template"
 os::cmd::expect_success "oadm policy add-cluster-role-to-user oauth-editor system:serviceaccount:logging:logging-deployer"
-os::cmd::expect_success "oc create configmap logging-deployer \
-    --from-literal enable-ops-cluster=${ENABLE_OPS_CLUSTER} \
+deployer_args="--from-literal enable-ops-cluster=${ENABLE_OPS_CLUSTER} \
     --from-literal use-journal=${USE_JOURNAL:-} \
     --from-literal journal-source=${JOURNAL_SOURCE:-} \
     --from-literal journal-read-from-head=${JOURNAL_READ_FROM_HEAD:-false}"
+if [ -n "${PUBLIC_MASTER_HOST:-}" ] ; then
+    deployer_args="$deployer_args --from-literal public-master-url=https://${PUBLIC_MASTER_HOST}:8443"
+fi
+if [ -n "${KIBANA_HOST:-}" ] ; then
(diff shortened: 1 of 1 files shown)


richm:
[test]
@sosiouxme @ewolinetz ptal

richm:
I don't recognize the failure - trying again [test]

ewolinetz:
so the intent is that we would be able to actually log into the kibana UI for when we run this locally?

richm:
@ewolinetz Yes.  It works well using `vagrant up --provider aws` because aws gives you an external hostname and external ip address.  There is still a bit of manual editing to create a hostname alias for the kibana and kibana-ops hostnames.

richm:
If you use a different provider like vbox, libvirt, or openstack then I'm not sure how the hostname/ip address resolution works.  With libvirt, you would still have to edit /etc/hosts on your host.  With openstack, you still need to assign a floating ip address, then edit /etc/hosts.

hack/testing/logging.sh
```diff
@@ -439,8 +470,21 @@ os::build:wait_for_end "test"
...
 ### test app added ###
 
+### kibana setup - router account, router, kibana user ###
+oc create serviceaccount router -n default
+oadm policy add-scc-to-user privileged system:serviceaccount:default:router
+oadm policy add-cluster-role-to-user cluster-reader system:serviceaccount:default:router
+oadm router --create --namespace default --service-account=router \
+     --credentials $MASTER_CONFIG_DIR/openshift-router.kubeconfig
+oc login --username=kibtest --password=kibtest
+oc login --username=system:admin
```
ewolinetz: can we add the os::cmd::expect_success back to this line? If we aren't doing `SETUP_ONLY` we'd need to fail out here since the rest of the test probably won't work if we can't log in as 'system:admin'

ewolinetz:
I'm trying to think if there would be a more appropriate place to put this logic since it doesn't really buy anything for our CI tests... that and nit otherwise LGTM

richm:
Right - this isn't really useful for CI (unless we have some sort of kibana CI - we could do something) - but logging.sh is useful for setting up a devenv environment with logging deployed on top of it.

ewolinetz:
@richm true, i guess it makes sense then. The alternative would be what, us refactoring out the set up portion to its own script or wrapping the logic to add the router and such in an `if` block?

richm:
@ewolinetz yeah - it would be nice to have 3 different scripts - one to set up openshift+logging, one to set up router+kibana, and one to run the tests

richm:
[test]

richm:
@ewolinetz ok to merge?

ewolinetz:
[merge]

PR #238, opened 2016-09-15
creating default index template that can be used prior to common data…

ewolinetz:
… index template going out

This is to help with customers currently suffering with heap usage.

General purpose of this is to drop the current shard usage of 5/1 for new indices down to 1/1 as the default (they can always up the number of replicas at a later time). This also should help with the amount of fielddata being loaded in and taking up heap. The order is being kept low so that it can be overridden at a later time.

@sosiouxme @richm @lukas-vlcek @portante @t0ffel 
Feedback greatly appreciated.

### DIFF
--- a/hack/templates/default-index-template.json
+++ b/hack/templates/default-index-template.json
@@ -1,0 +1,89 @@
+{
+  "aliases": {},
+  "template": "*",
+  "order": 1,
+  "settings": {
+    "index.refresh_interval": "5s",
+    "index.number_of_shards": 1,
+    "index.number_of_replicas": 1,
+    "index.load_fixed_bitset_filters_eagerly": "false",
+    "indices.fielddata.cache.size": "50%",
+    "indices.fielddata.cache.expire": "1h"
+  },
+  "mappings": {
+    "_default_": {
+      "_meta": {
+        "version": "2016.09.15.0"
+      },
+      "date_detection": false,
+      "dynamic_templates": [
+        {
+          "message_field": {
+            "mapping": {
+              "index": "analyzed",
+              "omit_norms": true,
+              "type": "string"
+            },
+            "match": "message",
+            "match_mapping_type": "string"
+          }
+        },
+        {
+          "string_fields": {
+            "mapping": {
+              "index": "analyzed",
+              "omit_norms": true,
+              "type": "string"
+            },
+            "match": "*",
+            "match_mapping_type": "string"
+          }
+        }
+      ],
+      "properties": {
+        "hostname": {
+          "doc_values": true,
+          "index": "not_analyzed",
+          "type": "string"
+        },
+        "level": {
+          "doc_values": true,
+          "index": "not_analyzed",
+          "type": "string"
+        },
+        "message": {
+          "doc_values": false,
+          "index": "analyzed",
+          "norms": {
+            "enabled": false
+          },
+          "type": "string"
+        },
+        "kubernetes_pod_name": {
+          "doc_values": false,
+          "index": "analyzed",
+          "norms": {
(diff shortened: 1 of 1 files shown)


lukas-vlcek:
To me the `"index.number_of_shards": 1` seems too low. You can not increase number of shards later after index has been created. May be it is helpful for existing user use cases but generally having more shards should improve indexing performance. With only a single shard all indexed data goes always to a single node, the indexing load can not be carried out/shared by multiple nodes. So this really depends on how many nodes these customers typically have. If they usually have only a single node in ES cluster then yea.

lukas-vlcek:
I do not see how you push the index template into ES cluster, did I miss any initialization script?

lukas-vlcek:
Actually, did not know the trick with `index.load_fixed_bitset_filters_eagerly`. 😄 Is this particular setting really helping you? (It probably apply only in case of using nested documents?)

ewolinetz:
@lukas-vlcek this will need to be pushed manually; i wanted some eyes on what I wanted to provide to a customer.  the `index.load_fixed_bitset_filters_eagerly` trick is something the performance team implemented to help with fielddata being loaded in unnecessarily.

ewolinetz:
@lukas-vlcek currently we have the default of '5' which is way too high for most use cases especially if we're going to have them bump up the number of replicas to `(cluster_size - 1)`

sosiouxme:
LGTM

I don't know how authoritative this is but: https://qbox.io/blog/optimizing-elasticsearch-how-many-shards-per-index

For what we expect logging to be (lots of indexing, occasional searches) it seems 1 shard is appropriate.

lukas-vlcek:
I think we need to measure to really know. There are several factors that can be in contradiction.

Pros of more shards on a single node:
- More shards should mean better indexing throughput. If we have enough CPU cores then I think ES can benefit from this and push more data to two shards in parallel than to a single shard (each shard is a standalone Lucene index). This seems to be supported by this comment [here](https://github.com/elastic/elasticsearch/issues/16828#issuecomment-189448777).
- There is probably still limit of how many documents a single shard can hold (see bottom [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/_basic_concepts.html#_shards_amp_replicas)). So with more shards you can hold more documents in single index on single node.

Cons of more shards on a single node:
- All shards end up persisted on disk, so if the disk is slow then more shards may not really help with indexing throughput.
- More shards during heavy indexing means more (actually a LOT) lucene segments on the disk, and segments are subject to background segment merge process. Segment merges will be heavier.
- More shards will probably require more Heap usage.

ewolinetz:
Right, basically for this case I'm trying to avoid all the above cons within a customer's environment. I think that the common scenario is they will benefit more from having less shards on their nodes than from having more... Given that's just based on two different customer cases, one external and one being operations.

lukas-vlcek:
@ewolinetz If this is going to help then give it a go and try to measure if possible. We should really keep an eye on how indexing throughput (and searching as well) is impacted by this. (Check [this forum thread](https://discuss.elastic.co/t/shards-per-cpu/)).

portante:
@ewolinetz, @lukas-vlcek, @jcantrill, @sosiouxme, @richm, where are we at with this?

jcantrill:
@portante It looks like @ewolinetz applied the do not merge label so I will defer to he and @lukas-vlcek to approve it going in.

ewolinetz:
The intent is not to merge this but have it available as a reference while we wait for the common data template changes to be available for customers. Once 3.4 is available we can reevaluate the need for this and discuss just closing this PR.

jcantrill:
@ewolinetz should we close?

PR #240, opened 2016-09-15
set up to use kibana - router sa, router, kibtest user

richm:
Set up the router sa, the router, and the kibtest user to allow the
use of kibana
add env var SETUP_ONLY - if "true", just exit after setting
up the logging components

(cherry picked from commit d0d3c074edf76695bfe37df22e5656097b3fb396)


richm:
[merge]

PR #254, opened 2016-10-14
Fluend Deployment Modifications Proposal

jcantrill:
Github issue as requested from https://bugzilla.redhat.com/show_bug.cgi?id=1384626#c4  This is a work in progress.  Please comment accordingly:

cc: @ewolinetz @richm @sosiouxme @smarterclayton @lukas-vlcek @pweil- @timothysc @jeremyeder @rflorenc @portante

### DIFF
--- a/docs/proposals/fluentd-limit-requests-to-apiserver.md
+++ b/docs/proposals/fluentd-limit-requests-to-apiserver.md
@@ -1,18 +1,28 @@
 # Modifications to Fluentd Metadata Retrieval Proposal
 
 ##Motivation
-In the current design (v3.4) of the EFK stack, there is a Fluentd pod deployed to each Openshift node. Fluentd is responsible for reading container logs, adding pod metadata (e.g. labels, namespace) to individual log entries, and sending the logs to the Elasticsearch cluster. The pod metadata is sourced by watching the Openshift API server for Pod information via the [Kubernetes Metadata Filter](https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter)[2]. Testing at larger scales of Openshift clusters has shown issues where Fluentd is unable to connect to the API server because it is limited by the **maxRequestsInFlight** configuration setting [1].  This results in logs not being shipped to the aggregator and thus being unavailable.
+In the current design (v3.4) of the EFK stack, there is a Fluentd pod deployed to each Openshift node. Fluentd is responsible for reading container logs, adding pod metadata (e.g. labels, annotations) to individual log entries, and sending the logs to the Elasticsearch cluster. The additional metadata, such as labels, augment log entries to allow users to intelligently query their deployed services.  The metadata acts to enrich the logs so users may logically group their services as a single, coherent application.
+
+The pod metadata is currently sourced by watching the Openshift API server for Pod information via the [Kubernetes Metadata Filter](https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter)[2]; Each watch maintains an open connection to the API server. Testing at larger scales of Openshift clusters has shown issues where Fluentd is unable to connect to the API server because it is limited by the **maxRequestsInFlight** configuration setting [1].  The current deployment topology at scale:
+
+* Results in logs not being shipped to the aggregator and thus being unavailable
+* Limits the number of cluster nodes because of the cap on the number of open connections
+
+This proposal addresses both the short and long term solutions to resolve the connection issues being experienced using the v3.4 EFK stack.
 
 ##Proposal - Short Term (3.4)
 1. Leave the value as-is and note that it must be adjusted to account for # of nodes plus some headroom (swag @ 10%?).
 1. Increase the default to our current maximum supported node count, plus some headroom.
 1. Increasing maxRequestsInflight upstream
+1. **COMPLETE** ~~Modify Fluentd Metadata plugin to cache by namespace and pod name only~~[2][4]
 
 ##Proposal - Long Term
 ###Option 1 - Admission Controller
-1. Create a simple admission controller which simply embeds the necessary meta-data as env-variables into the pods on submission.  In this way every pod that is run on the system will have its data locally.
-1. Mount the docker socket into the Fluentd pod such that it can query containers env-variables.
-1. Remove any watches from the Fluentd pods, thereby eliminating excessive traffic from the node.
+1. Create an admission controller which embeds the necessary metadata as envirnment variables into the pods on submission.  In this way every pod that is run on the system will have its data locally.
+1. Mount the docker socket into the Fluentd pod such that it can query container's environment variables.
+1. Remove any watches from the Fluentd pods to eliminating excessive traffic from the node.
+
+This option implies log messages will never be updated when labels and annotations change since docker containers are immutable once they start.  It additionally binds the log normalization to the container runtime.
 
 This will require:
 * Admission controller introduced to [Openshift Origin](https://github.com/openshift/origin)
@@ -23,22 +33,33 @@ This will require:
 Questions:
 1. How long is metadata available?  It is limited by Pod GC?
 1. Can we completely eliminate a filter[1] to provide metadata? Consider keeping some implementation that contacts API server as a fallback mechanism
+1. What is the acceptable 'staleness' of labels and annotations associated with logs?  Is it acceptible for logs to not receive label and annotation updates when a pod definition is updated?
 
-###Option 2 - Rsyslog collector to Fluentd (Augmentor)
-Utilize rsyslog to collect logs and forward them to Fluentd.  There is initial work to support this and would allow a 'vanilla' rsyslog implementation to work in conjunction with some smaller number of Fluentd pods to add metadata.
+###Option 2 - Rsyslog collector to Fluentd (Normalizer)
+Utilize rsyslog to collect logs and forward them to Fluentd.  There is initial work to support this and would allow a 'vanilla' rsyslog implementation to work in conjunction with some smaller number of Fluentd pods to add metadata.  This would reduce the affects of scaling by:
+
+1. Moving the normalization off the compute nodes
+2. Reducing the number of open connections by utilizing a smaller number of normalizers to enrich all the logs in lieu of each Fluentd instance enriching its own logs.
 
 This will require:
 * Deployment changes (e.g. deployer or ansible)
-* Packing and containerization of rsyslog?
+* ~~Packing and containerization of rsyslog~~
+  * Packaging and containerization tasks completed by CDA team
 
 Questions:
 1. Does this require any Fluentd plugin changes?
-1. Does this require modifications to Fluentd? (e.g. how it receives logs, ships logs)
-1. How many Fluentd instance are needed?
+  * A: syslog input filter exists but is not encrypted[3]
+2. Does this require modifications to Fluentd? (e.g. how it receives logs, ships logs)
+  * A: Only the input to Fluentd changes.  Everything downstream remains the same
+3. How many Fluentd instance are needed?
+  * A: TBD
 1. Where do Fluentd instance get deployed? Infra nodes only?
 
-###Option 3 - Fluentd (Collector) to Fluentd (Augmentor)
-This is similiar to **Option 2** but it utilizes Fluentd as the collector instead of rsyslog.
+###Option 3 - Fluentd (Collector) to Fluentd (Normalizer)
+This is similiar to **Option 2** but it utilizes Fluentd as the collector instead of rsyslog.  This scenerio becomes more complicated to scale and deploy because:
+* Multiple configmaps to maintain (e.g. gather fluentd, enrichment fluentd)
+* Potentially two daemonsets
+* Certain nodes potentially more resource intensive (e.g. cpu, network, memory)
 
 This will require:
 * Deployment Changes (e.g. deployer or ansible)
@@ -47,7 +68,7 @@ Questions:
 See questions from **Option 2**
 
 ###Option 4 - Metadata service
-Introduce a service to the deployment of EFK that is solely responsible for watching Pods and gathering metadata that is provided by the Kubernetes Metadata filter.  The service is queried by Fluentd to retrieve the required information.  The expectation is instead of having N Fluentd images contacting the API server that there are some significantly less number of 'Metadata services' that provide the same information with reduced request.
+Introduce a service to the deployment of EFK that is solely responsible for watching Pods and gathering metadata that is provided by the Kubernetes Metadata filter.  The service is queried by Fluentd to retrieve the required information.  The expectation is instead of having N Fluentd pods contacting the API server that there are some significantly less number of 'Metadata services' that provide the same information with reduced request.
 
 This will require:
 * New Fluend plugin to interact with the new service
@@ -58,6 +79,19 @@ Questions:
 1. What scale is required? X instances per N cluster nodes
 1. Impact on cluster infrastructure
 1. Do instances need to be constrained to infra nodes?
+
+###Option 5 - Kubelet Handler for Pod metadata
+Add a handler to the Kubelet to allow pod information to be queried by name or to instigate a watch.  Ideally, the endpoint would feature the same characteristics of the API master pod endpoint to allow reuse of the existing metadata filter [2].  This would completely eliminate connections to the master.
+
+This will require:
+* Changes in Openshift origin
+* Deployment changes
+
+Questions:
+1. Changes to security policy to allow specific roles to access endpoint?
+
 ## References
 [1] https://bugzilla.redhat.com/show_bug.cgi?id=1384626
 [2] https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter
+[3] http://docs.fluentd.org/articles/in_syslog
+[4] https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/44
--- a/docs/proposals/fluentd-limit-requests-to-apiserver.md
+++ b/docs/proposals/fluentd-limit-requests-to-apiserver.md
@@ -1,0 +1,63 @@
+# Modifications to Fluentd Metadata Retrieval Proposal
+
+##Motivation
+In the current design (v3.4) of the EFK stack, there is a Fluentd pod deployed to each Openshift node. Fluentd is responsible for reading container logs, adding pod metadata (e.g. labels, namespace) to individual log entries, and sending the logs to the Elasticsearch cluster. The pod metadata is sourced by watching the Openshift API server for Pod information via the [Kubernetes Metadata Filter](https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter)[2]. Testing at larger scales of Openshift clusters has shown issues where Fluentd is unable to connect to the API server because it is limited by the **maxRequestsInFlight** configuration setting [1].  This results in logs not being shipped to the aggregator and thus being unavailable.
+
+##Proposal - Short Term (3.4)
+1. Leave the value as-is and note that it must be adjusted to account for # of nodes plus some headroom (swag @ 10%?).
+1. Increase the default to our current maximum supported node count, plus some headroom.
+1. Increasing maxRequestsInflight upstream
+
+##Proposal - Long Term
+###Option 1 - Admission Controller
+1. Create a simple admission controller which simply embeds the necessary meta-data as env-variables into the pods on submission.  In this way every pod that is run on the system will have its data locally.
+1. Mount the docker socket into the Fluentd pod such that it can query containers env-variables.
+1. Remove any watches from the Fluentd pods, thereby eliminating excessive traffic from the node.
+
+This will require:
+* Admission controller introduced to [Openshift Origin](https://github.com/openshift/origin)
+* Deployment changes (e.g. deployer or ansible)
+* A new Fluentd plugin
+* Support setup (e.g. ci, packaging, github/enterprise repo)
+
+Questions:
+1. How long is metadata available?  It is limited by Pod GC?
+1. Can we completely eliminate a filter[1] to provide metadata? Consider keeping some implementation that contacts API server as a fallback mechanism
+
+###Option 2 - Rsyslog collector to Fluentd (Augmentor)
+Utilize rsyslog to collect logs and forward them to Fluentd.  There is initial work to support this and would allow a 'vanilla' rsyslog implementation to work in conjunction with some smaller number of Fluentd pods to add metadata.
+
+This will require:
+* Deployment changes (e.g. deployer or ansible)
+* Packing and containerization of rsyslog?
+
+Questions:
+1. Does this require any Fluentd plugin changes?
+1. Does this require modifications to Fluentd? (e.g. how it receives logs, ships logs)
+1. How many Fluentd instance are needed?
+1. Where do Fluentd instance get deployed? Infra nodes only?
+
+###Option 3 - Fluentd (Collector) to Fluentd (Augmentor)
+This is similiar to **Option 2** but it utilizes Fluentd as the collector instead of rsyslog.
+
+This will require:
+* Deployment Changes (e.g. deployer or ansible)
+
+Questions:
+See questions from **Option 2**
+
+###Option 4 - Metadata service
+Introduce a service to the deployment of EFK that is solely responsible for watching Pods and gathering metadata that is provided by the Kubernetes Metadata filter.  The service is queried by Fluentd to retrieve the required information.  The expectation is instead of having N Fluentd images contacting the API server that there are some significantly less number of 'Metadata services' that provide the same information with reduced request.
+
+This will require:
+* New Fluend plugin to interact with the new service
+* Creation and maintenance of the service code
+  * e.g. technology choice, github repo, deployment changes (ansible), enterprise repo, ci, packaging
+
+Questions:
+1. What scale is required? X instances per N cluster nodes
+1. Impact on cluster infrastructure
+1. Do instances need to be constrained to infra nodes?
+## References
+[1] https://bugzilla.redhat.com/show_bug.cgi?id=1384626
+[2] https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter
--- a/docs/proposals/fluentd-limit-requests-to-apiserver.md
+++ b/docs/proposals/fluentd-limit-requests-to-apiserver.md
@@ -10,21 +10,25 @@ The pod metadata is currently sourced by watching the Openshift API server for P
 
 This proposal addresses both the short and long term solutions to resolve the connection issues being experienced using the v3.4 EFK stack.
 ### Considerations
-1. OpenShift Logging is not the only consumer of OpenShift log messages.  There are a significant number of OpenShift consumers who desire to use their existing logging infrastructure: Splunk, Elasticsearc.  We need to normalize log entries with metadata irrespective of the final destination.  This may dictate the deployment topology of a long term solution
+1. OpenShift Logging is not the only consumer of OpenShift log messages.  There are customers who desire to use their existing logging infrastructure(e.g. Splunk, Elasticsearch).  Log entries need to be enriched with metadata irrespective of the final destination.
+2. It is speculated that there is a low probability that labels and annotations will change during the lifetime of a running pod. We should determine if 'stale' labels and annotations are acceptable
+3. Identify the tolerance for dropped message.
 
-##Proposal - Short Term (3.4)
+## Proposal - Short Term (3.4)
 1. Leave the value as-is and note that it must be adjusted to account for # of nodes plus some headroom (swag @ 10%?).
 1. Increase the default to our current maximum supported node count, plus some headroom.
 1. Increasing maxRequestsInflight upstream
 1. **COMPLETE** ~~Modify Fluentd Metadata plugin to cache by namespace and pod name only~~[2][4]
 
-##Proposal - Long Term
-###Option 1 - Admission Controller
+## Proposal - Long Term
+### Option 1 - Admission Controller
 1. Create an admission controller which embeds the necessary metadata as envirnment variables into the pods on submission.  In this way every pod that is run on the system will have its data locally.
 1. Mount the docker socket into the Fluentd pod such that it can query container's environment variables.
 1. Remove any watches from the Fluentd pods to eliminating excessive traffic from the node.
 
-This option implies log messages will never be updated when labels and annotations change since docker containers are immutable once they start.  It additionally binds the log normalization to the container runtime.
+#### Limitations
+1. Labels and annotations will never be updated since docker containers are immutable.
+1. It binds the log normalization to the container runtime.
 
 This will require:
 * Admission controller introduced to [Openshift Origin](https://github.com/openshift/origin)
@@ -37,7 +41,7 @@ Questions:
 1. Can we completely eliminate a filter[1] to provide metadata? Consider keeping some implementation that contacts API server as a fallback mechanism
 1. What is the acceptable 'staleness' of labels and annotations associated with logs?  Is it acceptible for logs to not receive label and annotation updates when a pod definition is updated?
 
-###Option 2 - Rsyslog collector to Fluentd (Normalizer)
+### Option 2 - Rsyslog collector to Fluentd (Normalizer)
 Utilize rsyslog to collect logs and forward them to Fluentd.  There is initial work to support this and would allow a 'vanilla' rsyslog implementation to work in conjunction with some smaller number of Fluentd pods to add metadata.  This would reduce the affects of scaling by:
 
 1. Moving the normalization off the compute nodes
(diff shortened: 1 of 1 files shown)


docs/proposals/fluentd-limit-requests-to-apiserver.md
```diff
@@ -0,0 +1,63 @@
...
+Questions:
+1. How long is metadata available?  It is limited by Pod GC?
+1. Can we completely eliminate a filter[1] to provide metadata? Consider keeping some implementation that contacts API server as a fallback mechanism
+
+###Option 2 - Rsyslog collector to Fluentd (Augmentor)
+Utilize rsyslog to collect logs and forward them to Fluentd.  There is initial work to support this and would allow a 'vanilla' rsyslog implementation to work in conjunction with some smaller number of Fluentd pods to add metadata.
+
+This will require:
+* Deployment changes (e.g. deployer or ansible)
+* Packing and containerization of rsyslog?
```
richm: Yes - the common logging team is already planning to do that.

docs/proposals/fluentd-limit-requests-to-apiserver.md
```diff
@@ -0,0 +1,63 @@
...
+
+Questions:
+1. Does this require any Fluentd plugin changes?
```
ewolinetz: This depends on how Fluentd receives logs from rsyslog. If it can reuse the secure-forward plugin then no changes necessary. Fluentd does provide a syslog input filter, but it looks like things would be transported in cleartext [1]

[1] http://docs.fluentd.org/articles/in_syslog

docs/proposals/fluentd-limit-requests-to-apiserver.md
```diff
@@ -0,0 +1,63 @@
...
+1. Does this require modifications to Fluentd? (e.g. how it receives logs, ships logs)
```
ewolinetz: We should only need to change the input for Fluentd, everything else from the metadata gathering and log enrichment to shipping to Fluentd would be the same.

docs/proposals/fluentd-limit-requests-to-apiserver.md
```diff
@@ -0,0 +1,63 @@
...
+1. Where do Fluentd instance get deployed? Infra nodes only?
+
+###Option 3 - Fluentd (Collector) to Fluentd (Augmentor)
+This is similiar to **Option 2** but it utilizes Fluentd as the collector instead of rsyslog.
+
+This will require:
+* Deployment Changes (e.g. deployer or ansible)
+
+Questions:
+See questions from **Option 2**
```
ewolinetz: This also becomes complicated to scale up and deploy for the reasons that we would have two different configmaps to maintain, one that simply gathers and one that enriches the data.

They would need to be scaled independently so that's potentially two daemonsets.

We could then have some nodes where there are two Fluentd instances deployed, one of which would be more resource intensive (cpu, network, memory).

docs/proposals/fluentd-limit-requests-to-apiserver.md
```diff
@@ -0,0 +1,63 @@
...
+
+###Option 4 - Metadata service
+Introduce a service to the deployment of EFK that is solely responsible for watching Pods and gathering metadata that is provided by the Kubernetes Metadata filter.  The service is queried by Fluentd to retrieve the required information.  The expectation is instead of having N Fluentd images contacting the API server that there are some significantly less number of 'Metadata services' that provide the same information with reduced request.
```
ewolinetz: Who would maintain this and where would this live?

docs/proposals/fluentd-limit-requests-to-apiserver.md
```diff
@@ -0,0 +1,63 @@
...
+
+##Proposal - Short Term (3.4)
+1. Leave the value as-is and note that it must be adjusted to account for # of nodes plus some headroom (swag @ 10%?).
+1. Increase the default to our current maximum supported node count, plus some headroom.
+1. Increasing maxRequestsInflight upstream
+
+##Proposal - Long Term
+###Option 1 - Admission Controller
+1. Create a simple admission controller which simply embeds the necessary meta-data as env-variables into the pods on submission.  In this way every pod that is run on the system will have its data locally.
+1. Mount the docker socket into the Fluentd pod such that it can query containers env-variables.
```
pweil-: Mounting the docker socket definitely carries risk.  Also, bypassing the api makes this container runtime dependent whereas an api would allow for retrieval of the information regardless of the container runtime.

docs/proposals/fluentd-limit-requests-to-apiserver.md
```diff
@@ -0,0 +1,63 @@
...
+
+###Option 4 - Metadata service
+Introduce a service to the deployment of EFK that is solely responsible for watching Pods and gathering metadata that is provided by the Kubernetes Metadata filter.  The service is queried by Fluentd to retrieve the required information.  The expectation is instead of having N Fluentd images contacting the API server that there are some significantly less number of 'Metadata services' that provide the same information with reduced request.
+
+This will require:
+* New Fluend plugin to interact with the new service
+* Creation and maintenance of the service code
+  * e.g. technology choice, github repo, deployment changes (ansible), enterprise repo, ci, packaging
+
+Questions:
```
jcantrill: I suspect we would maintain this, at least initially, since we are the initial consumer

docs/proposals/fluentd-limit-requests-to-apiserver.md
```diff
@@ -0,0 +1,63 @@
+# Modifications to Fluentd Metadata Retrieval Proposal
+
+##Motivation
+In the current design (v3.4) of the EFK stack, there is a Fluentd pod deployed to each Openshift node. Fluentd is responsible for reading container logs, adding pod metadata (e.g. labels, namespace) to individual log entries, and sending the logs to the Elasticsearch cluster. The pod metadata is sourced by watching the Openshift API server for Pod information via the [Kubernetes Metadata Filter](https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter)[2]. Testing at larger scales of Openshift clusters has shown issues where Fluentd is unable to connect to the API server because it is limited by the **maxRequestsInFlight** configuration setting [1].  This results in logs not being shipped to the aggregator and thus being unavailable.
```
timothysc: It's more than that really.  

It's adding an extra per-node load onto both the api-server and the node itself.

docs/proposals/fluentd-limit-requests-to-apiserver.md
```diff
@@ -0,0 +1,63 @@
...
+
+##Proposal - Short Term (3.4)
+1. Leave the value as-is and note that it must be adjusted to account for # of nodes plus some headroom (swag @ 10%?).
+1. Increase the default to our current maximum supported node count, plus some headroom.
+1. Increasing maxRequestsInflight upstream
```
timothysc: This needs to be done anyway.

docs/proposals/fluentd-limit-requests-to-apiserver.md
```diff
@@ -0,0 +1,63 @@
...
+
+##Proposal - Long Term
+###Option 1 - Admission Controller
+1. Create a simple admission controller which simply embeds the necessary meta-data as env-variables into the pods on submission.  In this way every pod that is run on the system will have its data locally.
```
timothysc: We're probably going to need this long term for ISV policy integration, but that is orthogonal to the primary purpose of this proposal.

docs/proposals/fluentd-limit-requests-to-apiserver.md
timothysc: Not a fan of this option.  It adds more complexity to the system, and will still indirectly cache api-server data albeit @ 1 vs N.

docs/proposals/fluentd-limit-requests-to-apiserver.md
timothysc: The kubelet could also pass on this data possibly out-of-band from docker.

jcantrill:
@richm is there a 5th option here? combining rsyslog with the use of the admission controller.  Maybe a simple rsyslog plugin compared to what we were initially trying to do with rsyslog to replace fluentd.  Has benefits of removing the dependency packaging issues you are experiencing with fluentd but does bind us to the container runtime identified by @pweil-

richm:
@jcantrill I don't know what "admission controller" means, so I can't say for sure. In any case, we still have a fluentd in the mix: node rsyslog -> fluentd (forward listener/server side) -> elasticsearch
Would we need to create some sort of "docker socket api" plugin for rsyslog to use it in conjunction with the admission controller?

Option 5 is this - change OpenShift/Kubernetes to add _all_ of the required metadata with the log record.  For example, here is the information we get with journald:
```
# The stream identification is encoded into the PRIORITY field as an
# integer: 6, or github.com/coreos/go-systemd/journal.Info, marks stdout,
# while 3, or github.com/coreos/go-systemd/journal.Err, marks stderr.
PRIORITY=6
CONTAINER_ID=b6cbb6e73c0a
CONTAINER_ID_FULL=b6cbb6e73c0ad63ab820e4baa97cdc77cec729930e38a714826764ac0491341a
CONTAINER_NAME=k8s_registry.a49f5318_docker-registry-1-hhoj0_default_ae3a9bdc-1f66-11e6-80a2-fa163e2fff3a_799e4035
```
where CONTAINER_NAME is parsed like this:
```
    # Field 2 is the container_hash, field 5 is the pod_id, and field 6 is the pod_randhex
    config_param :container_name_to_kubernetes_regexp,
                 :string,
                 :default => '^k8s_(?<container_name>[^\.]+)\.[^_]+_(?<pod_name>[^_]+)_(?<namespace>[^_]+)_[^_]+_[a-f0-9]{8}$'
```
The additional metadata could be added as additional fields:
```
POD_ID_FULL=....
NAMESPACE_ID_FULL=
POD_LABELS=....
POD_ANNOTATIONS=....
```
This would be trivial to process with fluentd or rsyslog.

docs/proposals/fluentd-limit-requests-to-apiserver.md
pweil-: I like that idea more.  And perhaps this already works (although returning a list and not a specific pod)

```
$ curl -k https://0.0.0.0:10250/pods
{"kind":"PodList","apiVersion":"v1","metadata":{},"items":[{"metadata":{"name":"nginx","namespace":"default","selfLink":"/api/v1/namespaces/default/pods/nginx","uid":"61ca8777-9237-11e6-bfb5-54ee752009cb","resourceVersion":"63","creationTimestamp":"2016-10-14T17:55:24Z","labels":{"name":"nginx"},"annotations":{"kubernetes.io/config.seen":"2016-10-14T13:55:31.298735728-04:00","kubernetes.io/config.source":"api"}},"spec":{"volumes":[{"name":"default-token-ksyfd","secret":{"secretName":"default-token-ksyfd","defaultMode":420}}],"containers":[{"name":"nginx","image":"nginx","ports":[{"containerPort":80,"protocol":"TCP"}],"resources":{},"volumeMounts":[{"name":"default-token-ksyfd","readOnly":true,"mountPath":"/var/run/secrets/kubernetes.io/serviceaccount"}],"terminationMessagePath":"/dev/termination-log","imagePullPolicy":"Always"}],"restartPolicy":"Always","terminationGracePeriodSeconds":30,"dnsPolicy":"ClusterFirst","serviceAccountName":"default","serviceAccount":"default","nodeName":"127.0.0.1","securityContext":{}},"status":{"phase":"Running","conditions":[{"type":"Initialized","status":"True","lastProbeTime":null,"lastTransitionTime":"2016-10-14T17:55:31Z"},{"type":"Ready","status":"True","lastProbeTime":null,"lastTransitionTime":"2016-10-14T17:55:34Z"},{"type":"PodScheduled","status":"True","lastProbeTime":null,"lastTransitionTime":"2016-10-14T17:55:31Z"}],"hostIP":"127.0.0.1","podIP":"172.17.0.2","startTime":"2016-10-14T17:55:31Z","containerStatuses":[{"name":"nginx","state":{"running":{"startedAt":"2016-10-14T17:55:33Z"}},"lastState":{},"ready":true,"restartCount":0,"image":"nginx","imageID":"docker-pullable://nginx@sha256:e40499ca855c9edfb212e1c3ee1a6ba8b2d873a294d897b4840d49f94d20487c","containerID":"docker://39e9b03e617c7d136ec132703f6bba673b12c8f4a11acd3bbf2edbab79eb6541"}]}}]}
```

If a handler was installed by default that was equivalent of the `GetPodByName` call that the `HostInterface` already supports it would definitely be handy.  This seems like a small change that would allow us to leverage anything the Kubelet uses for getting to the information.

sosiouxme: I think fluentd should still maintain a local cache of pod metadata, wherever it gets it from - we don't want to be making API calls on every line of logs, even if they are local API calls. Having that cache would take care of the case where a pod has just been deleted and we're reading its container logs, assuming it lived long enough for fluentd to read its metadata at least once.

docs/proposals/fluentd-limit-requests-to-apiserver.md
portante: Can we add the specific reason we need the metadata retrieval in the first place?

In a pod, the fluentd process has access to the `pod_name` and the `namespace_name`, but is missing the:
 * `pod_id` (or `metadata.uid` from k8s)
 * `host` (or `spec.nodeName`)
 * `labels` (or `metadata.labels`)
 * `annotations` (or `metadata.annotations`)

The fluentd process currently needs to fetch the metadata associated with the pod/namespace and then cache it to allow it to scale.

The current problem comes from the fact that the API server connection appears to be kept open the entire time, not just when an API call needs to be made.

docs/proposals/fluentd-limit-requests-to-apiserver.md
portante: Consider adding a fourth item here, fixing https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/issues/43 which will help the scaling a bit, and potentially allow us to have the fluentd pods close the connection to the API server.

docs/proposals/fluentd-limit-requests-to-apiserver.md
```diff
@@ -0,0 +1,63 @@
...
+* Deployment changes (e.g. deployer or ansible)
+* A new Fluentd plugin
+* Support setup (e.g. ci, packaging, github/enterprise repo)
+
+Questions:
+1. How long is metadata available?  It is limited by Pod GC?
+1. Can we completely eliminate a filter[1] to provide metadata? Consider keeping some implementation that contacts API server as a fallback mechanism
+
+###Option 2 - Rsyslog collector to Fluentd (Augmentor)
+Utilize rsyslog to collect logs and forward them to Fluentd.  There is initial work to support this and would allow a 'vanilla' rsyslog implementation to work in conjunction with some smaller number of Fluentd pods to add metadata.
```
portante: How does this solve the problem?

docs/proposals/fluentd-limit-requests-to-apiserver.md
portante: This would also let us keep the collectors lean, so that we could use the very lightweight rsyslog collector to just data off the nodes with little memory or cpu overhead, and allow the central metadata enriching pods (normalizers) do the heavy lifting to properly enrich the logs with the right metadata, while keeping the connections to the API servers low.

docs/proposals/fluentd-limit-requests-to-apiserver.md
portante: viaq might be a good place for the code, and it would run on the infra nodes next to the ES instance.

docs/proposals/fluentd-limit-requests-to-apiserver.md
```diff
@@ -0,0 +1,63 @@
+# Modifications to Fluentd Metadata Retrieval Proposal
```
portante: Can you add a clear problem statement like: "Current design of fluentd pods maintain open connections to the API server which limit the number of nodes because API server has a cap on the number of open connections.  The fluentd pods currently need the connection to the API server so that they can query for additional metadata to enrich the logs with labels, annotations, host, and pod ID."

docs/proposals/fluentd-limit-requests-to-apiserver.md
portante: How does this solve the problem of too many connections to the API server?  Is it because the admission controller does not need to keep the connection to the API server open?

docs/proposals/fluentd-limit-requests-to-apiserver.md
portante: Do labels and annotations change for a pod/namespace pair during the lifetime of a pod and its containers?

richm: The fluent-plugin-k8s assumes labels and annotations can change during the lifetime of a pod: https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/blob/master/lib/fluent/plugin/filter_kubernetes_metadata.rb#L375

docs/proposals/fluentd-limit-requests-to-apiserver.md
jimmidyson: API server connections (watches) are kept open to keep metadata that is only available currently from the API server (labels, annotations, etc) up to date & attached to log events as they happen. This is cached to reduce the number of API server calls, but as it's a watch it keeps a long lived connection alive to keep the metadata up to date.

If you wanted to remove the persistent connection, you could disable watch & set a reasonable TTL on the cache so that cached data is expired regularly. This will increase the number of short lived connections to the API server significantly though. Also this means the metadata should be considered stale between API calls.

docs/proposals/fluentd-limit-requests-to-apiserver.md
jimmidyson: This doesn't eliminate the need for a persistent connection for watch, but rather reduces the duplicate data in the cache.

docs/proposals/fluentd-limit-requests-to-apiserver.md
jimmidyson: This sounds like a reasonable approach, but obviously requires extra grunt for the normalizer. This would obviously need to be perf tested so that it doesn't become a bottleneck for ingestion.

docs/proposals/fluentd-limit-requests-to-apiserver.md
timothysc: B/c this watches are are being done outside of the core client I'm wondering what level of connection reuse exists.  During last experiments, we exhausted ports on our LB @ 500 Nodes.  

Also, b/c you are not using the native client in the 1.5 cycle folks will need to ensure that they are leveraging http2.

jimmidyson: Sorry? 1.5 will only support http2? Can you link to github issues for that please? There are clients in many languages that will simply not be able to migrate to http2 yet AFAIK.

> B/c the watches are are being done outside of the core client I'm wondering what level of connection reuse exists.

This uses the same Ruby client that ManageIQ uses (https://github.com/abonas/kubeclient) which in turn uses the http gem (https://github.com/httprb/http). This seems to support connection reuse so...

> During last experiments, we exhausted ports on our LB @ 500 Nodes.

Could you capture number of connections per node please? Also running something on a node to capture how many connections fluentd is making would be great too if possible.

jimmidyson: Looking at the REST client that https://github.com/abonas/kubeclient uses, https://github.com/rest-client/rest-client, shows up some issues around connection reuse. I'm not a big enough Ruby user to know in detail, but perhaps @abonas could comment or know someone who could help out? It might require changes to https://github.com/rest-client/rest-client to support this - seeing as this client is used in manageiq too that might very well be worth doing.

docs/proposals/fluentd-limit-requests-to-apiserver.md
timothysc: This can be built into the kubelet <> out of band dump.  As they are already maintaining a watch.

docs/proposals/fluentd-limit-requests-to-apiserver.md
timothysc: Just thinking this through, but why couldn't such a container be part of a ES-pod to ensure that Sink isn't 2 hops.

jimmidyson: That doesn't affect the amount of cpu/memory required of course, but would reduce network bandwidth consumed. I would prefer to be able to scale normalizer & store independently personally, but I'm sure a co-located setup would be suitable for some/most use cases.

We then have to think about how we handle back-pressure, etc. Would we want a kafka-esque setup here to handle this? This starts to make the deployment much more complicated & I would only want to do this if it is actually necessary.

docs/proposals/fluentd-limit-requests-to-apiserver.md
jcantrill: No.  Assuming the kubelet also supported 'watch' we could reuse the existing plugin and function as is.  What additional load would this add to our compute nodes?

docs/proposals/fluentd-limit-requests-to-apiserver.md
jcantrill: If the sole purpose of this service is to provide metadata, why wouldn't we simply use Option 2 or 3 which has Fluentd and could be configured to enrich logs already.  The service would need to watch which is what the fluend plugin does.  There is going to need to be some sort of tweaking to understand how many instances are required; same with fluentd as a plugin.  It may make sense to co-locate with ES, but there is no reason we could not do the same with Fluentd + k8e-meta.  I think this is my least favorite option as it introduces yet another thing that needs to be maintained and something that can already be done with fluentd now.

richm: We can horizontally scale fluentd to handle back-pressure.  And if that isn't enough, we could introduce a message queue, which @smarterclayton has said some customers want.

- small case: fluentd collector/normalizer per node -> elasticsearch
This is what we have now - it is simple, but only suitable for a low number of nodes.

- medium case: fluentd collector per node -> fluentd normalizer cluster -> elasticsearch
This is proposal 3) above - the fluentd collector on each node does no docker/kubernetes metadata processing, but instead uses secure_forward to send the data to a cluster of fluentd normalizers which adds the kubernetes annotations and has the persistent watcher connections to OpenShift API.

- large case: fluentd collector per node -> message queue -> fluentd normalizer cluster -> elasticsearch

jimmidyson:
I don't know about the timescales for implementing whatever this proposal decides, but if you're not already aware there are discussions on the future of logging in Kubernetes once CRI becomes stable (no idea of timelines for that) - see https://github.com/kubernetes/kubernetes/pull/33111. Would be good to just be able to switch out the collection piece (i.e. current fluentd on the node) & keep the rest of the pipeline stable once this stabilises if possible.

docs/proposals/fluentd-limit-requests-to-apiserver.md
sosiouxme: It makes the most sense to decentralize the metadata, meaning ensure that the necessary data is available on the node and no new per-node master API request is necessary to augment metadata. That will scale indefinitely and minimize complexity. Also it will support sending enriched logs directly from fluentd to other sources besides our blessed ES - an external ES or fluentd as some customers are requesting.

We should have the necessary metadata already in the kubelet. All we need to do is make it available to the node's fluentd pod (authentication needed and possibly changes to kubelet API). I would expect the kubelet to even get updates if the metadata changes.

The only real downside of decentralizing is that we're no longer talking to the authoritative source of information, so we risk having stale data. In practice, this isn't going to happen much - there's not much of a use case for changing a pod once it's been scheduled to a node, and if logs were given (for a brief period of a minute or two) outdated labels, I can live with that.

docs/proposals/fluentd-limit-requests-to-apiserver.md
jcantrill: +1

docs/proposals/fluentd-limit-requests-to-apiserver.md
jcantrill: added

docs/proposals/fluentd-limit-requests-to-apiserver.md
jcantrill: Attempted to integrate.  See if it is to your satisfaction

docs/proposals/fluentd-limit-requests-to-apiserver.md
jimmidyson: already fixed btw

docs/proposals/fluentd-limit-requests-to-apiserver.md
```diff
@@ -0,0 +1,107 @@
...
+1. Labels and annotations will never be updated since docker containers are immutable.
+1. It binds the log normalization to the container runtime.
+
+This will require:
+* Admission controller introduced to [Openshift Origin](https://github.com/openshift/origin)
+* Deployment changes (e.g. deployer or ansible)
+* A new Fluentd plugin
+* Support setup (e.g. ci, packaging, github/enterprise repo)
+
+Questions:
```
t0ffel: Can't container override its env vars?

docs/proposals/fluentd-limit-requests-to-apiserver.md
```diff
@@ -0,0 +1,107 @@
...
+* Deployment Changes (e.g. deployer or ansible)
+
+Questions:
+See questions from **Option 2**
+
+### Option 4 - Metadata service
+Introduce a service to the deployment of EFK that is solely responsible for watching Pods and gathering metadata that is provided by the Kubernetes Metadata filter.  The service is queried by Fluentd to retrieve the required information.  The expectation is instead of having N Fluentd pods contacting the API server that there are some significantly less number of 'Metadata services' that provide the same information with reduced requests.
+
+This will require:
+* New Fluend plugin to interact with the new service
```
t0ffel: https://github.com/ViaQ/fluent-plugin-openshift_metadata_input

docs/proposals/fluentd-limit-requests-to-apiserver.md
```diff
@@ -0,0 +1,107 @@
...
+
+* Results in logs not being shipped to the aggregator and thus being unavailable
+* Limits the number of cluster nodes because of the cap on the number of open connections
+
+This proposal addresses both the short and long term solutions to resolve the connection issues being experienced using the v3.4 EFK stack.
+### Considerations
+1. OpenShift Logging is not the only consumer of OpenShift log messages.  There are customers who desire to use their existing logging infrastructure(e.g. Splunk, Elasticsearch).  Log entries need to be enriched with metadata irrespective of the final destination.
+2. It is speculated that there is a low probability that labels and annotations will change during the lifetime of a running pod. We should determine if 'stale' labels and annotations are acceptable
```
t0ffel: fyi, there is an existing bug in fluentd plugin https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/issues/28

jcantrill: This would implies we could  have stale labels now since the server terminates watches after a known amount of time.  It doesn't answer the staleness question but it is something we should fix ASAP

docs/proposals/fluentd-limit-requests-to-apiserver.md
jcantrill: Containers can override image environment variables but that happens one time; they can not be modified after the container is running.

jcantrill: I'm not sure I understand what you are saying.  An application can consume an env var in an image.  Dockerfiles allow an image author to define the env var with a default.  Running the image, you can override the value of the env var.  OpenShift further supports this by allowing you to define env vars for the container of a pod that will be applied to the container with OpenShift starts the pod.

docs/proposals/fluentd-limit-requests-to-apiserver.md
smarterclayton: Stale is actually valuable - when I delete a pod, there is no more record of its labels.  But I still want its logs, and want the labels *the pod had* during its lifetime.

docs/proposals/fluentd-limit-requests-to-apiserver.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Proposal - Short Term (3.4)
+1. Leave the value as-is and note that it must be adjusted to account for # of nodes plus some headroom (swag @ 10%?).
+1. Increase the default to our current maximum supported node count, plus some headroom.
+1. Increasing maxRequestsInflight upstream
+1. **COMPLETE** ~~Modify Fluentd Metadata plugin to cache by namespace and pod name only~~[2][4]
+
+## Proposal - Long Term
+### Option 1 - Admission Controller
+1. Create an admission controller which embeds the necessary metadata as envirnment variables into the pods on submission.  In this way every pod that is run on the system will have its data locally.
```
smarterclayton: Unlikely.  If fluent needs more info, we need to be writing it from the system.

docs/proposals/fluentd-limit-requests-to-apiserver.md
```diff
@@ -0,0 +1,107 @@
...
+1. Mount the docker socket into the Fluentd pod such that it can query container's environment variables.
+1. Remove any watches from the Fluentd pods to eliminating excessive traffic from the node.
+
+#### Limitations
+1. Labels and annotations will never be updated since docker containers are immutable.
+1. It binds the log normalization to the container runtime.
+
+This will require:
+* Admission controller introduced to [Openshift Origin](https://github.com/openshift/origin)
```
smarterclayton: Needs to be a Kubernetes solution, not just OpenShift, and as mentioned I don't think an AC will work for this because it pollutes each pod.  Instead, we need to be able to get this data efficiently on the node from the container runtime or kubelet.

docs/proposals/fluentd-limit-requests-to-apiserver.md
```diff
@@ -0,0 +1,107 @@
...
+  * e.g. technology choice, github repo, deployment changes (ansible), enterprise repo, ci, packaging
+
+Questions:
+1. What scale is required? X instances per N cluster nodes
+1. Impact on cluster infrastructure
+1. Do instances need to be constrained to infra nodes?
+2. Is this option really add value over the collector to normalizer options?
+
+### Option 5 - Kubelet Handler for Pod metadata
+Add a handler to the Kubelet to allow pod information to be queried by name or to instantiate a watch.  Ideally, the endpoint would feature the same characteristics of the API master pod endpoint to allow reuse of the existing metadata filter [2].  This would completely eliminate connections to the master.
```
smarterclayton: This is probably the option most likely to fly in Kube - why shouldn't the local logging agent have access to rich metadata logging info efficiently?

timothysc: Provided we have all the data we need, then agreed.

docs/proposals/fluentd-limit-requests-to-apiserver.md
portante: @smarterclayton, so by stale we mean that an active pod that is not deleted still emitting logs gets its logs enriched with metadata for labels and annotations that are stale (from cached values).

docs/proposals/fluentd-limit-requests-to-apiserver.md
portante: @smarterclayton, do you mean to say that it is unlikely that every pod will have data locally using this method?  Won't we know ahead of time if `fluentd` needs more info?

docs/proposals/fluentd-limit-requests-to-apiserver.md
derekwaynecarr: It would force us to properly version and configure the kubelet served APIs.

jcantrill: @derekwaynecarr Are we likely receive pushback upstream to get this request because of that?

timothysc: @derekwaynecarr it's not a strict requirement, as we version our whole stack in lock-step.

sosiouxme: @portante Having a middle tier fluentd automatically doubles network traffic, and we won't gain that back by not sending metadata to the edge. Actually (nearly?) all the metadata is already at the edge in the kubelet, we just need access to it from fluentd.

derekwaynecarr: @jcantrill - do you have a spec of what the api would look like?  obviously, today we have the summary API in the kubelet today, but its pseudo-public and is heavily skewed to only include data needed for metrics/monitoring pipeline.

portante: @sosiouxme, not sure how it automatically doubles it.

Let's say we have N fluentd pods sending data to 1 ES service.  To keep it simple we have a base log payload of 256 bytes in each log message, and 100 bytes of pod/node local metadata added (hostname, timestamp, pod id, namespace id, etc.).  So each log message sent to ES will be 356 bytes total.  Now we add the metadata fluentd adds from kubernetes, pod name, namespace name, labels, and annotations, so let's say that it is 100 bytes.  So each log message sent across the wire name becomes 456 bytes in size.

With a centralized ES configuration, the fluentd pods would send only 356 bytes to the centralized normalizer, which lives in a pod next to the ES instance.  The normalizer would then in turn add the 100 bytes and send to the ES instance.  So now we have 100 bytes *per log message* NOT sent between all the fluentd pods and the infrastructure nodes.  That is M messages * N fluentd pods * 100 bytes saved of network bandwidth.  If the actual total of labels and annotations grows beyond 100 bytes (which is likely since they are arbitrary data sets), the savings is even greater.

If we then can use Kibi, were we store a record of pod id/namespace id -> (pod name, namespace name, labels, annotations) in ES as a separate record, allowing Kibi to enrich that metadata as a join operation on logs which only contain the pod id/namespace id pair, we save all the network traffic, and the API calls.

jcantrill:
Closing.  We can reopen if it needs to be merged but there is nothing specific here that is of use in identifying where we are as of release 3.7

PR #255, opened 2016-10-17
Addressing dynamic pvc being created even when set to false

ewolinetz:
@sosiouxme @richm PTAL

Currently always evaluates to "dynamic-", even when es_pvc_dynamic = false


richm:
lgtm

ewolinetz:
[merge]

PR #263, opened 2016-10-24
Added a Makefile with a target to build Docker images

stevekuznetsov:
In order to interface with the origin-ci-tool that will be
replacing the vagrant-openshfit repository, all actions that
a user (or the CI system) will want to do in this repository
should be supported with a `make` target. This decision to
use `make` as the API between users and the repository was
made with the express purpose of reducing the amount of
complexity that any tool would have to contain.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,4 @@
+# Build the Docker images for Origin Aggregated Logging
+build-images:
+	hack/build-images.sh
+.PHONY: build-images


stevekuznetsov:
The tool's repository is [here](https://github.com/stevekuznetsov/origin-ci-tool), and it exposes both a means by which a user can run an arbitrary `make` target as well as convenience functions for `build`, `install`, and `test`, of which the target in this PR introduces `build`. `install` will be supported with the Ansible-based deployment onto a running OpenShift cluster and `test` can be supported however you like.

jcantrill:
LGTM [merge]

stevekuznetsov:
@jcantrill I'm assuming the test run flaked out -- could you please re-tag for merge?

jcantrill:
[merge]

stevekuznetsov:
@jcantrill would you mind re-tagging this, please?

jcantrill:
[merge]

PR #264, opened 2016-10-25
WIP KBase information for splitting fluentd

ewolinetz:
@jcantrill @sosiouxme @timothysc @richm @t0ffel PTAL

I need to still test the whole solution e2e, but I'm looking for fact checking/word-smithing as the intent for this is to also live in a KBase article.

### DIFF
--- a/deployer/README.md
+++ b/deployer/README.md
@@ -1178,11 +1178,14 @@ the first created. You can check if the route in question is defined in multiple
 
 ## Splitting your Fluentd for high scale capacity (Advanced)
 
-In scenarios where the number of Fluentd nodes starts to exceed 100, the watching
-connection to Kubernetes that is used within Fluentd to collect pod metadata can
-cause the cluster to exceed the *maxRequestsInFlight* configuration setting. This
-can cause symptoms such as Fluentd being unable to connect to the Kubernetes service
-to collect metadata and preventing it from being able to process and ship logs.
+In scenarios where the number of nodes running Fluentd starts to exceed 100, the
+watching connection to Kubernetes that is used within Fluentd to collect pod
+metadata can cause the cluster to exceed the default value for the *maxRequestsInFlight*
+configuration setting of `400`. Fluentd uses two watches, one for pod
+changes and one for namespace changes. This means running Fluentd on around 200
+nodes, the cluster would start to experience symptoms such as Fluentd being unable
+to connect to the Kubernetes service to collect metadata and preventing it from
+being able to process and ship logs.
 
 The alternative to setting *maxRequestsInFlight* to be unlimited, potentially creating
 a larger load on the cluster, is to split up the Fluentd deployments into collector
@@ -1199,10 +1202,21 @@ can all remain running):
 
 Where "logging-infra-fluentd" is the nodeSelector configured for the Fluentd daemonset.
 
-Next, create the configmap that will be used for the Enriching Fluentd.
+Next, create the key/cert pair that will be used for each Fluentd to communicate
+with one another:
+
+    $ openssl req -new -newkey rsa:2048 -days 712 -x509 -keyout server.key -out server.crt -subj "/CN=logging-enrichment-fluentd/OU=OpenShift/O=Logging" -passout pass:kspass
+
+Once the cert and key are created, the logging-fluentd secret needs to add these
+generated files:
+
+    $ oc create secret generic logging-fluentd --dry-run=true --from-file server.crt --from-file server.key -o yaml | oc apply -f -
+
+After that, create the configmap that will be used for the Enriching Fluentd.
 
 First, generate a random key that will be used for communication between the two
-types of Fluentd pods:
+types of Fluentd pods, this will be the value for the `shared_key` configuration
+entry:
 
     $ tr -dc 'a-zA-Z0-9' < /dev/urandom | head -c 128
 
@@ -1216,15 +1230,21 @@ Once we have the random secret key, save the following to your local directory a
       @type secure_forward
       @label @INGRESS
 
-      self_hostname ${HOSTNAME}-enrichment
+      self_hostname logging-enrichment-fluentd
       bind 0.0.0.0
       port 24284
 
-      shared_key <our generated string>
+      shared_key <our generated string from above>
 
       secure yes
-      ca_cert_path        /etc/fluentd/keys/ca
-      ca_private_key_path /etc/fluentd/keys/key
+      enable_strict_verification yes
+      ca_cert_path        /etc/fluent/keys/server.crt
+      ca_private_key_path /etc/fluent/keys/server.key
+      ca_private_key_passphrase kspass
+
+      <client>
+        host logging-enrichment-fluentd
+      </client>
     </source>
 
     <label @INGRESS>
@@ -1255,7 +1275,7 @@ file created in the previous step as the source for the configmap:
 The next step will be to create the Deployment configuration that will be used to
 scale up the enrichment fluentd pods. Save the following file to your local file
 system as `enrichment-fluentd-dc.yaml`. Be sure to update the value of "K8S_HOST_URL"
-and "OPS_HOST" if necessary:
+and "OPS_HOST" as necessary:
 
     apiVersion: v1
     kind: "DeploymentConfig"
@@ -1263,12 +1283,12 @@ and "OPS_HOST" if necessary:
       name: "logging-enrichment-fluentd"
       labels:
         provider: openshift
-        component: "fluentd"
+        component: "enrichment-fluentd"
     spec:
       replicas: 0
       selector:
         provider: openshift
-        component: "fluentd"
+        component: "enrichment-fluentd"
       triggers: {}
       strategy:
         resources: {}
@@ -1282,13 +1302,13 @@ and "OPS_HOST" if necessary:
           name: enrichment-fluentd-elasticsearch
           labels:
             provider: openshift
-            component: "fluentd"
+            component: "enrichment-fluentd"
         spec:
           terminationGracePeriodSeconds: 300
           serviceAccountName: aggregated-logging-fluentd
           containers:
           - name: enrichment-fluentd-elasticsearch
-            image: ${IMAGE_PREFIX}logging-fluentd:${IMAGE_VERSION}
+            image: registry.access.redhat.com/openshift3/ose-logging-fluentd:latest
             imagePullPolicy: Always
             securityContext:
               privileged: true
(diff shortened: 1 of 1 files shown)


deployer/README.md
```diff
@@ -1175,3 +1175,260 @@ the first created. You can check if the route in question is defined in multiple
     logging     kibana-ops   kibana-ops.example.com              logging-kibana-ops
 
 (In this example there are no overlapping routes.)
+
+## Splitting your Fluentd for high scale capacity (Advanced)
```
portante: Nice, though this seems like a lot of steps, and providing an ansible playbook which does all that is described here, just adding a set of variables, would be really powerful.

deployer/README.md
```diff
@@ -1175,3 +1175,260 @@ the first created. You can check if the route in question is defined in multiple
...
+    $ oc label node --all logging-infra-fluentd-
+
+Where "logging-infra-fluentd" is the nodeSelector configured for the Fluentd daemonset.
+
+Next, create the configmap that will be used for the Enriching Fluentd.
+
+First, generate a random key that will be used for communication between the two
+types of Fluentd pods:
+
+    $ tr -dc 'a-zA-Z0-9' < /dev/urandom | head -c 128
```
portante: You might want to give this a shell variable name for the value, as a way to refer to it in the .conf file below.

deployer/README.md
```diff
@@ -1175,3 +1175,260 @@ the first created. You can check if the route in question is defined in multiple
...
+    ## source
+    <source>
+      @type secure_forward
+      @label @INGRESS
+
+      self_hostname ${HOSTNAME}-enrichment
+      bind 0.0.0.0
+      port 24284
+
+      shared_key <our generated string>
```
portante: Would it be worthwhile to say, "our generated secret key from above"?  And potentially reference the shell variable name here?

deployer/README.md
```diff
@@ -1175,3 +1175,260 @@ the first created. You can check if the route in question is defined in multiple
...
+      @include configs.d/dynamic/input-syslog-*.conf
+      @include configs.d/openshift/input-post-*.conf
+    ##
+
+    <label @INGRESS>
+      <match **>
+        @type secure_forward
+
+        self_hostname ${HOSTNAME}
+        shared_key <our generated string>
```
portante: Would it be worthwhile to say, "our generated secret key from above"? And potentially reference the shell variable name here?

deployer/README.md
```diff
@@ -1175,3 +1175,260 @@ the first created. You can check if the route in question is defined in multiple
...
+      </match>
+    </label>
+
+Next, since our collecting Fluentd expects to ship its logs to a service called
+'logging-enrichment-fluentd' that service will need to be created:
+
+    $ oc expose dc logging-enrichment-fluentd --port=24284
+
+At this point, the enrichment Fluentd pods can be scaled up. They do not need to
+be 1:1 with the Fluentd collectors, A ratio of 1:10 is likely a good start:
```
portante: Note that the fluentd process is not multi-threaded, so you might want to consider using http://docs.fluentd.org/articles/in_multiprocess in order to make use of the cores available to the pod.

deployer/README.md
```diff
@@ -1175,3 +1175,260 @@ the first created. You can check if the route in question is defined in multiple
     logging     kibana-ops   kibana-ops.example.com              logging-kibana-ops
 
 (In this example there are no overlapping routes.)
+
+## Splitting your Fluentd for high scale capacity (Advanced)
+
+In scenarios where the number of Fluentd nodes starts to exceed 100, the watching
```
t0ffel: 1. 'Fluentd nodes' - sounds weird to me. 'nodes running Fluentd' i think is better.
2. From what I understand each fluentd starts 2 watches. What's the default value for maxRequestsInFlight? 400?

deployer/README.md
```diff
@@ -1175,3 +1175,260 @@ the first created. You can check if the route in question is defined in multiple
...
+          name: enrichment-fluentd-elasticsearch
+          labels:
+            provider: openshift
+            component: "fluentd"
+        spec:
+          terminationGracePeriodSeconds: 300
+          serviceAccountName: aggregated-logging-fluentd
+          containers:
+          - name: enrichment-fluentd-elasticsearch
+            image: ${IMAGE_PREFIX}logging-fluentd:${IMAGE_VERSION}
```
t0ffel: this seem to have been taken from the template and needs to be updated.

I however think that it is better to provide template for the DC as well as for the service

deployer/README.md
```diff
@@ -1175,3 +1175,260 @@ the first created. You can check if the route in question is defined in multiple
...
+            imagePullPolicy: Always
+            securityContext:
+              privileged: true
```
t0ffel: this container shouldn't be privileged

deployer/README.md
```diff
@@ -1175,3 +1175,260 @@ the first created. You can check if the route in question is defined in multiple
...
+              limits:
+                cpu: 100m
+            ports:
+              - containerPort: 1095
+                hostPort: 1095
+                protocol: TCP
+                purpose: "to prevent more than one per node"
+                volumeMounts:
+                - name: runlogjournal
+                  mountPath: /run/log/journal
```
t0ffel: mount point is not needed

deployer/README.md
```diff
@@ -1175,3 +1175,260 @@ the first created. You can check if the route in question is defined in multiple
...
+                - name: varlog
+                  mountPath: /var/log
```
t0ffel: mount point is not needed

deployer/README.md
```diff
@@ -1175,3 +1175,260 @@ the first created. You can check if the route in question is defined in multiple
...
+                - name: varlibdockercontainers
+                  mountPath: /var/lib/docker/containers
```
t0ffel: mount point is not needed

deployer/README.md
```diff
@@ -1175,3 +1175,260 @@ the first created. You can check if the route in question is defined in multiple
...
+                  value: <logging-es | logging-es-ops>
+                - name: "OPS_PORT"
+                  value: 9200
+                - name: "OPS_CLIENT_CERT"
+                  value: /etc/fluent/keys/cert
+                - name: "OPS_CLIENT_KEY"
+                  value: /etc/fluent/keys/key
+                - name: "OPS_CA"
+                  value: /etc/fluent/keys/ca
+              volumes:
```
t0ffel: journal and /var/log volumes are not needed

deployer/README.md
```diff
@@ -1175,3 +1175,260 @@ the first created. You can check if the route in question is defined in multiple
...
+        type: Recreate
+      template:
+        metadata:
+          name: enrichment-fluentd-elasticsearch
+          labels:
+            provider: openshift
+            component: "fluentd"
+        spec:
+          terminationGracePeriodSeconds: 300
+          serviceAccountName: aggregated-logging-fluentd
```
t0ffel: Ideally it would probably be good to have a separate SA for collectors that are not `cluster-reader`. But I guess that's too much for the purpose..

deployer/README.md
```diff
@@ -1175,3 +1175,260 @@ the first created. You can check if the route in question is defined in multiple
...
+                volumeMounts:
+                - name: runlogjournal
+                  mountPath: /run/log/journal
+                - name: varlog
+                  mountPath: /var/log
+                - name: varlibdockercontainers
+                  mountPath: /var/lib/docker/containers
+                  readOnly: true
+                - name: config
+                  mountPath: /etc/fluent/configs.d/user
```
t0ffel: if fluent.conf gets mounted here we'll have:

/etc/fluent/configs.d/user/fluent.conf (which is locally enrichment-fluent.conf),

And this file has a particular include for example:
>    @include configs.d/openshift/system.conf

`/etc/fluent/configs.d/user/configs.d/openshift/system.conf` - shouldn't really exist if i'm not missing anything

deployer/README.md
```diff
@@ -1175,3 +1175,260 @@ the first created. You can check if the route in question is defined in multiple
...
+                  readOnly: true
+                - name: certs
+                  mountPath: /etc/fluent/keys
+                  readOnly: true
+                - name: dockerhostname
```
t0ffel: I don't know if all of this hostPath stuff is needed. My guess is that they are not really needed

deployer/README.md
```diff
@@ -1175,3 +1175,260 @@ the first created. You can check if the route in question is defined in multiple
...
+
+    <label @INGRESS>
+      <match **>
+        @type secure_forward
+
+        self_hostname ${HOSTNAME}
+        shared_key <our generated string>
+
+        secure yes
+        ca_cert_path        /etc/fluentd/keys/ca
```
t0ffel: so the same cert is used to talk between fluentd -> fluentd and fluentd -> ES?

deployer/README.md
ewolinetz: Agreed, that would be a nice thing to be able to configure with our logging role, if this is an approach we'll keep around

deployer/README.md
ewolinetz: Interesting, how would that work though if our only source [for the enrichment pods] is the single `secure-forward` input?

deployer/README.md
ewolinetz: My thought with saying '100' was that even if they wouldn't hit the default limit with maxRequestsInFlight, they would want to re-architect their logging solution to decrease the load it's putting on the API server in general. Would it just be appropriate to call that information out specifically? (2 watches per Fluentd and confirm the default value)

deployer/README.md
ewolinetz: My thought with not providing a template specifically is that we may not make this a permanent solution, so it would be easier to clean up afterwards? I can be convinced otherwise though

deployer/README.md
ewolinetz: I'll just need to make sure I phrase it in a way that they don't use the literal ${SHARED_KEY} like we have for ${HOSTNAME}...

deployer/README.md
ewolinetz: We also want this if we want to do a host port binding, this way we can restrict the number of enrichment fluentd pods to one per node

deployer/README.md
richm: I don't think it would work.  We should not recommend using in_multiprocess unless we can support it, either we provide it or allow the customer to install it, and actually try it out in house and verify exactly what the performance improvement is.  In the testing by @lukas-vlcek he could not figure out how to use it to gain an appreciable performance gain in the openshift aggregated logging case.

deployer/README.md
t0ffel: > if we don't specify that the container runs in the privileged SCC, it won't actually schedule the pod
I don't think so. I believe it'll still schedule the pod - seem to work for me.

> We also want this if we want to do a host port binding, this way we can restrict the number of enrichment fluentd pods to one per node
But why is this needed?

deployer/README.md
t0ffel: I personally find it easier to deal with templates.
For cleanup - if there is a label: "fluentd-role=normalizer" on template, dc and svc then the clean up is much easier.

deployer/README.md
```diff
@@ -1175,3 +1175,249 @@ the first created. You can check if the route in question is defined in multiple
...
+          - name: certs
+            secret:
+              secretName: logging-fluentd
+          - name: dockerhostname
+            hostPath:
+              path: /etc/hostname
+          - name: localtime
+            hostPath:
+              path: /etc/localtime
+          - name: dockercfg
```
t0ffel: I think this mountpoint is used to understand whether journal log driver is used. It shouldn't be needed here imho.

deployer/README.md
timothysc: 1/10 doesn't seem like that much of a win.  so on a 400 node cluster I need to deploy 40? 

Ideally I would scale with ES shards and schedule the enricher as part of the pod to minimize hops OTW.

ewolinetz: @timothysc i'm not sure of the ratio we would recommend/use, it may be possible to keep the enricher scale the same as the ES scale. I tried to phrase it with 'likely a good start' trying to defer on that fact. After performance testing has been done with this configuration, a better number can be given

deployer/README.md
ewolinetz: > > if we don't specify that the container runs in the privileged SCC, it won't actually schedule the pod

> I don't think so. I believe it'll still schedule the pod - seem to work for me.

Hmm ok, I'll have to retest that then. I thought that you are prevented from deploying a non-privileged pod with a privileged SA.


> > We  also want this if we want to do a host port binding, this way we can restrict the number of enrichment fluentd pods to one per node

> But why is this needed?

To restrict one enrichment pod to an OpenShift node, otherwise we could end up in a scenario where we have all of our enrichment pods on a single node.

deployer/README.md
ewolinetz: I was unable to get it to work reusing the same certs. Added steps to generate a cert/key pair just for fluentd -> fluentd comunication

ewolinetz:
Made changes based on testing configuration and comments so far

deployer/README.md
ewolinetz: this should still be possible since there are labels on the objects we're creating here `component=enrichment-fluentd`. This works on the svc and dc.

ewolinetz:
Bump -- is there any further discussion that needs to be made regarding this?

richm:
lgtm

deployer/README.md
```diff
@@ -1175,3 +1175,262 @@ the first created. You can check if the route in question is defined in multiple
...
+
+Next, create the key/cert pair that will be used for each Fluentd to communicate
+with one another:
+
+    $ openssl req -new -newkey rsa:2048 -days 712 -x509 -keyout server.key -out server.crt -subj "/CN=logging-enrichment-fluentd/OU=OpenShift/O=Logging" -passout pass:kspass
+
+Once the cert and key are created, the logging-fluentd secret needs to add these
+generated files:
+
+    $ oc create secret generic logging-fluentd --dry-run=true --from-file server.crt --from-file server.key -o yaml | oc apply -f -
```
sosiouxme: I would put the `dry-run` flag next to the `-o yaml` flag ... first reaction was "oh, he left the --dry-run flag in by accident"

deployer/README.md
```diff
@@ -1175,3 +1175,262 @@ the first created. You can check if the route in question is defined in multiple
...
+
+After that, create the configmap that will be used for the Enriching Fluentd.
+
+First, generate a random key that will be used for communication between the two
+types of Fluentd pods, this will be the value for the `shared_key` configuration
```
sosiouxme: `s/pods,/pods;/`

deployer/README.md
sosiouxme: How would a user detect if they didn't have enough enrichers for the traffic? Where might they start to see errors, or would log entry just get behind?

deployer/README.md
```diff
@@ -1175,3 +1175,262 @@ the first created. You can check if the route in question is defined in multiple
...
+
+Then, create the configmap that will be used for the Enriching Fluentd, using the
+file created in the previous step as the source for the configmap:
+
+    $ oc create configmap logging-fluentd-enrichment \
+        --from-file=fluent.conf=enrichment-fluent.conf
+
+The next step will be to create the Deployment configuration that will be used to
+scale up the enrichment fluentd pods. Save the following file to your local file
+system as `enrichment-fluentd-dc.yaml`. Be sure to update the value of "K8S_HOST_URL"
```
sosiouxme: I wouldn't call out `K8S_HOST_URL`, it shouldn't need editing.

deployer/README.md
ewolinetz: @lukas-vlcek @t0ffel is there any messages that we would see in Fluentd to indicate that it isn't able to keep up with the influx of events to the enrichment pods? Either by looking at the logs of the enrichment pods or of the ingress pods?

Would this be a good use of monitoring agent? http://docs.fluentd.org/articles/monitoring

t0ffel:
My only essential concern is that the pod is privileged, while the pod doesn't do anything that should require it to be privileged.

ewolinetz:
It would need to be privileged if its going to do a host port binding though, or should we not care if we have more than one enricher pod on the same node? I would think that could be a performance impact we'd want to avoid

richm:
We may want to have more than one enricher pod running on a node.  fluentd is not threaded and does not scale well for multiple CPUs.  If you have a node with multiple CPUs, you might actually achieve high CPU utilization by running multiple fluentd on the node.  And this will work with the enricher case (as oppposed to the collector case where it is difficult to run multiple fluentd and/or in_multiprocess because you can't easily divide reading data form the different sources among multiple processes).

ewolinetz:
@richm works for me. @t0ffel I'll verify that reusing the privileged SA with an unprivileged pod works and make the change

ewolinetz:
@richm @t0ffel removed privileged pod sections from enrichment DC, verified we can scale up to more than one replica on a host

deployer/README.md
lukas-vlcek: @ewolinetz I believe this is probably plugin specific. There might be some common configuration custom followed by all supported plugins (supported by core Fluentd develoeprs), but for community plugins I am afraid this is specific.

For example I realised that for Elasticsearch output plugin all responses from ES that are not `200` are ignored (i.e. dropped). This means that all bulk indexing requests (from Fluentd to ES) are dropped if they are rejected by ES. So there does not seem to be any Fluentd core functionality that plugins can re-use in such cases? See https://github.com/uken/fluent-plugin-elasticsearch/issues/105

As you mentioned it is worth investigating if Fluentd monitoring REST API can give us some information about back pressure or lack of processing resources. When I was investigating the above mentioned issue with ES bulk indexing I checked the this REST API (there is example of REST API output in the end of this comment https://github.com/ViaQ/integration-tests/issues/6#issuecomment-227431454) and is can expose info about internal queues. Again, not sure if other plugins expose the same information. But since the secure forward should be core supported plugin we could hope...

HTH

ewolinetz: So far with just one enricher and one collector, I added the monitoring endpoint and using `oc exec <pod> -- -- curl http://localhost:24220/api/plugins.json | python -m json.tool | grep buffer` I could see the buffer_total_queued_size increase and then go back to 0.

That may be a good indicator whether or not the enrichment pod is able to keep up, if we see our collector pods not be able to fully flush their buffers and the enrichment pods getting backed up in their queues.

For some reason on my system the non-operations ES plugin on the enricher pod isn't able to fully flush, same with the collector (this may be due to the enricher pod not fully flushing).

I believe the correct pattern, if this were to be graphed, would be a "saw" shape. If we see linear growth with that it, that may be an indication that we aren't able to keep up. I'll try to resolve this buffer not clearing (I don't see any errors in the Fluentd logs), and see if my theory holds true.

ewolinetz:
Added information on monitoring the Fluentd buffer usage

jcantrill:
@ewolinetz Did we ever produce this kbase?  I believe the info is probably still relevant.

portante:
@jcantrill, @richm, isn't this replaced by mux?  Would it be better to just point them at the mux playbook?

richm:
@portante yes, this is essentially what turned into mux

jcantrill:
@richm so we should just close this PR or is it still a viable kbase we wish to publish?

richm:
I think we should publish this for 3.5/1.5 and earlier releases

jcantrill:
@ewolinetz possible to fix this up with the current stack and add to our docs here?

jcantrill:
Closing now.  Resolvable in 4.x via pipelines when implemented.

PR #277, opened 2016-11-14
Sync master changes into enterprise branch

ewolinetz:
Other than needing to resolve conflicts...

@richm @sosiouxme @jcantrill PTAL


ewolinetz:
I want to make sure we don't have anything in here that should not be made available in enterprise

PR #278, opened 2016-11-14
removing all files except for readmes and deployer.yaml files

ewolinetz:
@sosiouxme @sdodson PTAL


sdodson:
LGTM, if we want to manage this differently that's fine with me too. I just need somewhere to pull the template from, though perhaps this is the last time we do that.

PR #283, opened 2016-11-16
testing upgrading to ES 2.4.1

ewolinetz:
@richm @lukas-vlcek PTAL


richm:
LGTM, but I would like to hear what @lukas-vlcek has to say, if anything, about the problems that may occur if we allow dots in the field names.

elasticsearch/Dockerfile
```diff
@@ -8,18 +8,19 @@ USER 0
...
+  OSE_ES_VER=2.4.1.0 \
   ES_CONF=/usr/share/elasticsearch/config \
   INSTANCE_RAM=512G \
   NODE_QUORUM=1 \
   RECOVER_AFTER_NODES=1 \
   RECOVER_EXPECTED_NODES=1 \
   RECOVER_AFTER_TIME=5m \
-  PLUGIN_LOGLEVEL=INFO
+  PLUGIN_LOGLEVEL=INFO \
+  ES_JAVA_OPTS="-Dmapper.allow_dots_in_name=true"
```
lukas-vlcek: Would it be better to setup `mapper.allow_dots_in_name` in standard [`elasticsearch.yml`](https://github.com/openshift/origin-aggregated-logging/blob/master/deployer/conf/elasticsearch.yml) file instead?

ewolinetz: not sure, i went with what the ES docs said to do... didn't try to see if it would work in `elasticsearch.yml`.

lukas-vlcek: I think it should be possible, the say in [Enabling support for dots in field names](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/dots-in-names.html#_enabling_support_for_dots_in_field_names) that:

> (ES_JAVA_OPTS can be set in the /etc/default/elasticsearch file if you’re using the Debian package, or in the /etc/sysconfig/elasticsearch file if you’re using the RPM.)

If it can be set in `elasticsearch.yml` then I would say let's move it in that file so that we have all setting in one place.

ewolinetz: Sounds good, i'll retest this with that in the config instead

lukas-vlcek:
@richm I think the only problem with allowing dots in fields when upgrading from older ES is well described here: https://www.elastic.co/guide/en/elasticsearch/reference/2.4/dots-in-names.html#_upgrading_fields_with_dots_to_5_x (see the "Dots can create conflicting fields"). But I am not sure if we can easily test this on our end 'cos it depends on data in indices.

Otherwise LGTM.

elasticsearch/Dockerfile
ewolinetz: providing it in the ES config did not take... on starting up I saw the "unable to upgrade the mappings for the index"...

I'll re-add it to the dockerfile

ewolinetz:
[merge]

elasticsearch/Dockerfile
lukas-vlcek: Interesting... :-( may be we shall file a bug with Elasticsearch.

PR #287, opened 2016-11-18
Bug 1396433 - Fluentd pop lots of dump an error event: error_class=NoMethodError error="undefined method `empty?'

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1396433
Never test a value for `empty?` directly - always use the `isempty`
method which can be safely used.
Added tests to exercise the code with top level values and nested
values that do not have the `empty?` method.


richm:
[test]
@jcantrill @ewolinetz @nhosoi  PTAL

nhosoi:
Your fix and new test cases look good to me.

richm:
[merge]

PR #295, opened 2016-11-29
Update fluentd.yaml

ebalsumgo:
Right now you do not set a memory limit on the Pods, which means they can grow exponentially since the project default is not set to have any limits (as per documentation). I am going with 512MB mostly by guessing based on what I have seen it consume on average in production environment

### DIFF
--- a/deployer/templates/fluentd.yaml
+++ b/deployer/templates/fluentd.yaml
@@ -63,6 +63,7 @@ objects:
             resources:
               limits:
                 cpu: 100m
+                memory: 512Mi
             volumeMounts:
             - name: runlogjournal
               mountPath: /run/log/journal


deployer/templates/fluentd.yaml
```diff
@@ -63,6 +63,7 @@ objects:
             resources:
               limits:
                 cpu: 100m
+                memory: 512Mi
```
ebalsumgo: Well I am sitting and waiting for it to hit it, but I think this will behave like any other part of OCP/Origin in that it will not have any more access to memory and the app may/may not crash (tends to happen with Java pods) and then get restarted by the RC/DC, or chug along but slowly (tends to happen with PHP pods). 

I just made the change in production and re-deployed one fluentD pod and am monitoring now, but looking at average usages, by eye not actually crunching numbers, most fluentd pods are consuming about 310MB, which seems reasonable enough, hence my 512Mi designation to give it room to grow.

richm:
[test]

ewolinetz:
LGTM.
CI tests are currently not working, I'll create a quick patch for that though.

deployer/templates/fluentd.yaml
ebalsumgo: For what its worth I deleted half of the fluentD pods and let them re-create with the new memory limit and have not seen them cross much above 100MB Ram utilization since yesterday. The other half are ranging from 300MB to 3GB but they have also been running for about 21days.

ewolinetz:
[test]

ewolinetz:
[test]

ewolinetz:
[merge]

deployer/templates/fluentd.yaml
portante: The fluentd pods don't use a lot of memory when the flow of logs is not interrupted.  But if the output to Elasticsearch backs up, then the queue can grow and memory will be used.

In hind-sight, the testing of this PR should have been to verify that the buffer queue used for each output filter combined does not exceed the 512 MB limit.

PR #296, opened 2016-11-30
Bug 1395170 - Logging upgrade mode: Upgrade pod log states "No matching indexes found - skipping update_for_common_data_model"

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1395170
If the admin-cert exists, upgrade will skip the uuid_migrate step
which sets up the cert/key needed to use curl for the common data
model upgrade code.  The fix is to call those shell functions as
needed if the variables and files do not exist.
This also allows test-upgrade.sh to be run standalone, outside of
the context of logging.sh, and tests specifically for the existing
admin-cert case by skipping the removeAdminCert step in the test.


richm:
[test]
@ewolinetz This should fix the bug and add a test to reproduce it

hack/testing/logging.sh
```diff
@@ -107,140 +107,12 @@ function cleanup()
...
-  echo $pod
-}
-
 trap "exit" INT TERM
 trap "cleanup" EXIT
 
 echo "[INFO] Starting logging tests at " `date`
 
-ensure_iptables_or_die
+#ensure_iptables_or_die
```
ewolinetz: should we update this to be `os::util::ensure::iptables_privileges_exist`?

hack/testing/test-upgrade.sh
```diff
@@ -233,7 +274,10 @@ function verifyUpgrade() {
       [[ -n "$(oc logs $UPGRADE_POD | grep 'Migration skipped for project '$project' - using common data model')" ]] && continue
       [[ -z "$(oc logs $UPGRADE_POD | grep 'Migration for project '$project': {"acknowledged":true}')" ]] && return 1
     done
+  fi
 
+  if [ $checkCDMMigrate = true ]; then
+    [[ -n "$(oc logs $UPGRADE_POD | grep 'Migration skipped for project oldindex.'${genUUID}'.'${indexDate}' - using common data model')" ]] && return 1
```
ewolinetz: should we also check that we can find ^{"acknowledged": true} in the deployer logs?

hack/testing/logging.sh
richm: Well, I was hoping that that fix + the image-import fix would be in a separate PR

hack/testing/test-upgrade.sh
richm: ack

richm:
[test]

hack/testing/logging.sh
ewolinetz: this was resolved in https://github.com/openshift/origin-aggregated-logging/pull/297

richm:
[test]

hack/testing/test-upgrade.sh
```diff
@@ -32,6 +52,26 @@ function dumpEvents() {
...
+
+  waitFor "[[ \"Running\" == \"\$(oc get pods -l component=es -o jsonpath='{.items[*].status.phase}')\" ]]" "$(( 3 * TIME_MIN ))" || return 1
+  [ -z "$esPod" ] && echo "Unable to find ES pod for recreating old index pattern" && return 1
+
+  # create an old index pattern
+  oc exec $esPod -- curl -s --cacert /etc/elasticsearch/secret/admin-ca --cert /etc/elasticsearch/secret/admin-cert --key /etc/elasticsearch/secret/admin-key \
+                         -XPUT "https://logging-es:9200/oldindex.${genUUID}.${indexDate}" -d '{ "settings": { "index": { "number_of_shards": 1, "number_of_replicas": 0 } } }'
+
+  # necessary?
+  oc exec $esPod -- curl -s --cacert /etc/elasticsearch/secret/admin-ca --cert /etc/elasticsearch/secret/admin-cert --key /etc/elasticsearch/secret/admin-key \
```
ewolinetz: we probably can remove this if you want... Im sure it isn't causing any harm though

richm:
[test]

richm:
[test]

richm:
[test]

richm:
@ewolinetz ptal

ewolinetz:
LGTM

richm:
[merge]

PR #300, opened 2016-12-01
Modify messages to try and improve visibility when investigating test…

jcantrill:
… failures

[test]

### DIFF
--- a/hack/testing/check-EFK-running.sh
+++ b/hack/testing/check-EFK-running.sh
@@ -43,7 +43,7 @@ function checkESStarted() {
   # If this instance detects a different master, it won't recover its own indices
   #  check for output from "[cluster.service " with "] detected_master ["
   local non_master=$(echo $cluster_service | grep "detected_master")
-# OR
+  # OR
   # instance is the master if logs have this:
   #  check for output from "[cluster.service " with "] new_master ["
   local master=$(echo $cluster_service | grep "new_master")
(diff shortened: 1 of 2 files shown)


hack/testing/check-EFK-running.sh
```diff
@@ -130,60 +130,83 @@ else
 fi
 
 TEST_DIVIDER="-------------------------------------------------------"
+
+function echo_divider() {
+  msg=${1:-""}
+  echo $TEST_DIVIDER
+  echo "$TEST_DIVIDER $msg"
```
richm: The test divider is pretty long - how about just echo the msg here?

richm: use
```
   echo "$TEST_DIVIDER ${1:-}"
```
That will print $1 if it is set, otherwise, it will print an empty string

hack/testing/check-EFK-running.sh
```diff
@@ -130,60 +130,83 @@ else
...
+  echo $TEST_DIVIDER
+}
+
+function echo_error() {
+  msg=${1:-""}
+  echo "[ERROR] $msg"
```
richm: see above - don't use msg, just use `${1:-}`

hack/testing/check-EFK-running.sh
```diff
@@ -130,60 +130,83 @@ else
...
+}
+
+function echo_info() {
+  msg=${1:-""}
+  echo "[INFO ] $msg" 
+}
+
+function echo_warn() {
+  msg=${1:-""}
+  echo "[WARN ] $msg"
```
richm: see above - don't use msg, just use `${1:-}`

jcantrill:
@richm @ewolinetz base on this run https://ci.openshift.redhat.com/jenkins/job/test-origin-aggregated-logging/1036/consoleText doesnt appear it provided any value in making the logs more reasonable.  I'm in favor of closing this PR

PR #301, opened 2016-12-01
add wait_for_fluentd_to_catch_up

richm:
There are sporadic test failures in tests that add a log entry
and wait for it to show up in elasticsearch.  One of the issues is
that it may take several minutes for fluentd to catch up with the
current system and container logs, especially when using the journal
as the log source.  I'm not sure why it takes so long, that is an
issue for future investigation.  But doing this "wait" up front before
the other tests should speed up those tests and eliminate some sporadic
test failures.

This also adds some useful testing functions to the shared util.sh
which we can move other tests to use.

It sometimes now (with origin 1.5?) takes 2 minutes for the test app
database to be available.

### DIFF
--- a/deployer/scripts/util.sh
+++ b/deployer/scripts/util.sh
@@ -318,3 +318,104 @@ function get_latest_pod() {
 
   echo $pod
 }
+
+# $1 - es pod name
+# $2 - es hostname (e.g. logging-es or logging-es-ops)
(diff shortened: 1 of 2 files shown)


richm:
@ewolinetz @jcantrill PTAL
[test]

ewolinetz:
LGTM

richm:
[merge]

PR #310, opened 2016-12-14
generate externally facing certs for external kibana access

richm:


richm:
[test]
@ewolinetz ptal

ewolinetz:
Is this/should this be documented?
Should we verify the hostname we see on the route as part of this test?

richm:
> Is this/should this be documented?

Yes.  https://github.com/openshift/origin-aggregated-logging/tree/master/deployer#create-deployer-secret
"kibana.crt - A browser-facing certificate for the Kibana route. If not supplied, the route is secured with the default router cert."
This commit just adds a test for this case.

> Should we verify the hostname we see on the route as part of this test?

I don't think we should require it.  There may be cases where the clients have the kibana hostname in some sort of DNS/hosts that isn't available within the OpenShift cluster.

ewolinetz:
LGTM

richm:
[merge]

PR #312, opened 2016-12-16
create external routes for ES; allow kibtest user ES _cluster access

richm:
* create external reencrypt routes for Elasticsearch - this will allow
username/token access from outside of the cluster
* add the kibtest user to the sg_role_admin to allow it to access the
  /_cluster/health and other monitoring endpoints
* increase the test app timeouts to avoid spurious test failures

These patches allow external monitoring tools (e.g.viaq/watches-cli) to
monitor ES.


richm:
@ewolinetz @jcantrill @lukas-vlcek PTAL

richm:
[test]

richm:
[test] - etcd . . .

ewolinetz:
just to make sure i'm not missing anything here... we are adding the kibtest user to be able to monitor that end point, but we aren't yet using it in this test correct?

richm:
> just to make sure i'm not missing anything here... we are adding the kibtest user to be able to monitor that end point, but we aren't yet using it in this test correct?

Correct.  Should I add a test for this?  try to curl the /_cluster/health endpoint?

ewolinetz:
> Correct. Should I add a test for this? try to curl the /_cluster/health endpoint?
Might as well

richm:
[merge]

PR #319, opened 2017-01-09
use openshift oid when generating certs

jcantrill:
This PR moves to an openshift provided OID.  It is a first pass at allowing us to use OID the way we wish to.

cc @ewolinetz @richm

I will add similiar changes to the ansible work as well

### DIFF
--- a/deployer/conf/elasticsearch.yml
+++ b/deployer/conf/elasticsearch.yml
@@ -58,6 +58,7 @@ path:
 searchguard:
   authcz.admin_dn:
   - CN=system.admin,OU=OpenShift,O=Logging
+  cert.oid: '1.3.6.1.4.1.2312.17.100.2.1'
   config_index_name: ".searchguard.${HOSTNAME}"
   ssl:
     transport:
--- a/deployer/scripts/util.sh
+++ b/deployer/scripts/util.sh
@@ -2,9 +2,9 @@
 
 function generate_JKS_chain() {
     dir=${SCRATCH_DIR:-_output}
-    ADD_OID=$1
-    NODE_NAME=$2
-    CERT_NAMES=${3:-$NODE_NAME}
+    NODE_NAME=$1
+    CERT_NAMES=${2:-$NODE_NAME}
+    OPENSHIFT_OID=${3:-''}
     ks_pass=${KS_PASS:-kspass}
     ts_pass=${TS_PASS:-tspass}
     rm -rf $NODE_NAME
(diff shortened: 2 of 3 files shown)


jcantrill:
[test]

deployer/conf/elasticsearch.yml
```diff
@@ -58,6 +58,7 @@ path:
 searchguard:
   authcz.admin_dn:
   - CN=system.admin,OU=OpenShift,O=Logging
+  cert.oid: '1.3.6.1.4.1.2312.17.100.2.1'
```
smarterclayton: you also need to be checking the value for this to be secure - it's not enough to see it has this oid.  You need to verify that the incoming cert.oid value as an extension == the value you have on the cert you were given.  Basically, that the other person has a cert signed where this extension value == your extension value.

Maybe I'm not seeing all the changes.

deployer/scripts/util.sh
```diff
@@ -14,8 +14,8 @@ function generate_JKS_chain() {
         extension_names="${extension_names},dns:${name}"
     done
 
-    if [ "$ADD_OID" = true ]; then
-        extension_names="${extension_names},oid:1.2.3.4.5.5"
+    if [ ! -z "$OPENSHIFT_OID// " ]; then
```
richm:
```
   if [ -n "$OPENSHIFT_OID" ]; then
```

deployer/conf/elasticsearch.yml
jcantrill: This change will only replace the default value we were using with the one provided, which is also baked into the self generated certs that we do now.  This only moves us away from using the default OID using the mechanism that is the default for SG.  We will need to produce an alternative impl to https://github.com/floragunncom/search-guard/pull/269/files#diff-69afaa0c3b3e2095d1cefe863b648fb7R42 in order for us to validate any extension value properly.  Additionally we will need to consume the cert service which we have yet to do.

deployer/scripts/util.sh
jcantrill: @richm  is that the same as "if the arg is not empty"...

richm: @jcantrill yes - if `-n` is true it means arg is defined and is not empty - this is the preferred way to check rather than using `! -z`.

jcantrill:
[test]

richm:
What do we need to do for this PR?  Move it to openshift-ansible?  Backport to 1.4?  Just close it?

jcantrill:
@ewolinetz I presume I can close this PR in favor of the work we are doing for the operator and the gen of certs on container start?

ewolinetz:
@jcantrill yes, i believe so

PR #320, opened 2017-01-11
Remove USER from deployer/Dockerfile

jcantrill:
Remove the USER dec from deployer/Dockerfile
in support of using as one off jks pod with ansible changes.  
cc @richm 

[test]


richm:
lgtm

jcantrill:
[merge]

PR #327, opened 2017-01-25
new-app params cannot be comma delimited

richm:
This is the fix for the
```
warning: --param no longer accepts comma-separated lists of values
```
errors.
This also updates the deployer base image to origin v1.5.0-alpha.2, which
uncovered the problem.
@jcantrill @ewolinetz ptal
[test]

### DIFF
--- a/deployer/Dockerfile
+++ b/deployer/Dockerfile
@@ -1,4 +1,4 @@
-FROM openshift/origin:v1.2.0
+FROM openshift/origin:v1.5.0-alpha.2
 
 MAINTAINER OpenShift Development <dev@lists.openshift.redhat.com>
 
--- a/deployer/scripts/install.sh
+++ b/deployer/scripts/install.sh
@@ -90,7 +90,7 @@ function initialize_install_vars() {
   # if env vars defined, get values from them
   # special-casing this as it's required anywhere we create a DC,
   # including both installs and upgrades. so ensure it's always set.
-  image_params="IMAGE_VERSION_DEFAULT=${image_version},IMAGE_PREFIX_DEFAULT=${image_prefix}"
+  image_params="--param IMAGE_VERSION_DEFAULT=${image_version} --param IMAGE_PREFIX_DEFAULT=${image_prefix}"
 } #initialize_install_vars()
 
 function procure_server_cert() {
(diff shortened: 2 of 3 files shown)


deployer/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM openshift/origin:v1.2.0
+FROM openshift/origin:v1.5.0-alpha.2
```
jcantrill: Do we really want to bump this far?  Might it be more prudent to use v1.4.0 or v1.4.1?  Maybe this is less important at the moment now that we are moving away from the deployer

jcantrill:
@richm  Can you also confirm the build-images and push-release script dont require changes to support the base image change.

jcantrill:
it otherwise LGTM

deployer/Dockerfile
ewolinetz: Agreed, we shouldn't need to really bump this unless we want to take advantage of some of the new oc/oadm features

deployer/scripts/install.sh
```diff
@@ -90,7 +90,7 @@ function initialize_install_vars() {
   # if env vars defined, get values from them
   # special-casing this as it's required anywhere we create a DC,
   # including both installs and upgrades. so ensure it's always set.
-  image_params="IMAGE_VERSION_DEFAULT=${image_version},IMAGE_PREFIX_DEFAULT=${image_prefix}"
+  image_params="--param IMAGE_VERSION_DEFAULT=${image_version} --param IMAGE_PREFIX_DEFAULT=${image_prefix}"
```
ewolinetz: I thought this was resolved already... I guess that was just in enterprise?

hack/templates/dev-builds.yaml
```diff
@@ -61,10 +61,10 @@ objects:
   spec:
     dockerImageRepository: openshift/origin
     tags:
-    - name: v1.2.0
+    - name: v1.5.0-alpha.2
```
ewolinetz: Same echo as above regarding version bumping

ewolinetz:
Deployer image versions aside, LGTM

richm:
The problem is that the upstream deployer pod image is based on a very, very old origin, but the downstream is based on the very latest openshift.  This means the operating environment for the scripts, and the oc commands, are very different depending on which version they are operating in.  This is why we see problems downstream that we do not see upstream.  We need to be consistent in the base image versions of the deploiyer for sanity.

ewolinetz:
@richm ack to that. Should we use the latest released 1.4.0 version then rather than the 1.5.0-alpha image?

richm:
@ewolinetz The downstream is using the latest 1.5/3.5 candidate.  If we want to be in sync with downstream we should use that.  Otherwise, I don't know if the behavior of `oc` is consistent between 1.4 and 1.5

ewolinetz:
@richm  good point, alright, I'm sold then. :shipit:

richm:
@jcantrill afaict builds work fine with the new image version

richm:
[test] again

jcantrill:
As long as we are satisfied that for the functionality we require that it works as expected, I am on board with the version bump. LGTM

richm:
[merge]

PR #331, opened 2017-02-08
Add deletion of configmaps to uninstaller

nak3:
uninstall: Add deletion of configmaps to uninstaller

Currently, logging uninstaller doesn't delete configmaps which could
be created by logging deployer. This patch added the deletion of
configmaps to the deleting process.

### DIFF
--- a/deployer/scripts/install.sh
+++ b/deployer/scripts/install.sh
@@ -7,6 +7,8 @@ function delete_logging() {
   oc process logging-support-template | oc delete -f - || :
   oc delete dc,rc,svc,routes,templates,daemonset --selector logging-infra
   oc delete is -l logging-infra=support
+  oc delete configmap logging-deployer || :
+  oc delete configmap -l logging-infra=support
   # secrets don't have label selectors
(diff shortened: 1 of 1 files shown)


jcantrill:
@nak3 The deployer is deprecated in master in favor of this ansible role: https://github.com/openshift/openshift-ansible/tree/master/roles/openshift_logging  It currently only exists now in support of 'oc cluster up'.  We can apply this patch to the 3.4/1.4 code stream but it is then beholden to the next targeted release for those versions which is currently unknown to me.

nak3:
> It currently only exists now in support of 'oc cluster up'.

@jcantrill Then, I think this patch should be merged for the `oc cluster up`.

jcantrill:
@sosiouxme @ewolinetz  Did we not leave configmaps hanging around so that you could 'reinstall' without needing to supply all your config information again?  If so, then this should probably be deleted only when some flag is set.

jcantrill:
I'm closing this PR as it is not relevant to the ansible work and unlikely to see production in 3.4 and earlier.

PR #332, opened 2017-02-14
use new rubygem fluentd viaq-data-model filter plugin

richm:
@jcantrill @ewolinetz PTAL
test
This will also require a change in openshift-ansible - the fluentd.j2
template in openshift_logging - need to rename filter-common-data-model
to filter-viaq-data-model


ewolinetz:
[merge]

richm:
[merge]

PR #334, opened 2017-02-22
[WIP] Seeding Kibana Dashboards

enoodle:
Pushing Kibana dashboards to the index.

### DIFF
--- a/elasticsearch/kibana_dashboards/manageiq_ops_overview.json
+++ b/elasticsearch/kibana_dashboards/manageiq_ops_overview.json
@@ -1,0 +1,13 @@
+{
+  "title": "ManageIQ Ops Overview",
+  "hits": 0,
+  "description": "",
+  "panelsJSON": "[{\"id\":\"hostnames\",\"type\":\"visualization\",\"panelIndex\":1,\"size_x\":7,\"size_y\":4,\"col\":1,\"row\":1}]",
+  "optionsJSON": "{\"darkTheme\":true}",
+  "uiStateJSON": "{}",
+  "version": 1,
+  "timeRestore": false,
+  "kibanaSavedObjectMeta": {
+    "searchSourceJSON": "{\"filter\":[{\"query\":{\"query_string\":{\"query\":\"*\",\"analyze_wildcard\":true}}}]}"
+  }
+}
--- a/elasticsearch/run.sh
+++ b/elasticsearch/run.sh
@@ -100,9 +100,9 @@ rm -f $LOG_FILE
 exit 1
 }
 
-verify_or_add_index_templates() {
+verify_or_add_index_entities() {
     wait_for_port_open
-    # Uncomment this if you want to wait for cluster becoming more stable before index template being pushed in.
+    # Uncomment this if you want to wait for cluster becoming more stable before index entity being pushed in.
     # Give up on timeout and continue...
     # curl -v -s -X GET \
     #     --cacert $secret_dir/admin-ca \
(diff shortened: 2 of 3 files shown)


enoodle:
cc @portante @ewolinetz @richm 
This is just a preview as we need to also push the visualizations and searches and the actual dashboard.

jcantrill:
[test]

jcantrill:
@enoodle Do you believe this will address https://trello.com/c/FRtolChB

elasticsearch/run.sh
```diff
@@ -100,42 +100,47 @@ rm -f $LOG_FILE
...
+                -d@$entity_file \
+                $ES_REST_BASEURL/$index_name/$entity
         fi
     done
     shopt -u failglob
 }
 
-verify_or_add_index_templates &
+verify_or_add_index_entities "index_templates" "template" "_template" &
+verify_or_add_index_entities "kibana_dashboards" "dashboard" ".kibana/dashboard" &
```
jcantrill: This will not exactly work in this context (I dont think) since the openshift ES plugin modifies the index used for each user to be a 'kibana profile'.  I think each user ultimately gets something like '.kibana.USERUUID'.

jcantrill: That happens here: https://github.com/fabric8io/openshift-elasticsearch-plugin/blob/master/src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexFilter.java#L56 so it will need to be modified in the plugin mostlikely to seed on demand

enoodle: @jcantrill Thanks, I was just wondering why I get success message but can't see the dashboards.
Maybe it would be easier to tell kibana to also look at a different index for dashboards? something like `.kibana-shared` where we would be able to put shared resources.
My concern is that seeding this on demand from this plugin might be problematic - would we duplicate any request there to both ".kibana.USERUUID" and ".kibana" but then it just propogates them to the rest of the "chain" , how would we merge the results? (I am just not sure how this "chain" works yet)

enoodle:
> @enoodle Do you believe this will address https://trello.com/c/FRtolChB

@jcantrill Yes, this is the goal I aim for when this will work and all the searchs/visualizations will be seeded too.

elasticsearch/run.sh
ewolinetz: i think we we'd want to do is also update when we are creating a user's kibana profile to copy the seeded dashboards into their profile here:
https://github.com/fabric8io/openshift-elasticsearch-plugin/blob/master/src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java#L174-L216

richm: @ewolinetz yes, and I think it is related to  https://github.com/fabric8io/openshift-elasticsearch-plugin/issues/63 - in that case, we also want to have the ability to provide files to load instead of using the hardcoded JSON blob index pattern.

jcantrill:
@enoodle We have another card that is in regards to sharing kibana dashboards and visualizations.  There needs to be design work on how to make that happen.  Sharing implies some form of restrictions an visiblity based on owners and what groups you are apart of.

elasticsearch/run.sh
enoodle: @ewolinetz, @jcantrill  From what I can understand from those two snippets:
- It seems that the contents of `.kibana` index are being copied in [1] to the user's kibana index
- Every push to the `.kibana` index is routed to a user index.

=> This PR is actually pushing the dashboard to `.kibana.USERHASH` for this admin user that has these credentials. I feel like we should actually find a way to bypass this KibanaUserReindexFilter for this and push directly to `.kibana`

[1]https://github.com/fabric8io/openshift-elasticsearch-plugin/blob/master/src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java#L197-L201

enoodle:
Currently this is waiting on https://trello.com/c/O6MCrGUx/454-ability-to-seed-kibana-dashboard-for-admins-logging-epic-ois-agl-exp

elasticsearch/kibana_dashboards/manageiq_ops_overview.json
```diff
@@ -0,0 +1,13 @@
...
+  "description": "",
+  "panelsJSON": "[{\"id\":\"hostnames\",\"type\":\"visualization\",\"panelIndex\":1,\"size_x\":7,\"size_y\":4,\"col\":1,\"row\":1}]",
+  "optionsJSON": "{\"darkTheme\":true}",
+  "uiStateJSON": "{}",
+  "version": 1,
+  "timeRestore": false,
+  "kibanaSavedObjectMeta": {
+    "searchSourceJSON": "{\"filter\":[{\"query\":{\"query_string\":{\"query\":\"*\",\"analyze_wildcard\":true}}}]}"
+  }
+}
```
lukas-vlcek: @enoodle Is this dashboard really making any specific information available? I haven't tried it myself but it looks pretty basic to me. The reason why I ask is that I myself run into issue that standard Kibana dashboards exports do not contain index patterns (which can contain other important data/objects used on dashboards) and index patterns are required to be recreated manually by users. Have you run into similar issue yourself as well?

enoodle: This was a work in progress waiting for the actual Dashboard to insert here but due to the problems with actually using/sharing it I never updated here. I will reopen/resubmit when we have a way to share dashboards between users in some manner.

PR #337, opened 2017-02-23
add utility to make it easier to get stuff from es

jcantrill:
Simplifies getting project index data from es. Call like:

`oc exec POD -- bash es_util`

I don't know how to pass ENV vars with the command so the only way to change any of them would be to provide actual args.  Suggestions?

### DIFF
--- a/elasticsearch/es_util
+++ b/elasticsearch/es_util
@@ -1,0 +1,17 @@
+#/bin/bash -euxo pipefail
+
+# Utility to grab documents from the ES
+# instance using the admin certs and keys
+
+QUERY=${QUERY:-""} 
+INDEX=${INDEX:-"project.*"}
+SIZE=${SIZE:-10}
+SORT=${SORT:-"@timestamp:desc"}
+
+BASE=${BASE:-https://localhost:9200}
+
+if [ -z "${QUERY:-}" ]; then
+  QUERY="$BASE/$INDEX/_search?size=$SIZE&sort=$SORT"
+fi
+
+curl -s -k --cert /etc/elasticsearch/secret/admin-cert --key /etc/elasticsearch/secret/admin-key $QUERY | python -mjson.tool
--- a/elasticsearch/Dockerfile
+++ b/elasticsearch/Dockerfile
@@ -40,6 +40,7 @@ ADD index_templates/ /usr/share/elasticsearch/index_templates/
 ADD install.sh ${HOME}/
 RUN ${HOME}/install.sh
 ADD run.sh ${HOME}/
+ADD es_util /usr/local/bin
 
 WORKDIR ${HOME}
 USER 1000


jcantrill:
[test]

jcantrill:
can pass ENV vars like:

 `oc exec logging-es-hhy11s1c-5-gh6kq -- bash -c "INDEX=.operations.* es_util"`

elasticsearch/es_util
```diff
@@ -0,0 +1,19 @@
...
+
+# Utility to grab documents from the ES
+# instance using the admin certs and keys
+
+QUERY=${QUERY:-""} 
+INDEX=${INDEX:-"project.*"}
+SIZE=${SIZE:-10}
+SORT=${SORT:-"@timestamp:desc"}
+
+BASE=https://logging-es:9200
```
richm: `BASE=${BASE:-https://localhost:9200}`
since you are running this inside the same pod that ES is running, you can use `localhost`.

elasticsearch/es_util
```diff
@@ -0,0 +1,19 @@
+#/bin/bash -x
```
richm: suggest -euxo pipefail

elasticsearch/es_util
```diff
@@ -0,0 +1,19 @@
...
+# instance using the admin certs and keys
+
+QUERY=${QUERY:-""} 
+INDEX=${INDEX:-"project.*"}
+SIZE=${SIZE:-10}
+SORT=${SORT:-"@timestamp:desc"}
+
+BASE=https://logging-es:9200
+
+if [ -z $QUERY ]; then
```
richm: `"${QUERY:-}"` when using -u.

elasticsearch/es_util
```diff
@@ -0,0 +1,19 @@
...
+  QUERY="$INDEX/_search?size=$SIZE\&sort=$SORT"
+fi
+
+URL="$BASE/$INDEX"
+
+curl -s -k --cert /etc/elasticsearch/secret/admin-cert --key /etc/elasticsearch/secret/admin-key $URL | python -mjson.tool
```
richm: Where do you use `$QUERY`?

elasticsearch/es_util
```diff
@@ -0,0 +1,19 @@
...
+
+QUERY=${QUERY:-""} 
+INDEX=${INDEX:-"project.*"}
+SIZE=${SIZE:-10}
+SORT=${SORT:-"@timestamp:desc"}
+
+BASE=https://logging-es:9200
+
+if [ -z $QUERY ]; then
+  QUERY="$INDEX/_search?size=$SIZE\&sort=$SORT"
```
richm: Do you want `$INDEX` in `$QUERY`?

ewolinetz:
would it be easier if we just created an alias for the `curl` and cert info? and then a user would still `-XGET 'https://localhost:9200/....'` and such

elasticsearch/es_util
jcantrill: updated

jcantrill:
@ewolinetz I'm not sure I understand what you are suggesting.

jcantrill:
@richm updated to address your comments

richm:
[merge]

PR #345, opened 2017-03-10
add the level field when reading system logs from journald

richm:
note that there is no level field when reading from /var/log/messages
@jcantrill ptal

### DIFF
--- a/fluentd/configs.d/openshift/filter-syslog-record-transform.conf
+++ b/fluentd/configs.d/openshift/filter-syslog-record-transform.conf
@@ -37,6 +37,7 @@
     # Plus we'd better avoid the file IO.
     hostname ${_HOSTNAME.eql?('localhost') ? (begin; File.open('/etc/docker-hostname') { |f| f.readline }.rstrip; rescue; _HOSTNAME; end) : _HOSTNAME}
     message ${record["MESSAGE"]}
+    level ${record["PRIORITY"]}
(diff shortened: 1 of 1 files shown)


fluentd/configs.d/openshift/filter-syslog-record-transform.conf
```diff
@@ -37,6 +37,7 @@
     # Plus we'd better avoid the file IO.
     hostname ${_HOSTNAME.eql?('localhost') ? (begin; File.open('/etc/docker-hostname') { |f| f.readline }.rstrip; rescue; _HOSTNAME; end) : _HOSTNAME}
     message ${record["MESSAGE"]}
+    level ${record["PRIORITY"]}
```
portante: In the common data model, I believe we look for the syslog priority numbers to be changed to the names defined by syslog, while adding names that don't map.

richm: @portante ok.  Then the problem is that it isn't clear how to translate PRIORITY numeric values to level string values.  See https://github.com/ViaQ/elasticsearch-templates/issues/35

jcantrill:
@richm What does that mean and what are the implications of not having it?  Otherwise LGTM

richm:
setting DO NOT MERGE until we can resolve the level issue with the common data model

fluentd/configs.d/openshift/filter-syslog-record-transform.conf
portante: I believe [Anton](https://github.com/t0ffel/fluentd-openshift/blob/3b2098c3bb4e722bc7a161648b17fedf865db46a/configs/node-collector/journal/filter-k8s-record-transform.conf) has a fix in progress for that.

PR #346, opened 2017-03-10
Client cert design proposal for container start

jcantrill:
Design proposal for https://trello.com/c/6a9bxQdX based on a discussion I had with @mwringe  and https://trello.com/c/QddRRfrS/427-5-adopt-using-cert-generation-service-logging-epic-ois-agl-exp#comment-58912b58b3005f71db0f81e6


docs/proposals/generate-jks-certs-on-container-start.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Container Resource Definition:
+Generated client certificates will be mounted as:
+```
+/etc/elasticsearch/secrets/client-cert/<ALIAS>/tls.{cert,key}
+```
+For example:
+```
+/etc/elasticsearch/secrets/client-cert
+  |- elasticsearch
```
sosiouxme: So would this be the mount point, and thus one secret for each client? Because I don't think you can get a directory hierarchy in a secret. If you don't want a secret per client then you may need to flatten this...

docs/proposals/generate-jks-certs-on-container-start.md
```diff
@@ -0,0 +1,80 @@
...
+### Container Starts
+In the ```run.sh``` before starting Elasticsearch:
+
+**Psuedo Code:**
+```
+  if $CLIENT_CERT_DIR has directories {
+    for each $DIR in $CLIENT_CERT_DIR {
+      basename = basename($DIR)
+      if $DIR has tls.cert & tls.key & exists?($CLIENT_CA_DIR/basename/ca.cert)) {
+        keytool import --alias=basename tls.cert tls.key --out_file=keystore
```
sosiouxme: FWIW keytool still doesn't support directly importing a key. You have to perform an intermediate step of converting to pkcs12 a la http://stackoverflow.com/a/17710626/262768

docs/proposals/generate-jks-certs-on-container-start.md
jcantrill: I'm expecting one secret per client if that is what we need.  Conceptually, we could reuse it for all the clients, but this was the best I could conceive for needing to separate them.

docs/proposals/generate-jks-certs-on-container-start.md
jcantrill: I recognize there may need to be some additional manipulation.  This is just the step for 'start with X and import Y' into the keystore

jcantrill: I'm just hoping between these three files: ca, key, cert, its enough to create whatever needs to end up in the key and trust stores.

mwringe: The keystore will need these 3 pieces of information, the truststore only needs the ca and the cert.

How many ca and certs you need for the truststore will depend on the number of components you need to communicate with

mwringe:
I would be tempted to have a config map which stores the certificate and ca, and a secret which contains the key.

The certificate and ca should be public information (any client gets this when accessing your endpoint) so it doesn't necessarily have to go in a secret (but it can of course still go in a secret if we want)

So if we want to have A use certificate based mutual authentication with B, then:

A needs to know: A's key, plus the certificate and ca for A & B
B needs to know: B's key, plus the certificate and ca for A & B

And if you have some other component talking to A, say something like manageIQ, then you can give them access to A's CA so that it can verify the endpoint, all without granting them access to the key at the same time.

richm:
> So if we want to have A use certificate based mutual authentication with B, then:

> A needs to know: A's key, plus the certificate and ca for A & B

A does not need to know the cert of B, if A is the TLS client and B is the TLS server, and A trusts the ca of B

> B needs to know: B's key, plus the certificate and ca for A & B

Likewise, B does not need to know the cert of A, if B trusts the ca of A

docs/proposals/generate-jks-certs-on-container-start.md
```diff
@@ -0,0 +1,80 @@
+# Generating JKS When Container Starts
+
+## Overview
+This proposal is the design to support [creating Java Keystores](https://trello.com/c/6a9bxQdX/) on container start from secrets that are mounted into the container.  This will:
+
+* Allow the logging stack to consume certificates provided by an OpenShift certificate generation service
+* Remove the dependency on Java for the OpenShift installer.
+* Remove the need for the OpenShift installer to generate certificates and Java Keystores
```
ewolinetz: s/Openshift installer/Deployer pod and Ansible role/

docs/proposals/generate-jks-certs-on-container-start.md
```diff
@@ -0,0 +1,80 @@
...
+  |    |- tls.cert
+  |    |- tls.key
+  |
+  |- curator
+  |    |- tls.cert
+  |    |- tls.key
+  |
+```
+
+Generated client CAs will be mounted as:
```
ewolinetz: This is the CA that we will use to create a truststore and have the components provide as they communicate with ES?

docs/proposals/generate-jks-certs-on-container-start.md
ewolinetz: The truststore should only need the CA so long as its the one used to sign all the other certs

ewolinetz: Should we also add verbage around the possibly future use of the go-lang tool that Jimmi wrote so that we can create a JKS directly using these certs?

ewolinetz:
> The certificate and ca should be public information (any client gets this when accessing your endpoint) so it doesn't necessarily have to go in a secret (but it can of course still go in a secret if we want)

I'm a little weary of this because that seems like it would be _more_ open to someone potentially impersonating Fluentd and pushing bogus data (or worse if they can then see the admin cert in ES). Although I'm not sure if configmaps are any more restricted than secrets, but to me this seems like the exact purpose of a secret.

Granted, I haven't looked closely enough at the cert generation service but where would the private keys be mounted? Or how would that work with mutual TLS?

docs/proposals/generate-jks-certs-on-container-start.md
jcantrill: I don't have a solid opinion on if we should reference the go-lang tool or not.  I know we still need to test to determine if it will produce a keystore that we can use.  It seems on the surface the SG plugin is particular about the certs it uses.  Feel free to suggest verbiage for us to add in.

docs/proposals/generate-jks-certs-on-container-start.md
jcantrill: These are the CA's generated by the service that are populated into a secret.  To my understanding, any service expecting traffic from this client will need the CA

jcantrill:
@mwringe Per the design suggested by @smarterclayton It suggests to me that we will get secrets that has the information required; this would be provided by the platform.  I don't fully understand how we would populate a config map or why.  I would expect that if we want a clients and servers to have access to the certs and keys we would just provide them access to the secrets.

mwringe:
> I would expect that if we want a clients and servers to have access to the certs and keys we would just provide them access to the secrets.

Assuming that the client is only interested in the CA certificate, then this should be fine. The only 'secret' part is really the key, the CA and public certificate are not really private information.

jcantrill:
Seems I have tacit approval from @sosiouxme.  Last call for any objections or considerations that need to be included..

jcantrill:
[merge]

jcantrill:
[merge] Ran locally and works fine.. Issue fixed in ansible?

PR #347, opened 2017-03-13
tool to use to check speed of logs to Fluentd and Elasticsearch from source

richm:
= Prerequisites
* Must be root or run with sudo: `$ sudo ./check-log-speed.sh`
* Must be on the host, not inside a pod
* Must be run with system:admin rights `oc login --username=system:admin`

= Usage
```
$ sudo ./check-log-speed.sh
```

= Output when using file based logs not journal
```
Elasticsearch is 8 seconds behind /var/log/messages
```
This calculates the timestamp of the last record in `/var/log/messages`
and compares that to the timestamp of the last .operations.* index record
in Elasticsearch.  In this example, it means there is about an 8 second
lag between when the record was written to `/var/log/messages` and when
it was stored in Elasticsearch.
```
Elasticsearch is 7 seconds behind Fluentd for /var/log/messages
```
This calculates the timestamp of the last record in `/var/log/messages`
and compares that to the timestamp of the last record in `/var/log/messages`
read by Fluentd.  In this example, it means there is about a 7 second lag
between when Fluentd read the record and stored it in Elasticsearch.
```
Fluentd is 1 seconds behind /var/log/messages
```
This calculates the timestamp of the last record in `/var/log/messages`
and compares that to the timestamp of the last record in `/var/log/messages`
read by Fluentd.  In this example, it means there is about a 1 second lag
between when Fluentd read the record and when the record was written to
`/var/log/messages`.
```
Fluentd is 206 bytes of offset behind /var/log/messages
```
This is the difference between the file length of `/var/log/messages`
and the last file offset in `/var/log/messages` read by Fluentd, in bytes.
```
Fluentd is 1 records behind /var/log/messages
```
This is number of records that Fluentd has not yet read from
`/var/log/messages`.
```
Elasticsearch is 10 seconds behind /var/log/containers/docker-registry-1-k5gs5_default_registry-ef6bdf268b256d33afa82ffe8f22c4193b63fa0eb55a1f239c30354dea8e0f6a.log
Elasticsearch is 10 seconds behind Fluentd for /var/log/containers/docker-registry-1-k5gs5_default_registry-ef6bdf268b256d33afa82ffe8f22c4193b63fa0eb55a1f239c30354dea8e0f6a.log
```
Similar to the above.  In addition to `/var/log/messages`, the script
reads every file in `/var/log/containers/*.log` and skips the empty files.
```
Skipped 15 empty files of total 32
```
This tells you how many files were examined in total, and how many of those
were skipped because they were empty.

If there are no records in Elasticsearch for a particular file, you will get
a message like this:
```
Elasticsearch has no records for /var/log/containers/docker-registry-1-k5gs5_default_registry-ef6bdf268b256d33afa82ffe8f22c4193b63fa0eb55a1f239c30354dea8e0f6a.log
```

If Fluentd has not read any records from a file, then the Fluentd differences
reported will be from the beginning of the file.

= Output when using journal based logs
Fluentd has only 1 position to keep track of in the journal, which it
stores as the journal cursor in `/var/log/journal.pos`.  Journal uses
microsecond resolution timestamps, so the times reported by the script
are in second.microsecond format.
```
Fluentd is 6.123456 seconds behind the journal
```
The script gets the timestamp of the last record in the journal, and the
timestamp of the current Fluentd journal cursor position, and compares
the two.  In this case, the last record read by Fluentd is 6.123456 seconds
older than the last record in the journal.
```
Fluentd is 5 records behind the journal
```
This means that there are 5 records in the journal that Fluentd has
not yet read.
```
The Elasticsearch operations index is 9.543210 seconds behind the journal
```
This means the timestamp on the last record written to the .operations.*
indices in Elasticsearch is 9.543210 seconds behind the timestamp of the
last operations record written to the journal.
```
The Elasticsearch project index is 7.343810 seconds behind the journal
```
This means the timestamp on the last record written to any of the
project.* indices in Elasticsearch is 7.343810 seconds behind the
timestamp of the last container record written to the journal.
```
The Elasticsearch operations index is 9.543210 seconds behind Fluentd
```
This means the timestamp on the last record written to the .operations.*
indices in Elasticsearch is 9.543210 seconds behind the timestamp of the
last operations record read by Fluentd.
```
The Elasticsearch project index is 7.343810 seconds behind Fluentd
```
This means the timestamp on the last record written to any of the
project.* indices in Elasticsearch is 7.343810 seconds behind the
timestamp of the last container record read by Fluentd.

### DIFF
--- a/hack/testing/check-log-speed.sh
+++ b/hack/testing/check-log-speed.sh
@@ -1,0 +1,329 @@
+#!/bin/bash
+
+set -euo pipefail
+
+function get_running_pod() {
+    # $1 is component for selector
+    oc get pods -l component=$1 | awk -v sel=$1 '$1 ~ sel && $3 == "Running" {print $1}' | head -1
+}
+
+function query_es_from_es() {
+    oc exec $1 -- curl --connect-timeout 1 -s -k \
+       --cert /etc/elasticsearch/secret/admin-cert --key /etc/elasticsearch/secret/admin-key \
+       https://localhost:9200/${2}*/${3}\?"$4"
+}
+
+# read JSON query arguments from stdin
+function query_es_from_es_json() {
+    oc exec -i $1 -- curl --connect-timeout 1 -s -k \
+       --cert /etc/elasticsearch/secret/admin-cert --key /etc/elasticsearch/secret/admin-key \
+       https://localhost:9200/${2}*/${3} -d@-
+}
+
+function get_url_path_of_record() {
+    # returns $index/$type/$_id
+    python -c "import json, sys; hsh = json.loads(sys.stdin.read())['hits']['hits'][0]; print '%(_index)s/%(_type)s/%(_id)s'.encode('utf-8') % hsh"
+}
+
+function get_last_field_from_es() {
+    query_es_from_es $1 $2 _search "size=1\&sort=@{timefield}:desc" | \
+        get_field_value_from_json $3
+}
+
+function get_last_url_from_es() {
+    query_es_from_es $1 $2 _search "size=1&sort=${timefield}:desc${3:+&$3}" | \
+        get_url_path_of_record
+}
+
+function get_field_value_from_record() {
+    python -c 'import json, sys; print json.loads(sys.stdin.read())["_source"]["'"$1"'"].encode("utf-8")'
+}
+
+function get_field_value_from_es_url() {
+    oc exec -i $1 -- curl --connect-timeout 1 -s -k \
+       --cert /etc/elasticsearch/secret/admin-cert --key /etc/elasticsearch/secret/admin-key \
+       https://localhost:9200/$2 | get_field_value_from_record "$3"
+}
+
+# $1 - shell command or function to call to test if wait is over -
+#      this command/function should return true if the condition
+#      has been met, or false if still waiting for condition to be met
+# $2 - shell command or function to call if we timed out for error handling
+# $3 - timeout in seconds - should be a multiple of $4 (interval)
+# $4 - loop interval in seconds
+function wait_until_cmd_or_err() {
+    let ii=$3
+    local interval=${4:-1}
+    while [ $ii -gt 0 ] ; do
+        $1 && break
+        sleep $interval
+        let ii=ii-$interval
+    done
+    if [ $ii -le 0 ] ; then
+        $2
+        return 1
+    fi
+    return 0
+}
+
+debug() {
+    if [ -n "${DEBUG:-}" ] ; then
+        echo "$@"
+    fi
+}
+
+verbose() {
+    if [ -n "${VERBOSE:-}" ] ; then
+        echo "$@"
+    fi
+}
+
+err() {
+    echo ERROR: "$@"
+}
+
+info() {
+    echo "$@"
+}
+
+info_same() {
+    verbose "$@"
+}
+
+info_diff() {
+    # $X is $N $units behind $Y
+    # or
+    # $X is up-to-date with $Y
+    if [ $2 = "0" -o $2 = "0.0" ] ; then
+        info_same $1 is up-to-date with $4 in $3
+    else
+        info $1 is $2 $3 behind $4
+    fi
+}
+
+es_pod=`get_running_pod es`
+es_ops_pod=`get_running_pod es-ops 2> /dev/null`
+es_ops_pod=${es_ops_pod:-$es_pod}
+
+if grep -q -- '--log-driver=journald' /etc/sysconfig/docker ; then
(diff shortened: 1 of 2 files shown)


richm:
@jcantrill @ewolinetz @nhosoi ptal

nhosoi:
It'd be helpful why the output for non-jourmal logs does not include the ES index type (operations vs. project)?
> Elasticsearch is 7 seconds behind Fluentd for /var/log/messages
> Elasticsearch is 10 seconds behind /var/log/containers/docker-registry-1-k5gs5_default_registry-...
while the output for journal logs does as follows:
> The Elasticsearch operations index is 9.54321 seconds ...
> The Elasticsearch project index is 7.343810 seconds ...

richm:
Right, sorry.  If there is no difference, there is no output.  If you want to see all of the "xxx is up-to-date with yyy" messages, set VERBOSE=1:
```
$ VERBOSE=1 ./check-log-speed.sh
```

jcantrill:
@richm  Why can we not run this from inside the fluentd pod? It would be useful to package this with fluentd and be able to tell people to 'oc exec fluentdpodonnode'

hack/testing/check-log-speed.sh
```diff
@@ -0,0 +1,329 @@
...
+    # or
+    # $X is up-to-date with $Y
+    if [ $2 = "0" -o $2 = "0.0" ] ; then
+        info_same $1 is up-to-date with $4 in $3
+    else
+        info $1 is $2 $3 behind $4
+    fi
+}
+
+es_pod=`get_running_pod es`
```
portante: We tried this on a two-node ES cluster and this returned two ES pod names, and then the rest of the script did not work for us.

richm: I can easily change it to use the first es pod it finds - will that work?

richm:
@portante please try the latest commit

portante:
@richm, thanks ... but since I cannot run `oc login --username=system:admin` (master URL fails with "`no route to host`"), I cannot verify this.

richm:
If you are root, try 
```
export KUBECONFIG=/etc/origin/master/admin.kubeconfig
then
oc login --username=system:admin
```

portante:
Thanks Rich, but on the infra node I am on, `/etc/origin/master/admin.kubeconfig` does not exist.

richm:
> @richm Why can we not run this from inside the fluentd pod? It would be useful to package this with fluentd and be able to tell people to 'oc exec fluentdpodonnode'

The problem is that inside the fluentd pod, afaik, we only have access to the fluentd cert and key, which do not have read/search access to Elasticsearch, only create index/write access.  So if we did that, we would lose the ability to compare with Elasticsearch.  And we can't run the command inside the elasticsearch pod because it doesn't have access to the log files/journal.

@ewolinetz Is my assumption correct that fluentd does not have access to the admin cert/key?

richm:
@portante Can you use the `oc` command at all, on any node?

hack/testing/check-log-speed.sh
ewolinetz: it should work given all our ES pods are currently master-data nodes. this may change in the future if we can configure them differently but we can cross that bridge when we get to it

ewolinetz:
> @jcantrill The problem is that inside the fluentd pod, afaik, we only have access to the fluentd cert and key, which do not have read/search access to Elasticsearch, only create index/write access

Correct

> @ewolinetz Is my assumption correct that fluentd does not have access to the admin cert/key?

Yes, it does not have access to it

richm:
For ease of distribution, I could make this part of the fluentd image e.g. in /opt/app-root/src/bin/check-log-speed.sh - but then you would have to

    oc rsync $fluentdpod:/opt/app-root/src/bin .

then `./check-log-speed.sh`

I could also remove the need to use `oc` when running the script - the user would have to provide the ES admin cert/key (e.g. in /etc/origin/logging/system.admin.{crt,key})

What would you like to do?

jcantrill:
I think providing as much tooling as is useful to us and them with the images make sense to me along with clear instructions on how to properly use it.

richm:
@jcantrill ok - it sounds like you are suggesting to put it in the fluentd image, and provide instructions about how to rsync it out and use it?
@portante What about the requirement of the `oc` command?  I think it will be tricky for users to figure out how to get the admin cert/key, find the cluster IP of ES, all without using `oc`.

richm:
Ok.  The current plan is to adapt the existing tool to work in at least a 2 machine master + node environment.

jcantrill:
Should we close this PR or fix up and push into master?

PR #348, opened 2017-03-14
update deployer readme

jcantrill:
Update readme docs to deprecate the deployer and point people to use openshift-ansible

### DIFF
--- a/docs/checking-efk-health.md
+++ b/docs/checking-efk-health.md
@@ -1,0 +1,254 @@
+# Checking EFK Health
+
+Determining the health of an EFK deployment and if it is running
+is described in the subsequent sections.
+
+## Fluentd
+
+Check Fluentd logs for the message that it has read in its config file:
+```
+2016-02-19 20:40:44 +0000 [info]: reading config file path="/etc/fluent/fluent.conf"
+```
+
+Verify that Fluentd has been able to start reading in log files
+by checking the contents of `/var/log/node.log.pos` and `/var/log/es-containers.log.pos`.
+The `node.log.pos` file is a position file that keeps track of where Fluentd is in reading the syslog log files.  
+The `es-containers.log.pos` is a similiar file to keep track of where fluentd is in reading the docker log files (i.e `/var/log/containers`).  
+If the OpenShift node is configured to use the systemd journal as the log source, the position file is `/var/log/journal.pos`.
+
+You can view the journal starting at the location where fluentd last read it by using a command like the following (as root):
+
+```
+$ journalctl -c `cat /var/log/journal.pos`
+```
+
+## Elasticsearch
+
+Elasticsearch logs provide information such as the number of indices it recovered upon starting up.
+```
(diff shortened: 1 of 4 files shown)


ewolinetz:
I feel like we should also be pointing to a playbook that can be used and provide an example inventory file. We shouldn't just point to the role

jcantrill:
@ewolinetz @richm  Updated to add a quickstart.  Thoughts?

docs/quickstart.md
```diff
@@ -0,0 +1,73 @@
...
+
+Following are instructions to standing up the logging stack.  Additional
+parameterization to further customize and alter the deployment can be found
+in the [openshift_logging role](https://github.com/openshift/openshift-ansible/tree/master/roles/openshift_logging).  
+
+## Ansible Setup
+These instructions assume you are familiar with Ansible and have met the minimal [installation requirements](http://docs.ansible.com/ansible/intro_installation.html) for the control machine and managed nodes.
+
+1. `git clone https://github.com/ansible/ansible`
+2. `cd ansible && git checkout v2.2.0.0-1`
```
ewolinetz: i think we actually would need to use `2.2.1.0` now

jcantrill: Sum up my IRC comments:

```
FATAL: Current Ansible version (2.2.1.0) is not supported
```

docs/checking-efk-health.md
```diff
@@ -0,0 +1,246 @@
...
+
+Check Fluentd logs for the message that it has read in its config file:
+```
+2016-02-19 20:40:44 +0000 [info]: reading config file path="/etc/fluent/fluent.conf"
+```
+
+Verify that Fluentd has been able to start reading in log files
+by checking the contents of `/var/log/node.log.pos` and `/var/log/es-containers.log.pos`.
+The `node.log.pos` file is a position file that keeps track of where Fluentd is in reading the syslog log files.  The `es-containers.log.pos` is a similiar file to keep track of where fluentd is in reading the docker log files (i.e `/var/log/containers`).  If the OpenShift node is configured to use the systemd journal as the log source, the position file is `/var/log/journal.pos`.
+
```
richm: You can view the journal starting at the location where fluentd last read it by using a command like the following (as root):
```
# journalctl -c `cat /var/log/journal.pos`
```

docs/quickstart.md
```diff
@@ -0,0 +1,73 @@
...
+3. `source hacking/env-setup`
+
+## Deploy Aggregated Logging
+
+1. Clone the Openshift installer
+```
+git clone https://github.com/openshift/openshift-ansible/tree/master/roles/openshift_logging
```
richm: git clone https://github.com/openshift/openshift-ansible

jcantrill:
@richm I made a few edits to bring the ansible code in line with master.  I believe we have a sufficient example inventory file that I tested against `oc cluster up` .  I'm going to merge based on your initial approval

jcantrill:
[merge]

PR #350, opened 2017-03-16
bug 1432389. Fix log throttle support for json-file driver

jcantrill:
The PR fixes the issue where the script was:
*  trying to evaluate a env setting that was not present on the DS
* fix the filename of the expected throttle file

### DIFF
--- a/fluentd/generate_throttle_configs.rb
+++ b/fluentd/generate_throttle_configs.rb
@@ -201,7 +201,7 @@ def get_project_pattern(name)
   project_pattern << "_*.log"
 end
 
-filename = "/etc/fluent/configs.d/user/throttle-conf.yaml"
+filename = "/etc/fluent/configs.d/user/throttle-config.yaml"
 
 parsed = ""
 parsed = YAML.load_file(filename) if File.exists?(filename)
@@ -211,11 +211,6 @@ excludeSyslog = false
 
 # We do not yet support throttling logs read from the journal
 unless ENV['USE_JOURNAL'] == "true"
-  filename = "#{ENV['THROTTLE_CONF_LOCATION']}/settings"
-
-  parsed = ""
-  parsed = YAML.load_file(filename) if File.exists?(filename)
-
   parsed.each { |project|
     name = project[0]
     options = project[1]


fluentd/generate_throttle_configs.rb
```diff
@@ -2,16 +2,35 @@
 require 'date'
 require 'logger'
 
+log_level = ENV['LOG_LEVEL'] || 'INFO'
```
ewolinetz: doesn't it make sense to have this default at 'WARN' given that's what we do for the rescue below?

fluentd/generate_throttle_configs.rb
```diff
@@ -2,16 +2,35 @@
...
+DEFAULT_FILENAME = "/etc/fluent/configs.d/user/throttle-conf.yaml"
+
+VALID_SETTINGS = {"read_lines_limit" => "number"}
 
 def get_file_name(name, isSyslog)
   ## file_name follows pattern: gen-#{name}-YYYYMMDD.conf ##
 
-  file_name = "/etc/fluent/configs.d/dynamic/input-docker-" unless isSyslog
-  file_name = "/etc/fluent/configs.d/dynamic/input-syslog-" if isSyslog
+  file_name = if isSyslog
```
ewolinetz: isn't this block basically the same as the above two lines?

fluentd/generate_throttle_configs.rb
```diff
@@ -201,39 +217,37 @@ def get_project_pattern(name)
...
-          excluded.push(get_project_pattern("openshift-infra"))
-
+          @log.debug("Found throttling settings for operations. Excluding projects: #{DEFAULT_OPS_PROJECTS}") 
+          DEFAULT_OPS_PROJECTS.each do |p|
+            excluded.push(get_project_pattern(p))
+          end
           excludeSyslog = true
           write_to_file(name, k, v, true)
+        else
+          excluded.push(get_project_pattern(name))
```
ewolinetz: isn't this the same as the removed line 228?

fluentd/generate_throttle_configs.rb
jcantrill: I set to INFO because of this log message https://github.com/openshift/origin-aggregated-logging/pull/350/files#diff-12cce37c2f5e5292a8adf8d11adae7b1R232 but maybe it makes sense to keep them the same.  I'm indifferent

fluentd/generate_throttle_configs.rb
jcantrill: it is but in a more ruby construct.

fluentd/generate_throttle_configs.rb
jcantrill: It is but in a more traditional if/else construct.  Reading the original logic, I believe it still is valid.

fluentd/generate_throttle_configs.rb
ewolinetz: I think it should be 'WARN' given that is what we are defaulting to if its not a valid setting. It should be the same as when it isn't provided

ewolinetz:
@jcantrill is the fix that is required to resolve the bz the ENV check just at the top of the file? It's not entirely clear to me based on this commit given there seems to also be stylized refactoring...

jcantrill:
@ewolinetz The real issue was because of https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/generate_throttle_configs.rb#L204-L217  This change was introduced last July.  It completely ignores the file from the confimap that is mounted into the container.  The only way throttle settings would work is to set the env var (which there is no default) and the file to be named 'settings'. 

I took the liberty to refactor some bits as well to provide some additional debug logging for future.

fluentd/generate_throttle_configs.rb
jcantrill: updated to WARN

jcantrill:
[test]

ewolinetz:
@jcantrill ack, could we just move the refactoring changes into their own commit just to make that more clear? otherwise LGTM

jcantrill:
@ewolinetz suggestions on an easy way to do?
`git stash`
make minimal changes - PR
`git stash pop`
`git apply` - PR

jcantrill:
[merge]

jcantrill:
[merge]

PR #353, opened 2017-03-20
bug 1431551. Seed SG config on start instead of in plugin

jcantrill:
This PR moves SG seeding to run.sh.  Blocked on https://github.com/fabric8io/openshift-elasticsearch-plugin/pull/68


elasticsearch/install.sh
```diff
@@ -8,10 +8,20 @@ ln -s /usr/share/elasticsearch /usr/share/java/elasticsearch
 /usr/share/elasticsearch/bin/plugin install -b com.floragunn/search-guard-ssl/${SG_SSL_VER}
 /usr/share/elasticsearch/bin/plugin install -b com.floragunn/search-guard-2/${SG_VER}
 /usr/share/elasticsearch/bin/plugin install io.fabric8/elasticsearch-cloud-kubernetes/${ES_CLOUD_K8S_VER}
-/usr/share/elasticsearch/bin/plugin install io.fabric8.elasticsearch/openshift-elasticsearch-plugin/${OSE_ES_VER}
+#/usr/share/elasticsearch/bin/plugin install io.fabric8.elasticsearch/openshift-elasticsearch-plugin/${OSE_ES_VER}
+/usr/share/elasticsearch/bin/plugin install file:///openshift-elasticsearch-plugin-2.4.1.5-SNAPSHOT.zip
```
portante: Is this the new name?

elasticsearch/Dockerfile
```diff
@@ -34,7 +34,7 @@ RUN rpm --import https://packages.elastic.co/GPG-KEY-elasticsearch && \
                 java-1.8.0-openjdk-headless \
                 elasticsearch-${ES_VER} && \
     yum clean all
-
+ADD openshift-elasticsearch-plugin-2.4.1.5-SNAPSHOT.zip /
```
portante: Is this the right name?

elasticsearch/install.sh
jcantrill: This is really temporary until we publish the plugin.  Ill update when that merges and we agree things are good

elasticsearch/Dockerfile
jcantrill: This line will be removed

elasticsearch/install.sh
ewolinetz: hopefully this will change once the plugin is released?

elasticsearch/Dockerfile
ewolinetz: this should change once the other PR is merged and released

elasticsearch/run.sh
```diff
@@ -100,8 +104,30 @@ rm -f $LOG_FILE
 exit 1
 }
 
+seed_searchguard(){
+    /usr/share/elasticsearch/plugins/search-guard-2/tools/sgadmin.sh \
+        -cd ${HOME}/sgconfig \
+        -i .searchguard.${HOSTNAME} \
```
ewolinetz: does this delete the index if it already exists? or does it just ignore the creation of it?

jcantrill: I do not believe so.  I think it simply updates the index.  The log produces for each file in the directory:

```
Will update 'internalusers' with /opt/app-root/src/sgconfig/sg_internal_users.yml
SUCC: Configuration for 'internalusers' created or updated
```

ewolinetz: gotcha, i was wondering where we provided the `Deletes SG index on start up if exists` logic is why... (taken from linked PR)

elasticsearch/install.sh
```diff
@@ -8,10 +8,20 @@ ln -s /usr/share/elasticsearch /usr/share/java/elasticsearch
...
 chmod -R o+rx /etc/elasticsearch
+chmod +x /usr/share/elasticsearch/plugins/search-guard-2/tools/sgadmin.sh
+
+PASSWD=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 20 | head -n 1)
+cat > ${HOME}/sgconfig/sg_internal_users.yml << CONF
+---
+  $(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 8 | head -n 1):
+    hash: $PASSWD
+CONF
+
```
richm: ok for now - but we should figure out how to get rid of the need for the sg_internal_users file/index

richm:
lgtm

elasticsearch/install.sh
jcantrill: I think we will be good here.  Per the documentation, the hash is supposed to be 'bcrypt'.  I presume if someone ever tried to use it there would be some sort of error while unencrypting.  Regardless, I would prefer to see some changes to SG which ignored this file if it was not configured for it.

lukas-vlcek:
Actually, I am thinking that we shall address node and cluster initialization in [Kubernetes readiness probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/) instead of `run.sh` script.

Notice that we will have to change how `number_of_shards` and `number_of_replicas` are setup and propagated on index creation:

- Starting with ES 5.x [index level settings are not allowed in `elasticsearch.yml`](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_50_settings_changes.html#_index_level_settings)
- Related issue: https://github.com/elastic/elasticsearch/issues/18073

jcantrill:
[test]

jcantrill:
Manually merged for 1.4 branch

PR #362, opened 2017-03-24
min master fix to avoid es split brain

jcantrill:
cc @portante

### DIFF
--- a/deployer/templates/es.yaml
+++ b/deployer/templates/es.yaml
@@ -81,6 +81,9 @@ objects:
                 name: INSTANCE_RAM
                 value: ${INSTANCE_RAM}
               -
+                name: MIN_MASTERS
+                value: ${MIN_MASTERS}
+              -
                 name: NODE_QUORUM
                 value: ${NODE_QUORUM}
               -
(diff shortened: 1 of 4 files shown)


deployer/templates/es.yaml
```diff
@@ -147,9 +150,13 @@ objects:
     description: "Numeric ID of a supplemental group for storage volume access."
     name: STORAGE_GROUP
     value: ${STORAGE_GROUP_DEFAULT}
+  -
+    name: MIN_MASTERS
```
ewolinetz: Can we get a `description` on here?

richm:
fyi, you don't need bc:
```
 +  es_min_masters=$(expr $es_cluster_size / 2 + 1)
```
expr is a bourne shell built-in

deployer/templates/es.yaml
```diff
@@ -147,9 +150,14 @@ objects:
     description: "Numeric ID of a supplemental group for storage volume access."
     name: STORAGE_GROUP
     value: ${STORAGE_GROUP_DEFAULT}
+  -
+    description: "The mimimum number of masters that must be present in the ES cluster for the master election process to complete"
```
portante: If we can, tying it back to the actual Elasticsearch parameter name it lands in would be really great, too.

jcantrill:
@richm good to know.  Started to modify and was reminded this is on 1.4.  Going to pass on fixing as it works and will not be used going forward.

ewolinetz:
for the sake of prosperity -- we missed adding an upgrade step/check in scripts/upgrade.sh for the DC patch

PR #365, opened 2017-03-28
Add document describing requirements to file an issue

jcantrill:
cc @portante

This is a work in progress to document what we would like when people file issues.  I think some of this bit we could automate and I think maybe @lukas-vlcek was playing with a utility or script we can ref to get information.

### DIFF
--- a/docs/issues.md
+++ b/docs/issues.md
@@ -1,0 +1,36 @@
+# Filing An Issue
+The Openshift Aggregated Logging integration is a multicomponent technology stack whose individual 
+parts must all work together to gather and archive logs.  It is important to include as much information 
+as possible when issues arise in order to properly debug and diagnose.  Run 
+**ADD SCRIPT OR UTILITY HERE** which will gather the desired information.  Alternatively, you can gather 
+them manually. 
+
+**NOTE:** It is your responsiblity as the reporter of an issue to redact any sensitive information from
+any and all artifacts that are included as part of the issue.
+
+Please include as much of the following to help us expedite a resolution:
+
+### OpenShift Cluster Details
+1. Cluster version (e.g. OCP v3.5)
+2. Cluster infrastructure (e.g. AWS, GCP)
+3. Number of infrastructure nodes and size (e.g. CPU, RAM)
+4. Number of compute nodes and size (e.g. CPU, RAM)
+5. Logging Image versions
+
+**ADD NOTE HERE ON HOW TO RETRIEVE IMAGE VERSION FROM OCP AND ORIGIN**
+
+### Installer Details
+1. How was logging installed (e.g. Openshift Ansible/Installer, deployer)
+2. Openshift installer version
+3. Ansible inventory file
+4. Ansible installer logs
+
+### Logging Component Details
+**MAYBE WE CAN GRAB SOME OF THESE BY LABEL**
+1. Elasticsearch logs: `SOME NON-OC LOGS COMMAND`
+2. Other component logs: `oc logs $PODNAME`
+3. Fluentd spec: `oc get ds logging-fluentd -o yaml`
+4. Other component specs: `oc get dc $DCNAME -o yaml`
+5. Configmaps: `oc get configmaps $CONFIGMAP -o yaml`
+6. Events: `oc get events $PODNAME -o yaml`
+7. Output of https://github.com/openshift/openshift-tools/blob/prod/scripts/monitoring/cron-send-logging-checks.py
--- a/README.md
+++ b/README.md
@@ -66,3 +66,8 @@ start and stop order.
 Determining the health of an EFK deployment and if it is running can be assessed
 by running the `check-EFK-running.sh` and `check-logs.sh` [e2e tests](hack/testing/).
 Additionally, see [Checking EFK Health](docs/checking-efk-health.md)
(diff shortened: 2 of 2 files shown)


jcantrill:
cc @pweil- Tag you in for good measuer

docs/issues.md
```diff
@@ -0,0 +1,26 @@
+# Filing An Issue
```
pweil-: can we add this as the GH template?

docs/issues.md
```diff
@@ -0,0 +1,26 @@
...
+2. Openshift installer version
+3. Ansible inventory file
+4. Ansible installer logs
+
+### Logging Component Details
+**MAYBE WE CAN GRAB SOME OF THESE BY LABEL**
+1. Elasticsearch logs: `SOME NON-OC LOGS COMMAND`
+2. Other component logs: `oc logs $PODNAME`
+3. Fluentd spec: `oc get ds logging-fluentd -o yaml`
+4. Other component specs: `oc get dc $DCNAME -o yaml`
```
pweil-: don't you want the actual pod definitions so that you can see all defaults that get applied?

docs/issues.md
```diff
@@ -0,0 +1,26 @@
...
+
+### Installer Details
+1. How was logging installed (e.g. Openshift Ansible/Installer, deployer)
+2. Openshift installer version
+3. Ansible inventory file
+4. Ansible installer logs
+
+### Logging Component Details
+**MAYBE WE CAN GRAB SOME OF THESE BY LABEL**
+1. Elasticsearch logs: `SOME NON-OC LOGS COMMAND`
```
pweil-: probably want a "don't forget to redact any sensitive info" reminder

docs/issues.md
```diff
@@ -0,0 +1,26 @@
...
+2. Other component logs: `oc logs $PODNAME`
+3. Fluentd spec: `oc get ds logging-fluentd -o yaml`
+4. Other component specs: `oc get dc $DCNAME -o yaml`
+5. Configmaps: `oc get configmaps $CONFIGMAP -o yaml`
+6. Events: `oc get events $PODNAME -o yaml`
```
pweil-: Are events good enough?  How often do you need master or kubelet logs when debugging?

docs/issues.md
jcantrill: We can.  I think we should also provide something as the BZ template too

docs/issues.md
jcantrill: Is that different then the DC?  Should the resulting POD be what is defined in the DC/DS?  Do we want both?

docs/issues.md
jcantrill: Master and node logs would be a good addition, though I have no concrete knowledge where to source those.

docs/issues.md
pweil-: in the DC/RC the runtime environment is not fully defined.  Upon submission all the defaulting code for the pod is run.  Things like ImagePullPolicy, SecurityContext, etc are then all filled in.  Basically anything that runs in defaulters and pod admission plugins will not be viewed in the DC.  Having both may be valuable.

docs/issues.md
pweil-: Usually `journalctl` or leave it up to the admin to figure it out

portante: Also include output from Wesley Hearns ES [health check](https://github.com/openshift/openshift-tools/blob/prod/scripts/monitoring/cron-send-logging-checks.py) [1].

[1] https://github.com/openshift/openshift-tools/blob/prod/scripts/monitoring/cron-send-logging-checks.py

docs/issues.md
```diff
@@ -0,0 +1,26 @@
...
+4. Number of compute nodes and size (e.g. CPU, RAM)
+5. Logging Image versions
+
+**ADD NOTE HERE ON HOW TO RETRIEVE IMAGE VERSION FROM OCP AND ORIGIN**
+
+### Installer Details
+1. How was logging installed (e.g. Openshift Ansible/Installer, deployer)
+2. Openshift installer version
+3. Ansible inventory file
+4. Ansible installer logs
```
ewolinetz: We should also add deployer logs (if applicable)

jcantrill:
It would be nice to package this tool into the image if it proves to be useful, but this alone could be a non-starter to doing that:
```
If you encounter OutOfMemoryExceptions, use the DIAG_JAVA_OPTS environment variable to set an -Xmx value greater than the standard 2g. Start with -Xmx4g and move up from there if necessary.
```
We would not have enough memory.

docs/issues.md
jcantrill: added

docs/issues.md
jcantrill: Added

docs/issues.md
jcantrill: Updated

richm:
We now have logging support in https://github.com/openshift/openshift-ansible/tree/master/roles/openshift_health_checker
https://github.com/openshift/openshift-ansible/tree/master/roles/openshift_health_checker/openshift_checks/logging

Should we focus our efforts to enhance that, or elsewhere?

jcantrill:
[merge]

PR #367, opened 2017-04-03
enable console logging for testing

richm:
There are several places in the CI test code that expect `oc logs $espod`
will return some value.  In openshift-ansible the default now is to disable
console logging for ES which will make `oc logs $espod` not work.  This
commit enables ES console logging for CI test purposes.

@jcantrill @ewolinetz @nhosoi @portante PTAL
[test]

### DIFF
--- a/hack/testing/init-log-stack
+++ b/hack/testing/init-log-stack
@@ -33,6 +33,7 @@ openshift_logging_image_prefix=$imageprefix
 openshift_logging_use_ops=${ENABLE_OPS_CLUSTER:-False}
 openshift_logging_fluentd_use_journal=${USE_JOURNAL:-}
 openshift_logging_fluentd_journal_read_from_head=${JOURNAL_READ_FROM_HEAD:-False}
+openshift_logging_es_log_appenders=['console']
 
 EOL
 


hack/testing/init-log-stack
```diff
@@ -33,6 +33,7 @@ openshift_logging_image_prefix=$imageprefix
 openshift_logging_use_ops=${ENABLE_OPS_CLUSTER:-False}
 openshift_logging_fluentd_use_journal=${USE_JOURNAL:-}
 openshift_logging_fluentd_journal_read_from_head=${JOURNAL_READ_FROM_HEAD:-False}
+openshift_logging_es_log_appenders=['file','console']
```
ewolinetz: do we want to have both file and console? can't we get away with just using 'console' for this PR?

jcantrill: we run the risk of the feedback loop but that seems to only be an extreme case.  console is probably sufficient.

portante: We should always be against using console when logs from the console get fed back into the Elasticsearch instance from which they were emitted.

ewolinetz: To clarify, this is only for testing. Hence me wondering if we need to set up the 'file' appender since I don't believe we have any tests that explicitly check the file-based logs of Elasticsearch.

ewolinetz:
other than my question on if `file` is necessary for testing LGTM

richm:
@ewolinetz latest patch only enables console, not file - afaict we do not need file logging for the CI testing

richm:
[test]

ewolinetz:
[merge]

richm:
[merge]

PR #368, opened 2017-04-06
fix 1439356. Add logic to upgrade for min master

jcantrill:


deployer/scripts/upgrade.sh
```diff
@@ -754,6 +772,9 @@ function upgrade_logging() {
         6)
           common_data_model=true
           ;;
+        7)
+          update_es_for_min_masters
+          ;;
         $LOGGING_VERSION)
```
richm: Do you need to change `LOGGING_VERSION=7` in the `deployer/Dockerfile`?

deployer/scripts/upgrade.sh
```diff
@@ -754,6 +772,9 @@ function upgrade_logging() {
         6)
           common_data_model=true
           ;;
+        7)
```
ewolinetz: to detect that this is the current version we'll also need to update the variable in the Dockerfile and `getDeploymentVersion`

deployer/scripts/upgrade.sh
jcantrill: bumped

deployer/scripts/upgrade.sh
jcantrill: added check

jcantrill:
@ewolinetz need to test but I hope this test is sufficient

hack/testing/test-upgrade.sh
```diff
@@ -243,6 +243,11 @@ function verifyUpgrade() {
   local checkMigrate=${2:-false}
   local checkCDMMigrate=${3:-false}
 
+### check min_master migration
```
ewolinetz: you will also need to create a function that will remove the `MIN_MASTER` entry before running the upgrade, otherwise this will pass simply because a new install provides it

hack/testing/test-upgrade.sh
```diff
@@ -323,6 +338,7 @@ removeFluentdConfigMaps
 removeEsCuratorConfigMaps
 removeAdminCert
 removeCurator
+removeMinMaster
```
ewolinetz: can you also add this for the second rerun? after that feel free to kick off a test

jcantrill: I'll have to manually test since we have no CI hooked up to 1.4

hack/testing/test-upgrade.sh
jcantrill: added.

jcantrill:
@richm Updated to use NODE_QUORUM since that is the same value calculated in earlier release

richm:
lgtm

jcantrill:
[test]

jcantrill:
@ewolinetz  @richm  Added a test but we dont have CI attached so it wont do much good.  This will also need to be adjusted based on the backports that @richm was doing since the logging version conflicts

richm:
[test]

richm:
I think we either have to 
* modify the CI test to work with the latest devenv - but then we need to be careful that we don't accidentally introduce some code that won't work with 1.4
* modify the CI test to work with released 1.4/3.4 images - not trivial
* create/find a 1.4 devenv image - ???

jcantrill:
@richm CI would be the best option, but until we have master fixed up I think this is going to be difficult.  The other option instead of using a 1.4 devenv is to modify the 1.4 setup to just use a 1.4 image or binaries that already exist.  We car more about a running 1.4 cluster then the actual devenv env

jcantrill:
[test]

deployer/scripts/upgrade.sh
```diff
@@ -58,6 +58,12 @@ function getDeploymentVersion() {
     return
   fi
 
+  # check for ES MIN_MASTER env var
+  if [ -z "$(oc get configmap logging-elasticsearch -o yaml | grep 'MIN_MASTERS')" ]; then
```
richm: This means "if MIN_MASTERS is not in the configmap, return 8" - is this as intended?

deployer/scripts/upgrade.sh
```diff
@@ -849,6 +876,7 @@ function upgrade_logging() {
   # 5 -- update ES for 2.x
   # 6 -- add aliases for common data model
   # 7 -- add index patterns for Kibana for common data model
+  # 8 -- min_master config change
```
richm: you don't need this any more

deployer/scripts/upgrade.sh
```diff
@@ -897,6 +925,9 @@ function upgrade_logging() {
           add_index_pattern_config
           add_common_data_model_plugin
           ;;
+        8)
+          update_es_for_min_masters
+          ;;
```
richm: you don't need this any more

hack/testing/test-upgrade.sh
```diff
@@ -293,6 +298,9 @@ function verifyUpgrade() {
   local checkMigrate=${2:-false}
   local checkCDMMigrate=${3:-false}
 
+### check configmap patched for MIN_MASTERS
+  [[ -z "$(oc get configmap logging-elasticsearch | grep MIN_MASTERS)" ]] && return 1
```
richm: I think you want to use `-n` here 
The way this currently reads is "if the configmap does not have MIN_MASTERS, error"

richm:
ship it

PR #369, opened 2017-04-06
add normal user for access control checking

richm:
The previous commit did had some leftover debugging code.
@jcantrill @ewolinetz PTAL
[test]

### DIFF
--- a/hack/testing/logging.sh
+++ b/hack/testing/logging.sh
@@ -142,7 +142,7 @@ os::cmd::expect_success 'oadm router'
 ######### logging specific code starts here ####################
 
 os::cmd::expect_success "oadm new-project logging --node-selector=''"
-os::cmd::expect_success "oc project logging"
+os::cmd::expect_success "oc project logging > /dev/null"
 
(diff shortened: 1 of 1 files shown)


hack/testing/logging.sh
```diff
@@ -142,7 +142,7 @@ os::cmd::expect_success 'oadm router'
 ######### logging specific code starts here ####################
 
 os::cmd::expect_success "oadm new-project logging --node-selector=''"
-os::cmd::expect_success "oc project logging"
+os::cmd::expect_success "oc project logging > /dev/null"
```
ewolinetz: just out of curiosity is this just to suppress the output that we may already be on this project? e.g. this has no impact on the expect_success

jcantrill:
LGTM

hack/testing/logging.sh
richm: @ewolinetz correct - it just reduces the noise in our tests - has no impact on expect_success.

richm:
[merge]

PR #371, opened 2017-04-07
Change timeout of probe script

### LINKED ISSUE #370: Readiness probe timeout should be set less than 1sec
Elasticsearch readiness probe script was introduced in https://github.com/openshift/origin-aggregated-logging/pull/358

There is known issue that prevents probe script timeout control from Docker/Kubernetes. See https://github.com/openshift/openshift-ansible/pull/3733#pullrequestreview-31314038

The timeout has to be controlled from within the probe script itself and it is probably safer to set it to less than 1 sec.

I do not know if we control version of curl that is used by probe script but be aware that (taken from `man curl`):

````
-m, --max-time <seconds>
              Maximum time in seconds that you allow the whole operation to take.  This is useful for preventing your batch jobs from hanging
              for  hours  due to slow networks or links going down.  Since 7.32.0, this option accepts decimal values, but the actual timeout
              will decrease in accuracy as the specified timeout increases in decimal precision.  See also the --connect-timeout option.

              If this option is used several times, the last one will be used.
````

Notice: "`Since 7.32.0, this option accepts decimal values, ...`"

### LINKED ISSUE #372: Set readiness probe timeout to 4 sec
We shall keep readiness probe timeout below readiness probe `periodSeconds` value.
As of writing we are starting with the following probe config (see https://github.com/openshift/openshift-ansible/pull/3733):

    initialDelaySeconds: 5
    timeoutSeconds: 4
    periodSeconds: 5

This gives us opportunity to set the timeout to `4` seconds.

Notice: **We have to control the probe script timeout within the script itself**, we can not rely on the `timeoutSeconds` configuration value due to this issue: https://github.com/kubernetes/kubernetes/issues/26895

### PR DESCRIPTION
**Pull request by lukas-vlcek**
Closes #372

### DIFF
--- a/elasticsearch/probe/readiness.sh
+++ b/elasticsearch/probe/readiness.sh
@@ -20,7 +20,7 @@
 ES_REST_BASEURL=https://localhost:9200
 EXPECTED_RESPONSE_CODE=200
 secret_dir=/etc/elasticsearch/secret
-max_time=${max_time:-1}
+max_time=${max_time:-4}
 
 response_code=$(curl -s -X HEAD \
     --cacert $secret_dir/admin-ca \

### CODE REVIEW

File: elasticsearch/probe/readiness.sh [hunk 1]
```diff
@@ -20,7 +20,7 @@
 ES_REST_BASEURL=https://localhost:9200
 EXPECTED_RESPONSE_CODE=200
 secret_dir=/etc/elasticsearch/secret
-max_time=${max_time:-1}
+max_time=${max_time:-0.8}
```
rhcarvalho: I don't understand why are you changing it from 1s to 0.8?! Based on what?

rhcarvalho: A too short timeout might cause more trouble than a longer timeout. In a cluster under load / slow traffic consecutive readiness probes could fail and incorrectly take down Elasticsearch when it is otherwise functional.

rhcarvalho: By "take down" I mean the Service would stop routing requests to the pods that failed the readiness probe.

lukas-vlcek:
@rhcarvalho 

- This script is configured to be used as a readiness probe in Openshift and the `timeoutSeconds` was set to `5`.
- As you pointed out earlier the `timeoutSeconds` configuration is ignored for readiness probes meaning if readiness probe is setup up then it will be called ever `1` sec and we can not do anything about that (did I got it right?)
- Originally the timeout in this script was setup to `1` sec as well which you recommended me to lower down a bit to avoid overlapping probe calls
- So I change the timeout in this script from `1` to `0.8`

HTH

lukas-vlcek:
And yes I agree. Timeout of `0.8` seconds opens the question if we should use readiness probes at all. To me this seems very questionable.

lukas-vlcek:
Discussed with @richm: May be we can reimplement the readiness probe into go in which case the `timeoutSeconds` configuration for the probe will not be ignored?

lukas-vlcek:
Updated the timeout to `4` seconds.

ewolinetz:
bump on this review since its blocking another pr

jcantrill:
[test]

jcantrill:
[merge]

PR #373, opened 2017-04-07
Bug 1439504 - Kibana show raw json in "message" field for pods messages in json format

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1439504
The problem is that the k8s record transformer filter was wiping
out the "message" field added by the k8s meta filter when reading
logs from the journal.  The fix is to preserve that field if present.
@jcantrill @ewolinetz PTAL
[test]


richm:
[merge]

PR #375, opened 2017-04-10
Bug 1439504 - Kibana show raw json in "message" field for pods messages in json format

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1439504
The problem is that the k8s record transformer filter was wiping
out the "message" field added by the k8s meta filter when reading
logs from the journal.  The fix is to preserve that field if present.

Do not set openshift_release in CI test if OS_VERSION is unset.
Instead, let ansible figure out the version from the installed software.

(cherry picked from commit 9a6274451d850c4d3e42ba35781b607659255c02)
@jcantrill @ewolinetz PTAL backport to release-1.5


richm:
@jcantrill or @ewolinetz since this is for release-1.5 I can't merge - can one of you guys merge this?  Thanks!

PR #378, opened 2017-04-14
Add support for mux - secure_forward service

richm:
See the docs/proposals/mux-logging-service.md for details
@jcantrill @ewolinetz @nhosoi @lukas-vlcek @t0ffel @portante
[test]


docs/proposals/mux-logging-service.md
```diff
@@ -0,0 +1,305 @@
+Introduction
```
jcantrill: Is this intended to more then a 'proposal'?  Do we expect to debate its introduction and design?  If not, I would move it to live under the 'docs' directory.

docs/proposals/mux-logging-service.md
```diff
@@ -0,0 +1,305 @@
...
+*mux* can also be exposed as an external service.  This allows the ability to
+set up an OpenShift cluster to act as a central logging service for other
+projects and products outside of the cluster.
+
+### Server Side Configuration ###
+
+There is a deploymentConfig - `logging-mux`.  This is the central logging
+service.  There is also a `logging-mux` Service which exposes the
+secure_forward listener port using `externalIPs`.  There is a Secret
+`logging-mux` which holds the server `cert` and `key`, the `ca` cert, and the
```
jcantrill: Side note.  Need to refactor this I suspect when client cert generation becomes available to us.

docs/proposals/mux-logging-service.md
```diff
@@ -0,0 +1,305 @@
...
+essentially the same as in the logging-fluentd configmap.
+
+There are environment variables for the logging-mux DC config:
+
+* `USE_MUX` - `true`/`false` - default is `false` - if `true`, the logging-mux
+  will act as a service inside the cluster for the Fluentd collectors running
+  on each node.  This allows the per-node Fluentd collectors to read and ship
+  logs off the node as fast as possible to the logging-mux service, which will
+  normalize and enrich with Kubernetes metadata, and store in Elasticsearch.
+* `USE_MUX_SERVICE` - `true`/`false` - default is `false` - if `true`, the
```
jcantrill: We should consider making this more explicit like 'ALLOW_EXTERNAL_RECORDS'.  Something that represents permission to use the service from somplace unexpected.

docs/proposals/mux-logging-service.md
```diff
@@ -0,0 +1,305 @@
...
+
+There are several new Ansible parameters which can be used with the
+`openshift_logging` role:
+
+* `openshift_logging_use_mux` - `True`/`False` - default `False` - if
+  `True`, create the mux Service, DeploymentConfig, Secrets, ConfigMap, etc.
+* `openshift_logging_use_mux_client` - `True`/`False` - default `False` - if
+  `True`, configure the Fluentd collectors running on each node to send the raw
+  logs to mux instead of directly to Elasticsearch
+* `openshift_logging_use_mux_service` - `True`/`False` - default `False` - if
```
jcantrill: Same comment as previously

docs/proposals/mux-logging-service.md
```diff
@@ -0,0 +1,305 @@
...
+
+These logs will be written to Elasticsearch in an index with a name like the
+following: `project.$namespace_name.$namespace_uuid.YYYY.MM.DD`
+* `project` is the literal string "project"
+* `$namespace_name` is the name of the OpenShift namespace
+* `$namespace_uuid` is the UUID of the OpenShift namespace - fluentd will look
+  this up from OpenShift/Kubernetes using the kubernetes metadata filter
+* `YYYY.MM.DD` is the year, month, and day
+
+The namespaces used in the tag must be added ahead of time to OpenShift - for
```
jcantrill: Is there a scenario where this is not already the case?  You can not run a container without it being part of a project which must exist ahead of time.

docs/proposals/mux-logging-service.md
```diff
@@ -0,0 +1,305 @@
...
+`project.namespacename.namespaceuuid.YYYY.MM.DD`, similar to how OpenShift
+container records are written.
+
+### Details ###
+
+mux *does not* read the logs/journal on the system - it does not mount
+`/var/log`, etc.  This is why it is a DeploymentConfig and not a DaemonSet.  As
+such, mux does not need to mount `/var/log`, `/var/lib`, and `/run/log`.
+However, since it does need to know the absolute time, it still needs to mount
+`/etc/localtime` from the host.
```
jcantrill: Is this the only way to get the proper time from the host?  @pweil- ?

docs/proposals/mux-logging-service.md
```diff
@@ -0,0 +1,305 @@
...
+
+The `logging-mux` secret is mounted on `/etc/fluent/muxkeys` and holds the mux
+TLS server `cert` and `key`, as well as the `ca` cert and the `shared_key`.
+
+These records are sent to the `@MUX` label.  The first thing here is this
+filter:
+
+    <filter kubernetes.var.log.containers.**>
+
+Since mux has no way to know if the Kubernetes records were read from
```
jcantrill: Consider making this more an affirmative statement of what we are doing.  Is it significant to know we use one or the other?  I think its more significant to know: what it does, how it varies based on a given condition.

docs/proposals/mux-logging-service.md
```diff
@@ -0,0 +1,305 @@
...
+`$1` is the value of the `mux_namespace_name` field.  This tag format was
+chosen because it will match the k8s meta filter plugin match
+
+    <filter kubernetes.**>
+
+and will match that plugin's tag match pattern (if for some reason it is using the
+`json-file` method), but it will not match any of the other kubernetes log
+record filters.
+
+Then all records are passed to the `@INGRESS` label, where there are two
```
jcantrill: We should consider adding a flow diagram

fluentd/run.sh
```diff
@@ -14,29 +14,74 @@ docker_uses_journal() {
...
-        if [ -d /var/log/journal ] ; then
-            export JOURNAL_SOURCE=/var/log/journal
-        else
-            export JOURNAL_SOURCE=/run/log/journal
+if [ "${USE_MUX_SERVICE:-}" = true ] ; then
+    # mux service implies mux
+    export USE_MUX=true
+fi
+
+if [ "${USE_MUX:-}" != "true" ] ; then
```
jcantrill: Is it intentional that the bool is quoted where the one before and afterwards are not?

ewolinetz:
Is this a specific usage of secure-forward with a specific fluentd config?

docs/proposals/mux-logging-service.md
richm: I want to emphasize this point for users of common logging that aren't otherwise users of OpenShift, that may be unfamiliar with how OpenShift works.  Also, these namespaces will only be used for authn/authz/access control - there will be no other resources (pods, dc, etc.) associated with these namespaces.

docs/proposals/mux-logging-service.md
richm: In this section I wanted to list some of the design decisions and why they were made.

docs/proposals/mux-logging-service.md
richm: ok

fluentd/run.sh
richm: no - in bash in a comparison, `true` is the same as `"true"`

richm:
> Is this a specific usage of secure-forward with a specific fluentd config?

@ewolinetz Yes.  You could consider that this is a specific implementation of your enricher proposal, with the additional ability to be a secure_forward listener for log records coming in from outside of the cluster.

fluentd/run.sh
```diff
@@ -14,29 +14,74 @@ docker_uses_journal() {
...
 
 IPADDR4=`/usr/sbin/ip -4 addr show dev eth0 | grep inet | sed -e "s/[ \t]*inet \([0-9.]*\).*/\1/"`
 IPADDR6=`/usr/sbin/ip -6 addr show dev eth0 | grep inet6 | sed "s/[ \t]*inet6 \([a-f0-9:]*\).*/\1/"`
 export IPADDR4 IPADDR6
 
 CFG_DIR=/etc/fluent/configs.d
-ruby generate_throttle_configs.rb
+if [ "${USE_MUX:-}" = "true" ] ; then
+    # copy our standard mux configs to the openshift dir
+    cp $CFG_DIR/input-*-mux.conf $CFG_DIR/openshift
```
t0ffel: imho, this is very hard to maintain. Meaning that in order to understand what config files are in a specific fluentd pod you can't just look in the folder in the repository (neither you can simply look in the configmap).

My suggestion would be to have a separate folder w/ the mux-specific setup.
I think the logic that builds all the configuration files should be outside of container runtime execution as much as possible.

richm: @t0ffel Can you be more specific?  I don't understand what you mean.  Do you mean - have a subfolder of fluent/configs.d called "mux"?  And copy files into that folder when mux is enabled, and remove files from that folder when mux is disabled?

richm: I'm trying to do 2 things here 1) preserve as much as possible the ability to reuse the existing fluentd image without drastic changes and 2) rely on the fluentd feature that missing wildcard files are not an error

t0ffel: @richm I think it is easier if configmaps contain all the configuration running fluentd process.
i.e. ansible builds configmaps based on `USE_MUX`, then fluentd simply mounts the configmap - doesn't have any run time logic associated w/ copying around the configs.

For config folder structure, I'd advocate for the layout that I have here: https://github.com/t0ffel/fluentd-openshift/tree/master/configs/node-collector but I think that's a different conversation - not in scope of this PR.

richm: ok - so all of the mux config should be configmaps, as separate files in the logging-mux configmap, mounted at fluent/configs.d/user

richm: > i.e. ansible builds configmaps based on USE_MUX, then fluentd simply mounts the configmap - doesn't have any run time logic associated w/ copying around the configs.

@t0ffel This would require redesigning the way fluentd works too.  I think that is outside the scope of this change.  That is, my proposed commit does not change the way logging/fluentd works now - files are copied around at runtime now - I'm not changing that.  My proposed change does allow you to override the default mux config files by using the logging-mux configmap, but those files are still copied around at runtime.  My change does not make it any harder to maintain or debug.  For example, in the current code before my change, the only way to find out what fluentd configs are being used is to oc rsh or exec into the fluentd pod and ls/cat the files under /etc/fluent.  I don't see any easy way to convert to a configmap-only, no copying scheme for mux without either major changes to fluentd, or diverging the mux and fluentd configuration.

richm:
Any more comments?  If not, I'd like to get this merged.

ewolinetz:
@richm so this is purely a post-installation configuration that needs to be done? Is the intent that both fluentd and mux would be hosted on OCP at the same time? If so is mux also a daemonset?

Currently it looks like we would not be able to do that as they use the same daemonset and env vars cause it to use a different configuration. I'm wondering if it would make more sense to push some logic also into the ansible role (is there a PR already open for that?)

richm:
> so this is purely a post-installation configuration that needs to be done?

The intention is that all of this works in conjunction with https://github.com/openshift/openshift-ansible/pull/3932 - so that you can install OpenShift + logging + mux in one shot.  Of course you could configure all of this post-install as well.

> Is the intent that both fluentd and mux would be hosted on OCP at the same time?

Yes.  For example, fluentd could be configured with MUX_CLIENT=true, in which case it would send its raw logs to mux for filtering and enrichment.  In this case, mux is configured to get the logs from the node fluentds, enrich them, filter them, and send them to elasticsearch.  mux can also be configured to do this, and serve requests from outside the cluster, at the same time.  Or not.

> If so is mux also a daemonset?

No.  mux is a deploymentconfig - since it doesn't read /var/log or journal, it isn't tied to a node, so doesn't need to be a daemonset.

Yes, there is an openshift-ansible PR https://github.com/openshift/openshift-ansible/pull/3932

fluentd/run.sh
```diff
@@ -14,29 +14,74 @@ docker_uses_journal() {
...
-            export JOURNAL_SOURCE=/run/log/journal
+if [ "${MUX_ALLOW_EXTERNAL:-}" = "true" ] ; then
+    # mux service implies mux
+    export USE_MUX=true
+fi
+
+if [ "${USE_MUX:-}" != "true" ] ; then
+    if [ "${USE_JOURNAL:-}" != false ] ; then
+        if [ -z "${JOURNAL_SOURCE:-}" ] ; then
+            if [ -d /var/log/journal ] ; then
```
t0ffel: are you sure that existence of the directory guarantees that logs go to that dir? why not just listen to both always?

richm: @t0ffel no - but this PR does not change that behavior - please open a separate issue, or suggest a way that fluentd can know which directory to listen to, or suggest a way that fluent can listen to both /var/log/journal and /run/log/journal from the host.

richm:
> btw. is there any intention to move the configs to openshift-ansible repo? Right now it's not easy to track configuration changes spanned between 2 repos.

I don't know.  @jcantrill that might be a good candidate for a tech debt issue

richm:
[test]

richm:
We're just waiting for https://github.com/openshift/openshift-ansible/pull/3932 to merge . . .

richm:
[merge]

PR #382, opened 2017-04-19
Add Elasticsearch memory_lock

### LINKED ISSUE #206: Make sure swapping is disabled for Elasticsearch
Especially for the performance reasons we need to make sure Elasticsearch is not subject of swapping.

We need to focus on two levels here:
- At the OS/Elasticsearch level. This means we can use techniques recommended [here](https://www.elastic.co/guide/en/elasticsearch/reference/1.5/setup-configuration.html#setup-configuration-memory) to disable swapping at the OS level
- At the Docker level using `docker run [OPTIONS]`. See
  - https://docs.docker.com/v1.10/engine/reference/run/#runtime-constraints-on-resources
  - https://docs.docker.com/v1.10/engine/reference/run/#swappiness-constraint

lukas-vlcek:
Closes #206

### DIFF
--- a/deployer/conf/elasticsearch.yml
+++ b/deployer/conf/elasticsearch.yml
@@ -1,6 +1,8 @@
 cluster:
   name: ${CLUSTER_NAME}
 
+bootstrap.memory_lock: true
+
 script:
   inline: on
   indexed: on


lukas-vlcek:
Notice, this is applicable to Elasticsearch 2.4.0 and newer. In previous ES versions the setting was called `bootstrap.mlockall`.

See [2.4.0 breaking changes](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/breaking-changes-2.4.html).

jcantrill:
@lukas-vlcek  Can you please also update the run.sh script to set min and max heap to be the same.  This seems like a reasonable PR to append it too.

lukas-vlcek:
@jcantrill My take on this is https://github.com/openshift/origin-aggregated-logging/pull/389

lukas-vlcek:
@jcantrill @ewolinetz @richm @portante

As far as I know we should be able to rely on disabled swap on the host going forward (via `swapoff -a`). In such case the question is if we need to bring this change into master. (That is the reason why I created new PR for `Xms==Xmx` thing)

IMO it does not hurt to bring this into master as long as we are on ES 2.x. However, once we migrate to ES 5.x we may need to revert this change due to [bootstrap check](https://www.elastic.co/guide/en/elasticsearch/reference/5.0/_memory_lock_check.html). I do not know if that bootstrap check will fail in this case.

jcantrill:
[test]

richm:
Does this change need for elasticsearch to be running in privileged mode?

portante:
@richm, good catch.  I would think so.  @lukas-vlcek, did you try this out under openshift somewhere?

jcantrill:
[merge]

portante:
@jcantrill, I don't think this should be merged.

jcantrill:
@portante What is the reason we wish to hold off merging this change?

richm:
@jcantrill Lukáš needs to do more testing to confirm that this actually works, and how to configure the machine (e.g. ulimit, etc.) in order to make it work.  I think in ES 2.x it will silently fail if the machine is not set up correctly, and in ES 5.x it will hard fail.  @lukas-vlcek

lukas-vlcek:
@jcantrill Added "Do not merge" label for now.

We really need to get to the bottom of this and deliver clear documentation (backed by testing) around this setting. When we started looking at this (#206) it was not known OCP would decide to disable swap at the OS level so we need to reflect new situation now.

Even as for assumption that it will failing hard with ES 5.x - I am still not 100% sure about this. There is bootstrap check that will fail if mlockall is required and ES will not be able to lock the memory at startup. However, failed bootstrap checks should be ignored if ES runs in [**development mode**](https://www.elastic.co/guide/en/elasticsearch/reference/5.0/bootstrap-checks.html#_development_vs_production_mode) - i.e. "if ES does not bind transport to an external interface" (so does it apply if we run ES in container and [bind it](https://github.com/openshift/origin-aggregated-logging/blob/49b11bf22734ea704262e09b26e1ddb85f737e39/deployer/conf/elasticsearch.yml#L22) to `0.0.0.0`?).

portante:
@jcantrill, I don't think this commit is necessary, given we should be running without swap for Elasticsearch, and we don't want to run a privileged container just so that ES can lock all memory.

lukas-vlcek:
It makes sense to add/improve documentation around setup and use of memory lock setting for ES for users that have reason to use it. But there should be a different ticket opened for this. Closing for now.

PR #385, opened 2017-04-19
bug 1441369. Allow setting of kibana memory limits

jcantrill:


jcantrill:
Dependent on merge of https://github.com/fabric8io/openshift-auth-proxy/pull/13 to update the submodule

deployer/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM openshift/origin:v1.2.0
+FROM openshift/origin:v1.4.1
```
richm: double plus good

deployer/scripts/upgrade.sh
```diff
@@ -938,8 +962,15 @@ function upgrade_logging() {
   if oc get configmap logging-elasticsearch -o yaml | grep -q MIN_MASTER ; then
      update_es_for_min_masters
   fi
-  scaleUp
 
+  kibanaDC=$(oc get dc logging-kibana -o yaml)
+  kibanaOpsDC=$(oc get dc logging-kibana-ops -o yaml)
+  if [ -z $(echo $kibanaDC | grep -q KIBANA_MEMORY_LIMIT) ] || [ -z $(echo $kibanaDC | grep -q KIBANA_PROXY_MEMORY_LIMIT) ] ||
+     [ -z $(echo $kibanaOpsDC | grep -q KIBANA_MEMORY_LIMIT) ] || [ -z $(echo $kibanaOpsDc | grep -q KIBANA_PROXY_MEMORY_LIMIT) ] ; then
```
richm: `grep -q` has no output - so the condition will always be `true`

I suggest this:
```
if echo "$kibanaDC" | grep -q KIBANA_MEMORY_LIMIT && echo "$kibanaDC" | grep -q KIBANA_PROXY_MEMORY_LIMIT && echo "$kibanaOpsDC" | grep -q KIBANA_MEMORY_LIMIT && echo "$kibanaOpsDC" | grep -q KIBANA_PROXY_MEMORY_LIMIT ; then
  echo memory limits already updated
else
  update_kibana_memory_limits
fi
```

jcantrill:
@richm had to add a bunch of env vars because of the changes made to the auth proxy.  It could no longer find the secrets because the defaults are not absolute paths.  Grrrr

PR #386, opened 2017-04-20
make CI work for 1.4

jcantrill:
@richm  this is the code im having issue with the tests passing in CI.  any help you could add would be appreciated.

### DIFF
--- a/hack/testing/logging.sh
+++ b/hack/testing/logging.sh
@@ -250,7 +250,7 @@ fi
 
 # start fluentd
 os::cmd::try_until_success "oc get daemonset logging-fluentd" "$(( 1 * TIME_MIN ))"
-os::cmd::expect_success "oc label node --all logging-infra-fluentd=true"
+os::cmd::expect_success "oc label node --all logging-infra-fluentd=true --overwrite"
 
 # the old way with dc's
 # # scale up a fluentd pod
@@ -275,6 +275,7 @@ os::cmd::expect_success "oadm policy add-cluster-role-to-user cluster-admin kibt
 os::cmd::expect_success "oc project logging"
 # also give kibtest access to cluster stats
 espod=`get_running_pod es`
+os::cmd::try_until_text "oc logs $espod" "Seeded the searchguard ACL index" "$(( 3 * TIME_MIN ))"
 wait_for_es_ready $espod logging-es 30
 oc exec $espod -- curl -s -k --cert /etc/elasticsearch/secret/admin-cert \
    --key /etc/elasticsearch/secret/admin-key \
@@ -286,6 +287,7 @@ oc exec $espod -- curl -s -k --cert /etc/elasticsearch/secret/admin-cert \
     python -mjson.tool
 if [ "$ENABLE_OPS_CLUSTER" = "true" ] ; then
     esopspod=`get_running_pod es-ops`
+    os::cmd::try_until_text "oc logs $esopspod" "Seeded the searchguard ACL index" "$(( 3 * TIME_MIN ))"
     wait_for_es_ready $esopspod logging-es-ops 30
     oc exec $esopspod -- curl -s -k --cert /etc/elasticsearch/secret/admin-cert \
        --key /etc/elasticsearch/secret/admin-key \
--- a/hack/testing/logging.sh
+++ b/hack/testing/logging.sh
@@ -13,15 +13,25 @@ set -o nounset
 set -o pipefail
 
 STARTTIME=$(date +%s)
+GIT_BRANCH=${CI_GIT_BRANCH:-release-1.4}
 # assume this script is being run from openshift/origin-aggregated-logging/hack/testing, and
 # origin is checked out in openshift/origin
 OS_ROOT=${OS_ROOT:-$(dirname "${BASH_SOURCE}")/../../../origin}
 # use absolute path
 pushd $OS_ROOT
 OS_ROOT=`pwd`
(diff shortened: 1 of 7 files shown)


richm:
What errors do you get when you run this?

hack/testing/logging.sh
```diff
@@ -7,7 +7,9 @@
 # This scripts starts the OpenShift server with a default configuration.
 # The OpenShift Docker registry and router are installed.
 # It will run the logging tests
-
+if [ ! -z ${DEBUG} ] ; then
```
richm:
```
if [ -n "${DEBUG:-}" ] ; then
```

hack/testing/logging.sh
```diff
@@ -112,37 +117,55 @@ trap "cleanup" EXIT
...
 export ARTIFACT_DIR=${ARTIFACT_DIR:-${TMPDIR:-/tmp}/origin-aggregated-logging/artifacts}
 os::util::environment::use_sudo
-os::util::environment::setup_all_server_vars "origin-aggregated-logging/"
+#os::util::environment::setup_all_server_vars "origin-aggregated-logging/"
 
 os::log::system::start
 
+OCP_VERSION="v1.4.1"
 export KUBELET_HOST=$(hostname)
+export CLUSTER_UP_DIR=/tmp/origin-aggregated-logging/cluster
```
richm: does /tmp/origin-aggregated-logging/cluster exist at this point?

hack/testing/logging.sh
```diff
@@ -112,37 +117,55 @@ trap "cleanup" EXIT
...
+export KUBECONFIG="${ADMIN_KUBECONFIG}"
 
-os::start::configure_server
+os::cmd::expect_success "docker pull openshift/origin:${OCP_VERSION}"
+
+CLUSTER_UP="oc cluster up --version=${OCP_VERSION} --host-data-dir=${ETCD_DATA_DIR} --host-config-dir=${SERVER_CONFIG_DIR}"
+
+#os::start::configure_server
 if [ -n "${KIBANA_HOST:-}" ] ; then
+    os::cmd::expect_success "$CLUSTER_UP"
```
richm: This runs `oc cluster up --version=${OCP_VERSION} ...`

hack/testing/logging.sh
```diff
@@ -112,37 +117,55 @@ trap "cleanup" EXIT
...
+    oc cluster up --version=v1.4.1 --host-data-dir=${ETCD_DATA_DIR} --host-config-dir=${SERVER_CONFIG_DIR}
```
richm: This runs the same oc cluster up again?

jcantrill:
[test]

hack/testing/logging.sh
jcantrill: Are you suggesting maybe I need a mkdir statement? It works on my AMI, but that could just be a fluke

hack/testing/logging.sh
jcantrill: Yes

hack/testing/logging.sh
jcantrill: Should be removed... never made it through this code path

hack/testing/logging.sh
richm: I just didn't see where this is created - but if it is created then don't worry about it.

richm:
@jcantrill the error is coming from here: https://github.com/jcantrill/origin-aggregated-logging/blob/2d93f6d9b9c9ee047de0615b6803667bd2a2ad9b/hack/testing/logging.sh#L303

```
++ oc get dc -o name -l component=es-ops
No resources found.
+ ops_dc=
+ os::cmd::expect_success 'oc patch         -p '\''{"spec":{"template":{"spec":{"containers":[{"name":"elasticsearch","securityContext":{"privileged": true}}]}}}}'\'''
```

The es-ops dc is not found.

jcantrill:
[test]

hack/testing/logging.sh
```diff
@@ -162,7 +173,7 @@ fi
 if [ $NOSETUP = 1 ] ; then
```
richm: change this to
```
if [ "$NOSETUP" = 1 ] ; then
```

richm: because of this:
```
./logging.sh: line 145: [: =: unary operator expected
```

jcantrill:
@richm @stevekuznetsov Do we see any value in pursuing this?  I ideally I would like for 1.4 to work with our CI but I have concern there is a lot of effort here to make this work with minimal gain.

stevekuznetsov:
1.4 doesn't use the Ansible playbooks for setup, right? Kind of the point of the stuff I was trying to do was to make the tests _only_ care about the things they are testing, not the set-up or the cluster or whatever. So if I succeeded in that, we should be able to slap those tests into this branch and, barring functionality differences, they should be fine. Are you seeing that other work is necessary, too?

jcantrill:
@stevekuznetsov I have not additionally evaluated but it was slow go on my last attempts.  1.4 is not ansible based, correct.

stevekuznetsov:
Yeah, it's hard to say. Supporting 1.4 and 1.5 was definitely in the design space for the changes. I think it would make more sense to focus right now on #443. Once that's finished, move the tests from `master` verbatim back to the old branches and then make edits to the test logic as necessary.

jcantrill:
Closing as this seems like a pipe dream... and less important for now

PR #390, opened 2017-04-25
allow to set ES startup timeout in DC; errexit will exit when numerical eval to 0

richm:
This allows the ES startup timeout to be set from the DC, and fixes
some eval and other numeric evaluation problems

### DIFF
--- a/elasticsearch/run.sh
+++ b/elasticsearch/run.sh
@@ -9,8 +9,8 @@ fi
 export KUBERNETES_AUTH_TRYKUBECONFIG="false"
 ES_REST_BASEURL=https://localhost:9200
 LOG_FILE=elasticsearch_connect_log.txt
-RETRY_COUNT=300		# how many times
-RETRY_INTERVAL=1	# how often (in sec)
+RETRY_COUNT=${RETRY_COUNT:-300}		# how many times
+RETRY_INTERVAL=${RETRY_INTERVAL:-1}	# how often (in sec)
 
 retry=$RETRY_COUNT
 max_time=$(( RETRY_COUNT * RETRY_INTERVAL ))	# should be integer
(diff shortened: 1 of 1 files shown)


ewolinetz:
since this is going into 1.4 should there also be an update to the ES template where these ENV values will be available? or are we going to recommend patching them afterwards/make these more secretive?

elasticsearch/run.sh
```diff
@@ -85,7 +85,7 @@ wait_for_port_open() {
 do
     echo -n "."
     sleep $RETRY_INTERVAL
-    (( retry -= 1 ))
+    (( retry -= 1 )) || :
```
lukas-vlcek: What this `|| :` magic does? Is it emoticon?

ewolinetz: lets you ignore any errors basically

lukas-vlcek: +1

richm: The `expr` command, and numeric evaluations done inside `(( .... ))`, if they return a value of `0`, will be interpreted by the shell as an error return.  Since we run with `set -e`, this will cause the script to exit with an error.  Since the only way these numeric operations can fail in these cases is to return `0` (as opposed to say a divide by 0), the fix is to just ignore any errors returned from these by using `|| :` which means `if the previous command returned an error, do nothing`.

lukas-vlcek: Amazing to see emoticons had a real meaning before they [become](https://www.w3.org/2005/Incubator/emotion/) a thing.
Thanks Rich.

richm:
[merge]

jcantrill:
@richm Do we need to rework this PR for master and backporting to various releases?

richm:
I think there have been other changes since this one to improve es run.sh that we should take instead of this one - so I can just close and withdraw this PR

jcantrill:
@richm should we close this?

PR #391, opened 2017-04-26
Make local ES node REST point URL configurable

lukas-vlcek:
This PR makes the readiness probe URL configurable and one can override it using `${ES_REST_BASEURL}` env var. The default value is newly set to `https://logging-es:9200`.

In general, if cURL command is run against invalid URL (or the DNS is not working correctly) then we can get `${response code} == "000"`.

See example:
````bash
$ cat test.sh
#!/bin/bash
URL=$1
response_code=$(curl -s -v -X HEAD \
    --max-time 1 \
    -w '%{response_code}' \
    "${URL}") 
echo Response code: ${response_code}
````
````bash
# Invalid URL
# ===============================
$ ./test.sh https://dummy
* Rebuilt URL to: https://dummy/
* Could not resolve host: dummy
* Closing connection 0
Response code: 000

# Valid URL
# ===============================
$ ./test.sh  https://www.google.com
* Rebuilt URL to: https://www.google.com/
*   Trying 2a00:1450:400e:809::2004...
* connect to 2a00:1450:400e:809::2004 port 443 failed: Connection refused
*   Trying 216.58.211.100...
* Connected to www.google.com (216.58.211.100) port 443 (#0)

(... truncated ...)

* Operation timed out after 282 milliseconds with 0 out of 259 bytes received
* Closing connection 0
Response code: 302
````

@richm @jcantrill @ewolinetz PTAL, general questions for you below:

- Does the default value `https://logging-es:9200` represent the local ES node in this context?
- Shall we change the name `${ES_REST_BASEURL}` of the env var to something better?
- I would like to add `-v` to the cURL call so that we can see in the logs (`oc describe pod ...`) possible reason why readiness probe fails, but I am afraid this can be expensive change (if I am not mistaken readiness probe get called every few sec according to configuration, it is not stopped after first successful pass, or is it? //cc @rhcarvalho), other option is to print verbose info only if the test fails, but that seem to be more complicated in bash.

### DIFF
--- a/elasticsearch/probe/readiness.sh
+++ b/elasticsearch/probe/readiness.sh
@@ -17,7 +17,7 @@
 #
 
 # TODO: try re-use code from ./run.sh
-ES_REST_BASEURL=https://localhost:9200
+ES_REST_BASEURL=${ES_REST_BASEURL:-https://logging-es:9200}
 EXPECTED_RESPONSE_CODE=200
 secret_dir=/etc/elasticsearch/secret
 max_time=${max_time:-4}
@@ -33,6 +33,6 @@ response_code=$(curl -s -X HEAD \
 if [ ${response_code} == ${EXPECTED_RESPONSE_CODE} ]; then
   exit 0
 else
-  echo "Elasticsearch node is not ready to accept HTTP requests yet [response code: ${response_code}]"
+  echo "Elasticsearch node is not ready to accept HTTP requests at ${ES_REST_BASEURL} [response code: ${response_code}]"
   exit 1
 fi


richm:
I would like to understand why `https://localhost:9200` works in some environments but not in others.  I would strongly prefer to use `https://localhost:9200` so we don't have to change it for `logging-es` and `logging-es-ops`.  Does the readiness probe run in the es pod?  Does it run in the es container?  How does it work?

lukas-vlcek:
These are all good questions and we need to figure out, that is why I opened this PR.

rhcarvalho:
> if I am not mistaken readiness probe get called every few sec according to configuration, it is not stopped after first successful pass, or is it? //cc @rhcarvalho

You are not mistaken. Both liveness and readiness probes are executed periodically throughout the life time of a pod.

rhcarvalho:
@lukas-vlcek did you try using the simpler HTTP based probe? I suspect that is not enough because of the certificates?

Looking at the [docs](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/) and [code](https://github.com/openshift/origin/blob/33052f6494d53dc563f8ee80babc419b24cfe79d/vendor/k8s.io/kubernetes/pkg/api/v1/types.go#L1249-L1269), looks like the only auth method we can use with the HTTP probe is via headers... would that suffice here?

lukas-vlcek:
@rhcarvalho I am thinking that in the future we may want to improve the probe test to pass only after the service (Elasticsearch cluster) has been initialized correctly. In such case the HTTP probe would not be enough, so if possible I would like to find solution for command type probe.

lukas-vlcek:
Actually it turns out that `localhost` is ok in this context. Moreover, if I run curl against `localhost:9200` and there is nothing listening yet on that port then I also get `000`.

````
$ curl -s -v -X HEAD \
> --max-time 1 \
> -w '%{response_code}' \
> https://localhost:9200
* Rebuilt URL to: https://localhost:9200/
*   Trying ::1...
* connect to ::1 port 9200 failed: Connection refused
*   Trying 127.0.0.1...
* connect to 127.0.0.1 port 9200 failed: Connection refused
* Failed to connect to localhost port 9200: Connection refused
* Closing connection 0
000
````
So it is expected that curl check in readiness probe will yield `000` until ES is started. I am going to close this ticket for now. We can consider adding some more debug info into readiness probe but that should be addressed in different ticket.

PR #392, opened 2017-04-26
fixes for test env vars and paths

richm:
There have been some changes to the test init procedure which was
wiping out our env. vars.
[test]

### DIFF
--- a/hack/testing/logging.sh
+++ b/hack/testing/logging.sh
@@ -71,8 +71,13 @@ fi
 # use a few tools from the deployer
 source "$OS_O_A_L_DIR/deployer/scripts/util.sh"
 
+# have to set these here so setup_tmpdir_vars will not give them bogus values
+export LOG_DIR=${LOG_DIR:-${TMPDIR:-/tmp}/origin-aggregated-logging/logs}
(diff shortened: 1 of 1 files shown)


hack/testing/logging.sh
```diff
@@ -119,10 +124,8 @@ trap "exit" INT TERM
 trap "cleanup" EXIT
 
 # override LOG_DIR and ARTIFACTS_DIR
-export LOG_DIR=${LOG_DIR:-${TMPDIR:-/tmp}/origin-aggregated-logging/logs}
-export ARTIFACT_DIR=${ARTIFACT_DIR:-${TMPDIR:-/tmp}/origin-aggregated-logging/artifacts}
 os::util::environment::use_sudo
-os::util::environment::setup_all_server_vars "origin-aggregated-logging/"
```
ewolinetz: we don't need to call it like this anymore because of `OS_TMP_ENV_SET`?

richm: we still need to call it, but without the `"origin-aggregated-logging/"` because that is set by `OS_TMP_ENV_SET` above.

ewolinetz:
[merge]

PR #398, opened 2017-05-01
Fluentd file logging

richm:
@jcantrill @ewolinetz @nhosoi several fixes needed to make logging and logging CI work after the recent k8s 1.6.1 rebase
[test]

### DIFF
--- a/fluentd/configs.d/openshift/filter-k8s-meta.conf
+++ b/fluentd/configs.d/openshift/filter-k8s-meta.conf
@@ -5,4 +5,5 @@
   ca_file /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
   include_namespace_id true
   use_journal "#{ENV['USE_JOURNAL'] || 'false'}"
+  container_name_to_kubernetes_regexp '^(?<name_prefix>[^_]+)_(?<container_name>[^\._]+)(\.(?<container_hash>[^_]+))?_(?<pod_name>[^_]+)_(?<namespace>[^_]+)_[^_]+_[^_]+$'
 </filter>
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -105,5 +105,8 @@ else
     echo > $CFG_DIR/dynamic/es-ops-copy-config.conf
 fi
 
-
-exec fluentd $fluentdargs
+if [[ $DEBUG ]] ; then
+    exec fluentd $fluentdargs > /var/log/fluentd.log 2>&1
+else
+    exec fluentd $fluentdargs
+fi
--- a/hack/testing/check-EFK-running.sh
+++ b/hack/testing/check-EFK-running.sh
@@ -339,11 +339,11 @@ done
 
 echo $TEST_DIVIDER
 echo "Checking if ES contains common data model index templates"
-for pod in $(oc get pods -l component=es -o name | sed 's,pod/,,'); do
+for pod in $(oc get pods -l component=es -o jsonpath='{.items[*].metadata.name}'); do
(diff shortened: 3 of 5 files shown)


hack/testing/check-EFK-running.sh
```diff
@@ -339,11 +339,11 @@ done
 
 echo $TEST_DIVIDER
 echo "Checking if ES contains common data model index templates"
-for pod in $(oc get pods -l component=es -o name | sed 's,pod/,,'); do
+for pod in $(oc get pods -l component=es -o jsonpath='{.items[*].metadata.name}'); do
```
ewolinetz: 👍 to finally changing this

fluentd/configs.d/openshift/filter-k8s-meta.conf
```diff
@@ -5,4 +5,5 @@
   ca_file /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
   include_namespace_id true
   use_journal "#{ENV['USE_JOURNAL'] || 'false'}"
+  container_name_to_kubernetes_regexp '^(?<name_prefix>[^_]+)_(?<container_name>[^\._]+)(\.(?<container_hash>[^_]+))?_(?<pod_name>[^_]+)_(?<namespace>[^_]+)_[^_]+_[^_]+$'
```
ewolinetz: this is the same as the new default, correct?

fluentd/run.sh
```diff
@@ -105,5 +105,8 @@ else
     echo > $CFG_DIR/dynamic/es-ops-copy-config.conf
 fi
 
-
-exec fluentd $fluentdargs
+if [[ $DEBUG ]] ; then
+    exec fluentd $fluentdargs > /var/log/fluentd.log 2>&1
```
ewolinetz: Why are we just writing our logs out to /var/log/fluentd.log if we're debugging? Is it possible to move this to the fluentdargs instead since we are adding `-vv` there.

fluentd/configs.d/openshift/filter-k8s-meta.conf
richm: yes

fluentd/run.sh
richm: The problem is using filters like this for debugging, when you are trying to see the record "in flight":
```
<filter **>
  @type stdout
</filter>
```
This causes terrible feedback loops in fluentd/es.  The only solution is to _not_ output the log to stdout.

richm:
> With this change, we can keep using the current fluent-plugin-kubernetes_metadata_filter-0.26.3?

@nhosoi yes

richm:
[test] - see if this is a test flake

fluentd/run.sh
richm: @ewolinetz This doesn't work:
```
output='-vv > log 2>&1'
...
exec fluentd $output
```
This does work, but we have to use eval:
```
eval "exec $output"
```
I suppose we could do something like this too:
```
exec fluentd > $stdout 2> $stderr
```
and set `stdout=/dev/stdout` and `stderr=/dev/stderr` by default.  Not sure if `stdout=/var/log/fluentd.log` and `stderr=/var/log/fluentd.log` will work as expected.  I can do some file descriptor magic.

ewolinetz: i'd rather avoid needing to `eval` or make it look more complicated than necessary just to keep it a single line.

I was also confusing the `[[ $VERBOSE ]]` check with this one. I'm fine with the current implementation in this PR

richm:
[merge]

richm:
[merge] flake

richm:
[merge]

PR #399, opened 2017-05-01
fix 1.5 ci

jcantrill:


jcantrill:
[test]

hack/testing/init-log-stack
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 OS_ANSIBLE_VERBOSITY=${OS_ANSIBLE_VERBOSITY:-"vv"}
-OS_VERSION_STRING=${OS_VERSION:+openshift_release=$OS_VERSION}
+OS_VERSION_STRING="openshift_release=v1.5.0"
```
richm: What happens if you don't change this?

hack/testing/prep-host
```diff
@@ -7,7 +7,7 @@ ANSIBLE_RPM=${ANSIBLE_RPM:-ansible}
 RUAMEL_YAML_RPM=${RUAMEL_YAML_RPM:-python-ruamel-yaml}
 
 OS_ANSIBLE_REPO=${OS_ANSIBLE_REPO:-https://github.com/openshift/openshift-ansible}
-OS_ANSIBLE_BRANCH=${OS_ANSIBLE_BRANCH:-master}
+OS_ANSIBLE_BRANCH=release-1.5
```
richm:
```
OS_ANSIBLE_BRANCH=${OS_ANSIBLE_BRANCH:-release-1.5}
```
so that we can run tests + override the ansible branch

hack/testing/init-log-stack
jcantrill: I don't know. I assume since we are using the same job it will come in as an env var like '3.6' but I didnt want to take that chance.

richm: The default is that `OS_VERSION` is not set at all.  openshift is already installed on the system.  ansible should just figure out (facts, etc.) what `openshift_release` is already installed.

hack/testing/prep-host
jcantrill: the only issue with allowing you to set the branch is that we are using the jenkins job on master so OS_ANSIBLE_BRANCH will always be populated with master

richm:
[test] flake

jcantrill:
[test] flake?

richm:
[test] flake again

jcantrill:
[test]

jcantrill:
[test]

hack/testing/prep-host
```diff
@@ -32,3 +40,6 @@ do
 done
 
 export PATH=/usr/local/bin:$PATH
+
+sudo sed -i -e 's/--selinux-enabled//g'  /etc/sysconfig/docker
```
jcantrill: @stevekuznetsov  found out that removing this from the docker config allowed me to pass my server start problem

jcantrill:
[test]

richm:
ship it

jcantrill:
[merge]

hack/testing/prep-host
stevekuznetsov: Turning SELinux off can be a temporary band-aid but we need to look into why it was causing you failures. Can you reproduce the server start problem in a VM and paste here `sudo ausearch -m avc`?

stevekuznetsov: I would not expect this fix to be permanent.

PR #408, opened 2017-05-10
look for /etc/docker/daemon.json for log-driver

richm:
docker now uses /etc/docker/daemon.json or /etc/sysconfig/docker to
configure the log-driver.  Use both of these, preferring daemon.json,
to figure out which log driver docker is using.  The tests can use
`docker info` instead, but this is not available inside the fluentd
container.
This patch relies on https://github.com/openshift/openshift-ansible/pull/4136
to mount /etc/docker inside the fluentd container.
@jcantrill @ewolinetz @skuznets ptal

### DIFF
--- a/deployer/scripts/util.sh
+++ b/deployer/scripts/util.sh
@@ -546,3 +546,23 @@ function wait_for_fluentd_to_catch_up() {
     echo END wait_for_fluentd_to_catch_up at `date -u --rfc-3339=ns`
     return $rc
 }
+
+docker_uses_journal() {
+    # note the unintuitive logic - in this case, a 0 return means true, and a 1
+    # return means false
+    # need to be able to handle cases like
+    # OPTIONS='--log-driver=json-file ....' # or use --log-driver=journald
+    # if "log-driver" is set in /etc/docker/daemon.json, assume that it is
+    # authoritative
+    # otherwise, look for /etc/sysconfig/docker
+    if type -p docker > /dev/null && sudo docker info | grep -q 'Logging Driver: journald' ; then
+        return 0
+    elif grep -q '^[^#].*"log-driver":' /etc/docker/daemon.json 2> /dev/null ; then
+        if grep -q '^[^#].*"log-driver":.*journald' /etc/docker/daemon.json 2> /dev/null ; then
+            return 0
+        fi
+    elif grep -q "^OPTIONS='[^']*--log-driver=journald" /etc/sysconfig/docker 2> /dev/null ; then
+        return 0
+    fi
+    return 1
+}
--- a/hack/testing/init-log-stack
+++ b/hack/testing/init-log-stack
@@ -2,6 +2,7 @@
 
 OS_ANSIBLE_VERBOSITY=${OS_ANSIBLE_VERBOSITY:-"vv"}
 OS_VERSION_STRING=${OS_VERSION:+openshift_release=$OS_VERSION}
+USE_JOURNAL_VAR=${USE_JOURNAL:+openshift_logging_fluentd_use_journal=${USE_JOURNAL}}
 
 ANSIBLE_PLAYBOOK_ARGS=${ANSIBLE_PLAYBOOK_ARGS:-""}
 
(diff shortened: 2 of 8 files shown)


richm:
[test] - let's see what happens . . .

richm:
@stevekuznetsov ptal

deployer/scripts/util.sh
```diff
@@ -546,3 +546,23 @@ function wait_for_fluentd_to_catch_up() {
     echo END wait_for_fluentd_to_catch_up at `date -u --rfc-3339=ns`
     return $rc
 }
+
+docker_uses_journal() {
+    # note the unintuitive logic - in this case, a 0 return means true, and a 1
```
stevekuznetsov: Depending on who you are writing this for, this _is_ intuitive in that this is how Bash return codes always work ;)

deployer/scripts/util.sh
```diff
@@ -546,3 +546,23 @@ function wait_for_fluentd_to_catch_up() {
...
+    # return means false
+    # need to be able to handle cases like
+    # OPTIONS='--log-driver=json-file ....' # or use --log-driver=journald
+    # if "log-driver" is set in /etc/docker/daemon.json, assume that it is
+    # authoritative
```
stevekuznetsov: If you're assuming this is the authority and you find it, but do not find `journald` in there, shouldn't you return false?

deployer/scripts/util.sh
richm: ack - just making sure

deployer/scripts/util.sh
richm: @stevekuznetsov it does return false - it should fall through to line 567 and `return 1` - false

deployer/scripts/util.sh
richm: I'm writing this for bash haters . . .

deployer/scripts/util.sh
stevekuznetsov: As written it will execute the `elif` on 564, no?

stevekuznetsov: Specifically, what will the daemon do when `log-driver: json-file` is in `/etc/docker/daemon.json` and `--log-driver=journald` is in `/etc/sysconfig/docker`?

richm: In this case, the `elif grep -q '^[^#].*"log-driver":' /etc/docker/daemon.json` at line 560 will return `true`, so the flow will go to line 561, where the `if grep -q '^[^#].*"log-driver":.*journald' /etc/docker/daemon.json` will return `false`.  The flow will then fall through to line 567 and return `false`.  Which means, if `log-driver` is set in `/etc/docker/daemon.json`, but the value is not `journald`, return `false`, which is the expected behavior.

richm:
yep - tests will fail until https://github.com/openshift/openshift-ansible/pull/4136 is merged . . .

deployer/scripts/util.sh
stevekuznetsov: Not sure I can reproduce that fall-through behavior:

```sh
#!/bin/bash

echo "log-driver: json-file" > /tmp/daemon.json
echo "OPTIONS='--log-driver=journald" > /tmp/docker

if grep -q '^[^#].*"log-driver":' /tmp/daemon.json 2> /dev/null ; then
    if grep -q '^[^#].*"log-driver":.*journald' /tmp/daemon.json 2> /dev/null ; then
        exit 0
    fi
elif grep -q "^OPTIONS='[^']*--log-driver=journald" /tmp/docker 2> /dev/null ; then
    exit 0
fi
exit 1
```

```
$ /tmp/richm.sh; echo $?
0
```

hack/testing/init-log-stack
```diff
@@ -31,7 +32,7 @@ openshift_logging_master_public_url=https://${PUBLIC_MASTER_HOST:-localhost}:844
 
 openshift_logging_image_prefix=$imageprefix
 openshift_logging_use_ops=${ENABLE_OPS_CLUSTER:-False}
-openshift_logging_fluentd_use_journal=${USE_JOURNAL:-}
+$USE_JOURNAL_VAR
```
ewolinetz: what's this doing?

richm: If `USE_JOURNAL` is explicitly set, then explicitly set `openshift_logging_fluentd_use_journal=$USE_JOURNAL` in the inventory.
If `USE_JOURNAL` is not set, then omit `openshift_logging_fluentd_use_journal` from the inventory, so that we will exercise the ansible code paths where `openshift_logging_fluentd_use_journal` is not defined.

richm: as opposed to doing something like 
```
openshift_logging_fluentd_use_journal=
```
because I don't know if that is allowed

ewolinetz: ack, i didn't see that this was in the creation of the inventory file

richm:
[test]

richm:
[merge]

PR #410, opened 2017-05-12
Remove deployer subtree now that we use ansible

portante:
Now that the we deploy via ansible, we no longer need the deployment
configuration subtree, as everything should be handled by the ansible
process.


jcantrill:
Closing in favor of #664

PR #411, opened 2017-05-12
Remove deployer subtree now that we use ansible

portante:
Now that the we deploy via ansible, we no longer need the deployment
configuration subtree, as everything should be handled by the ansible
process.


richm:
I suggest moving `deployer/scripts/util.sh` to `hack/testing`, and `fluentd.yaml` to `hack/templates`.  They are still used by the testing code.  You will also have to edit logging.sh and anywhere else `deployer/scripts/util.sh` is used.

richm:
looks like #410 is a dup of this PR?

jcantrill:
We can not do this until we have support in 'oc cluster up' to use playbook2image to install logging.  Additionally, there is another PR I have started that moves some of the relevant document bit to other places

portante:
@jcantrill, @richm, so it looks like `oc cluster up` is failing to get logging right because it is using the deployer here instead of ansible.  These settings, configs, etc. are not working.  So it can't hurt to remove them for the release-1.5 branch.

@richm, this is a duplicate of #410, just applied to the 1.5 branch.  I'll make the suggested changes.

portante:
@richm, @jcantril, I rebased this and updated to account for the testing code a bit.  It still needs a bit of work to reconcile everything, but it seems like we have a chance to remove the deployer and get rid of duplicate code between `origin-aggregated-loging` and `openshift-ansible`.

jcantrill:
@portante We would love to move them to the ansible image, however,

1. I know of no plan to migrate the openshift-ansible image to 1.5/3.5.  @sdodson ?
2. I am only now working on making 'oc cluster up' compatible with 1 (https://github.com/openshift/origin/pull/14338). It's a possible backport candidate but with the image it is not useful.

sdodson:
> I know of no plan to migrate the openshift-ansible image to 1.5/3.5. @sdodson ?

If you need it to clean things up we can surely make it happen.

jcantrill:
@sdodson If this image was made available for 1.5 which is ansible based, it would allows to remove the deployer entirely for logging.  Additionally, if there were some way to make it available for 1.4, we could make the 'oc cluster up' code also ansible based which would just run the deployer behind the scene. Having it in 1.5 would be the most desirable.

jcantrill:
@richm I don't see where we can remove the deployer for 1.5.  We will always have scenarios where, in order to support 'oc cluster up', a version of the deployer will have to exist.  There will be users that have an older version of 'oc' that is not capable of doing anything but installs using the deployer.  I merged the changes to the README but we should consider modifying this PR to capture any missed changes or close it completely.

portante:
@jcantrill

jcantrill:
@portante This PR [1] fixes 'oc cluster up' to use ansible, but it is only valid for the 3.6 origin code stream.  How to we support users who want to use oc where they dont have a version that can use ansible?

This PR[2] backports the documentation changes identifiying the deployer is no longer supported.  We could provide additional documentation that it is still used for 'oc cluster up'

[1] https://github.com/openshift/origin/pull/14338
[2] https://github.com/openshift/origin-aggregated-logging/pull/484

jcantrill:
Closing this PR as the deployer  is still in use for 'oc cluster up' for versions of oc prior to 3.6

portante:
@jcantrill, so are you then going to maintain duplicate data between openshift-ansible repo and this repo for 1.5?

jcantrill:
The only duplication is the green install scenario which is more or less modifying the es config file used by es and deployed via the deployer.  If we have users who do not heed our deprecation warning then IMO they take issues upon themselves.  We will only be making updates as required to support 'oc cluster up' which is by nature a development tool with limited expectations and does not support an 'upgrade' path

portante:
I'll leave my branch around so that when this bites us, we can consider it again. =)

PR #412, opened 2017-05-13
some Elasticsearch run.sh improvements and logging.sh improvements

richm:
It was very difficult to read the Elasticsearch logs with the `.`
characters from the http check interspersed with messages from ES
itself.  This commit makes the ES run.sh logging compatible with
the output from ES itself.  You can look for "container.run" to
see only the logs from the run.sh script.  This also allows for
using a DEBUG option to see the details from the various checks, and
using an ERROR level to hide most of the noise.
One interesting problem I found is that if wait_for_fluentd_to_catch_up
ran before fluentd was fully reading from the log sources, that
test could add entries that _fluentd would never read_.  The solution
is to look for the position files before writing something that you
expect fluentd to read.
@jcantrill @ewolinetz @nhosoi @lukas-vlcek @stevekuznetzov
[test]

### DIFF
--- a/deployer/scripts/util.sh
+++ b/deployer/scripts/util.sh
@@ -490,6 +490,7 @@ function test_count_err() {
 # $2 - command to call to pass the uuid_es_ops
 # $3 - expected number of matches
 function wait_for_fluentd_to_catch_up() {
+    local starttime=`date +%s`
     echo START wait_for_fluentd_to_catch_up at `date -u --rfc-3339=ns`
     local es_pod=`get_running_pod es`
     local es_ops_pod=`get_running_pod es-ops`
@@ -499,10 +500,12 @@ function wait_for_fluentd_to_catch_up() {
     local uuid_es=`uuidgen`
     local uuid_es_ops=`uuidgen`
     local expected=${3:-1}
-    local timeout=600
+    local timeout=300
 
     add_test_message $uuid_es
+    echo added es message $uuid_es
     logger -i -p local6.info -t $uuid_es_ops $uuid_es_ops
+    echo added es-ops message $uuid_es_ops
 
     local rc=0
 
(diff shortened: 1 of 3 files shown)


deployer/scripts/util.sh
```diff
@@ -514,7 +517,7 @@ function wait_for_fluentd_to_catch_up() {
     else
         echo failed - $FUNCNAME: not found $expected record project logging for $uuid_es after $timeout seconds
         echo "Checking journal for $uuid_es..."
-        if journalctl | grep $uuid_es ; then
+        if sudo journalctl | grep $uuid_es ; then
```
jcantrill: What has changed that we need suod now?

richm: I was getting some errors that not all log entries could be read.  Is it a problem to use sudo here?

richm: @jcantrill ok to merge?

jcantrill:
[merge]

elasticsearch/run.sh
```diff
@@ -39,75 +82,75 @@ if [[ "${INSTANCE_RAM:-}" =~ $regex ]]; then
...
     if [ -r "${mem_file}" ]; then
         max_mem="$(cat ${mem_file})"
         if [ ${max_mem} -lt ${num} ]; then
             ((num = ${max_mem}))
-            echo "Setting the maximum allowable RAM to $(($num / BYTES_PER_MEG))m which is the largest amount available"
+            warn "Setting the maximum allowable RAM to $(($num / BYTES_PER_MEG))m which is the largest amount available"
         fi
     else
-        echo "Unable to determine the maximum allowable RAM for this host in order to configure Elasticsearch"
+        error "Unable to determine the maximum allowable RAM for this host in order to configure Elasticsearch"
```
ewolinetz: 👍  (to all the changes from `echo`)

PR #413, opened 2017-05-14
Don't rely on the com.redhat.viaq.common type in tests

stevekuznetsov:
Instead of relying on a specific message about an udpated mapping event
for the `com.redhat.viaq.common` type in the ElasticSearch pod logs, we
should look for the `/_cat/indices` API to return to us a valid set of
indices, which is a functionally equivalent test and is more straight-
forward.

[test]
/cc @richm


stevekuznetsov:
@richm good to merge?

richm:
[merge]

PR #414, opened 2017-05-14
[WIP] Remove `os.Exec()` calls from `check-logs.go`

stevekuznetsov:
If we are to be using Golang for some of the work in the log
functionality test, we should make sure to delegate only the
work that is not possible or not simple in Bash. The calls to
`os.Exec()` are perhaps more convoluted in the Golang program
than they are in Bash, so this change separates them out into
the shell script and keeps only the JSON parsing and filtering
logic.

TODO:

 - [ ] determine how to correctly escape the messages, if we need to
 - [ ] once we get the test functional (with actual messages showing up) test that this works as expected

/cc @richm


hack/testing/extract_messages.go
```diff
@@ -0,0 +1,79 @@
...
+		message := record.Fields.Message[0]
+
+		if journal == "true" {
+			// escape certain characters that were being interpreted by bash
+			message = strings.Replace(message, `\`, `\\`, -1)
+			message = strings.Replace(message, `"`, `\"`, -1)
+			message = strings.Replace(message, "`", `\`+"`", -1)
+			message = strings.Replace(message, `$`, `\$`, -1)
+		} else {
+			// escape certain characters that were being interpreted by grep -e "" or bash
```
stevekuznetsov: If we use `grep --fixed-strings` I am pretty sure we do not need any of the following escapes.

hack/testing/extract_messages.go
```diff
@@ -0,0 +1,79 @@
...
+		fmt.Prtintln("[ERROR] No messages found in JSON response")
+		os.Exit(1)
+	}
+
+	for _, record := range elasticSearchData.Hits.Hits {
+		// for each message, we need to check the logs
+		message := record.Fields.Message[0]
+
+		if journal == "true" {
+			// escape certain characters that were being interpreted by bash
```
stevekuznetsov: We might want to look into a Golang dependency that knows how to escape shell characters to make sure we don't need to keep maintaining this list.

test/cluster/functionality.sh
```diff
@@ -95,8 +95,27 @@ for elasticsearch_pod in $( oc get pods --selector component="${OAL_ELASTICSEACH
...
 			os::log::info "Cheking for index ${index} with Kibana pod ${kibana_pod}..."
-			# As we're checking system log files, we need to use `sudo`
-			os::cmd::expect_success "sudo -E VERBOSE=true go run '${OS_O_A_L_DIR}/hack/testing/check-logs.go' '${kibana_pod}' '${elasticsearch_api}' '${index}' '${index_search_path}' '${query_size}' '${test_user}' '${test_token}' '${test_ip}'"
+			elasticsearch_data="$(
+				oc exec "${kibana_pod}" -- curl --silent \
+                                                --key /etc/kibana/keys/key   \
+                                                --cert /etc/kibana/keys/cert \
+                                                --cacert /etc/kibana/keys/ca \
+                                                --header "X-Forwarded-For: ${test_ip}"         \
+                                                --header "X-Proxy-Remote-User: ${test_user}"   \
```
stevekuznetsov: As mentioned before it would be good to break out this authz test from the log test, but that can be done separately.

hack/testing/extract_messages.go
richm: e.g. like `shellquote` in python.

The bigger problem is - why are we mixing Go and Bash?  Why not just have go read in the files/journal and do the string comparison?

stevekuznetsov: Right - that was the impetus for this PR. I think I just interpreted it the other way - use Go for the JSON parsing and shell for the rest. Could go either way really but why write a search function when you've got `grep`?

richm: Ok.

richm:
@stevekuznetsov what's the status of this PR?  ready to test/merge?

stevekuznetsov:
@richm I would want to get the test functional (figure out why it's finding no messages and make it find messages so it succeeds in the current state) before moving forward with this. I don't think I have the domain knowledge to tackle that prerequisite task.

test/cluster/functionality.sh
```diff
@@ -95,8 +95,27 @@ for elasticsearch_pod in $( oc get pods --selector component="${OAL_ELASTICSEACH
...
+                                                --header "Authorization: Bearer ${test_token}" \
+                                                --data "fields=message"     \
+                                                --data "size=${query_size}" \
+                                                --data "q=hostname:$( uname --nodename )" \
```
richm: Does it work if you omit this line?

stevekuznetsov:
[test]

stevekuznetsov:
OK, this is in whatever broken state we have -- 

```
[INFO] Cheking for index .operations with Kibana pod logging-kibana-ops-1-t74pk...
exit status 1
Running test/cluster/functionality.sh:114: executing 'sudo journalctl 'MESSAGE=[ERROR] No messages found in JSON response'' expecting success...
[ERROR] PID 52030: test/cluster/functionality.sh:97: `go run "${OS_O_A_L_DIR}/hack/testing/extract_messages.go" <<< "${elasticsearch_data}"` exited with status 1.
[INFO] 		Stack Trace: 
SUCCESS after 0.037s: test/cluster/functionality.sh:114: executing 'sudo journalctl 'MESSAGE=[ERROR] No messages found in JSON response'' expecting success
[INFO] 		  1: test/cluster/functionality.sh:97: `go run "${OS_O_A_L_DIR}/hack/testing/extract_messages.go" <<< "${elasticsearch_data}"`
[INFO]   Exiting with code 1.
```

@richm does someone with domain knowledge want to dig in and actually determine what is going on?

stevekuznetsov:
@richm I'll close this -- if someone wants to do the maintenance I'll keep the branch around.

/close

PR #415, opened 2017-05-15
bug 1437952. unmount /var/lib/docker/container/*/shm to allow pods to…

jcantrill:
… terminate
This PR fixes 1437952 by umounting the sub directory on start

cc @eparis

### DIFF
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -139,6 +139,11 @@ else
     rm -f $CFG_DIR/openshift/input-pre-debug.conf
 fi
 
+# bug https://bugzilla.redhat.com/show_bug.cgi?id=1437952
+# pods unable to be terminated because fluentd has them busy
+echo "umounts of dead containers will fail. Ignoring..."
+umount /var/lib/docker/containers/*/shm || :
+
 if [[ $DEBUG ]] ; then
     exec fluentd $fluentdargs > /var/log/fluentd.log 2>&1
 else


eparis:
I think https://bugzilla.redhat.com/show_bug.cgi?id=1450554 might actually give us a permanent solution. But it requires a new version of docker and doesn't get things moving now.

We might also need to unmount things from /var/lib/docker/devicemapper/mnt/* but we know this is a good start!

LGTM

eparis:
@rhvgoyal

jcantrill:
companion PR for 1.4 #416

stevekuznetsov:
@jcantrill I don't think it would impact me -- nothing major is in flight right now, and if it were I'd rebase on top of your change anyway. Feel free to add stuff.

jcantrill:
@richm  Trying to test this further, I have several questions:

* How does fluentd add new files to tail?  Is it controled by the 'refresh_interval' parameter and the plugin cycles that often to add/remove files
* How would this change resolve new containers since wouldnt it be necessary to mount additional containers as they come online?  Shouldn't we see the same issue? for new containers

I am unable to verify that fluentd is actually reading what it should.  I have hello-openshift with 3 pods but see only a single entry in the position file for a pod that no longer exists.  My fluentd pod gives me a number of warnings:
`2017-05-15 20:55:56 +0000 [warn]: buffer flush took longer time than slow_flush_log_threshold: plugin_id="object:16befd0" elapsed_time=22.000149293 slow_flush_log_threshold=20.0`

related?

richm:
fluentd uses `in_tail` to read from `/var/log/containers/*`: http://docs.fluentd.org/v0.12/articles/in_tail

It does something like `inotify /var/log/containers` to find out when new files are added to that directory, so new files are added instantaneously.

I don't know enough about how docker works to know if we will see the same problem when containers are added as they come online.  fluentd isn't doing an explicit `mount /var/lib/containers/xxx/shm` when container `xxx` comes online.

I don't know what that `slow_flush` error is about.  Could be related, I don't know.

fluentd/run.sh
```diff
@@ -139,6 +139,11 @@ else
     rm -f $CFG_DIR/openshift/input-pre-debug.conf
 fi
 
+# bug https://bugzilla.redhat.com/show_bug.cgi?id=1437952
+# pods unable to be terminated because fluentd has them busy
+echo "umounts of dead containers will fail. Ignoring..."
+umount /var/lib/docker/containers/*/shm ||
```
richm: use `|| :` - the `:` colon is the bash `null command`

eparis:
I just wrote: https://github.com/eparis/unmount-below

Which maybe make it easy for us to do:

./unmount-below --dry-run=false /var/lib/docker/containers
./unmount-below --dry-run=false /var/lib/docker/devicemapper/

Now I'm not sure how to deliver it or anything. Just pointing out something i'm playing with while we search for the best solution :-(

rhvgoyal:
For devicemapper just doing detached/lazy umout might be sufficient.

syscall.Umount(MNT_DETACH)

jcantrill:
[merge] based on conversation with @richm and @portante

jcantrill:
[merge] based on conversation with @portante and @richm

PR #416, opened 2017-05-15
bug 1437952. unmount /var/lib/docker/container/*/shm to allow pods to…

jcantrill:
… terminate

cc @pweil- @portante

### DIFF
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -60,5 +60,9 @@ else
     echo > $CFG_DIR/dynamic/es-ops-copy-config.conf
 fi
 
+# bug https://bugzilla.redhat.com/show_bug.cgi?id=1437952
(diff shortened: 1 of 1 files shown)


fluentd/run.sh
```diff
@@ -60,5 +60,9 @@ else
     echo > $CFG_DIR/dynamic/es-ops-copy-config.conf
 fi
 
+# bug https://bugzilla.redhat.com/show_bug.cgi?id=1437952
+# pods unable to be terminated because fluentd has them busy
+echo "umounts of dead containers will fail. Ignoring..."
+umount /var/lib/docker/containers/*/shm ||
```
richm: `|| :`
You need the `:` colon at the end.  That is the `null command` in bash.

PR #400, opened 2017-05-01
adding test-mux.sh, filter-pre-mux-test-client.conf

nhosoi:
fixing cleanup in test-fluentd-forward.sh

The logging.sh and related CI tests use a logging-fluentd-template.
The mux env vars need to be added to the fluentd template yaml file.

PR #417, opened 2017-05-15
Parameterizing fluentd buffer_queue_limit and buffer_size_limit and increase the default value

nhosoi:
This pull request contains PR #400, which is not merged yet.
The CI test passed on CentOS Linux release 7.3.1611.


richm:
you should also add these buffer parameters to the secure_forward configurations and/or mux configurations.  the `secure_forward` plugin is a buffer plugin, and so should be able to use those parameters as well.

nhosoi:
I had to remove the original branch test-mux-1 and recreate it to rebase.
Now, new, rebased test-mux-1 is made.  But this PR does not automatically pick it up?
Should I newly submit pr again?

PR #418, opened 2017-05-16
Parameterizing fluentd buffer_queue_limit and buffer_size_limit; increasing the default values

nhosoi:
Introducing environment variables: BUFFER_QUEUE_LIMIT, BUFFER_SIZE_LIMIT

They are set to these parameters:
      buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT']}"
      buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT']}"
    
Default values are increased as follows:
      BUFFER_QUEUE_LIMIT: 256 -> 1024
      BUFFER_SIZE_LIMIT: 8388608 -> 16777216

### DIFF
--- a/deployer/templates/fluentd.yaml
+++ b/deployer/templates/fluentd.yaml
@@ -159,6 +159,10 @@ objects:
               value: ${USE_MUX_CLIENT}
             - name: "MUX_ALLOW_EXTERNAL"
               value: ${MUX_ALLOW_EXTERNAL}
+            - name: "BUFFER_QUEUE_LIMIT"
+              value: ${BUFFER_QUEUE_LIMIT}
+            - name: "BUFFER_SIZE_LIMIT"
+              value: ${BUFFER_SIZE_LIMIT}
           volumes:
           - name: runlogjournal
             hostPath:
@@ -332,3 +336,11 @@ parameters:
   description: 'Configure MUX SERVER (false|true).'
   name: MUX_ALLOW_EXTERNAL
   value: "false"
+-
+  description: 'Fluentd buffer queue limit (0.12 only)'
+  name: BUFFER_QUEUE_LIMIT
+  value: "1024"
+-
+  description: 'Fluentd buffer size limit'
+  name: BUFFER_SIZE_LIMIT
+  value: "16777216"
--- a/fluentd/configs.d/filter-pre-mux-client.conf
+++ b/fluentd/configs.d/filter-pre-mux-client.conf
@@ -5,6 +5,8 @@
   shared_key "#{File.open('/etc/fluent/muxkeys/shared_key') do |f| f.readline end.rstrip}"
   secure yes
   ca_cert_path /etc/fluent/muxkeys/ca
+  buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT']}"
+  buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT']}"
   <server>
     host logging-mux
     port "#{ENV['FORWARD_LISTEN_PORT'] || '24284'}"
--- a/fluentd/configs.d/mux-post-input-filter-tag.conf
+++ b/fluentd/configs.d/mux-post-input-filter-tag.conf
@@ -15,8 +15,6 @@
 <match **>
   @type rewrite_tag_filter
   @label @INGRESS
-  buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT']}"
-  buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT']}"
   rewriterule1 mux_need_k8s_meta ^false$ mux
   rewriterule2 mux_namespace_name (.+) kubernetes.mux.var.log.containers.mux-mux.mux-mux_$1_mux-0123456789012345678901234567890123456789012345678901234567890123.log
 </match>
--- a/fluentd/configs.d/mux-post-input-filter-tag.conf
+++ b/fluentd/configs.d/mux-post-input-filter-tag.conf
@@ -15,6 +15,8 @@
 <match **>
   @type rewrite_tag_filter
   @label @INGRESS
+  buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT']}"
+  buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT']}"
   rewriterule1 mux_need_k8s_meta ^false$ mux
   rewriterule2 mux_namespace_name (.+) kubernetes.mux.var.log.containers.mux-mux.mux-mux_$1_mux-0123456789012345678901234567890123456789012345678901234567890123.log
 </match>
--- a/fluentd/configs.d/openshift/es-copy-config.conf
+++ b/fluentd/configs.d/openshift/es-copy-config.conf
@@ -19,5 +19,7 @@
       reload_on_failure false
       flush_interval 5s
       max_retry_wait 300
-      disable_retry_limit
+      disable_retry_limit true
+      buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT']}"
+      buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT']}"
     </store>
--- a/fluentd/configs.d/openshift/es-ops-copy-config.conf
+++ b/fluentd/configs.d/openshift/es-ops-copy-config.conf
@@ -19,5 +19,7 @@
       reload_on_failure false
       flush_interval 5s
       max_retry_wait 300
-      disable_retry_limit
+      disable_retry_limit true
+      buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT']}"
+      buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT']}"
     </store>
--- a/fluentd/configs.d/openshift/output-es-config.conf
+++ b/fluentd/configs.d/openshift/output-es-config.conf
@@ -19,5 +19,7 @@
       reload_on_failure false
       flush_interval 5s
       max_retry_wait 300
-      disable_retry_limit
+      disable_retry_limit true
+      buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT']}"
+      buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT']}"
     </store>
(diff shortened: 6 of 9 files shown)


deployer/templates/fluentd.yaml
```diff
@@ -159,6 +159,10 @@ objects:
               value: ${USE_MUX_CLIENT}
             - name: "MUX_ALLOW_EXTERNAL"
               value: ${MUX_ALLOW_EXTERNAL}
+            - name: "BUFFER_QUEUE_LIMIT"
```
jcantrill: the change needs to be added to https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging/templates/fluentd.j2 since the deployer is deprecated and no longer supported.

deployer/templates/fluentd.yaml
```diff
@@ -332,3 +336,11 @@ parameters:
   description: 'Configure MUX SERVER (false|true).'
   name: MUX_ALLOW_EXTERNAL
   value: "false"
+-
+  description: 'Fluentd buffer queue limit (0.12 only)'
```
jcantrill: Add these defaults to openshift-ansible and please add appropriate notes with the unit (e.g. bytes, kb)

fluentd/configs.d/filter-pre-mux-client.conf
```diff
@@ -5,6 +5,8 @@
   shared_key "#{File.open('/etc/fluent/muxkeys/shared_key') do |f| f.readline end.rstrip}"
   secure yes
   ca_cert_path /etc/fluent/muxkeys/ca
+  buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT']}"
```
jcantrill: Does this need to go into ansible: https://github.com/openshift/openshift-ansible/tree/master/roles/openshift_logging/files

Same question regarding all the config.d files

deployer/templates/fluentd.yaml
richm: We still use this template for CI testing, unless @stevekuznetsov is going to get rid of that.  So it needs to be in both places.

fluentd/configs.d/filter-pre-mux-client.conf
richm: No, none of these conf files are in ansible currently.  The only fluentd conf files we have in ansible are the main fluent.conf and the secure_forward.conf for sending logs outside of the cluster.

deployer/templates/fluentd.yaml
stevekuznetsov: I'm not sure I fully understand the work involved with removing it but it seemed like you thought that was the appropriate thing to do?

richm: The work is going to be in the various test-*.sh scripts that use the logging-fluentd-template.  They should modify the daemonset directly.

stevekuznetsov: OK. @jcantrill let's update the template for now, removing it will be on the TODO but not something that's already finished so no use anticipating it here :)

fluentd/configs.d/filter-pre-mux-client.conf
ewolinetz: Also to expand on rich's response. There are all contained within the fluentd image, so they won't need to be added to the roles at all

ewolinetz:
This looks really good. Thanks @nhosoi !
Is there a plan on being to specify this at a more granular level in the future? (Similar to how we did the initial ingestion throttling)

nhosoi:
@ewolinetz Thanks!  Regarding the future plan "Similar to how we did the initial ingestion throttling", I'm not even sure if it is possible or not for now..., but it'd be nice if we could decide the optimal value set depending on the hardware configuration, especially the available memory size.  Do you think there is a way to do such autotuning for the fluentd bot?

richm:
[test]

richm:
[test] again - see if this is a flake

fluentd/configs.d/mux-post-input-filter-tag.conf
```diff
@@ -15,6 +15,8 @@
 <match **>
   @type rewrite_tag_filter
   @label @INGRESS
+  buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT']}"
+  buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT']}"
```
richm: I think these should be removed.

richm:
@nhosoi looks like there is a problem with test-fluent-forward - could be related to my comment above, not sure

richm:
[test]

richm:
[test]

fluentd/configs.d/mux-post-input-filter-tag.conf
nhosoi: @richm, let me double check...  "these" means the buffer_.*_limit in fluentd/configs.d/mux-post-input-filter-tag.conf?  
Any other ones?
Sorry & Thanks!

richm: @nhosoi yes, that's what I mean - I already fixed it in the latest commit

richm:
[merge]

fluentd/configs.d/filter-pre-mux-client.conf
portante: Late review, but are these okay defaults?  If we have 1,024 * 16 MB chunks in the queue, how much memory does fluentd use then?  Isn't that 16 GB of memory?

portante: Wouldn't 1 MB chunks be sufficient, with a `flush_interval` of 1 s to try to keep the queue short?

Additionally, if we need a bigger queue, then we'll want to use the `file` buffer instead of `memory` buffer.

nhosoi: > Late review, but are these okay defaults? If we have 1,024 * 16 MB chunks in the queue, how much memory does fluentd use then? Isn't that 16 GB of memory?

> Wouldn't 1 MB chunks be sufficient, with a flush_interval of 1 s to try to keep the queue short?
I agree 16 GB of memory is too large.  The default value set given by fluentd is buffer_chunk_size=8m, buffer_queue_limit=256, i.e., 2GB, in total.

Indeed, we have to make sure the buffer_chunk_limit should be able to be flushed to the ElasticSearch in flush_interval.  If flush_interval is 1 sec (1/5 of the current value), what would be the optimal size of chunk size?  1mb (1/8 of the current value)?

Probably, we should provide some formula to determine the value set of "buffer_chunk_limit, buffer_queue_limit, and flush_interval" based on the available (or specified) physical memory size?  For instance, if we choose 2 out of 3 parameters, the last one is automatically chosen not to exceed the available resource?

Another thing we'd like to consider is the fluentd parameters are shared between the client side fluentd and the mux.  The requirements for the default value set to the client side fluentd and the mux are different.  The mux should have enough capacity to collect the logs from client side fluentd's and send them to the ElasticSearch.  That's one of the motivations for making these parameters configurable.  We don't think one set does not fit all...

That is, we need to have 2 sets of default values (sorry, this is not very accurate.  Of course, the default value set is just one, but I expect we could put them in the env file in the openshift-ansible.)  And there are 3 type of use cases of fluentd, the standalone type, the client side fluentd and the mux.  (could there be something else we should be aware of?)  I think it's ideal that the standalone and the mux could share one default value set; the client side could have another which values are less than the ones in the other set to prevent overflow problem on the mux.  (I'm afraid it may add more complexity to the openshift-ansible...)

> Additionally, if we need a bigger queue, then we'll want to use the file buffer instead of memory buffer.

Currently, some customers are facing "queue size exceeds limit" with the default size 256.  We need to have some way to increase buffer_queue_limit.  Are you suggesting we should provide a method to increase buffer_queue_limit AND switch to use the file buffer?  Or we want to use the file buffer all the time?

portante: @nhosoi, is it really a concern to think a buffer chunk size of even 16 MB would take longer than 1 second to transmit?  Can't a 1 Gb/s link can transfer 125 MB in one second?  It does not seem to be the case that a low `flush_interval` will cause some kind of problem with an inability to transmit a chunk in that window.

If we are going to stick with the `memory` buffer, which is arguably a bad idea, we'll want to probably use no more than 512 MB of memory for a fluentd pod so that that pod is not obtrusive in its memory consumption.  With a low `flush_interval`, it is possible that chunks are queued to be flushed before they are filled.  So then the `buffer_queue_limit` and `buffer_chunk_limit` serve to form the upper-bound of memory usage.

If a `buffer_queue_limit` is reached, then an exception is raised, and the `json-file` reader or `journald` reader should stop reading more data until it can place data into the buffer queue.  Let's assert that that is the behavior we want going forward.

nhosoi: @portante
> If a buffer_queue_limit is reached, then an exception is raised, and the json-file reader or journald reader should stop reading more data until it can place data into the buffer queue. Let's assert that that is the behavior we want going forward.
Sorry for my ignorance...  This feature is going to be in the newer version of fluentd?

> If we are going to stick with the memory buffer, which is arguably a bad idea, ...
I googled around and found configuring with the file buffer does not show much performance degradation, which is promising.  And that's the only a concern, isn't it?

Do we want to make the buffering type configurable?  Or permanently switching to the file buffer is good enough?

> With a low flush_interval, it is possible that chunks are queued to be flushed before they are filled.
I think that's the status we are hoping.  To realize it, I'd think we would like to guarantee that the ElasticSearch has enough CPU and memory.  Another thought is when we have 2 ElasticSearch pods (one for project; another for .operations), setting 2 to num_threads and see how speeds up the flushing.  (But fluentd would require more resources...)
http://docs.fluentd.org/v0.12/articles/out_file#numthreads

portante: @nhosoi, the behavior of what to do with an exception is documented in the [stable version of fluentd, 0.12](http://docs.fluentd.org/v0.12/articles/buffer-plugin-overview), see description of `buffer_queue_full_action`.

Not too concerned about the performance, but more concerned about lost messages if fluentd is restarted and unable to flush it buffers when using the `memory` buffer mechanism.

I don't think we should make the buffering type configurable unless we get demand for it, so switching to `file` buffer should be considered a permanent switch for now.

Certainly tracking future work tuning the number of threads for fluentd behavior of the `out_forward`, `out_secure_forward`, and the `out_elasticsearch` (see [Buffered Output Options](https://github.com/uken/fluent-plugin-elasticsearch#buffered-output-options)) output plugins would be great.

PR #423, opened 2017-05-19
Add a readiness probe to the Kibana image

stevekuznetsov:
In order to ensure that the Kubernetes machinery can determine when the
Kibana Pods are becoming ready, we need to add a readiness probe to the
Containers that make up those pods. The Kibana readiness probe simply
hits the base URL at `http://localhost:5601/` and expects a 200.

/cc @richm @ewolinetz @lukas-vlcek

### DIFF
--- a/kibana/probe/readiness.sh
+++ b/kibana/probe/readiness.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+#
+# Copyright 2017 Red Hat, Inc. and/or its affiliates
+# and other contributors as indicated by the @author tags.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+KIBANA_REST_BASEURL=http://localhost:5601
+EXPECTED_RESPONSE_CODE=200
+max_time="${max_time:-4}"
+
+response_code="$(
+    curl --silent                          \
+         --request HEAD                    \
+         --max-time "${max_time}"          \
+         --write-out '%{response_code}'    \
+         "${KIBANA_REST_BASEURL}/"
+)"
+
+if [ "${response_code}" == "${EXPECTED_RESPONSE_CODE}" ]; then
+    exit 0
+else
+    echo "Kibana node is not ready to accept HTTP requests yet [response code: ${response_code}]"
+    exit 1
+fi
--- a/kibana/Dockerfile
+++ b/kibana/Dockerfile
@@ -24,6 +24,7 @@ RUN rpm --import https://packages.elastic.co/GPG-KEY-elasticsearch && \
                 kibana-${KIBANA_VER} && \
     yum clean all
 
+ADD probe/ /usr/share/kibana/probe/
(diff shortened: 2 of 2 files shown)


stevekuznetsov:
@richm should I update the `hack/templates/logging.yml` as well to add the readiness probe?

lukas-vlcek:
Seems that `curl localhost:5601/api/status` yields useful data. For a highlevel status we might consider checking:

    $ curl -s localhost:5601/api/status | jq .status.overall.state
    "green"

stevekuznetsov:
@lukas-vlcek good call -- are we OK with adding `jq` to the image just for this purpose? I can do some ugly Bash to get around it otherwise!

kibana/probe/readiness.sh
```diff
@@ -0,0 +1,43 @@
...
+)"
+
+status="$(
+    curl --silent                          \
+         --request HEAD                    \
+         --max-time "${max_time}"          \
+         "${KIBANA_REST_BASEURL}/api/status"
+)"
+
+if [ "${response_code}" == "${EXPECTED_RESPONSE_CODE}" ] && grep -q "green" <<<"${status}"; then
```
ewolinetz: Does the full status being green depend on it being able to connect to ES? And is that what we want to do?

stevekuznetsov: Unclear -- if we do not think this is necessary we can drop that part. Unless there are worries that just hitting the base URL is not enough, I would suggest we start with that.

richm:
Which is worse - the probe reporting that Kibana is "ready" when it might not be able to show results from Elasticsearch, or the probe never reporting that Kibana is "ready" because the cluster is in some strange state that prevents Kibana from returning "green"?

stevekuznetsov:
To be clear this probe will allow us to do two things:
 - rely better on things like `kubectl rollout status` that depend on the `ReplicationController` and/or `Pod` state
 - place the Kibana pods into the load balancer pool for the Kibana service only when they are ready

So given that the status quo is load-balancing to all the Kibana pods even when they're not ready, a false positive will still be an improvement as it will be less of a false positive than before. A false negative could be impactful to end users if they had previously been able to hit non-"green" pods and do anything useful with them in that state. Only the status bit should be a regression.

lukas-vlcek:
At this point I am not sure what exactly it means when Kibana is not in green status.

AFAIK Kibana can include several plugins (it contains some plugins/modules out of the box and further it can include additional plugins installed by users - like the [Sense plugin](https://www.elastic.co/guide/en/sense/current/installing.html)) as soon as single plugin is not green the top level Kibana status is probably not green either. But, this does not have to mean Kibana is not usable and unable to serve web traffic, further it can still be used to display status page which can be helpful for debugging the problem.

In particular, when Kibana is unable to connect to ES it is in Red status, see attached screenshot:
screen shot 2017-05-19 at 20 16 51

So IMO just from the point of receiving traffic by Kibana pod I do not think we need to worry about particular Kibana status, just checking 200 response should be enough.

But I am not able to comment on the other aspects pointed out by @stevekuznetsov in [previous comment](#issuecomment-302740204). That is currently beyond my Kubernetes/OpenShift knowledge.

stevekuznetsov:
> So IMO just from the point of receiving traffic by Kibana pod I do not think we need to worry about particular Kibana status, just checking 200 response should be enough.

I agree -- if we find this does not suffice, we can update in the future.

stevekuznetsov:
Removed the Health API check re[test]

stevekuznetsov:
@richm @lukas-vlcek any comments before merge?

richm:
[merge]

stevekuznetsov:
The `yum makecache` failed:

```
Cannot retrieve metalink for repository: epel/x86_64. Please verify its path and try again
```

@richm could you please re-tag for merge

richm:
[merge]

PR #424, opened 2017-05-19
allow tests to run against released software with NOSETUP flag to logging.sh

richm:
If logging.sh is run with the NOSETUP argument, it will skip the set up
of OpenShift and logging components, and will assume they are already
present.
In addition, clean up and make consistent the use of the VERBOSE and CLUSTER
environment variables, and document the current logging.sh variables.
@jcantrill @stevekuznetsov @nhosoi @lukas-vlcek PTAL
[test]

### DIFF
--- a/hack/testing/README.md
+++ b/hack/testing/README.md
@@ -38,7 +38,47 @@ to have been created (based on the logs of the Elasticsearch pods).
 This Go script will query the appropriate ES pod (if an OPS cluster was used)
 and check the local log files for the each result message.
 
-## check-curator.sh
+## Environment values for testing
+
+In addition to providing the argument `true` to `e2e-test.sh` to indicate that
+an OPS cluster is used, the following environment values can be used to
+change behavior of the scripts.
+
+The format for the following would be
+`ENV_VAR1=value [ENV_VAR2=value2] ./e2e-test.sh`
+
+```
+VERBOSE                 | If set, the scripts will print out each command as executed | default: "" (unset)
+KIBANA_CLUSTER_SIZE     | The num of Kibana pod replicas | default: 1
+KIBANA_OPS_CLUSTER_SIZE | The num of Kibana Ops pod replicas if OPS cluster is used | default: 1
+ES_CLUSTER_SIZE         | The num of Elasticsearch pod replicas | default: 1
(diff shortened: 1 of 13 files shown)


hack/testing/README.md
```diff
@@ -38,7 +38,47 @@ to have been created (based on the logs of the Elasticsearch pods).
...
+
+In addition to providing the argument `true` to `e2e-test.sh` to indicate that
+an OPS cluster is used, the following environment values can be used to
+change behavior of the scripts.
+
+The format for the following would be
+`ENV_VAR1=value [ENV_VAR2=value2] ./e2e-test.sh`
+
+```
+VERBOSE                 | If set, the scripts will print out each command as executed | default: "" (unset)
```
stevekuznetsov: You probably do not want to mix `set -o xtrace` with `os::cmd` -- the latter is designed to give you all the information you need. If you run with `set -o xtrace` you get to see the (amazingly verbose) gory detail of the `os::cmd` implementation and I'm not sure anyone needs that.

hack/testing/README.md
```diff
@@ -38,7 +38,47 @@ to have been created (based on the logs of the Elasticsearch pods).
...
+```
+
+# logging.sh
+The script logging.sh is the main CI test driver.  There are a number of
+environment variables that can control its behavior.  In addition, giving the
+`NOSETUP` argument to the script will cause it to not set up OpenShift and the
+logging components.  Instead, it will assume these have already been set up,
+and will just configure the environment to run the tests, then run the tests.
+```
+USE_JOURNAL             | If set to "true" or "false", explicitly enable or disable using the journald as the logging source | default: unset - fluentd will use whatever is configured for the system
```
stevekuznetsov: We should separate arguments like this that apply only to startup from those like the Operations Cluster stuff which can be added post-facto

richm:
@stevekuznetsov I'll mark this PR as `do not merge` for the time being.  I agree that this should be more modular.  However, there are still a couple of issues that need to be addressed with using the check scripts against an arbitrary deployment.  The check scripts still expect that Elasticsearch uses the console for output and will fail otherwise.  That should be changed to use the ES API directly, or something like `oc exec $espod -- cat /elasticsearch/logging-es/logs/logging-es.log`.  And the whole `USE_CLUSTER` business needs to be changed so that the check scripts detect whether or not the `OPS` components are being used.

PR #426, opened 2017-05-19
Change startup polling logic to hit APIs and check pod readiness

stevekuznetsov:
The current approach of looking through container logs to determine if
the ElasticSearch or Kibana node is a specific state is not only fragile
as there is no compatibility guarantee on the log messages but also not
the most useful as it requires that a cluster is set up to log to the
console, which is not default in production clusters.

Specifically, to query for Kibana and ElasticSearch readiness we should
be able to (in a perfect world) simply wait for the `oc rollout status`
check to succeed, as a full rollout of the `DeploymentConfig` will
require that the containers report a ready state. However, there are two
issues with that approach today:

 - Kibana does not have a readiness probe set
 - the OpenShift-Ansible installer creates `DeploymentConfig`s that set
   the desired replica count to 0, which always causes the rollout to
   succeed as there is no work to do. When these deployments are scaled
   in the future, the rollout is not updated

Both of these issues are being worked on, but in the interim we can
continue checking for Kibana and ElasticSearch readiness in this test
by issuing the same `curl` commands that the `ReadinessProbe` will. This
check can be relaxed as the other changes merge.

[test]
/cc @richm

### DIFF
--- a/deployer/README.md
+++ b/deployer/README.md
@@ -13,7 +13,7 @@
   * [Upgrading your EFK stack](#upgrading-your-efk-stack)
   * [Uninstall and Reinstall](#uninstall-and-reinstall)
 * [Using Kibana](#using-kibana)
-* [Adjusting ElasticSearch After Deployment](#adjusting-elasticsearch-after-deployment)
+* [Adjusting Elasticsearch After Deployment](#adjusting-elasticsearch-after-deployment)
 * [Checking EFK Health](#checking-efk-health)
 * [Troubleshooting](#troubleshooting)
 
@@ -22,7 +22,7 @@
 The deployer pod can enable deploying the full stack of the aggregated
 logging solution with just a few prerequisites:
 
-1. Sufficient volumes defined for ElasticSearch cluster storage.
+1. Sufficient volumes defined for Elasticsearch cluster storage.
 2. A router deployment for serving cluster-defined routes for Kibana.
 
 The deployer generates all the necessary certs/keys/etc for cluster
@@ -101,15 +101,15 @@ but will still precede the corresponding template values within the deployer pod
 You will need to specify the hostname at which Kibana should be
 exposed to client browsers, and also the master URL where client
 browsers will be directed for authenticating to OpenShift. You should
-read the [ElasticSearch](#elasticsearch) section below before choosing
-ElasticSearch parameters for the deployer. These and other parameters
+read the [Elasticsearch](#elasticsearch) section below before choosing
+Elasticsearch parameters for the deployer. These and other parameters
 are available:
 
 * `kibana-hostname`: External hostname where web clients will reach Kibana
 * `public-master-url`: External URL for the master, for OAuth purposes
-* `es-cluster-size`: How many instances of ElasticSearch to deploy. At least 3 are needed for redundancy, and more can be used for scaling.
-* `es-instance-ram`: Amount of RAM to reserve per ElasticSearch instance (e.g. 1024M, 2G). Defaults to 8GiB; must be at least 512M (Ref.: [ElasticSearch documentation](https://www.elastic.co/guide/en/elasticsearch/guide/current/hardware.html#_memory).
-* `es-pvc-size`: Size of the PersistentVolumeClaim to create per ElasticSearch ops instance, e.g. 100G. If empty, no PVCs will be created and emptyDir volumes are used instead.
+* `es-cluster-size`: How many instances of Elasticsearch to deploy. At least 3 are needed for redundancy, and more can be used for scaling.
+* `es-instance-ram`: Amount of RAM to reserve per Elasticsearch instance (e.g. 1024M, 2G). Defaults to 8GiB; must be at least 512M (Ref.: [Elasticsearch documentation](https://www.elastic.co/guide/en/elasticsearch/guide/current/hardware.html#_memory).
+* `es-pvc-size`: Size of the PersistentVolumeClaim to create per Elasticsearch ops instance, e.g. 100G. If empty, no PVCs will be created and emptyDir volumes are used instead.
 * `es-pvc-prefix`: Prefix for the names of PersistentVolumeClaims to be created; a number will be appended per instance. If they don't already exist, they will be created with size `es-pvc-size`.
 * `es-pvc-dynamic`: Set to `true` to have created PersistentVolumeClaims annotated such that their backing storage can be dynamically provisioned (if that is available for your cluster).
 * `storage-group`: Number of a supplemental group ID for access to Elasticsearch storage volumes; backing volumes should allow access by this group ID (defaults to 65534).
@@ -234,7 +234,7 @@ deployer creates and how to change them.
 ### Ops cluster
 
 If you set `enable-ops-cluster` to `true` for the deployer, fluentd
-expects to split logs between the main ElasticSearch cluster and another
+expects to split logs between the main Elasticsearch cluster and another
 cluster reserved for operations logs (which are defined as node system
 logs and the projects `default`, `openshift`, and `openshift-infra`). Thus
 a separate Elasticsearch cluster, a separate Kibana, and a separate
@@ -242,20 +242,20 @@ Curator are deployed to index, access, and manage operations logs. These
 deployments are set apart with the `-ops` included in their names. Keep
 these separate deployments in mind while reading the following.
 
-### ElasticSearch
+### Elasticsearch
 
-The deployer creates the number of ElasticSearch instances specified by
-`es-cluster-size`. The nature of ElasticSearch and current Kubernetes
+The deployer creates the number of Elasticsearch instances specified by
+`es-cluster-size`. The nature of Elasticsearch and current Kubernetes
 limitations require that we use a different scaling mechanism than the
 standard Kubernetes scaling.
 
 Scaling a standard deployment (a Kubernetes ReplicationController)
 to multiple pods currently mounts the same volumes on all pods in the
-deployment. However, multiple ElasticSearch instances in a cluster
+deployment. However, multiple Elasticsearch instances in a cluster
 cannot share storage; each pod requires its own storage. Work is under
 way to enable specifying multiple volumes to be allocated individually
 to instances in a deployment, but for now the deployer creates multiple
-deployments in order to scale ElasticSearch to multiple instances. You
+deployments in order to scale Elasticsearch to multiple instances. You
 can view the deployments with:
 
     $ oc get dc --selector logging-infra=elasticsearch
@@ -279,7 +279,7 @@ as directed below.
 By default, the deployer creates an ephemeral deployment in which all
 of a pod's data will be lost any time it is restarted. For production
 use you should specify a persistent storage volume for each deployment
-of ElasticSearch. The deployer parameters with `-pvc-` in the name should
+of Elasticsearch. The deployer parameters with `-pvc-` in the name should
 be used for this. You can either use a pre-existing set of PVCs (specify
 a common prefix for their names and append numbers starting at 1, for
 example with default prefix `logging-es-` supply PVCs `logging-es-1`,
@@ -311,7 +311,7 @@ you would like to use, you can set it with:
 
 #### Node selector
 
-ElasticSearch can be very resource-heavy, particularly in RAM, depending
+Elasticsearch can be very resource-heavy, particularly in RAM, depending
 on the volume of logs your cluster generates. Per Elastic's guidance,
 all members of the cluster should have low latency network connections
 to each other.  You will likely want to direct the instances to dedicated
@@ -342,7 +342,7 @@ oc patch dc/logging-es-{unique name} -p '{"spec":{"template":{"spec":{"nodeSelec
 Recall that the default scheduler algorithm will spread pods to different
 nodes (in the same region, if regions are defined). However this can
 have unexpected consequences in several scenarios and you will most
-likely want to label and specify designated nodes for ElasticSearch.
+likely want to label and specify designated nodes for Elasticsearch.
 
 #### Cluster parameters
 
@@ -362,7 +362,7 @@ should be sufficient, unless you need to scale ES after deployment.
 
 Fluentd is deployed as a DaemonSet that deploys replicas according
 to a node label selector (which you can choose; the default is
-`logging-infra-fluentd`). Once you have ElasticSearch running as
+`logging-infra-fluentd`). Once you have Elasticsearch running as
 desired, label the nodes to deploy Fluentd to in order to feed
 logs into ES. The example below would label a node named
 'ip-172-18-2-170.ec2.internal' using the default Fluentd node selector.
@@ -803,14 +803,14 @@ Here is some information specific to the aggregated logging deployment.
 1. Login is performed via OAuth2, as with the web console. The default certificate
 authentication used for the admin user isn't available, but you can create
 other users and make them cluster admins.
-2. Kibana and ElasticSearch have been customized to display logs only
+2. Kibana and Elasticsearch have been customized to display logs only
 to users that have access to the projects the logs came from. So if you login
 and have no access to anything, be sure your user has access to at least one
 project. Cluster admin users should have access to all project logs as
 well as host logs.
-3. To do anything with ElasticSearch and Kibana, Kibana has to have
+3. To do anything with Elasticsearch and Kibana, Kibana has to have
 defined some index patterns that match indices being recorded in
-ElasticSearch. This should already be done for you, but you should be
+Elasticsearch. This should already be done for you, but you should be
 aware how these work in case you want to customize anything. When logs
 from applications in a project are recorded, they are indexed by project
 name and date in the format `project.name.YYYY-MM-DD`. For matching a project's
@@ -830,16 +830,16 @@ user to be able to view these logs, run `$ oc edit configmap/logging-elasticsear
 change the value of `openshift.operations.allow_cluster_reader` to `true` and
 restart your ES cluster.
 
-# Adjusting ElasticSearch After Deployment
+# Adjusting Elasticsearch After Deployment
 
-If you need to change the ElasticSearch cluster size after deployment,
-DO NOT just scale existing deployments up or down. ElasticSearch cannot
+If you need to change the Elasticsearch cluster size after deployment,
+DO NOT just scale existing deployments up or down. Elasticsearch cannot
 scale by ordinary Kubernetes mechanisms, as explained above. Each instance
 requires its own storage, and thus under current capabilities, its own
 deployment. The deployer defined a template `logging-es-template` which
-can be used to create new ElasticSearch deployments.
+can be used to create new Elasticsearch deployments.
 
-Adjusting the scale of the ElasticSearch cluster
+Adjusting the scale of the Elasticsearch cluster
 typically requires adjusting cluster parameters that
 vary by cluster size. [Elastic documentation discusses these
 issues](https://www.elastic.co/guide/en/elasticsearch/guide/current/_important_configuration_changes.html)
@@ -852,15 +852,15 @@ and existing deployments when changing the cluster size.
 Changing cluster parameters (or any parameters/secrets, really) requires
 re-deploying the instances. In order to minimize resynchronization
 between the instances as they are restarted, we advise halting traffic to
-ElasticSearch and then taking down the entire cluster for maintenance. No
+Elasticsearch and then taking down the entire cluster for maintenance. No
 logs will be lost; Fluentd simply blocks until the cluster returns.
 
-Halting traffic to ElasticSearch requires scaling down Kibana and removing node labels for Fluentd:
+Halting traffic to Elasticsearch requires scaling down Kibana and removing node labels for Fluentd:
 
     $ oc label node --all logging-infra-
     $ oc scale rc/logging-kibana-1 --replicas=0
 
-Next scale all of the ElasticSearch deployments to 0 similarly.
+Next scale all of the Elasticsearch deployments to 0 similarly.
 
     $ oc get rc --selector logging-infra=elasticsearch
     $ oc scale rc/logging-es-... --replicas=0
@@ -1059,7 +1059,7 @@ with a non-zero exit code, and no deployed pods, e.g.:
     NAME                           READY     STATUS             RESTARTS   AGE
     logging-es-2e7ut0iq-1-deploy   1/1       ExitCode:255       0          1m
 
-(In this example, the deployer pod name for an ElasticSearch deployment is shown;
+(In this example, the deployer pod name for an Elasticsearch deployment is shown;
 this is from ReplicationController `logging-es-2e7ut0iq-1` which is a deployment
 of DeploymentConfig `logging-es-2e7ut0iq`.)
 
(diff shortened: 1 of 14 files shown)


stevekuznetsov:
@richm any comments before merge?

test/cluster/functionality.sh
```diff
@@ -48,14 +48,15 @@ elasticsearch_api="$( oc get svc "${OAL_ELASTICSEACH_SERVICE}" -o jsonpath='{ .m
...
+	os::cmd::try_until_text "oc exec ${kibana_pod} -c kibana -- curl -s --request HEAD --write-out '%{response_code}' http://localhost:5601/" "200" "$(( 10*TIME_MIN ))"
+	os::cmd::expect_success_and_text "oc get pod ${kibana_pod} -o jsonpath='{ .status.containerStatuses[?(@.name==\"kibana\")].ready }'" "true"
+	os::cmd::expect_success_and_text "oc get pod ${kibana_pod} -o jsonpath='{ .status.containerStatuses[?(@.name==\"kibana-proxy\")].ready }'" "true"
 done
 
 for elasticsearch_pod in $( oc get pods --selector component="${OAL_ELASTICSEACH_COMPONENT}" -o jsonpath='{ .items[*].metadata.name }' ); do
 	os::log::info "Testing ElasticSearch pod ${elasticsearch_pod} for a successful start..."
-	os::cmd::try_until_text "oc logs ${elasticsearch_pod}" "\[cluster\.service\s*\]" "$(( 10*TIME_MIN ))"
-	os::cmd::try_until_text "oc logs ${elasticsearch_pod}" "\[node\s*\]\s*\[.*\]\s*started" "$(( 10*TIME_MIN ))"
+	os::cmd::try_until_text "oc exec ${elasticsearch_pod} -- curl -sk --cert /etc/elasticsearch/secret/admin-cert --key /etc/elasticsearch/secret/admin-key -X HEAD -w '%{response_code}' https://localhost:9200/" '200' "$(( 10*TIME_MIN ))"
```
richm: It may take several minutes between curl returning 200 and the "cluster.service" and "node started" to appear in the logs, so this check is not nearly as strict, and possibly could report "ready" when it isn't so ready.  But I'm not really sure what to do about this, if there is some API call we can make that will return the same information.  Perhaps @portante or @lukas-vlcek can suggest something.

The name of the product is `Elasticsearch` not `ElasticSearch`.

Can you use this opportunity to change `ELASTICSEACH` to `ELASTICSEARCH`?

stevekuznetsov: > It may take several minutes between curl returning 200 and the "cluster.service" and "node started" to appear in the logs

I would expect whatever we `curl` here to be identical to what we are doing in the readiness check -- we wait below to see indices show up, so we just need the service to be started here. What does "ready" mean to you?

stevekuznetsov:
Spelling commits generated with:

```sh
sed -i 's/ElasticSearch/Elasticsearch/g' $( find * -type f )
sed -i 's/ELASTICSEACH/ELASTICSEARCH/g' $( find * -type f )
```

stevekuznetsov:
@richm so entirely aside from the question of what "ready" should or should not mean, you are correct that the check for a 200 on `:9200` is necessary but not sufficient for the test to run:

```
[INFO] Testing ElasticSearch pod logging-es-3hz559ps-1-fmlwc for a successful start...
Running /data/src/github.com/openshift/origin-aggregated-logging/test/cluster/functionality.sh:58: executing 'oc exec logging-es-3hz559ps-1-fmlwc -- curl -sk --cert /etc/elasticsearch/secret/admin-cert --key /etc/elasticsearch/secret/admin-key -X HEAD -w '%{response_code}' https://localhost:9200/' expecting any result and text '200'; re-trying every 0.2s until completion or 600.000s...
SUCCESS after 0.428s: /data/src/github.com/openshift/origin-aggregated-logging/test/cluster/functionality.sh:58: executing 'oc exec logging-es-3hz559ps-1-fmlwc -- curl -sk --cert /etc/elasticsearch/secret/admin-cert --key /etc/elasticsearch/secret/admin-key -X HEAD -w '%{response_code}' https://localhost:9200/' expecting any result and text '200'; re-trying every 0.2s until completion or 600.000s
Running /data/src/github.com/openshift/origin-aggregated-logging/test/cluster/functionality.sh:59: executing 'oc get pod logging-es-3hz559ps-1-fmlwc -o jsonpath='{ .status.containerStatuses[?(@.name=="elasticsearch")].ready }'' expecting success and text 'true'...
SUCCESS after 0.254s: /data/src/github.com/openshift/origin-aggregated-logging/test/cluster/functionality.sh:59: executing 'oc get pod logging-es-3hz559ps-1-fmlwc -o jsonpath='{ .status.containerStatuses[?(@.name=="elasticsearch")].ready }'' expecting success and text 'true'
[INFO] Checking that ElasticSearch pod logging-es-3hz559ps-1-fmlwc recovered its indices after starting...
[FATAL] ElasticSearch pod logging-es-3hz559ps-1-fmlwc isn't master and was unable to detect a master
```

I will try to determine what API endpoints signify that the node has started and ensure that this test also waits for those.

richm:
egads - we had ElasticSearch all over the place . . .
Yeah, the test isn't going to pass until we can replace all of those `oc logs $espod` calls with the appropriate API call . . .

test/cluster/functionality.sh
```diff
@@ -43,41 +56,45 @@ test_token="$( oc whoami -t )"
...
+	os::cmd::expect_success_and_text "oc get pod ${elasticsearch_pod} -o jsonpath='{ .status.containerStatuses[?(@.name==\"elasticsearch\")].ready }'" "true"
 
-	os::log::info "Checking that ElasticSearch pod ${elasticsearch_pod} recovered its indices after starting..."
-	if oc logs "${elasticsearch_pod}" | grep -E "\[cluster\.service\s*\]" | grep -q "new_master"; then
+	os::log::info "Checking that Elasticsearch pod ${elasticsearch_pod} recovered its indices after starting..."
+	os::cmd::try_until_text "curl_es '${elasticsearch_pod}' '/_cluster/state/master_node' -X HEAD -w '%{response_code}'" '200' "$(( 10*TIME_MIN ))"
+	es_master_id="$( curl_es "${elasticsearch_pod}" "/_cluster/state/master_node" | jq --raw-output '.master_node' )"
+	es_pod_node_id="$( curl_es "${elasticsearch_pod}" "/_nodes/_local" | jq '.nodes' | jq --raw-output 'keys[0]' )"
+	if [[ "${es_master_id}" == "${es_pod_node_id}" ]]; then
 		os::cmd::expect_success_and_text "oc logs ${elasticsearch_pod}" "\[gateway\s*\]\s*\[.*\]\s*recovered\s*\[[0-9]*\]\s*indices into cluster_state"
```
stevekuznetsov: @richm I have been able to get the "is this node a master?" check translated to API calls. However, I do not fully understand what this specific log message means and/or how we can get at it from the API.

test/cluster/functionality.sh
```diff
@@ -43,41 +56,45 @@ test_token="$( oc whoami -t )"
...
 	elif oc logs "${elasticsearch_pod}" | grep -E "\[cluster\.service\s*\]" | grep -q "detected_master"; then
```
stevekuznetsov: @richm similarly it is unclear how to tell if, when the node is not a master, that it has detected one in the cluster. Seeing as the other possibility here is that the node has failed, do we need this test? won't it also fail the next step where we look for indices?

deployer/README.md
```diff
@@ -852,15 +852,15 @@ and existing deployments when changing the cluster size.
...
 between the instances as they are restarted, we advise halting traffic to
-ElasticSearch and then taking down the entire cluster for maintenance. No
+Elasticsearch and then taking down the entire cluster for maintenance. No
 logs will be lost; Fluentd simply blocks until the cluster returns.
 
-Halting traffic to ElasticSearch requires scaling down Kibana and removing node labels for Fluentd:
+Halting traffic to Elasticsearch requires scaling down Kibana and removing node labels for Fluentd:
 
     $ oc label node --all logging-infra-
     $ oc scale rc/logging-kibana-1 --replicas=0
```
kargakis: Scaling the RCs is not going to work anymore - you should direct all actions on DCs.

stevekuznetsov: @kargakis can you open a stand-alone issue for this on the repo?

kargakis: https://github.com/openshift/origin-aggregated-logging/pull/428

test/cluster/functionality.sh
stevekuznetsov: Look like `GET /_recovery` will give us a field `{ .<index-id>.shards.[*].stage }` which we can wait for `"done"` -- but that's data per-index so we'd need the name of the indices we expect to be recovered _a priori_

test/cluster/functionality.sh
```diff
@@ -28,6 +28,20 @@
...
 
+# curl_es hits the Elasticsearch API from within a pod.
+function curl_es() {
+	local pod="$1"
+	local endpoint="$2"
+	shift; shift
+	local args=( "${@:-}" )
+
+	local secret_dir="/etc/elasticsearch/secret/"
+	oc exec "${pod}" -- curl --silent --insecure "${args[@]}" \
```
richm: Why is it necessary to use `args`?  Can you just do
```
oc exec "${pod}" -- curl --silent --insecure "$@" \
```

test/cluster/functionality.sh
```diff
@@ -28,6 +28,20 @@
 source "${OS_ROOT}/hack/lib/init.sh"
 os::util::environment::setup_time_vars
 
+# curl_es hits the Elasticsearch API from within a pod.
+function curl_es() {
```
richm: This function is a superset of `query_es_from_es` - https://github.com/openshift/origin-aggregated-logging/blob/master/deployer/scripts/util.sh#L406 - were you planning on replacing `query_es_from_es` with this function?

test/cluster/functionality.sh
```diff
@@ -73,7 +73,10 @@ for elasticsearch_pod in $( oc get pods --selector component="${OAL_ELASTICSEARC
 	os::cmd::expect_success_and_text "oc get pod ${elasticsearch_pod} -o jsonpath='{ .status.containerStatuses[?(@.name==\"elasticsearch\")].ready }'" "true"
 
 	os::log::info "Checking that Elasticsearch pod ${elasticsearch_pod} recovered its indices after starting..."
-	if oc logs "${elasticsearch_pod}" | grep -E "\[cluster\.service\s*\]" | grep -q "new_master"; then
+	os::cmd::try_until_text "curl_es '${elasticsearch_pod}' '/_cluster/state/master_node' -w '%{response_code}'" "}200$" "$(( 10*TIME_MIN ))"
+	es_master_id="$( curl_es "${elasticsearch_pod}" "/_cluster/state/master_node" | jq --raw-output '.master_node' )"
```
richm: Does the use of `jq` introduce another dependency?  We've been using `python -mjson.tool` to parse JSON, or even `python -c 'import sys,json; dct = json.load(sys.stdin); ...'` for more intensive processing.

test/cluster/functionality.sh
richm: I don't either - @ewolinetz or @sosiouxme ?

test/cluster/functionality.sh
richm: hmm - not sure

test/cluster/functionality.sh
sosiouxme: Not certain but I think that's the message you see when the node has finished startup (which includes recovery after an outage) and is ready for use. If it hasn't logged this then it's not really ready. Not sure if there's a health check for readiness on the ES API?

test/cluster/functionality.sh
sosiouxme: `GET _cat/master` should give you the master according to that node.

test/cluster/functionality.sh
stevekuznetsov: You could, I just prefer to translate positional arguments into human-readable names ASAP.

stevekuznetsov: You could, I just prefer to translate positional arguments into human-readable names ASAP.

test/cluster/functionality.sh
stevekuznetsov: I'll DRY this out -- good call.

test/cluster/functionality.sh
stevekuznetsov: `jq` always has and will continue to be on our test machines, but this would potentially introduce a dependency for local developers. I like it as it's easy to read and works much like the Kubernetes JSONPath. I can try to use local Python HEREDOC if you are concerned about this dependency.

test/cluster/functionality.sh
stevekuznetsov: @sosiouxme Is there a part of the ES API doc that goes over assumptions for `/_cat` endpoints? The doc for the `/_cat/master` endpoint does strongly suggest you're getting the opinion of the specific node you hit, but is that written in more explicit language somewhere?

Also, once we know that the node is not a master and that there is something at `/_cat/master` to indicate it has found one to follow, how do we go about this index recovery bit? Do we know _which_ indices are recovered into `cluster_state`? Where are they recovered from? Can we just wait for them?

test/cluster/functionality.sh
ewolinetz: The only correction I would change to Luke's comment is that it may not necessarily be that ES has actually completed its recovery but rather its pulled in all the metadata and may still recover in an asynch manner

sosiouxme:
>
>
> @sosiouxme <https://github.com/sosiouxme> Is there a part of the ES API
> doc that goes over assumptions for /_cat endpoints? The doc for the
> /_cat/master endpoint does strongly suggest you're getting the opinion of
> the specific node you hit, but is that written in more explicit language
> somewhere?
>

Not that I know of, we took it from ops without really examining it.


> Also, once we know that the node is not a master and that there is
> something at /_cat/master to indicate it has found one to follow, how do
> we go about this index recovery bit? Do we know *which* indices are
> recovered into cluster_state? Where are they recovered from? Can we just
> wait for them?
>
>
My guess would be you don't actually care much about which indices, but
rather whether recovery has completed and the node is ready for requests.
If index recovery status is actually of interest, the /_recovery endpoint
<https://www.elastic.co/guide/en/elasticsearch/reference/2.1/indices-recovery.html>
looks like it provides that.
It's not obvious to me that the same information is available via the API.
/_cluster/state?local=true
<https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-state.html>
and /_cluster/health?local=true
<https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-health.html>
look
like they might be useful but I'm guessing they might be impacted by
cluster-wide index recovery, and a cluster can be partly operational even
when tons of shards are red.

stevekuznetsov:
@sosiouxme I was looking at the `/_recovery` API -- some questions I posted in a comment (https://github.com/openshift/origin-aggregated-logging/pull/426#discussion_r117635255):

> Looks like `GET /_recovery` will give us a field `{ .<index-id>.shards.[*].stage }` which we can wait for `"done"` -- but that's data per-index so we'd need the name of the indices we expect to be recovered _a priori_

stevekuznetsov:
@sosiouxme and without knowing at least some indices to wait for, the obvious issue would be we wait for "all indices" to recover but none are ever posted so our test succeeds incorrectly.

stevekuznetsov:
@sosiouxme @richm also before we go crazy trying to re-write that test without logging calls ...

In the next couple of lines we wait for the `project.*` or `.operations.*` indices to be populated. This specific block of tests is trying to be a lot more fine-grained in detecting failures for ES to come up at all. If we just remove it, the next block waiting for those indices to be populated will surely fail as well. So, if there is no simple or obvious way to test for this without log grepping, and we don't want to be doing log grepping, we could just remove this test and pay the price of having a little coarser test failures. The pod logs, etc, would all still be there anyway, we just couldn't at first blush say "the pod was the master but could not recover indices". We'd still be able to say that the pod was or was not the master, and if neither we could still fail.

test/cluster/functionality.sh
richm: @stevekuznetsov jq looks pretty cool.  I think it should be fine.  The Fedora dependencies are quite small.  Is `jq` generally available on CentOS and RHEL, or are special repos required?

richm: The nice thing about `python -mjson.tool` is that both `python` and the `json` python library are available everywhere on every platform.

richm:
I think we should leave in the checks for the `project.*` and `.operations.*` indices.  For a brand new install (which will almost always be the case with CI testing), this is a good test that fluentd is working and is able to connect and send logs to Elasticsearch.  For an existing installation, not so much, since there will almost always be old indices present.

test/cluster/functionality.sh
stevekuznetsov: > `jq` generally available on CentOS and RHEL, or are special repos required?

Yes, it's available on RHEL and CentOS just fine. It looks like it has just one dependency -- `oniguruma`. 

<details>
<summary> Here is an install log from a RHEL machine </summary>

```
Loaded plugins: amazon-id, rhui-lb, search-disabled-repos
Resolving Dependencies
--> Running transaction check
---> Package jq.x86_64 0:1.5-1.el7 will be installed
--> Processing Dependency: libonig.so.2()(64bit) for package: jq-1.5-1.el7.x86_64
--> Running transaction check
---> Package oniguruma.x86_64 0:5.9.5-3.el7 will be installed
--> Finished Dependency Resolution

Dependencies Resolved

================================================================================
 Package    Arch    Version      Repository                                Size
================================================================================
Installing:
 jq         x86_64  1.5-1.el7    centos-paas-sig-openshift-origin15-rpms  152 k
Installing for dependencies:
 oniguruma  x86_64  5.9.5-3.el7  centos-paas-sig-openshift-origin15-rpms  127 k

Transaction Summary
================================================================================
Install  1 Package (+1 Dependent package)

Total download size: 280 k
Installed size: 906 k
Downloading packages:
--------------------------------------------------------------------------------
Total                                              331 kB/s | 280 kB  00:00     
Running transaction check
Running transaction test
Transaction test succeeded
Running transaction
  Installing : oniguruma-5.9.5-3.el7.x86_64                                 1/2 
  Installing : jq-1.5-1.el7.x86_64                                          2/2 
  Verifying  : oniguruma-5.9.5-3.el7.x86_64                                 1/2 
  Verifying  : jq-1.5-1.el7.x86_64                                          2/2 

Installed:
  jq.x86_64 0:1.5-1.el7                                                         

Dependency Installed:
  oniguruma.x86_64 0:5.9.5-3.el7                                                

Complete!
```

</details>

stevekuznetsov:
> I think we should leave in the checks for the `project.*` and `.operations.*` indices. 

Right, sorry if this was not clear. What I meant to ask was:

 - today, we look to see if the node we are `oc exec`-ing into is the master in the ES cluster; if it is, we also check that it has the `"recovered \d+ indices into cluster state"` log message
 - today, we then futhermore wait until the `project.*` and `.operations.*` indices are populated
 - it seems very hard to replicate the index recovery logic with the API
 - if we just remove the index recovery line, we will still get a failure if the indices are recovered, as the next bit (waiting for `project.*` and `.operations.*`) will fail, but we lose the ability to say that an install failed because a master could not recover indices

So my question is -- unless we find a simple way to query the API for recovered indices (are there internal or housekeeping indices with well-known stable names we could look for?), can we just drop this check? A failure to recover indices on a master will become a failure to get the `project.*` and `.operations.*` indices to show up, but a developer will be able to dig into why that happened without too much issue (as the log for the `project.*` and `.operations.*` failure will show no indices returned).

test/cluster/functionality.sh
richm: Ok - so you have to add the `centos-paas-sig-openshift-origin15-rpms` repo to get it.  That means in a pure RHEL environment, it should be available in some `atomic-openshift-*` repo.

stevekuznetsov: Not sure why `yum` chose that repo specifically, but the package is in `epel`:

```
$ yum list available --showduplicates jq
Available Packages
jq.x86_64  1.5-1.el7  centos-paas-sig-openshift-origin15-rpms 
jq.x86_64  1.5-1.el7  centos-paas-sig-openshift-origin36-rpms 
jq.x86_64  1.5-1.el7  epel
```

sosiouxme:
EPEL is not supported so customers shouldn't be using it. I wouldn't rely
on them having jq installed just so this can run if all you need is a json
parser.

stevekuznetsov:
> EPEL is not supported so customers shouldn't be using it. I wouldn't rely on them having jq installed just so this can run if all you need is a json parser.

ACK. I'll switch over to the inline Python stuff.

ewolinetz:
So I think (and we'd likely have to check the src on it to confirm) that the elected master printing out it `Recovered \d+ indices` is just an indication that it is now ready. In a new cluster that number is always `[0]`.

If fluentd is not running that is the last line you should see in the master logs (so long as searchguard isn't causing issues).

Maybe we should separate out the check that ES is started and was able to get to a point where it knows how many indices there are to recover and that it is accepting connections; and checking whether or not Fluentd has pushed data up to ES and it has processed it so that it can output that it has updated metadata for and/or created an index named `project.*`.

stevekuznetsov:
We're seemingly failing on startup? 

```
May 22 17:24:58 ip-172-18-13-96.ec2.internal dockerd-current[4509]: {"type":"response","@timestamp":"2017-05-22T21:24:58Z","tags":[],"pid":1,"method":"get","statusCode":404,"req":{"url":"/2bfac536-41be-4f4d-9083-316ee8ce83ff","method":"get","headers":{"user-agent":"curl/7.29.0","host":"localhost:5601","accept":"*/*"},"remoteAddress":"127.0.0.1","userAgent":"127.0.0.1"},"res":{"statusCode":404,"responseTime":14,"contentLength":9},"message":"GET /2bfac536-41be-4f4d-9083-316ee8ce83ff 404 14ms - 9.0B"}
Found 2bfac536-41be-4f4d-9083-316ee8ce83ff in journal
Error: found 0 for project .operations message bd493552-6739-497b-a882-603e27688978 - expected 1
{
    "error": {
        "reason": "No feature for name [_count\\]",
        "root_cause": [
            {
                "reason": "No feature for name [_count\\]",
                "type": "illegal_argument_exception"
            }
        ],
        "type": "illegal_argument_exception"
    },
    "status": 400
}
{
    "error": {
        "reason": "No feature for name [_search\\]",
        "root_cause": [
            {
                "reason": "No feature for name [_search\\]",
                "type": "illegal_argument_exception"
            }
        ],
        "type": "illegal_argument_exception"
    },
    "status": 400
}
failed - wait_for_fluentd_to_catch_up: not found 1 record project .operations for bd493552-6739-497b-a882-603e27688978 after 300 seconds
Checking journal for bd493552-6739-497b-a882-603e27688978...
May 22 17:24:58 ip-172-18-13-96.ec2.internal bd493552-6739-497b-a882-603e27688978[28137]: bd493552-6739-497b-a882-603e27688978
Found bd493552-6739-497b-a882-603e27688978 in journal
END wait_for_fluentd_to_catch_up took 833 seconds at 2017-05-22 21:38:49.006478284+00:00
```

Not sure the changes here could fail the start-up phase?

richm:
This looks odd:
```
[_count\\]
```
Looks like it has a spurious trailing `\`

stevekuznetsov:
Gotta love dumb Bash quoting rules -- @richm looks like we are good to go now.

richm:
@stevekuznetsov lgtm, but I think at least someone else should approve like @jcantrill or @ewolinetz

jcantrill:
[merge]

PR #428, opened 2017-05-20
Fixes in the deployer readme

kargakis:
Signed-off-by: Michail Kargakis <mkargaki@redhat.com>

### DIFF
--- a/deployer/README.md
+++ b/deployer/README.md
@@ -292,7 +292,7 @@ You may instead choose to add volumes manually to deployments with the
 (which is actually recommended by Elastic in order to take advantage of
 local disk performance):
 
-    $ oc volume dc/logging-es-rca2m9u8 \
+    $ oc set volume dc/logging-es-rca2m9u8 \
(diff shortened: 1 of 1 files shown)


deployer/README.md
```diff
@@ -1067,11 +1066,7 @@ Deployment failure can happen for a number of transitory reasons, such as
 the image pull taking too long, or nodes being unresponsive. Examine the
 deployer pod logs for possible reasons; but often you can simply redeploy:
 
-    $ oc deploy --latest logging-es-2e7ut0iq
-
-Or you may be able to scale up the existing deployment:
```
kargakis: I've removed this suggestion since the RC at this point is marked as Failed. Scaling it up does not work anymore, I would expect, because of the use of informers in the DC controller (since 3.3 or 3.4, can never remember off the top of my head).

jcantrill:
Manually merging this change since its to the deployer readme.  Hopefully we can drop this dir completely once we move the util scripts

PR #429, opened 2017-05-22
add ability to expose Elasticsearch as an external route

richm:
Now that https://github.com/openshift/openshift-ansible/commit/b61044dfa3669d79bd5e99c846ad4d10de172583
has merged, we can use that code to set up ES routes rather than the
hack in logging.sh
@jcantrill @ewolinetz @stevekuznetsov @lukas-vlcek @nhosoi @noodle PTAL
[test]


stevekuznetsov:
Nice, this really simplifies that Bash!

richm:
[merge]

PR #431, opened 2017-05-23
skip umount in mux mode

richm:
mux does not mount anything from the host, so we can skip the umount
[test]


richm:
[merge]

PR #432, opened 2017-05-23
disable k8s meta plugin for mux client

richm:
One of the reasons for using fluentd as a mux client is to avoid
hitting the API server - this patch makes sure that the fluentd k8s
meta plugin is explicitly disabled.


richm:
@jcantrill @ewolinetz @lukas-vlcek @nhosoi PTAL
[test]

richm:
[merge]

PR #441, opened 2017-05-26
Turn off console logging; add file logging

portante:


deployer/conf/elasticsearch-logging.yml
```diff
@@ -1,16 +1,25 @@
...
+
   # reduce the logging for aws, too much is logged under the default INFO
   com.amazonaws: WARN
 
   io.fabric8.elasticsearch: ${PLUGIN_LOGLEVEL}
   io.fabric8.kubernetes: ${PLUGIN_LOGLEVEL}
 
+  # aws will try to do some sketchy JMX stuff, but its not needed.
+  com.amazonaws.jmx.SdkMBeanRegistrySupport: ERROR
+  com.amazonaws.metrics.AwsSdkMetrics: ERROR
```
lukas-vlcek: @portante Why do we need this? Isn't this relevant to Elastic AWS plugin only? (https://stackoverflow.com/a/35752768)

portante: This brings the `logging.yml` file up-to-date with what is released in the 2.4.x of Elasticsearch itself.  These settings don't harm anything either, and allows us to easily diff with upstream without seeing too many differences to wade through.

lukas-vlcek: ok, I see. Thanks.

jcantrill:
LGTM.  Unless there are any other objects, I can manually merge.  Might we have an associated BZ?

deployer/conf/elasticsearch-logging.yml
```diff
@@ -1,16 +1,25 @@
 # you can override this using by setting a system property, for example -Des.logger.level=DEBUG
 es.logger.level: INFO
-rootLogger: ${es.logger.level}, console
+rootLogger: ${es.logger.level}, file
 logger:
   # log action execution errors for easier debugging
   action: WARN
 
+  # deprecation logging, turn to DEBUG to see them
+  deprecation: WARN, deprecation_log_file
```
richm: Not sure what this comment and level setting mean.  Does it mean no deprecations will be logged unless you set the DEBUG level?

deployer/conf/elasticsearch-logging.yml
```diff
@@ -1,16 +1,25 @@
...
   com.amazonaws: WARN
 
   io.fabric8.elasticsearch: ${PLUGIN_LOGLEVEL}
   io.fabric8.kubernetes: ${PLUGIN_LOGLEVEL}
 
+  # aws will try to do some sketchy JMX stuff, but its not needed.
+  com.amazonaws.jmx.SdkMBeanRegistrySupport: ERROR
+  com.amazonaws.metrics.AwsSdkMetrics: ERROR
+
+  org.apache.http: INFO
```
richm: What is the default log level for this?  By changing it to INFO, are we going to cause a lot more stuff to be logged to the ES logs, that we need to take into consideration for sizing?

deployer/conf/elasticsearch-logging.yml
portante: It means that you will only see deprecation log events in the file that are of a log level `WARN` or higher.  So most likely nothing since we don't mix versions of clients within the cluster.

FWIW, I run using `INFO` on my v2.4.x staging instance and I have seen nothing at all.

deployer/conf/elasticsearch-logging.yml
portante: I believe this is logging information for the HTTP client operations, where one ES makes an HTTP request to some other entity.

I don't think it makes a change to the log level, at least on my staging box I don't see any additional logs.

richm:
lgtm - @jcantrill will need to merge for 1.4 branch

jcantrill:
Updated downstream in: http://pkgs.devel.redhat.com/cgit/rpms/logging-deployment-docker/commit/?h=rhaos-3.4-rhel-7&id=f197b3fd541e27ea8e189404c52f3d8f099ba7ef

PR #448, opened 2017-05-31
Ensure gathering admin token doesn't switch projects

stevekuznetsov:
When determining the admin user and token, we can accidentally end up
switching projects if the admin user and the test user are not in the
same context. By recording the current project before switching over and
returning to that project afterword, we ensure that we do not make this
mistake.

[test]
[test_installer]

### DIFF
--- a/deployer/scripts/util.sh
+++ b/deployer/scripts/util.sh
@@ -373,11 +373,13 @@ function get_latest_pod() {
 
 # set the test_token, test_name, and test_ip for token auth
 function get_test_user_token() {
+    local current_project; current_project="$( oc project -q )"
     oc login --username=${LOG_ADMIN_USER:-${1:-admin}} --password=${LOG_ADMIN_PW:-${2:-admin}} > /dev/null
     test_token="$(oc whoami -t)"
     test_name="$(oc whoami)"
     test_ip="127.0.0.1"
     oc login --username=system:admin > /dev/null
+    oc project "${current_project}" > /dev/null
 }
 
 # $1 - kibana pod name


deployer/scripts/util.sh
```diff
@@ -373,11 +373,13 @@ function get_latest_pod() {
 
 # set the test_token, test_name, and test_ip for token auth
 function get_test_user_token() {
+    local current_project; current_project="$( oc project -q )"
     oc login --username=${LOG_ADMIN_USER:-${1:-admin}} --password=${LOG_ADMIN_PW:-${2:-admin}} > /dev/null
     test_token="$(oc whoami -t)"
     test_name="$(oc whoami)"
     test_ip="127.0.0.1"
     oc login --username=system:admin > /dev/null
+    oc project "${current_project}"
```
richm: `oc project "${current_project}" > /dev/null`

stevekuznetsov: done

deployer/scripts/util.sh
```diff
@@ -373,11 +373,13 @@ function get_latest_pod() {
 
 # set the test_token, test_name, and test_ip for token auth
 function get_test_user_token() {
+    local current_project; current_project="$( oc project -q )"
     oc login --username=${LOG_ADMIN_USER:-${1:-admin}} --password=${LOG_ADMIN_PW:-${2:-admin}} > /dev/null
     test_token="$(oc whoami -t)"
     test_name="$(oc whoami)"
     test_ip="127.0.0.1"
     oc login --username=system:admin > /dev/null
+    oc project "${current_project}" > /dev/null
```
jcantrill: why do we not just specify the token with every oc command?  this would make us agnostic of the settings in kubeconfig?

stevekuznetsov: It's much better for tests to expect to neither change projects nor users. The only reason we need this right now is for token based auth inside kibana pod for es contact. I want to get rid of that everywhere we don't actually need it and keep one auth test where we do it in a controlled manner.

richm:
[merge]

PR #450, opened 2017-05-31
Add a new single entrypoint factored to exclude setup

stevekuznetsov:
Vendor origin/hack/lib at 375c727

---

[carry] Remove references to extraneous scripts

---

[carry] Ensure that $OS_ROOT = $OS_O_A_L_DIR

---

Add the _output/ directory to the gitignore

The Origin tooling will put script output into `_output/scripts`, which
is all information that should not be checked in to source control.

---

Add a new single entrypoint factored to exclude setup

In order to robustly develop test suites that run against clusters that
are not set up by scripts in this repository, we need to approach the
problem from the outset by factoring all test suites not to do their own
set up.

---


stevekuznetsov:
[test]

hack/testing/entrypoint.sh
```diff
@@ -0,0 +1,74 @@
...
+function run_suite() {
+	local test="$1"
+	suite_name="$( basename "${test}" '.sh' )"
+	os::test::junit::declare_suite_start "test/setup/${suite_name}"
+	os::cmd::expect_success "oc login -u system:admin"
+	os::cmd::expect_success "oc project logging"
+	os::test::junit::declare_suite_end
+
+	os::log::info "Logging test suite ${suite_name} started at $( date )"
+	if ! "${test}" "yes"; then
```
richm: What is this `yes` argument?  If this is whether or not the "ops" cluster exists, then
- would prefer to use `true` instead of `yes`
- would prefer to get rid of it altogether, and just have the test figure out if the ops cluster exists and if it should be tested

richm: And - would prefer to rewrite the `if` clause without the `!` -
```
if "${test}" "yes" ; then 
   test succeeded case
else
  test failed case
fi
```

richm:
otherwise, looks good

stevekuznetsov:
I'll use an explanatory variable -- that is for the OPS cluster. I wanted to remove that on a per-test basis following this, but not in this PR.

stevekuznetsov:
@richm comments addressed... once the tests here are green we can move forward. The vendor of the Bash stuff is not as clean as I'd like, but as we move forward that will get better. At least for now it keeps us isolated from Bash stuff changing in Origin and breaking things here.

stevekuznetsov:
@richm OK, tests look good. Once this merges I'll switch over the test_installer job to use it

richm:
[merge]

stevekuznetsov:
@richm guess we're in stage-cut. Hmm. These changes are for tooling, not the project, so there's no chance they can have any issues with impacting what is on Online. You have any qualms around another tag with [severity: blocker] to get around the merge gate?

richm:
[merge][severity: blocker]

richm:
@stevekuznetsov 
```
./test-curator.sh: line 392: runtime: command not found
```

richm:
[merge][severity: blocker]

stevekuznetsov:
Whoops, got a commit in here from a different branch.

stevekuznetsov:
Sorry about that, @richm can you re-tag for merge?

richm:
[merge][severity: blocker]

PR #452, opened 2017-06-01
globally remove imagePullPolicy: Always

jim-minter:
I'd like "imagePullPolicy: Always" to be removed, on the grounds that it makes development harder, and it should not be needed.  Is it possible?

### DIFF
--- a/deployer/deployer.yaml
+++ b/deployer/deployer.yaml
@@ -125,7 +125,6 @@ items:
     spec:
       containers:
       - image: ${IMAGE_PREFIX}logging-deployment:${IMAGE_VERSION}
-        imagePullPolicy: Always
         name: deployer
         volumeMounts:
         - name: empty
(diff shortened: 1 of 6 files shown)


jcantrill:
The deployer is deprecated in master.  These changes will need to be applied to the openshift_logging_* roles of ansible if they have not already been applied.  @ewolinetz @portante  thoughts on wanting to apply this in general?  Make it configurable in ansible?

jim-minter:
@jcantrill is there a note anywhere indicating that the deployer is deprecated? I couldn't find one.

jcantrill:
There is an open PR to modify the docs and correct, but the deployer only exists to support 'oc cluster up' which will be moved to an alternative in the forthcoming release

ewolinetz:
The reason we set the `imagePullPolicy` as `always` is so that we can ensure that when a deployment for logging is done the most recently published images are used. If it is desired that we have a way to not have this enabled then it should be an installation option, or we should document how to remove it manually.

richm:
Can we close this PR?  or move it to openshift-ansible?

jcantrill:
Closing this PR.  @jim-minter please open a PR against openshift-ansible if still appropriate.

PR #455, opened 2017-06-01
Sleep until the Curator delete time instead of a flat time

stevekuznetsov:
If we instruct Curator to delete indices at a specified time, we only
need to wait until that time comes to detect that Curator starts the
deletion task, instead of waiting for the amount of time in the future
that we originally used to determine when curator should delete.

[test]
[test_installer]

/cc @richm


stevekuznetsov:
This does cut down some time @richm 

```
sleeping 224 seconds to see if runhour and runminute are working . . .
sleeping 222 seconds to see if runhour and runminute are working . . .
```
```
sleeping 204 seconds to see if runhour and runminute are working . . .
sleeping 209 seconds to see if runhour and runminute are working . . .
```

richm:
[merge][severity: blocker]

PR #457, opened 2017-06-01
Bumping up default time and changing variable naming for ENV exposure

ewolinetz:
Will also need to make update in openshift-ansible for the readiness probe timeout and expose this env var in the ES DC

### DIFF
--- a/elasticsearch/probe/readiness.sh
+++ b/elasticsearch/probe/readiness.sh
@@ -20,7 +20,7 @@
 ES_REST_BASEURL=https://localhost:9200
 EXPECTED_RESPONSE_CODE=200
 secret_dir=/etc/elasticsearch/secret
-max_time=${max_time:-4}
+max_time=${READINESS_PROBE_TIMEOUT:-30}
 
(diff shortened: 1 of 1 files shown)


richm:
I think we need to merge a change to the DC at the same time: https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging_elasticsearch/templates/es.j2#L103
```
            timeoutSeconds: 4
            periodSeconds: 5
```
I think the way it works is that the timeoutSeconds is the same as the timeout inside the probe script, and periodSeconds must be greater than that.  @lukas-vlcek please comment

ewolinetz:
@richm https://github.com/openshift/openshift-ansible/pull/4351

ewolinetz:
@richm @lukas-vlcek bump on this

ewolinetz:
[merge]

ewolinetz:
i'll also make this change for the enterprise image(s) that apply

stevekuznetsov:
[test_installer]

ewolinetz:
re[merge]

PR #453, opened 2017-06-01
use openshift/origin:v1.5.1 as base for deployer

jim-minter:
the oc in openshift/origin:v1.5.0-alpha.2 does not work against origin master (oc deploy --latest fails).  released/non-alpha oc 1.4.1, 1.5.0 and 1.5.1 are all good - use the latest.

PR #421, opened 2017-05-18
bug 1420217. Update ES plugin that squashes stack on start

jcantrill:
blocked on https://github.com/fabric8io/openshift-elasticsearch-plugin/pull/77

PR #460, opened 2017-06-02
1.5 backports

jcantrill:
This PR brings in relevant changes
* to backport significant fixes from master to 1.5
* deployer changes in support of 'oc cluster up' only

cc @portante


jcantrill:
[test]

jcantrill:
Not exactly same change but should be sufficient to fix #453

deployer/scripts/install.sh
```diff
@@ -312,7 +312,7 @@ function generate_es_template(){
     --param ES_RECOVER_EXPECTED_NODES=${es_recover_expected_nodes} \
     --param ES_RECOVER_AFTER_TIME=${es_recover_after_time} \
     --param STORAGE_GROUP_DEFAULT=${storage_group} \
-    $image_params
+    "$image_params"
```
richm: Is this code, and the following upgrade.sh and test-upgrade.sh code, even called for 1.5?

deployer/scripts/upgrade.sh
```diff
@@ -810,6 +956,14 @@ function upgrade_logging() {
...
           common_data_model=true
           ;;
+<<<<<<< HEAD
+=======
+        7)
+          common_data_model=true
+          add_index_pattern_config
+          add_common_data_model_plugin
+          ;;
+>>>>>>> eddc26f... Bug 1440855 - Missing aliases after upgrade
```
richm: Need to fix this

deployer/scripts/install.sh
jcantrill: Probably not.  Any of the deployer related code is in support of 'oc cluster up' which in theory would not be performing upgrade IMO.  Not sure if you can even tell 'oc cluster up' to 'upgrade'

deployer/scripts/upgrade.sh
```diff
@@ -731,6 +731,7 @@ function update_for_common_data_model() {
   count=$(get_list_of_proj_uuid_indices | wc -l)
   if [ $count -eq 0 ] ; then
       echo No matching indexes found - skipping update_for_common_data_model
+      rm -f $indices
```
richm: I don't know if any of this is needed. If customers upgrade to the latest 1.4 before upgrading to 1.5, this won't be necessary.  If customers are upgrading from 1.3 -> 1.5, I have no idea if that will work.

At any rate, I don't think any of this code is used anymore in 1.5.

jcantrill:
also needs: https://github.com/openshift/origin-aggregated-logging/pull/421

jcantrill:
@lukas-vlcek I would expect the configuration of the probe to be in 1.5... but given it may not be, maybe I should back out the readiness probe.

jcantrill:
[test]

deployer/scripts/util.sh
```diff
@@ -304,58 +304,6 @@ function wait_for_builds_complete()
     return 0
 }
 
-get_error_pod() {
-    # $1 is component for selector
-    oc get pods -l component=$1 | awk -v sel=$1 '$1 ~ sel && ($3 == "Error" || $3 == "CrashLoopBackOff") {print $1}'
-}
-
-wait_for_pod_ACTION() {
```
wozniakjan: Deleting this function appears to be a reason for [Origin Aggregated Logging Test](https://ci.openshift.redhat.com/jenkins/job/test_pull_request_origin_aggregated_logging/99/console) failure, it is still used in `test-es-copy.sh`, `test-fluentd-forward.sh` and `test-viaq-data-model.sh`

wozniakjan: Deleting this function appears to be a reason for [Origin Aggregated Logging Test](https://ci.openshift.redhat.com/jenkins/job/test_pull_request_origin_aggregated_logging/99/console) failure, it is still used in `test-es-copy.sh`, `test-fluentd-forward.sh` and `test-viaq-data-model.sh`

jcantrill:
@richm I was able to stand up a cluster and confirm log flow manually.  Additionally I was able to deploy logging to the cluster using the deployer.  Merging as I am satisfied.

PR #461, opened 2017-06-02
Introducing environment variables: MUX_CPU_LIMIT and MUX_MEMORY_LIMIT…

nhosoi:
… with the default value set {500m, 2Gi}.

Setting them to openshift_logging_mux_cpu_limit and openshift_logging_mux_memory_limit
in the inventory file, respectively.

Setting BUFFER_QUEUE_LIMIT and BUFFER_SIZE_LIMIT to openshift_logging_mux_buffer_queue_limit
and openshift_logging_mux_buffer_size_limit, as well.

Switching to buffer_type file
buffer_path /var/fluentd/file_buffer_*
buffer_chunk_limit 1m
buffer_queue_limit 1024
flush_interval 1s

### DIFF
--- a/deployer/templates/fluentd.yaml
+++ b/deployer/templates/fluentd.yaml
@@ -343,4 +343,4 @@ parameters:
 -
   description: 'Fluentd buffer size limit'
   name: BUFFER_SIZE_LIMIT
-  value: "16777216"
+  value: "1048576"
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -59,7 +59,12 @@ IPADDR6=`/usr/sbin/ip -6 addr show dev eth0 | grep inet6 | sed "s/[ \t]*inet6 \(
 export IPADDR4 IPADDR6
 
 export BUFFER_QUEUE_LIMIT=${BUFFER_QUEUE_LIMIT:-1024}
-export BUFFER_SIZE_LIMIT=${BUFFER_SIZE_LIMIT:-16777216}
+export BUFFER_SIZE_LIMIT=${BUFFER_SIZE_LIMIT:-1048576}
+export MUX_CPU_LIMIT=${MUX_CPU_LIMIT:-500m}
+export MUX_MEMORY_LIMIT=${MUX_MEMORY_LIMIT:-2Gi}
+
+mkdir /var/fluentd
+chmod 0700 /var/fluentd
 
 CFG_DIR=/etc/fluent/configs.d
 if [ "${USE_MUX:-}" = "true" ] ; then
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -58,13 +58,11 @@ IPADDR4=`/usr/sbin/ip -4 addr show dev eth0 | grep inet | sed -e "s/[ \t]*inet \
 IPADDR6=`/usr/sbin/ip -6 addr show dev eth0 | grep inet6 | sed "s/[ \t]*inet6 \([a-f0-9:]*\).*/\1/"`
 export IPADDR4 IPADDR6
 
-export BUFFER_QUEUE_LIMIT=${BUFFER_QUEUE_LIMIT:-1024}
 export BUFFER_SIZE_LIMIT=${BUFFER_SIZE_LIMIT:-1048576}
 export MUX_CPU_LIMIT=${MUX_CPU_LIMIT:-500m}
 export MUX_MEMORY_LIMIT=${MUX_MEMORY_LIMIT:-2Gi}
-
-mkdir /var/fluentd
-chmod 0700 /var/fluentd
+export FLUENTD_CPU_LIMIT=${FLUENTD_CPU_LIMIT:-100m}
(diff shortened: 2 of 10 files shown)


nhosoi:
Could you please take a look, @richm, @portante and @jcantrill?
Thanks.

richm:
@nhosoi The changing of the flush interval, and adding file buffer support, should be in a separate commit

richm:
What is the relationship between the buffer_queue_limit, buffer_size_limit, and RAM consumed?  is it `buffer_queue_limit * buffer_size_limit == RAM bytes used`?

nhosoi:
> The changing of the flush interval, and adding file buffer support, should be in a separate commit
Ok. Making another PR and dropping them from this one.

> buffer_queue_limit * buffer_size_limit == RAM bytes used?
At most, I believe.  Watching the behaviour with "buffer_type file", the buffer file grows/shrinks/disappears.  But not sure if it is in memory.
And if you have multiple output (e.g., es and es-ops), buffer_queue_limit * buffer_size_limit * number_of_output.

richm:
> Watching the behaviour with "buffer_type file", the buffer file grows/shrinks/disappears.

Right.  That's a different issue.

My general concern is that if configure buffering, we need to make sure that we have enough space for the buffer.

For example, if you are using memory buffering (the default): if you have 512MB RAM for the mux pod, and there are two buffers (one for es, one for es-ops), then each buffer can use no more than 256MB RAM, minus some RAM for fluentd itself and other overhead, so probably 200MB for each buffer or something like that.

If using file buffering, we need to make sure there is enough disk space for the file buffer.  There are some other considerations for file buffering as well.  We want to be able to use a persistent volume for the file buffer, rather than the default which is to use the internal container file system.  That might be a hostMount, or a PV/PVC configured in openshift.

So how many bytes (memory or file) does buffer_queue_limit and buffer_size_limit use?

deployer/templates/fluentd.yaml
```diff
@@ -343,4 +343,4 @@ parameters:
 -
   description: 'Fluentd buffer size limit'
   name: BUFFER_SIZE_LIMIT
-  value: "16777216"
+  value: "1048576"
```
jcantrill: Not required as this file is not relevant in master/3.6

richm: it's only used for testing, and only until we get rid of the need to use the logging-fluentd-template and rely directly on the daemonset config.

stevekuznetsov:
[test_installer]

stevekuznetsov:
[test]

richm:
[test]

richm:
[test]

richm:
[test]

richm:
[test]

richm:
kibana build [test] flake?

fluentd/run.sh
```diff
@@ -58,13 +58,11 @@ IPADDR4=`/usr/sbin/ip -4 addr show dev eth0 | grep inet | sed -e "s/[ \t]*inet \
 IPADDR6=`/usr/sbin/ip -6 addr show dev eth0 | grep inet6 | sed "s/[ \t]*inet6 \([a-f0-9:]*\).*/\1/"`
 export IPADDR4 IPADDR6
 
-export BUFFER_QUEUE_LIMIT=${BUFFER_QUEUE_LIMIT:-1024}
 export BUFFER_SIZE_LIMIT=${BUFFER_SIZE_LIMIT:-1048576}
 export MUX_CPU_LIMIT=${MUX_CPU_LIMIT:-500m}
 export MUX_MEMORY_LIMIT=${MUX_MEMORY_LIMIT:-2Gi}
```
richm: I don't think these MUX variables need to be exported

fluentd/run.sh
```diff
@@ -58,13 +58,11 @@ IPADDR4=`/usr/sbin/ip -4 addr show dev eth0 | grep inet | sed -e "s/[ \t]*inet \
...
-
-mkdir /var/fluentd
-chmod 0700 /var/fluentd
+export FLUENTD_CPU_LIMIT=${FLUENTD_CPU_LIMIT:-100m}
+export FLUENTD_MEMORY_LIMIT=${FLUENTD_MEMORY_LIMIT:-512Mi}
```
richm: I don't think these need to be exported

fluentd/run.sh
nhosoi: Ok.  I removed them.  Thanks!

fluentd/run.sh
nhosoi: ditto.

richm:
[test]

richm:
```
/data/src/github.com/openshift/origin-aggregated-logging/hack/testing/init-log-stack: line 20: SET_MUX_CPU_LIMIT: unbound variable
```

nhosoi:
Thank you, @richm.
I pushed the fix.

nhosoi:
@richm, I removed the old test-mux-0 branch, squashed the commits into one, rebased with the latest master, then pushed to test-mux-0.  Sorry, I think I mistakenly closed this pr.  Is it ok to open a new pr?

PR #462, opened 2017-06-02
Retry the searchguard init until it succeeds

portante:
We repeatedly call the sgadmin script until it successfully returns,
sleeping 10 seconds between retries.

Partial fix for BZ #1457642

### DIFF
--- a/elasticsearch/run.sh
+++ b/elasticsearch/run.sh
@@ -128,30 +128,30 @@ wait_for_port_open() {
         --max-time $max_time \
         -o $LOG_FILE -w '%{response_code}' \
         $ES_REST_BASEURL) || test $response_code != "200"
-do
-    sleep $RETRY_INTERVAL
-    (( retry -= 1 )) || :
-    if (( retry == 0 )) ; then
-        timeouted=true
-        break
-    fi
-done
+    do
+        sleep $RETRY_INTERVAL
+        (( retry -= 1 )) || :
+        if (( retry == 0 )) ; then
+            timeouted=true
+            break
+        fi
+    done
 
-if [ $timeouted = true ] ; then
-    error "Timed out waiting for Elasticsearch to be ready"
-else
+    if [ $timeouted = true ] ; then
+        error "Timed out waiting for Elasticsearch to be ready"
+    else
+        rm -f $LOG_FILE
+        info Elasticsearch is ready and listening at $ES_REST_BASEURL
+        return 0
+    fi
+    cat $LOG_FILE
     rm -f $LOG_FILE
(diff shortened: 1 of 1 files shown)


jcantrill:
Ref https://github.com/openshift/openshift-ansible/pull/4364

jcantrill:
[test]

elasticsearch/run.sh
```diff
@@ -162,13 +162,11 @@ seed_searchguard(){
...
-      info "Seeded the searchguard ACL index"
-    else
-      error "Error seeding the searchguard ACL index"
-      exit 1
-    fi
+    do
+        warn "Error seeding the searchguard ACL index... retrying in 10 seconds"
+        sleep 10
+    done
+    info "Seeded the searchguard ACL index"
```
richm: Don't we need a timeout?

portante: The timeout today is killing us.  If the sgadmin script does not eventually succeed, then we are left in a lurch.  Nothing can move forward.  This way, one can correct the environment if possible to enable the sgadmin script to succeed (like fixing bad indices, or other pods causing a problem, etc.).

richm: Ok, but infinite loops are generally bad.  How about a 24 hour timeout?

portante: Is it an infinite loop if the loop ends when `sgadmin` succeeds?

If we don't do it this way, then all we are doing is lengthening the timeout to an arbitrary point that will work for some cases and not for others, right?

The problem today is that it can take a very long to load up all the Elasticsearch indices from slow storage mediums, where ES pods are under-powered, or the load on the node on which they are executing is too high.  Without a successful `sgadmin` init of a new index, the current ES pod will not allow fluentd or kibana access.  Eventually logs will be dropped by fluentd, and Kibana users are given very confusing behavioral messages.

Most of the time we have seen ES after 10 to 20 minutes load all the indices to the point where it exits its `red` state, and then we can manually run `sgadmin` to get things flowing.

Regardless of timeout, the logging system does not work at all if sgadmin does not properly get run on ES pod startup.

If there is another more elegant way to solve this, let's do it.

richm: > Is it an infinite loop if the loop ends when sgadmin succeeds?

It is an infinite loop if sgadmin never succeeds.

richm: Are you saying that a timeout of 1 week is unreasonable?

richm: I'm not going to approve this patch, but I'm not going to disapprove either.  If someone else is willing to approve, I have no objections.

jcantrill:
closing in favor or #469

PR #464, opened 2017-06-06
Update future log test, enable it in the entrypoint

stevekuznetsov:

/cc @richm 
[test]


test/future_dated_log.sh
```diff
@@ -0,0 +1,54 @@
...
+# control over the date and format so have to use sudo to
+# write directly to /var/log/messages
+message_date="$( date -u +"%b %d %H:%M:%S" --date="1 hour hence" )"
+message_uuid="$( uuidgen )"
+message="${message_date} localhost ${message_uuid}: ${message_uuid} message from test/future_dated_log"
+echo "${message}" | ${USE_SUDO:+sudo} tee -a /var/log/messages >/dev/null
+
+for elasticsearch_pod in $( oc get pods --selector component="${OAL_ELASTICSEARCH_COMPONENT}" -o jsonpath='{ .items[*].metadata.name }' ); do
+	os::log::info "Ensuring Elasticsearch pod ${elasticsearch_pod} persisted the message added to /var/log/messages..."
+	os::cmd::try_until_text "curl_es '${elasticsearch_pod}' '/.operations*/_count?q=systemd.u.SYSLOG_IDENTIFIER:${message_uuid}' | python -c 'import json, sys; print json.load(sys.stdin)[\"count\"];'" "1"
```
stevekuznetsov: Unclear how you feel about this -- before, there was a call-stack about six deep in this specific call. Clearly there needs to be a less verbose way to `curl` ES from within ES, but I was not certain that the complexity that was here before was useful. Made it hard to understand the test.

richm: Works for me. I realize someone else has to maintain this code other than me :-)

test/future_dated_log.sh
```diff
@@ -0,0 +1,54 @@
...
+
+# NOTE: can't use `logger` for this because we need complete
+# control over the date and format so have to use sudo to
+# write directly to /var/log/messages
+message_date="$( date -u +"%b %d %H:%M:%S" --date="1 hour hence" )"
+message_uuid="$( uuidgen )"
+message="${message_date} localhost ${message_uuid}: ${message_uuid} message from test/future_dated_log"
+echo "${message}" | ${USE_SUDO:+sudo} tee -a /var/log/messages >/dev/null
+
+for elasticsearch_pod in $( oc get pods --selector component="${OAL_ELASTICSEARCH_COMPONENT}" -o jsonpath='{ .items[*].metadata.name }' ); do
```
richm: messages written to `/var/log/messages` will _only_ be in the es-ops, not in es.  So you only have to query es-ops.  Of course, when not using the OPS cluster, es will be the same as es-ops.

test/future_dated_log.sh
richm: You are using `${message_uuid}` inside a single quote string - not sure how `try_until_text` will eval this string - are you sure `${message_uuid}` will be expanded to its value?

stevekuznetsov: Yeah, it's a little bit of a weird API but such is life with Bash

The double quote expands it all first, so you get `curl es 'pod-name=here'` -- then that gets `eval`ed inside of the `os::cmd` utility.

test/future_dated_log.sh
stevekuznetsov: Right, so this is why I run the test as such from `hack/testing/test-datetime-future.sh`:

```sh
if [[ $# -eq 1 ]]; then
    # There is an ops cluster set up, so we
    # need to expect to see the log entry in
    # the ops Elasticsearch.
    OAL_ELASTICSEARCH_COMPONENT="es-ops" "${OS_O_A_L_DIR}/test/future_dated_log.sh"
else
    OAL_ELASTICSEARCH_COMPONENT="es" "${OS_O_A_L_DIR}/test/future_dated_log.sh"
fi
```

stevekuznetsov:
@richm can you help me figure out what is going wrong in that test?

```
running test test-curator.sh
configmap "logging-curator" deleted
configmap "logging-curator" created
deploymentconfig "logging-curator" scaled
deploymentconfig "logging-curator" scaled
Error: the curator pod should be in the error state
logging-curator-1-sq55q
Error: did not find the correct error message
error: expected 'logs (POD | TYPE/NAME) [CONTAINER_NAME]'.
POD or TYPE/NAME is a required argument for the logs command
See 'oc logs -h' for help and examples.
The project name length must be less than or equal to 63 characters.  This is too long: [this-project-name-is-far-far-too-long-this-project-name-is-far-far-too-long-this-project-name-is-far-far-too-long-this-project-name-is-far-far-too-long]
configmap "logging-curator" deleted
configmap "logging-curator" created
deploymentconfig "logging-curator" scaled
deploymentconfig "logging-curator" scaled
[ERROR] PID 4255: /data/src/github.com/openshift/origin-aggregated-logging/logging.sh:303: `echo running test $test` exited with status 1.
[INFO] 		Stack Trace: 
[INFO] 		  1: /data/src/github.com/openshift/origin-aggregated-logging/logging.sh:303: `echo running test $test`
[INFO]   Exiting with code 1.
```

richm:
@stevekuznetsov not sure why the curator test is so fragile lately . . . curator has incorrect configuration - the test expects the pod to be in the Error or CrashLoopBackoff state, but it either isn't running, or isn't in that state.  Could be a timing issue.
[test]

stevekuznetsov:
@richm do we have tests running on `master`? I'll try to bisect where it started...

stevekuznetsov:
@richm also for what it's worth, it looks like a `logging.sh` issue only

richm:
[test]

richm:
[test]

richm:
another curator [test] flake?

richm:
[merge]

richm:
[test]

richm:
[merge]

richm:
curator timing flake [merge]

richm:
[merge]

richm:
[merge]

richm:
I guess bash doesn't like this
```
running test test-datetime-future.sh
./../../hack/lib/init.sh: line 29: os::util::absolute_path: readonly function
/data/src/github.com/openshift/origin-aggregated-logging/hack/lib/build/environment.sh: line 68: os::build::environment::create: readonly function
/data/src/github.com/openshift/origin-aggregated-logging/hack/lib/build/environment.sh: line 82: os::build::environment::release::workingdir: readonly function
/data/src/github.com/openshift/origin-aggregated-logging/hack/lib/build/environment.sh: line 95: os::build::environment::cleanup: readonly function
/data/src/github.com/openshift/origin-aggregated-logging/hack/lib/build/environment.sh: line 138: os::build::environment::start: readonly function
/data/src/github.com/openshift/origin-aggregated-logging/hack/lib/build/environment.sh: line 187: os::build::environment::withsource: readonly function
/data/src/github.com/openshift/origin-aggregated-logging/hack/lib/build/environment.sh: line 218: os::build::environment::run: readonly function
/data/src/github.com/openshift/origin-aggregated-logging/hack/lib/build/constants.sh: line 5: OS_BUILD_ENV_GOLANG: readonly variable
```

stevekuznetsov:
I think `logging.sh` makes some bad decisions re: `source hack/lib./init.sh` -- looking

stevekuznetsov:
We have logic for making that a non-issue in Origin, I wonder if the way that `logging.sh` is trying to override the `$TMPDIR` stuff is confusing it. Also not clear why the `source` doesn't also fail on `check-EFK-running.sh` or `check-logs.sh` calls/

richm:
[test]

stevekuznetsov:
@richm I think it was because we were `source`-ing  the test scripts instead of just running them.

stevekuznetsov:
Unclear what this means ...

```
[INFO] Ensuring Fluentd pod logging-fluentd-z4j2x timezone matches node timezone...
Running test/future_dated_log.sh:23: executing 'oc exec logging-fluentd-z4j2x -- date +%z' expecting success and text '-0400'...
FAILURE after 0.350s: test/future_dated_log.sh:23: executing 'oc exec logging-fluentd-z4j2x -- date +%z' expecting success and text '-0400': the output content test failed
-0400

There was no error output from the command.
```

stevekuznetsov:
Oh... `grep "-whatever"` gets interpreted as a flag. Yuck.

stevekuznetsov:
Going to change `date +%z` to `date +%Z` so we don't have to hack around that.

richm:
[merge]

PR #465, opened 2017-06-06
Add a script to automatically update the hack/lib

stevekuznetsov:
We vendor the Origin Bash libraries at hack/lib. We will need to update
that copy every so often while also keep intact the commits we intend to
carry on top. The `hack/update-lib.sh` script aims to automate that
process to some extent.

/cc @richm here is a simple script -- we should not run it until we have https://github.com/openshift/origin/pull/14437 merged


richm:
[merge]

PR #469, opened 2017-06-07
bug 1457642. Fix SG timeout

jcantrill:
This PR fixes SG timing out when it takes an extended period of time getting to yellow state:

* Adds loop to continue to try and seed
* Adds logging to identify the number of red indicies
* refactors es_util to make it easier to specify alternate queries

This PR overrides #462

### DIFF
--- a/elasticsearch/utils/es_seed_acl
+++ b/elasticsearch/utils/es_seed_acl
@@ -1,0 +1,41 @@
+#!/bin/bash -e
+
+source "logging"
+
+MAX_WAIT_SEED=${MAX_WAIT_SEED:-$((60*60*24*7))} #one week
+
+info "Seeding the searchguard ACL index.  Will wait up to $MAX_WAIT_SEED seconds."
+
+config=/usr/share/elasticsearch/config/elasticsearch.yml
+index=$(python -c "import yaml; print yaml.load(open('/usr/share/elasticsearch/config/elasticsearch.yml'))['searchguard']['config_index_name']" | xargs -i sh -c "echo {}")
+
+now=0
+numretries=0
+loginterval=${SG_ADMIN_LOG_INTERVAL:-100}
+while ! /usr/share/elasticsearch/plugins/openshift-elasticsearch/sgadmin.sh \
+    -cd ${HOME}/sgconfig \
+    -i $index \
+    -ks /etc/elasticsearch/secret/searchguard.key \
+    -kst JKS \
+    -kspass kspass \
+    -ts /etc/elasticsearch/secret/searchguard.truststore \
+    -tst JKS \
+    -tspass tspass \
+    -nhnv \
+    -icl ${DEBUG:+-dg} && [ ${now} -lt ${MAX_WAIT_SEED} ]
+do
+    if ! expr $numretries % $loginterval ; then
+        warn "Error seeding the searchguard ACL index... retrying in 10 seconds - $numretries retries so far"
+        warn "Seeding will continue to fail until the cluster status is YELLOW"
+        info "Remaining red indices: $(QUERY=_cat/indices es_util | grep "^red" | wc -l)"
+    fi
+    numretries=$((numretries+1))
+    sleep 10
+    now=$((now+40)) #wait plus 30s timeout from sgadmin
+done
+
+if [ ${now} -ge ${MAX_WAIT_SEED} ]; then
+    error "Giving up trying to seed ACL"
+    exit 1
+fi
+info "Seeded the searchguard ACL index"
--- a/elasticsearch/Dockerfile
+++ b/elasticsearch/Dockerfile
@@ -42,6 +42,7 @@ ADD install.sh ${HOME}/
 RUN ${HOME}/install.sh
 ADD run.sh ${HOME}/
 COPY utils/** /usr/local/bin/
+RUN ln -s /usr/local/bin/logging ${HOME}/logging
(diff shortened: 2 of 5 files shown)


jcantrill:
@richm Does this satisfy your concerns regarding the need for a timeout?

elasticsearch/utils/es_seed_acl
```diff
@@ -0,0 +1,32 @@
...
+    -kspass kspass \
+    -ts /etc/elasticsearch/secret/searchguard.truststore \
+    -tst JKS \
+    -tspass tspass \
+    -nhnv \
+    -icl ${DEBUG:+-dg} && [ ${now} -lt ${MAX_WAIT_SEED} ]
+do
+    warn "Error seeding the searchguard ACL index... retrying in 10 seconds"
+    warn "Seeding will continue to fail until the cluster status is YELLOW"
+    info "Remaining red indicies: $(QUERY=_cat/indicies es_util | grep "^red" | wc -l)"
```
richm: `indices`

portante: And you need to check that node quorum has been reached, as well.

elasticsearch/utils/es_seed_acl
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash -e
+
+source "logging"
+
+MAX_WAIT_SEED=${MAX_WAIT_SEED:-$((60*60*24))}
```
richm: should this be 1 week?

elasticsearch/utils/es_seed_acl
```diff
@@ -0,0 +1,32 @@
...
+    -tst JKS \
+    -tspass tspass \
+    -nhnv \
+    -icl ${DEBUG:+-dg} && [ ${now} -lt ${MAX_WAIT_SEED} ]
+do
+    warn "Error seeding the searchguard ACL index... retrying in 10 seconds"
+    warn "Seeding will continue to fail until the cluster status is YELLOW"
+    info "Remaining red indicies: $(QUERY=_cat/indicies es_util | grep "^red" | wc -l)"
+    sleep 10
+    now=$((now+40)) #wait plus 30s timeout from sgadmin
```
richm: The timeout is 30 seconds from sgadmin?  Then we will get this error message every 40 seconds?  That's an awful lot of spam if sgadmin is having problems.  Change it so that it only prints the message every hour or so, and print the number of retry attempts made in that interval.

elasticsearch/utils/es_seed_acl
```diff
@@ -0,0 +1,32 @@
...
+done
+
+if [ ${now} -ge ${MAX_WAIT_SEED} ]; then
+    error "Giving up trying to seed ACL"
```
portante: How does this help us in the field?  If it takes longer than the timeout, are the ES pods still up and running?  How does an administrator know that this condition has occurred?  Is it reported some where in the `oc get pods` output?  If we are not going to stop the ES pod itself, then we'll be right back to what we have today, no?

elasticsearch/utils/es_seed_acl
jcantrill: @portante educate me what you wish to do

elasticsearch/utils/es_seed_acl
jcantrill: This is to appease @richm suggestion we have a timeout and this provides a message why we are giving up.  I agree in practice a timeout is not useful really given it when we give up it means:
* cluster restarting the pod will could cause the ES cluster to take longer to recover indicies
* the es cluster is not functional until we seed

You would see this if you 'oc logs' for the pod in question.

jcantrill:
[test]

elasticsearch/utils/es_seed_acl
richm: Also - if the script exits with 1 when it times out, then the pod state will be `Error` or `CrashLoopBackoff` or some other error condition, as reported by `oc get pods` among others.

elasticsearch/utils/es_seed_acl
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash -e
+
+source "logging"
+
+MAX_WAIT_SEED=${MAX_WAIT_SEED:-$((60*60*24*7))} #one week
```
richm: ack

richm:
[test]

jcantrill:
@richm squashed your commit into mine but it doesnt look like it gave you any credit.  Thanks for the changes.

richm:
[test]

richm:
[test]

portante:
So after one week, the attempt to initialize sgadmin will fail.  Then what?

 * What problem does the timeout solve?  How are we documenting this to the users?
 * How are we going to tell support folks to deal with this?
 * Why are we providing a path that requires additional direct intervention by the operator?

Recording these questions here so the previous discussion in #462 is not lost when somebody wants to track down the behavior and why a timeout was maintained.

portante:
@jcantrill, I don't agree with the timeout, but the rest of the changes are fine.  We have figured out a way to get rid of the need to initialize sgadmin at the beginning of each ES pod's startup, so this will eventually be a non-issue.

jcantrill:
[merge]

jcantrill:
backport 1.5 in #519

jcantrill:
backport 1.4 in #524

PR #472, opened 2017-06-07
Update to common data model 1.6.0

richm:
Make sure the `level` field uses `notice`, `err`, `info`, etc. string
values rather than numbers or any other value.  If the value to be
mapped is unknown, the value of `unknown` will be used as the level
field value.
Add `DATA_VERSION=1.6.0` env. var.
Use FLUENTD_VERSION + DATA_VERSION to construct the version used in
the pipeline_metadata.
Add the `systemd` fields to container records read from journald.
Add new common data model fields and namespaces to viaq plugin.
@jcantrill @lukas-vlcek @nhosoi @portante PTAL
[test]


richm:
curator flake? [merge]

richm:
[merge]

PR #474, opened 2017-06-08
Save Kibana metadata in its own namespace; use message

portante:
Instead of throwing out the metadata in the log entry metadata record,
we move it to its own Kibana namespace.

We also move to using the `message` field instead of using `log`.

### DIFF
--- a/fluentd/configs.d/openshift/filter-kibana-transform.conf
+++ b/fluentd/configs.d/openshift/filter-kibana-transform.conf
@@ -2,7 +2,9 @@
   @type record_transformer
   enable_ruby
   <record>
-    log ${(err rescue nil) || (msg rescue nil) || record['MESSAGE'] || log}
+    message ${(err rescue nil) || (msg rescue nil) || record['MESSAGE'] || log}
+    kibana {"req":"${record['req']}","res":"${record['res']}","name":"${record['name']}","level":"${record['level']}","v":"${record['v']}","pid":"${record['pid']}"}
   </record>
-  remove_keys req,res,msg,name,level,v,pid,err
+  # remove the keys that have been moved to their new names above.
+  remove_keys msg,MESSAGE,log,err,req,res,name,level,v,pid
 </filter>


portante:
This is an attempt to address: https://bugzilla.redhat.com/show_bug.cgi?id=1459943

richm:
@portante Thanks, but
* this would break any/all customers that rely on the kibana fields being top level fields - so we would need to make this behavior configurable and default to _off_
* we need a namespace in viaq/elasticsearch-templates - and a build, release, etc.
* The problem with QE is that they keep attempting to test the fix, but the images with the fix haven't been built yet

richm:
> The problem with QE is that they keep attempting to test the fix, but the images with the fix haven't been built yet

Sorry, I got the wrong bz

richm:
> this would break any/all customers that rely on the kibana fields being top level fields - so we would need to make this behavior configurable and default to off

No, that's not correct.  AFAICT upon further investigation, it does not take away any fields.

richm:
so all we really need is a namespace/template

richm:
Here is a full kibana log message from the pod log:
```
{
    "@timestamp": "2017-06-10T01:51:22Z",
    "message": "HEAD / 200 2ms - 9.0B",
    "method": "head",
    "pid": 1,
    "req": {
        "headers": {
            "accept": "*/*",
            "host": "localhost:5601",
            "user-agent": "curl/7.29.0"
        },
        "method": "head",
        "remoteAddress": "127.0.0.1",
        "url": "/",
        "userAgent": "127.0.0.1"
    },
    "res": {
        "contentLength": 9,
        "responseTime": 2,
        "statusCode": 200
    },
    "statusCode": 200,
    "tags": [],
    "type": "response"
}
```

jcantrill:
[(?) Add Kibana Namespace for metadata](https://trello.com/c/Rt66l7qM/574-add-kibana-namespace-for-metadata)

jcantrill:
Closing for now.  PR is referenced in RFE request.

PR #479, opened 2017-06-09
use SG index defined in ES config

jcantrill:
This PR
* modified seeding to get the SG index from the ES config

cc @stevekuznetsov

### DIFF
--- a/elasticsearch/run.sh
+++ b/elasticsearch/run.sh
@@ -151,9 +151,13 @@ exit 1
 
 seed_searchguard(){
     info Seeding the searchguard ACL index
+
+    config=/usr/share/elasticsearch/config/elasticsearch.yml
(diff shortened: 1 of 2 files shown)


jcantrill:
[test]

elasticsearch/run.sh
```diff
@@ -151,9 +151,13 @@ exit 1
 
 seed_searchguard(){
     info Seeding the searchguard ACL index
+
+    config=/usr/share/elasticsearch/config/elasticsearch.yml
```
richm: I think this path is different in origin vs. enterprise?

jcantrill: probably... i feel like there is a bunch of settings that should probably be moved to some sort of ENV script

richm:
[merge]

jcantrill:
backport 1.5 in #519

PR #482, opened 2017-06-13
1. Introducing following environment variables with the initial value…

nhosoi:
… for fluentd.

  MUX_CPU_LIMIT: 500m
  MUX_MEMORY_LIMIT: 2Gi
  FLUENTD_CPU_LIMIT: 100m
  FLUENTD_MEMORY_LIMIT: 512Mi
They are configurable by openshift-ansible, as well.
  openshift_logging_fluentd_cpu_limit: 500m
  openshift_logging_fluentd_memory_limit: 2Gi
  openshift_logging_mux_cpu_limit: 500m
  openshift_logging_mux_memory_limit: 2Gi

2. Introducing an environment variable BUFFER_SIZE_LIMIT for the fluentd buffer_chunk_limit.
   Using the given BUFFER_SIZE_LIMIT, the available RAM size and number of output,
   BUFFER_QUEUE_LIMIT is calculated.
   The BUFFER_SIZE_LIMIT value and the BUFFER_QUEUE_LIMIT value are set to buffer_chunk_limit
   and buffer_queue_limit, respectively.

Note: replacement for pr #461

### DIFF
--- a/deployer/templates/fluentd.yaml
+++ b/deployer/templates/fluentd.yaml
@@ -343,4 +343,4 @@ parameters:
 -
   description: 'Fluentd buffer size limit'
   name: BUFFER_SIZE_LIMIT
-  value: "16777216"
+  value: "1048576"
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -58,8 +58,11 @@ IPADDR4=`/usr/sbin/ip -4 addr show dev eth0 | grep inet | sed -e "s/[ \t]*inet \
 IPADDR6=`/usr/sbin/ip -6 addr show dev eth0 | grep inet6 | sed "s/[ \t]*inet6 \([a-f0-9:]*\).*/\1/"`
 export IPADDR4 IPADDR6
 
-export BUFFER_QUEUE_LIMIT=${BUFFER_QUEUE_LIMIT:-1024}
-export BUFFER_SIZE_LIMIT=${BUFFER_SIZE_LIMIT:-16777216}
+BUFFER_SIZE_LIMIT=${BUFFER_SIZE_LIMIT:-1048576}
+MUX_CPU_LIMIT=${MUX_CPU_LIMIT:-500m}
+MUX_MEMORY_LIMIT=${MUX_MEMORY_LIMIT:-2Gi}
+FLUENTD_CPU_LIMIT=${FLUENTD_CPU_LIMIT:-100m}
+FLUENTD_MEMORY_LIMIT=${FLUENTD_MEMORY_LIMIT:-512Mi}
 
 CFG_DIR=/etc/fluent/configs.d
 if [ "${USE_MUX:-}" = "true" ] ; then
@@ -100,6 +103,33 @@ if [ "${USE_MUX_CLIENT:-}" = "true" ] ; then
 else
     rm -f $CFG_DIR/openshift/filter-pre-mux-client.conf
 fi
+
+if [ "${USE_MUX:-}" = "true" ] ; then
+    TOTAL_MEMORY_LIMIT=`echo $MUX_MEMORY_LIMIT |  sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc`
+else
+    TOTAL_MEMORY_LIMIT=`echo $FLUENTD_MEMORY_LIMIT |  sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc`
+fi
+BUFFER_SIZE_LIMIT=`echo $BUFFER_SIZE_LIMIT |  sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc`
+if [ $BUFFER_SIZE_LIMIT -eq 0 ]; then
+    BUFFER_SIZE_LIMIT=1048576
+fi
+
+DIV=1
+if [ "$ES_HOST" != "$OPS_HOST" ] || [ "$ES_PORT" != "$OPS_PORT" ] ; then
+    # using ops cluster
+    DIV=`expr $DIV \* 2`
+fi
+if [ "${USE_MUX_CLIENT:-}" = "true" ] ; then
+    DIV=`expr $DIV \* 2`
+fi
+
+# MEMORY_LIMIT per buffer
+MEMORY_LIMIT=`expr $TOTAL_MEMORY_LIMIT / $DIV`
+BUFFER_QUEUE_LIMIT=`expr $MEMORY_LIMIT / $BUFFER_SIZE_LIMIT`
+if [ $BUFFER_QUEUE_LIMIT -eq 0 ]; then
+    BUFFER_QUEUE_LIMIT=1024
+fi
+export BUFFER_QUEUE_LIMIT BUFFER_SIZE_LIMIT
 
 OPS_COPY_HOST="${OPS_COPY_HOST:-$ES_COPY_HOST}"
 OPS_COPY_PORT="${OPS_COPY_PORT:-$ES_COPY_PORT}"
--- a/fluentd/Dockerfile
+++ b/fluentd/Dockerfile
@@ -23,6 +23,7 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
       ruby-devel \
       libcurl-devel \
       make \
+      bc \
       iproute && \
     yum clean all
 RUN mkdir -p ${HOME} && \
--- a/fluentd/configs.d/filter-pre-mux-client.conf
+++ b/fluentd/configs.d/filter-pre-mux-client.conf
@@ -5,8 +5,9 @@
   shared_key "#{File.open('/etc/fluent/muxkeys/shared_key') do |f| f.readline end.rstrip}"
   secure yes
   ca_cert_path /etc/fluent/muxkeys/ca
-  buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT']}"
-  buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT']}"
+  flush_interval 5s
+  buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
+  buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT'] || '1m' }"
   <server>
     host logging-mux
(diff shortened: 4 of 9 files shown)


richm:
[merge]

fluentd/run.sh
```diff
@@ -101,6 +104,33 @@ else
...
+fi
+BUFFER_SIZE_LIMIT=`echo $BUFFER_SIZE_LIMIT |  sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc`
+if [ $BUFFER_SIZE_LIMIT -eq 0 ]; then
+    BUFFER_SIZE_LIMIT=1048576
+fi
+
+DIV=1
+if [ "$ES_HOST" != "$OPS_HOST" ] || [ "$ES_PORT" != "$OPS_PORT" ] ; then
+    # using ops cluster
+    DIV=`expr $DIV \* 2`
```
portante: Don't we always have one fluentd pod in the non-mux case that sends always has _two_ output queues, one for ops logs and one for apps logs?  In the ops case, don't we just send it to different ES instance?

When using mux, won't the collector only have one queue for forwarding?

fluentd/run.sh
```diff
@@ -101,6 +104,33 @@ else
...
+fi
+if [ "${USE_MUX_CLIENT:-}" = "true" ] ; then
+    DIV=`expr $DIV \* 2`
```
portante: In the MUX case won't still only ever have 2 queues regardless of ops vs non-ops cases?

fluentd/run.sh
```diff
@@ -101,6 +104,33 @@ else
...
+fi
+
+# MEMORY_LIMIT per buffer
+MEMORY_LIMIT=`expr $TOTAL_MEMORY_LIMIT / $DIV`
+BUFFER_QUEUE_LIMIT=`expr $MEMORY_LIMIT / $BUFFER_SIZE_LIMIT`
+if [ $BUFFER_QUEUE_LIMIT -eq 0 ]; then
```
portante: What are the scenarios where `BUFFER_QUEUE_LIMIT` becomes zero?  A quick test shows that `expr 1 / 2` returns `0` in bash.  If this then is the case where the `BUFFER_SIZE_LIMIT` (which is the fluentd `buffer_chunk_limit`, I believe) is greater than it would seem we have a configuration error, and we should not default to `1024` for the queue size.  That might be outside of the supported memory configuration.

deployer/templates/fluentd.yaml
```diff
@@ -343,4 +343,4 @@ parameters:
 -
   description: 'Fluentd buffer size limit'
```
nhosoi: From my understanding, the environment variable BUFFER_SIZE_LIMIT is initially set by ansible. (Both fluentd and mux, defaults to 1m)
openshift_logging/defaults/main.yml:openshift_logging_fluentd_buffer_size_limit: 1m
openshift_logging_fluentd/templates/fluentd.j2:          value: "{{ openshift_logging_fluentd_buffer_size_limit }}"
openshift_logging_mux/defaults/main.yml:openshift_logging_mux_buffer_size_limit: 1m
openshift_logging_mux/templates/mux.j2:          value: "{{ openshift_logging_mux_buffer_size_limit }}"

Then, fluentd/run.sh checks BUFFER_SIZE_LIMIT again and if it's not set, reinitialize it to 1m and it's set to buffer_chunk_limit as well as used for calculating buffer_queue_limit.

I'm also wondering deployer/templates/fluentd.yaml is in-use or not...

richm: > I'm also wondering deployer/templates/fluentd.yaml is in-use or not...

No, it is not.  Ignore everything under `deployer/` except `scripts`, and those are only used for testing.

fluentd/run.sh
richm: Actually, yes, that's true.  When sending output to Elasticsearch, we always have 2 plugins and 2 output queues, even when not using an ops cluster and `logging-es` and `logging-es-ops` point to the same host.

nhosoi: The fluentd config files are capable to go through output-es-config.conf (apps), output-es-ops-config.conf (ops) and their copies es-copy-config.conf (apps), es-ops-copy-config.conf (ops).

So, the count of output could be 1 through 4, I'd think.  And Mux can have the same flexibility?

fluentd/run.sh
richm: Any time we send output to Elasticsearch there will always be 2 queues - one for es and one for es-ops, even when `logging-es` and `logging-es-ops` are the same host.

fluentd/run.sh
richm: in any case where `expr` is expected to return a value of `0`, you should "catch" that error in the shell like this:
```
MEMORY_LIMIT=`expr $TOTAL_MEMORY_LIMIT / $DIV` || :
```
otherwise, it will generate an error if using `set -e`

fluentd/run.sh
nhosoi: Hmmm, we have no mechanism to restrict just mux, I think.  In the same manner, it could have up to 4 outputs?

richm: I think we can assume in the MUX case there will be 2 queues, one for es, and one for es-ops.

Of course, in any case, fluentd, or mux, if the user specifies additional output destinations by customizing fluent.conf or secure-forward.conf, then we have no way to know how many output queues there are.  In that case, we will need to document for the user how to proceed, and allow the user to either provide the number of output destinations, which we will then use to calculate the buffer limits and queue sizes, or the user will need to be able to provide these and override our calculations.

nhosoi: Ok. We could introduce another switch to force to override the calculations.  shall I open a card?

fluentd/run.sh
nhosoi: Thanks for the advice, @richm .  I'm adding " || : " to the currently working PR.

fluentd/run.sh
richm: @nhosoi yes, please

richm:
Also note that fluentd will allow a record even when the record size in bytes is greater than the chunk limit: https://github.com/fluent/fluentd/blob/v0.12.37/lib/fluent/buffer.rb#L221
```
        if data.bytesize > @buffer_chunk_limit
          $log.warn "Size of the emitted data exceeds buffer_chunk_limit."
          $log.warn "This may occur problems in the output plugins ``at this server.``"
          $log.warn "To avoid problems, set a smaller number to the buffer_chunk_limit"
          $log.warn "in the forward output ``at the log forwarding server.``"
          ### TODO
          # raise BufferChunkLimitError, "received data too large"
        end

        # chunk unique id is generated in #new_chunk
        nc = new_chunk(key)
        ... operation proceeds normally ...
```
This is the source of the warning message we sometimes see in the fluentd logs: "Size of the emitted data exceeds buffer_chunk_limit".

What this means is the chunk_limit * queue size is not strictly enforced by fluentd - the actual memory usage could be greater than that, depending on how many large records there are.  In this case, if the user expects many such records, we will need to advise the user to raise the chunk_limit and lower the queue size.

nhosoi:
Or increasing the resource memory limit for fluentd pod or the file buffer size?

fluentd/run.sh
portante: @Why do we need two output queues to the same elasticsearch instance?  Why not determine the index name during the pipeline earlier and just pull that name from the metadata so we can share the same queue?

nhosoi: If sending to the same elasticsearch instance (the same ES_HOST && ES_PORT), we have just one output queue.

fluentd/run.sh
portante: The more we discuss this, the more I wonder if we should simplify this a bit:
 * always use mux
 * fluentd clients have one queue to mux
 * fluentd output queue is `file` based
 * fluentd clients have a large chunk size (16MB), queue size of 100 (default), both settable
 * fluentd memory size calculated to be enough to run ruby
 * mux has one queue per ES instance
 * mux output queue(s) are `file` based
 * mux also have large chunk size (32 MB, a bit bigger for ES payload efficiency), settable
 * mux queue size defaults to 100, settable
 * mex memory size calculated to be enough to run ruby

fluentd/run.sh
portante: @nhosoi, each output configuration file when included adds its own queue, does it not?

richm: I think we always have two output queues.  At least, we always have two separate instances of the fluent-plugin-elasticsearch - one for logging-es, and one for logging-es-ops.  Two plugins == two separate output queues, unless the plugin itself is smart enough to combine them when it sees the output is the same.

@portante yes, we could calculate the index name ahead of time and use a single fluent-plugin-elasticsearch when there is no separate ops cluster.  But I think that is outside the scope of this PR.

fluentd/run.sh
richm: > always use mux
this is the last as in it depends on everything that follows _and_ having a successful scale test

> fluentd clients have one queue to mux
check - already in

> fluentd output queue is file based
this PR

> re: large chunk size
Does it matter what the chunk size is, or is the more relevant number the chunk size * the queue size?  For example, does the fluent-plugin-elasticsearch format and send each chunk as a separate bulk index operation, or does it go through all of the chunks in the queue and submit all of them as one bulk index operation?

> mux output queue(s) are file based
this PR

fluentd/run.sh
nhosoi: Yikes.  You are right...  Then, I have to get rid of this if expression and always initialize to DIV=2 unless secure_forward.
if [ "$ES_HOST" != "$OPS_HOST" ] || [ "$ES_PORT" != "$OPS_PORT" ] ; then

Do we want to have another card for "we could calculate the index name ahead of time and use a single fluent-plugin-elasticsearch when there is no separate ops cluster."?

richm:
> Or increasing the resource memory limit for fluentd pod or the file buffer size?

Yes.

fluentd/run.sh
richm: Yes, another card, please.

fluentd/run.sh
nhosoi: This PR means this one?
https://github.com/openshift/origin-aggregated-logging/pull/525

Please note that this #482 is already merged...

nhosoi: And I'm updating #525 based upon the discussions made in this PR.

richm: yes, sorry, I meant #525

portante: > Does it matter what the chunk size is, or is the more relevant number the chunk size * the queue size?

It looks like it does.  One chuck is used to fill a bulk request to elasticsearch.  So if the chunk is not big enough, we'll perform too many Elasticsearch operations.  If the flush interval is too short, then we may not completely fill a chunk before flushing it.

PR #486, opened 2017-06-14
elasticsearch: write heap dump to HEAP_DUMP_LOCATION

wozniakjan:
Allows persistent storage to be used for heap dump

### DIFF
--- a/elasticsearch/run.sh
+++ b/elasticsearch/run.sh
@@ -152,4 +152,8 @@ verify_or_add_index_templates() {
 
 verify_or_add_index_templates &
 
+HEAP_DUMP_LOCATION="${HEAP_DUMP_LOCATION:-/elasticsearch/persistent/hdump.prof}"
+info Setting heap dump location "$HEAP_DUMP_LOCATION"
+export JAVA_OPTS="${JAVA_OPTS:-} -XX:HeapDumpPath=$HEAP_DUMP_LOCATION"
+
 exec /usr/share/elasticsearch/bin/elasticsearch --path.conf=$ES_CONF --security.manager.enabled false


elasticsearch/run.sh
```diff
@@ -216,4 +216,8 @@ verify_or_add_index_templates() {
 
 verify_or_add_index_templates &
 
+if [ -v HEAP_DUMP_LOCATION ]; then
```
richm: What does `-v` do?  Are  you trying to test if the variable HEAP_DUMP_LOCATION is set to a non-empty value?  If so, use 
```
   if [ -n "${HEAP_DUMP_LOCATION:-}" ] ; then
```

elasticsearch/run.sh
```diff
@@ -216,4 +216,8 @@ verify_or_add_index_templates() {
...
+    info Setting heap dump location "$HEAP_DUMP_LOCATION"
+    export JAVA_OPTS="${JAVA_OPTS:-} -XX:HeapDumpPath=$HEAP_DUMP_LOCATION"
```
richm: If HeapDumpPath is not set in JAVA_OPTS, will Elasticsearch dump the heap?  If so, then we need to set this _always_, and the default should be `/elasticsearch/persistent` unless that is defined in a variable somewhere.

elasticsearch/run.sh
wozniakjan: they should have equivalent behaviour in bash 4.2 and newer, but I will fix according to your suggestion

https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html

elasticsearch/run.sh
wozniakjan: if `HeapDumpPath` is unset, java will write the dump file to the root partition (which was causing some issues)

[this](https://github.com/openshift/openshift-ansible/pull/4451) PR sets the `HEAP_DUMP_LOCATION` variable in openshift-ansible to `/elasticsearch/persistent`

richm: ok.  then I suggest something like this - first, remove the `if/fi` lines, then
```
HEAP_DUMP_LOCATION=${HEAP_DUMP_LOCATION:-/elasticsearch/persistent/hdump.prof}
info Setting heap dump location "$HEAP_DUMP_LOCATION"
export JAVA_OPTS="${JAVA_OPTS:-} -XX:HeapDumpPath=$HEAP_DUMP_LOCATION" 
```
Because we need this to always be set

wozniakjan: thanks for the review comments, updated as you suggested

richm:
[test]

richm:
[merge]

wozniakjan:
regarding the failed merge results, this code is fixing the 1369914 bug

richm:
[merge][severity: bug]

jcantrill:
@wozniakjan please squash into a single commit and ref a bz if there is one like "bug XXXX. provide..."

wozniakjan:
@jcantrill thanks for review, commits are squashed and message is updated

jcantrill:
[merge][severity:blocker]

PR #490, opened 2017-06-20
bump kibana-proxy to 0.0.24

jcantrill:


jcantrill:
@richm I'm just going to merge this as the backports for 1.5 included the memory fix already but this additional commit was a subsequent issue logged by the community to resolve a change that was introduced with the memory limits change.

PR #492, opened 2017-06-20
Ensure jUnit XML is generated on failure

stevekuznetsov:
If `os::test::junit::reconcile_output` is not trapped on `EXIT`, we will
not generate valid jUnit XML on a test failure as the output file will
be in a corrupt state.

[test]

/cc @richm @jcantrill this is why we were not getting jUnit output and why it was hard to tell what failed in logging tests sometimes -- with this it will be much quicker to tell.


stevekuznetsov:
Looks like some weird flake on the pre-install checks. re[test]

stevekuznetsov:
Successful test but we needed https://github.com/openshift/aos-cd-jobs/pull/351 to get jUnit re[test]

richm:
[merge][severity: bug]

PR #494, opened 2017-06-20
fix how deployer checks ES logs

jcantrill:
This PR fixes how the deployer checks ES logs.  A change was introduced to remove console from the appenders which would cause upgrade to fail

### DIFF
--- a/deployer/scripts/upgrade.sh
+++ b/deployer/scripts/upgrade.sh
@@ -124,11 +124,15 @@ function getPodName() {
 }
 
 function checkESStarted() {
-
   local pod=$1
   local cluster_service
+  local log_cmd="oc log $pod"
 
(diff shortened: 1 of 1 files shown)


deployer/scripts/upgrade.sh
```diff
@@ -124,11 +124,15 @@ function getPodName() {
 }
 
 function checkESStarted() {
-
   local pod=$1
   local cluster_service
+  local log_cmd="oc log $pod"
+
+  if $(oc get configmap logging-elasticsearch -o yaml | grep -oEq "rootLogger:.*(file)") ; then
```
richm:
```
 if oc get configmap logging-elasticsearch -o yaml | grep -oEq "rootLogger:.*(file)" ; then
```
That is, you don't need the `$()`

PR #501, opened 2017-06-21
Allow deployer to use an intermediate CA

ruromero:
Deployer is not able to generate all the internal certificates when an Intermediate CA is provided. Keytool will fail with the following error

```
echo 'Import back to keystore (including CA chain)'
/bin/keytool -import -file /etc/deploy/scratch/ca.crt -keystore /etc/deploy/scratch/system.admin.jks -storepass kspass -noprompt -alias sig-ca
Import back to keystore (including CA chain)
Certificate was added to keystore
/bin/keytool -import -file /etc/deploy/scratch/system.admin.crt -keystore /etc/deploy/scratch/system.admin.jks -storepass kspass -noprompt -alias system.admin
keytool error: java.lang.Exception: Failed to establish chain from reply
```

### DIFF
--- a/deployer/scripts/util.sh
+++ b/deployer/scripts/util.sh
@@ -68,6 +68,7 @@ function generate_JKS_chain() {
         -keystore $dir/keystore.jks \
         -storepass $ks_pass \
         -noprompt \
+        -trustcacerts \
         -alias $NODE_NAME
 
     echo "Import CA to truststore for validating client certs"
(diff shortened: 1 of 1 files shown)


jcantrill:
@sosiouxme or @ewolinetz  Any objects to why would not want to do this?  @ruromero was able to resolve his Kibana issue for the case where custom certs were being provided

sosiouxme:
This will help if the intermediate cert provided is signed by a chain starting with a public root cert from the JDK `cacerts` store (in whichever context this is running - in 1.4, the context of the deployer pod). I don't see how enabling this can hurt, so :+1:

Probably better would be providing the entire chain back to the root in the CA file. I think it's just a matter of concatenating all the certs in the file provided, but if it's something more than that we should consider an RFE for enabling it.

PR #502, opened 2017-06-22
fix get_test_user_token; use openshift_logging_kibana_hostname

richm:
get_test_user_token always returns the admin credentials, so it always
works.  This makes it use the command line arguments. This also adds tests
to make sure a normal user cannot see a project it is not a member of, and
cannot see .operations index.
This also uses `openshift_logging_kibana_hostname` instead of
`openshift_hosted_logging_hostname` which is not supported by
playbooks/byo/openshift-cluster/openshift-logging.yml
@jcantrill @nhosoi PTAL
[test]

### DIFF
--- a/deployer/scripts/util.sh
+++ b/deployer/scripts/util.sh
@@ -375,7 +375,7 @@ function get_latest_pod() {
 # set the test_token, test_name, and test_ip for token auth
 function get_test_user_token() {
     local current_project; current_project="$( oc project -q )"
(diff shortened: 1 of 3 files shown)


ewolinetz:
LGTM

richm:
[merge][severity:blocker]

richm:
looks like a flake - could not provision aws machine [merge][severity:blocker]

PR #510, opened 2017-06-28
Don't require imagebuilder for hack/build-images.sh

stevekuznetsov:
None of the Logging images require the functionality that is provided by
`imagebuilder`, so we do not need to require it. The Origin image build
utilities should be able to fall back to `docker build` as normal.

[test]

/cc @richm sorry, this was a mistake on my end in the script on the first pass


richm:
[test]

richm:
[test]

richm:
[merge][severity:blocker]

PR #511, opened 2017-06-28
bug 1465464. Kibana memory use underscore instead of dash

jcantrill:
Modifies the memory switch to use underscores instead of dashes per:

[1] https://github.com/nodejs/node/issues/7937
[2] https://github.com/elastic/kibana/issues/9006


jcantrill:
Compatible change to #512

PR #516, opened 2017-07-03
check user/project membership; --all to get all of users projects

richm:
Make sure that the user can view logs in the given projects by
iterating the projects and checking `oc policy can-i get pods/log`
for each project.
If the `--all` flag is specified instead of the list of projects,
look up all of the projects for the given user.
@jcantrill @portante PTAL
[test]

### DIFF
--- a/hack/enable-kibana-msearch-access.sh
+++ b/hack/enable-kibana-msearch-access.sh
@@ -90,7 +90,11 @@ function curl_es_input() {
 
 usage() {
     echo Usage: $0 user-name [project-1 ... project-N]
+    echo  or
+    echo Usage: $0 user-name --all
     echo If no projects are specified, the msearch access will be removed.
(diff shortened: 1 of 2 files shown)


hack/enable-kibana-msearch-access.sh
```diff
@@ -90,7 +90,11 @@ function curl_es_input() {
 
 usage() {
     echo Usage: $0 user-name [project-1 ... project-N]
+    echo  or
```
jcantrill: Generally, LGTM, but have you thought about reusing https://github.com/openshift/origin-aggregated-logging/blob/master/elasticsearch/utils/logging for log message?

richm: @jcantrill yes, except this script has to run outside of the elasticsearch pod.

richm:
[merge]

PR #517, opened 2017-07-04
Update fluentd to latest v0.14.18

amitkumarj441:
Signed-off-by : Amit Kumar Jaiswal <amitkumarj441@gmail.com>


richm:
Unfortunately we can't yet move to fluentd 0.14.  Hopefully in the next major (3.7) release.

amitkumarj441:
@richm Sure. Thanks for letting me know about this info.

portante:
Also, according to http://www.fluentd.org/download, 0.12 is considered "stable" and ready for production use.  When 0.14 moves to that, we should consider it then.

cosmo0920:
Now, Fluentd v1 , which is just bumped up version from v0.14, became stable release and v0.12 became old stable.

richm:
maybe when rhel8/centos8 is available we can switch - but for now, we're stuck on ruby 2.0 :-(

PR #518, opened 2017-07-06
script to add dashboards and other Kibana UI objects

richm:
The script es_load_kibana_ui_objects is used to load dashboards and
other Kibana UI objects for the given user.  Usage:

oc exec $espod -- es_load_kibana_ui_objects user-name

That is, it exists inside the Elasticsearch and ES-OPS pod, and must
be run inside those pods.  Additionally, it requires some indices and
other objects set up by the OpenShift Elasticsearch plugin, so the user
must login to Kibana (or Elasticsearch) before using this script.

This will also add an index pattern for `project.*` and load the
necessary index pattern file.

This relies on some additions to the viaq schema which is provided
by https://github.com/ViaQ/elasticsearch-templates/pull/50
@jcantrill @portante PTAL
[test]


jcantrill:
> Additionally, it requires some indices and
other objects set up by the OpenShift Elasticsearch plugin
What items are we talking about here?  Anything the plugin creates, it flushes from its cache and the ACLs every 60 seconds

elasticsearch/utils/es_load_kibana_ui_objects
```diff
@@ -0,0 +1,67 @@
...
+
+if [ "$resp_code" != 200 ] ; then
+    error Could not find kibana index $kibindex for user $1: $resp_code
+    exit 1
+fi
+
+info Adding Kibana dashboards and other UI objects for user $1 index $kibindex
+
+INDEX_PATTERN=${INDEX_PATTERN:-project.*}
+INDEX_PATTERN_FILE=${INDEX_PATTERN_FILE:-/usr/share/elasticsearch/index_patterns/com.redhat.viaq-openshift.index-pattern.json}
```
jcantrill: We should consider refactoring '/usr/share/elasticsearch/' into a var like ES_HOME.  If I ever get the reconciliation PR complete I think this will simply the impact and additional modifications we need to make.

test/kibana_dashboards.sh
```diff
@@ -0,0 +1,63 @@
...
+
+LOG_ADMIN_USER=${LOG_ADMIN_USER:-admin}
+LOG_ADMIN_PW=${LOG_ADMIN_PW:-admin}
+
+if oc get users "$LOG_ADMIN_USER" > /dev/null 2>&1 ; then
+    os::cmd::debug Using existing user $LOG_ADMIN_USER
+else
+    os::cmd::info Creating cluster-admin user $LOG_ADMIN_USER
+    current_project="$( oc project -q )"
+    os::cmd::debug "$( oc login --username=$LOG_ADMIN_USER --password=$LOG_ADMIN_PW )"
```
jcantrill: I don't believe this is necessary since when adding roles to a user, a user doesn't need to exist.  I'm pretty certain the association is by name and userpolicy is applied to users with that name.  There is no 'user' store per se.

elasticsearch/utils/es_load_kibana_ui_objects
richm: ok - but there are likely other places which will need to change

richm:
[test]

richm:
> What items are we talking about here? Anything the plugin creates, it flushes from its cache and the ACLs every 60 seconds

@jcantrill It doesn't appear to require the acls, just the `defaultIndex` entry, and possibly some of the index patterns

richm:
https://github.com/ViaQ/elasticsearch-templates/releases/tag/0.0.8
[merge]

richm:
[merge][severity: blocker]

richm:
flake [merge][severity:blocker]

PR #519, opened 2017-07-06
1.5 additional backports

jcantrill:
bug 1420217. Update ES plugin that squashes stack on start
bug 1457642. Fix SG timeout
use SG index defined in ES config
bump openshift-elasticsearch-plugin to 2.4.4.9
bump kibana-proxy to 0.1.0 which uses node 4.7.2


jcantrill:
Tested manually and ES starts as expected

PR #524, opened 2017-07-11
bug 1449378. Fix SG timeout

jcantrill:
This is a backport of #469 to fix timeout of SG.  It also brings in some of the basic utils introduced later which may be of use to us


jcantrill:
@richm Updated to remove `es_load_kibana_ui_objects` since this is a feature we are adding for 3.6

PR #525, opened 2017-07-12
Fluentd,Mux: make buffer_type configurable; By default, buffer_type file

nhosoi:
Description for calculating buffer_queue_limit
  0) The buffer type is initialized in the ansible inventory file as
       openshift_logging_fluentd_buffer_type: memory|file
       openshift_logging_mux_buffer_type: memory|file
  1) buffer_type: memory
     The available memory size is given via environment varirables FLUENTD_
     MEMORY_LIMIT or MUX_MEMORY_LIMIT, which are initialized by ansible and
     set in daemonset and deploymentconfig as "resource: limits.memory",
     respectively.
     The size is multiplied by 3/4, then divided by the number of output.
     The result is the memory buffer size.  The size is divided by the chunk
     size (by default 1m) and the result is set to buffer_queue_limit.
  2) buffer_type: file
     The available volume size is retrieved from the given FILE_BUFFER_PATH.
       Fluentd uses the existing hostmount "/var/log/fluentd" for the path.
       Mux allows to choose one of the 3 types in the ansible inventory:
         openshift_logging_mux_file_buffer_storage_type: pvc|hostmount|emptydir
       then the path is set to openshift_logging_mux_file_buffer_mount_path.
       It is default to "/mux/filebufferstorage".
     The available size of the volume is multiplied by 1/4 and the result is
     used for the total size.
     Against the total size, file buffer type allows to set an upper limit,
     FILE_BUFFER_LIMIT, which is initialized in the ansible inventory as
       openshift_logging_fluentd_file_buffer_limit  (default to 1G)
       openshift_logging_mux_file_buffer_limit  (default to 2G)
     Note that FILE_BUFFER_LIMIT size is per output.
     Comparing the total size from the volume and (FILE_BUFFER_LIMIT * output),
     the smaller one is chosen for calculating buffer_queue_limit.

CI test: Using daemonset config instead of template in test-es-copy.sh.
         adding PV/PVC test case for test-mux.

### DIFF
--- a/fluentd/configs.d/openshift/es-copy-config.conf
+++ b/fluentd/configs.d/openshift/es-copy-config.conf
@@ -20,7 +20,8 @@
       flush_interval 5s
       max_retry_wait 300
       disable_retry_limit true
-      buffer_type "#{ENV['BUFFER_TYPE']}"
+      buffer_type "#{ENV['BUFFER_TYPE'] || 'memory'}"
+      buffer_path "#{ENV['FILE_BUFFER_PATH_ES_COPY'] || '/dev/null'}"
       buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
       buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT'] || '1m' }"
     </store>
--- a/fluentd/configs.d/openshift/es-copy-config.conf
+++ b/fluentd/configs.d/openshift/es-copy-config.conf
@@ -20,6 +20,7 @@
       flush_interval 5s
       max_retry_wait 300
       disable_retry_limit true
+      buffer_type "#{ENV['BUFFER_TYPE']}"
       buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
       buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT'] || '1m' }"
     </store>
--- a/fluentd/configs.d/openshift/es-ops-copy-config.conf
+++ b/fluentd/configs.d/openshift/es-ops-copy-config.conf
@@ -20,6 +20,7 @@
       flush_interval 5s
       max_retry_wait 300
       disable_retry_limit true
+      buffer_type "#{ENV['BUFFER_TYPE']}"
       buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
       buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT'] || '1m' }"
     </store>
--- a/fluentd/configs.d/openshift/es-ops-copy-config.conf
+++ b/fluentd/configs.d/openshift/es-ops-copy-config.conf
@@ -20,7 +20,8 @@
       flush_interval 5s
       max_retry_wait 300
       disable_retry_limit true
-      buffer_type "#{ENV['BUFFER_TYPE']}"
+      buffer_type "#{ENV['BUFFER_TYPE'] || 'memory'}"
+      buffer_path "#{ENV['FILE_BUFFER_PATH_ESOPS_COPY'] || '/dev/null'}"
       buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
       buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT'] || '1m' }"
     </store>
--- a/fluentd/configs.d/openshift/output-es-config.conf
+++ b/fluentd/configs.d/openshift/output-es-config.conf
@@ -20,6 +20,7 @@
       flush_interval 5s
       max_retry_wait 300
       disable_retry_limit true
+      buffer_type "#{ENV['BUFFER_TYPE']}"
       buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
       buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT'] || '1m' }"
     </store>
--- a/fluentd/configs.d/openshift/output-es-config.conf
+++ b/fluentd/configs.d/openshift/output-es-config.conf
@@ -20,7 +20,8 @@
       flush_interval 5s
       max_retry_wait 300
       disable_retry_limit true
-      buffer_type "#{ENV['BUFFER_TYPE']}"
+      buffer_type "#{ENV['BUFFER_TYPE'] || 'memory'}"
+      buffer_path "#{ENV['FILE_BUFFER_PATH_ES']|| '/dev/null'}"
       buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
       buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT'] || '1m' }"
     </store>
--- a/fluentd/configs.d/openshift/output-es-ops-config.conf
+++ b/fluentd/configs.d/openshift/output-es-ops-config.conf
@@ -20,6 +20,7 @@
       flush_interval 5s
       max_retry_wait 300
       disable_retry_limit true
+      buffer_type "#{ENV['BUFFER_TYPE']}"
       buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
       buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT'] || '1m' }"
     </store>
--- a/fluentd/configs.d/openshift/output-es-ops-config.conf
+++ b/fluentd/configs.d/openshift/output-es-ops-config.conf
@@ -20,7 +20,8 @@
       flush_interval 5s
       max_retry_wait 300
       disable_retry_limit true
-      buffer_type "#{ENV['BUFFER_TYPE']}"
+      buffer_type "#{ENV['BUFFER_TYPE'] || 'memory'}"
+      buffer_path "#{ENV['FILE_BUFFER_PATH_ESOPS'] || '/dev/null'}"
       buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
       buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT'] || '1m' }"
     </store>
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -106,16 +106,19 @@ else
 fi
 
 # How many outputs?
-DIV=1
-if [ "$ES_HOST" != "$OPS_HOST" ] || [ "$ES_PORT" != "$OPS_PORT" ] ; then
-    # using ops cluster
-    DIV=`expr $DIV \* 2`
-fi
 if [ "${USE_MUX_CLIENT:-}" = "true" ] ; then
-    DIV=`expr $DIV \* 2`
+    # mux client has just one output.
+    DIV=1
+else
+    # fluend usually has 2 putputs.
+    DIV=2
+    if [ "$ES_COPY" = "true" ]; then
+        DIV=`expr $DIV \* 2`
+    fi
 fi
 
 if [ "$BUFFER_TYPE" = "file" -a -z "$FILE_BUFFER_PATH" ]; then
+    BUFFER_TYPE=memory
     echo "ERROR: File is set to buffer_type but FILE_BUFFER_PATH is empty.  Using memory buffer"
 fi
 
@@ -124,7 +127,6 @@ FILE_BUFFER_PATH_ESOPS=""
 FILE_BUFFER_PATH_ES_COPY=""
 FILE_BUFFER_PATH_ESOPS_COPY=""
 # Default buffer_type: file if FILE_BUFFER_PATH is specified.
-BUFFER_TYPE=${FILE_BUFFER_PATH:+file}
 if [ "$BUFFER_TYPE" = "file" ]; then
     if [ ! -d $FILE_BUFFER_PATH ]; then
         mkdir -p $FILE_BUFFER_PATH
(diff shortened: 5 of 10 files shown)


jcantrill:
[(?) Use `file` buffer instead of `memory` buffer for fluentd[logging][epic-ois-agl-perf]](https://trello.com/c/XpreI533/509-use-file-buffer-instead-of-memory-buffer-for-fluentdloggingepic-ois-agl-perf)

fluentd/configs.d/openshift/es-copy-config.conf
```diff
@@ -20,6 +20,7 @@
       flush_interval 5s
       max_retry_wait 300
       disable_retry_limit true
+      buffer_type "#{ENV['BUFFER_TYPE']}"
```
richm: suggest using a default of `memory` here just in case:
```
     buffer_type "#{ENV['BUFFER_TYPE'] || 'memory'}"
```

fluentd/configs.d/openshift/es-ops-copy-config.conf
```diff
@@ -20,6 +20,7 @@
       flush_interval 5s
       max_retry_wait 300
       disable_retry_limit true
+      buffer_type "#{ENV['BUFFER_TYPE']}"
```
richm: suggest using a default of `memory` here just in case:
```
     buffer_type "#{ENV['BUFFER_TYPE'] || 'memory'}"
```

fluentd/configs.d/openshift/output-es-config.conf
```diff
@@ -20,6 +20,7 @@
       flush_interval 5s
       max_retry_wait 300
       disable_retry_limit true
+      buffer_type "#{ENV['BUFFER_TYPE']}"
```
richm: suggest using a default of `memory` here just in case:
```
     buffer_type "#{ENV['BUFFER_TYPE'] || 'memory'}"
```

fluentd/configs.d/openshift/output-es-ops-config.conf
```diff
@@ -20,6 +20,7 @@
       flush_interval 5s
       max_retry_wait 300
       disable_retry_limit true
+      buffer_type "#{ENV['BUFFER_TYPE']}"
```
richm: suggest using a default of `memory` here just in case:
```
     buffer_type "#{ENV['BUFFER_TYPE'] || 'memory'}"
```

fluentd/run.sh
```diff
@@ -123,13 +112,58 @@ if [ "${USE_MUX_CLIENT:-}" = "true" ] ; then
...
 
+# Default buffer_type: file if FILE_BUFFER_PATH is specified.
+FILE_BUFFER_PATH=${FILE_BUFFER_PATH:-""}
+if [ "$BUFFER_TYPE" = "file" -a "$FILE_BUFFER_PATH" != "" ] ; then
+    if [ ! -d $FILE_BUFFER_PATH ]; then
+        mkdir -p $FILE_BUFFER_PATH
+    fi
+    # Get the available disk size; use 1/4 of it
+    DF_LIMIT=$(df -B1 $FILE_BUFFER_PATH | grep -v Filesystem | awk '{print $2}')
+    DF_LIMIT=${DF_LIMIT:-0}
```
richm: I think we should print a very loud error message here - 0 space means there will be serious problems

fluentd/run.sh
```diff
@@ -123,13 +112,58 @@ if [ "${USE_MUX_CLIENT:-}" = "true" ] ; then
...
+    DF_LIMIT=$(expr $DF_LIMIT / 4)
+    # Given buffer limit per output
+    TOTAL_LIMIT=$(echo ${FILE_BUFFER_LIMIT:-0} | sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc)
+    # In total
+    TOTAL_LIMIT=$(expr $TOTAL_LIMIT \* $DIV)
+    # Available disk space is less than the specified size.
+    if [ $DF_LIMIT -gt 0 -a $DF_LIMIT -lt $TOTAL_LIMIT ]; then
+        TOTAL_LIMIT=$DF_LIMIT
```
richm: Should print a warning here that available disk space $DF_LIMIT is less than requested $TOTAL_LIMIT

fluentd/run.sh
```diff
@@ -123,13 +112,58 @@ if [ "${USE_MUX_CLIENT:-}" = "true" ] ; then
...
+    fi
+    # Adding the buffer_path to each output config
+    for esconf in $CFG_DIR/openshift/es-*.conf $CFG_DIR/openshift/output-es-*.conf; do
+        estmp=$(basename $esconf)
+        esfile=${estmp%.*}
+        cat $esconf | sed "/buffer_path .*/d" > $esconf.tmp
+        cat $esconf.tmp | sed "/buffer_type/ a\
+\ \ \ \ \ \ buffer_path $FILE_BUFFER_PATH/$esfile" > $esconf
```
richm: Can we just add buffer_path to every conf file? e.g. something like
```
    buffer_path "#{ENV['FILE_BUFFER_PATH'] || '/dev/null'}"
```
or will that cause problems if using `buffer_type memory`?

fluentd/run.sh
```diff
@@ -123,13 +112,58 @@ if [ "${USE_MUX_CLIENT:-}" = "true" ] ; then
     DIV=`expr $DIV \* 2`
 fi
 
+# Default buffer_type: file if FILE_BUFFER_PATH is specified.
```
richm: Then should you have some code like 
```
BUFFER_TYPE=${FILE_BUFFER_PATH:+file}
```
which means "set BUFFER_TYPE to file if FILE_BUFFER_PATH is set, and if FILE_BUFFER_PATH is not set, BUFFER_TYPE will be empty"

fluentd/run.sh
```diff
@@ -123,13 +112,58 @@ if [ "${USE_MUX_CLIENT:-}" = "true" ] ; then
...
+FILE_BUFFER_PATH=${FILE_BUFFER_PATH:-""}
+if [ "$BUFFER_TYPE" = "file" -a "$FILE_BUFFER_PATH" != "" ] ; then
```
richm: where is BUFFER_TYPE set?

richm: also, should it be an error if BUFFER_TYPE = file but no FILE_BUFFER_PATH was given?

fluentd/run.sh
nhosoi: I was thinking it would be a problem for buffer_type memory.  But since this is just a warning, there may be no harm.  I'm testing with your suggested buffer_path (+ each config name).
   buffer_path "#{ENV['FILE_BUFFER_PATH']/output_es_config || '/dev/null'}"
=========================================================
2017-07-13 17:10:20 -0400 [warn]: parameter 'buffer_path' in <store>
  @type elasticsearch_dynamic
  host logging-es-ops
  port 9200
  scheme https
   [...]
  buffer_type memory
  buffer_path /dev/null
  buffer_queue_limit 48
  buffer_chunk_limit 16777216
</store> is not used.

fluentd/run.sh
nhosoi: BUFFER_TYPE is set in the env section of daemonset and deploymentconfig, respectively...

defaults/main.yml:openshift_logging_fluentd_buffer_type: "file"
templates/fluentd.j2:        - name: "BUFFER_TYPE"
templates/fluentd.j2:          value: "{{ openshift_logging_fluentd_buffer_type | lower }}"

defaults/main.yml:openshift_logging_mux_buffer_type: "file"
templates/mux.j2:        - name: "BUFFER_TYPE"
templates/mux.j2:          value: "{{ openshift_logging_mux_buffer_type | lower }}"

> also, should it be an error if BUFFER_TYPE = file but no FILE_BUFFER_PATH was given?
Right. Currently, the buffer type is silently reset to memory if no FILE_BUFFER_PATH is given.
I'm adding this error.  (Or warning is better?)
echo "ERROR: File is set to buffer_type but FILE_BUFFER_PATH is empty.  Using memory buffer"

fluentd/run.sh
nhosoi: Is it acceptable to prepare these hardcoded environmentvariables in run.sh?
+        FILE_BUFFER_PATH_ES=${FILE_BUFFER_PATH}/es
+        FILE_BUFFER_PATH_ESOPS=${FILE_BUFFER_PATH}/esops
+        FILE_BUFFER_PATH_ES_COPY=${FILE_BUFFER_PATH}/escopy
+        FILE_BUFFER_PATH_ESOPS_COPY=${FILE_BUFFER_PATH}/esopscopy
Each one is used in the corresponding output conf file.
For instance, configs.d/openshift/output-es-ops-config.conf
+      buffer_path "#{ENV['FILE_BUFFER_PATH_ESOPS'] || '/dev/null'}"

nhosoi: And confirmed the "[warn]: parameter 'buffer_path' in .... is not used." is just a warning and there's no issue for the functionality.

fluentd/configs.d/openshift/es-copy-config.conf
nhosoi: Done.

fluentd/configs.d/openshift/es-ops-copy-config.conf
nhosoi: Done.

fluentd/configs.d/openshift/output-es-config.conf
nhosoi: Done.

fluentd/configs.d/openshift/output-es-ops-config.conf
nhosoi: Done.

fluentd/run.sh
nhosoi: Done.

fluentd/run.sh
nhosoi: Done.

fluentd/run.sh
nhosoi: Done.

nhosoi:
Thanks for your reviews, @richm !!  Could you please take another look?  Thanks!

richm:
[test]

richm:
We need to have a destructive test, to see if the file buffering actually works.  That is, we need to kill fluentd/mux while it has records buffered but not sent, to see if restarting fluentd/mux will read and send the unsent records from the file buffer.  That is the main point of using a file buffer.

nhosoi:
@richm To add the destructive test, I'd like to learn how to stop then restart it.  Could you please tell me the way how to?  This is what I'm hoping to add to the CI test.
1.  stop ES
2.  generate logs for fluentd/mux to process
3.  stop mux
4.  restart ES
5.  restart mux
6.  check the logs are stored in the ES.

richm:
@nhosoi The simplest way to stop and restart is to just `oc delete pod $pod` where `$pod` is the name of a fluentd or mux pod.  That will stop the pod ASAP, then OpenShift will restart the pod.

Why do you need to stop and start ES?

nhosoi:
@richm The file buffer disappears immediately once the contents are sent to the ES...  So, I think I need the control to have such a window -- the ES is down and the mux is up -- to verify the PV is available independent from the mux pod.

nhosoi:
@richm Or just forcing the mux to fail to send the logs to ES could achieve the goal?  For instance, if I set something bogus to ES_HOST and/or ES_PORT with "oc set env", then the mux would piling the logs in the file buffer (with errors in the fluentd log), I'd think.
      host "#{ENV['ES_HOST']}"
      port "#{ENV['ES_PORT']}"
Then, if restarting the mux with the correct ES_HOST/ES_PORT resumes the logging flow without no missing logs, we could say the PV is functioning?

richm:
@nhosoi yes, sounds good

fluentd/run.sh
```diff
@@ -59,10 +59,8 @@ IPADDR6=`/usr/sbin/ip -6 addr show dev eth0 | grep inet6 | sed "s/[ \t]*inet6 \(
 export IPADDR4 IPADDR6
 
 BUFFER_SIZE_LIMIT=${BUFFER_SIZE_LIMIT:-1048576}
-MUX_CPU_LIMIT=${MUX_CPU_LIMIT:-500m}
-MUX_MEMORY_LIMIT=${MUX_MEMORY_LIMIT:-2Gi}
-FLUENTD_CPU_LIMIT=${FLUENTD_CPU_LIMIT:-100m}
 FLUENTD_MEMORY_LIMIT=${FLUENTD_MEMORY_LIMIT:-512Mi}
+MUX_MEMORY_LIMIT=${MUX_MEMORY_LIMIT:-2Gi}
```
portante: Don't we need different memory limits between `file` and `memory` buffer queue options?  Won't this be yet another combination of behaviors we'll have to test?  If we use `file` only, won't we be able to determine what the operational memory is for fluentd for all on-disk sizes and just one one value?

nhosoi: If we remove in-memory option entirely, if the disk is not available, should run.sh get failed and fluentd/mux fail to start?  Currently, memory buffer is used for fallback, as well...

nhosoi:
Additional patch has been pushed for:
-   Fixing the base output queue number (DIV) -- mux client has one; fluentd/mux has 2, by default.
-   Adding a Permanent Volume test to test-mux.sh.

@richm, @portante, if you could include the patch in your review, I'd appreciate it.

richm:
[test]

fluentd/run.sh
richm: I really want to have the option of using an in-memory buffer, even if we do not document it, and/or support it for customer use, if only for testing/PoC.  File buffer will be the default going forward, and will be the one that we CI test and fully support.

portante: In the scenario where we only have `file` buffer, we'd have a default location for the on-disk buffer that would be inside the pod.  If the user provides a PV, or requests the container use a PVC to be auto-created, then we'd use that.

richm: The on-disk buffer inside the pod is useless for the real use case, because you want the data to be persistent despite losing the pod.  I suppose it would be useful for performance testing.

In any case, I insist that we do not drop support for memory buffers.

richm:
[test]

richm:
[test]

richm:
[test]

fluentd/configs.d/openshift/es-copy-config.conf
```diff
@@ -20,6 +20,8 @@
       flush_interval 5s
       max_retry_wait 300
       disable_retry_limit true
+      buffer_type "#{ENV['BUFFER_TYPE'] || 'memory'}"
+      buffer_path "#{ENV['FILE_BUFFER_PATH_ES_COPY'] || '/dev/null'}"
```
portante: If a user sets `BUFFER_TYPE` and then fails to set `FILE_BUFFER_PATH_ES_COPY`, and `FILE_BUFFER_PATH_ESOPS_COPY`, and all the others, won't they still default to `/dev/null`?  The `BUFFER_TYPE` parameter affects all uses of a fluentd `buffer`, so we might want to handle that case.

fluentd/run.sh
```diff
@@ -107,32 +105,89 @@ else
...
+        # 75 % of 2G
+        TOTAL_LIMIT=1610612736
+    else
+        # 75 % of 512M
+        TOTAL_LIMIT=402653184
+    fi
+fi
+
+BUFFER_SIZE_LIMIT=$(echo $BUFFER_SIZE_LIMIT |  sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc)
+BUFFER_SIZE_LIMIT=${BUFFER_SIZE_LIMIT:-1048576}
```
portante: This calculation now needs to account for how many tags we have in flight when tags land in the elasticsearch output plugins.  I think we need to consider a change to tags first so that this calculation will be correct.

fluentd/configs.d/openshift/es-copy-config.conf
nhosoi: In run.sh, if BUFFER_TYPE is set to file and FILE_BUFFER_PATH is not given, then I was thinking to reset the type to memory.  I noticed that @richm modified the error message as follows.  But I cannot find the "exit" code. 
```
 if [ "$BUFFER_TYPE" = "file" -a -z "$FILE_BUFFER_PATH" ]; then
-    BUFFER_TYPE=memory
-    echo "ERROR: File is set to buffer_type but FILE_BUFFER_PATH is empty.  Using memory buffer"
+    echo "ERROR: BUFFER_TYPE is set to file but FILE_BUFFER_PATH is empty.  Please specify a FILE_BUFFER_PATH.  Exiting."
 fi
```
Shouldn't we call "exit 1" after the "Exiting." error message?

richm: yes

jcantrill:
Closing in deference to #544

PR #526, opened 2017-07-12
bug 1470368. Use DC_NAME for the searchguard index

jcantrill:
Use DCNAME for SG Index instead of hostname


jcantrill:
Tested:
* Fresh Install
* Edit configmap to replace DC_NAME with HOSTNAME
** Edit dc to add MIN_MASTERS=1
** Verified upgrade ads DC_NAME and ES server starts

PR #528, opened 2017-07-13
port curator to use hack/lib scripts

richm:
Move curator to test/curator.sh in order to run with the new
entrypoint.sh.  Make the old hack/testing script a wrapper
around the new script.
@jcantrill @stevekuznetsov PTAL
[test]


test/curator.sh
```diff
@@ -0,0 +1,379 @@
...
+        os::log::debug "$( oc rollout latest dc/logging-curator${ops:-} )"
+    fi
+    # wait until redeployed
+    os::log::debug "$( oc rollout status -w dc/logging-curator${ops:-} )"
+    # get new pod
+    curpod=`get_running_pod curator${ops:-}`
+}
+
+uses_config_maps() {
+    oc get dc logging-curator -o yaml | grep -q -i configmap:
```
stevekuznetsov: Should be a `jsonpath`

test/curator.sh
```diff
@@ -0,0 +1,379 @@
...
+        os::log::debug "$( oc delete secret curator-config${ops:-} )" || os::log::debug "no such secret curator-config${ops:-} - ignore"
+        os::log::debug "$( oc secrets new curator-config${ops:-} settings=$1 )"
+        os::log::debug "$( oc volumes dc/logging-curator${ops:-} --add --type=secret --secret-name=curator-config${ops:-} --mount-path=/etc/curator --name=curator-config --overwrite )"
+    fi
+    restart_curator ${2:-}
+}
+
+if uses_config_maps ; then
+    origconfig=`mktemp`
+    os::log::debug "$( oc get configmap logging-curator -o yaml > $origconfig )" || :
```
stevekuznetsov: Why is this in a `os::log::debug`>

test/curator.sh
```diff
@@ -0,0 +1,379 @@
...
+    if [ -n "${esopspod:-}" ] ; then
+        delete_indices $esopspod
+    fi
+    if [ -n "${origconfig:-}" -a -f $origconfig ] ; then
+        os::log::debug "$( oc delete configmap logging-curator )"
+        sleep 1
+        os::log::debug "$( oc create -f $origconfig )"
+        sleep 1
+    else
+        os::log::debug "$( oc delete secret curator-config )"
```
stevekuznetsov: Why are these in debugs?

test/curator.sh
```diff
@@ -0,0 +1,379 @@
...
+this-project-name-is-far-far-too-long-this-project-name-is-far-far-too-long-this-project-name-is-far-far-too-long-this-project-name-is-far-far-too-long:
+  delete:
+    days: 1
+EOF
+    update_config_and_restart $curtest errors
+    sleep 1
+    rc=0
+    # curator pod should be in error state
+    errpod=`get_error_pod curator`
+    if ! os::cmd::expect_success "test -n '$errpod'" ; then
```
stevekuznetsov: These are also cases where the `if` doesn't make sense -- let `os::cmd ` handle the failure if you want the failure to fail the job

richm:
@stevekuznetsov re: use of os::log::debug - those commands just produce too much noise - that output is really only useful if you are trying to debug some strange failure - ideally, that output would captured in a separate log used for debugging 

What I would really like is that the console output of the test only displays the absolute minimum, basic information to let me know what's going on in the test, and pass/fail.  If I then need additional information, I should be able to drill down, rather than having to wade through 500MB of output.

jcantrill:
[merge]

richm:
[test]

richm:
[merge][severity:blocker]

jcantrill:
[merge]

stevekuznetsov:
Nice!

PR #529, opened 2017-07-14
bug 1468987: kibana OOM

wozniakjan:
The javascript engine V8 used by nodejs has heap split to 4 different spaces. Setting `max_old_space_size` to half of what the container has available so other heap spaces have some available memory. This should prevent the container from getting OOM killed.

The issue occured originally with kibana-proxy but since both use nodejs, it is fixed here as well as a preventative measure.


richm:
[test]

richm:
possible [test] flake

jcantrill:
[merge][severity:blocker]

PR #532, opened 2017-07-17
do not dump env vars unless VERBOSE is set

richm:
do not dump env vars unless VERBOSE is set
this makes the fluentd logs a bit too noisy
You can always get the list of env vars in the fluentd process like this:
```
```
@jcantrill @nhosoi PTAL
[test]


nhosoi:
Looks good.

richm:
[merge][severity:blocker]

PR #534, opened 2017-07-19
Bug 1467963 There are no Kibana dashboards for container admins

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1467963
Additional fixes:
* Add kibana_ui_objects directory to downstream Dockerfile
* Other fixes to sync upstream and downstream Dockerfiles
* use ES_HOME instead of hardcoded path
* check for missing files and directories
@jcantrill PTAL
[test]
need to merge this asap

### DIFF
--- a/elasticsearch/utils/es_load_kibana_ui_objects
+++ b/elasticsearch/utils/es_load_kibana_ui_objects
@@ -37,9 +37,9 @@ fi
 info Adding Kibana dashboards and other UI objects for user $1 index $kibindex
 
 INDEX_PATTERN=${INDEX_PATTERN:-project.*}
-INDEX_PATTERN_FILE=${INDEX_PATTERN_FILE:-/usr/share/java/elasticsearch/index_patterns/com.redhat.viaq-openshift.index-pattern.json}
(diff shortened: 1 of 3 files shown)


richm:
@nhosoi @portante @ewolinetz PTAL - need to merge this downstream ASAP

elasticsearch/utils/es_load_kibana_ui_objects
```diff
@@ -50,6 +50,11 @@ cat $INDEX_PATTERN_FILE | \
 info Adding the Kibana UI objects . . .
 {
     for file in $KIBANA_UI_OBJECTS_DIR/*.json ; do
+        if [ ! -f "$file" ] ; then
```
portante: So this is the case where the glob does not match anything so `file` gains the value of `$KIBANA_UI_OBJECTS_DIR/*.json`, right?

richm: @portante right

richm:
[merge][severity:blocker]

richm:
@portante For `docker build`, and perhaps for other container build systems, it is not allowed to override the ENV variables on the command line e.g. to build with a different version of an elasticsearch plugin without having to change the Dockerfile or install.sh.  Instead, you must use ARG instead of ENV to have parameters that you can override at build time.

PR #539, opened 2017-07-22
update downstream versions of fluentd packages

richm:
update downstream versions of fluentd packages
[test]


richm:
[merge][severity:blocker]

PR #546, opened 2017-07-25
add MUX_CLIENT_MODE; full_no_k8s_meta to do processing on fluentd

richm:
Adds new fluentd env var MUX_CLIENT_MODE - default value is unset
If `minimal`, this is the same behavior as `USE_MUX_CLIENT true`.  If
`full_no_k8s_meta`, then fluentd will perform *all* of the data processing
and filtering *except* the Kubernetes metadata annotation.  The records will
be sent to mux for Kubernetes metadata annotation before sending the logs to
Elasticsearch.  In this case, it is assumed you want to deploy mux to be as
lightweight as possible, and move the processing burden to the individual
nodes.

For container logs, we preserve the CONTAINER_NAME and CONTAINER_ID_FULL
fields when the records are sent from fluentd to mux so that mux can
grab the k8s metadata.

For operations logs, we assume no further processing is required, so we
retag the records in order to bypass further processing and go to the
operations output.

(cherry picked from commit 9d55cdcdda74a23555cb060a2c6e223a83023dbc)
@jcantrill PTAL


richm:
[merge][severity:blocker]

PR #551, opened 2017-07-26
bug 1469711: kibana OOM

jcantrill:
The javascript engine V8 used by nodejs has heap split to 4 different
spaces. Setting `max_old_space_size` to half of what the container
has available so other heap spaces have some available memory. This
should prevent the container from getting OOM killed.

The issue occured originally with kibana-proxy but since both use
nodejs, it is fixed here as well as a preventative measure.

(cherry picked from commit 4cb131929ba887aaea840c5357ad06e2fb750929)


jcantrill:
1.5 backport  of #529

PR #553, opened 2017-07-26
mux service always check for k8s metadata

richm:
Previously, mux would only look to see if the incoming record needed
k8s metadata if operating in MUX_ALLOW_EXTERNAL=true mode, for
accepting records from outside of the cluster.  Now that we have the
ability to run fluentd with MUX_CLIENT_MODE=full_no_k8s_meta, we can
get records coming from inside the cluster without k8s metadata.
This commit changes the flow so that mux will always check incoming
records for k8s metadata.
This has the side effect of getting rid of the need for MUX_ALLOW_EXTERNAL
processing in mux since it will just always assume this case now, and
this gets rid of the file mux-post-input-filter-tag.conf since that
functionality has moved into input-post-forward-mux.conf
@jcantrill @portante PTAL
[test]


nhosoi:
May I ask this change is for the performance improvement effort?

And could remove the check for the eliminated environment variable MUX_ALLOW_EXTERNAL?
```
diff --git a/hack/testing/logging.sh b/hack/testing/logging.sh
index c283e88..91c13a0 100755
--- a/hack/testing/logging.sh
+++ b/hack/testing/logging.sh
@@ -38,10 +38,9 @@ USE_LOCAL_SOURCE=${USE_LOCAL_SOURCE:-false}
 TEST_PERF=${TEST_PERF:-false}
 ES_VOLUME=${ES_VOLUME:-/var/lib/es}
 ES_OPS_VOLUME=${ES_OPS_VOLUME:-/var/lib/es-ops}
-export MUX_ALLOW_EXTERNAL=${MUX_ALLOW_EXTERNAL:-false}
 export USE_MUX_CLIENT=${USE_MUX_CLIENT:-false}
 export USE_MUX=${USE_MUX:-false}
-if [ "$MUX_ALLOW_EXTERNAL" = true -o "$USE_MUX_CLIENT" = true ] ; then
+if [ "$USE_MUX_CLIENT" = true ] ; then
     export USE_MUX=true
 fi
 
diff --git a/hack/testing/test-mux.sh b/hack/testing/test-mux.sh
index 5add552..e603cad 100755
--- a/hack/testing/test-mux.sh
+++ b/hack/testing/test-mux.sh
@@ -16,8 +16,8 @@ if ! type get_running_pod > /dev/null 2>&1 ; then
     . ${OS_O_A_L_DIR:-../..}/deployer/scripts/util.sh
 fi
 
-if [ "$USE_MUX_CLIENT" == "false" -o "$MUX_ALLOW_EXTERNAL" == "false" ]; then
-    echo "Skipping -- This test requires both USE_MUX_CLIENT and MUX_ALLOW_EXTERNAL are true."
+if [ "$USE_MUX_CLIENT" == "false" ]; then
+    echo "Skipping -- This test requires USE_MUX_CLIENT is true."
     exit 0
 fi
```

richm:
> May I ask this change is for the performance improvement effort?

Yes, but indirectly.  That is - we want the fluentd on each node to do as much processing as possible, and have mux only do the k8s metadata processing.

richm:
[test]

deployer/templates/fluentd.yaml
```diff
@@ -329,13 +323,9 @@ parameters:
...
   value: "false"
 -
-  description: 'Configure MUX CLIENT (false|true).'
-  name: USE_MUX_CLIENT
-  value: "false"
--
-  description: 'Configure MUX SERVER (false|true).'
-  name: MUX_ALLOW_EXTERNAL
-  value: "false"
+  description: 'Configure MUX CLIENT MODE (minimal|full_no_k8s_meta).'
```
portante: Can we use a description like, "Configure fluentd collector MUX CLIENT MODE ..."

portante: Can we consider renaming `full_no_k8s_meta` to be something like, `local_xform`, `local_enrich`, `local`, `normalize`, `preprocess`, or something else that does not name an internal-to-the-code action?

docs/mux-logging-service.md
```diff
@@ -41,30 +41,27 @@ There are environment variables for the logging-mux DC config:
...
-  `secure_forward` to send logs to mux.
-* `MUX_CLIENT_MODE` - `minimal`/`full_no_k8s_meta` - default is unset - if
-  `minimal`, this is the same behavior as `USE_MUX_CLIENT true`.  If
-  `full_no_k8s_meta`, then fluentd will perform *all* of the data processing
+* `MUX_CLIENT_MODE` - `minimal`/`full_no_k8s_meta` - default is unset
+  If `full_no_k8s_meta`, then fluentd will perform *all* of the data processing
   and filtering *except* the Kubernetes metadata annotation.  The records will
   be sent to mux for Kubernetes metadata annotation before sending the logs to
   Elasticsearch.  In this case, it is assumed you want to deploy mux to be as
   lightweight as possible, and move the processing burden to the individual
```
portante: Perhaps word the last part, "and move as much of the processing burden as possible to the individual fluentd collector pods."

docs/mux-logging-service.md
```diff
@@ -41,30 +41,27 @@ There are environment variables for the logging-mux DC config:
...
 
 These are environment variables for the logging-fluentd daemonset config:
 
-* `USE_MUX_CLIENT` - `true`/`false` - default is `false` - if `true`, fluentd
-  will not send logs directly to Elasticsearch, instead it will use
-  `secure_forward` to send logs to mux.
-* `MUX_CLIENT_MODE` - `minimal`/`full_no_k8s_meta` - default is unset - if
-  `minimal`, this is the same behavior as `USE_MUX_CLIENT true`.  If
-  `full_no_k8s_meta`, then fluentd will perform *all* of the data processing
+* `MUX_CLIENT_MODE` - `minimal`/`full_no_k8s_meta` - default is unset
```
portante: Is there any indication that the `minimal` mode could be supported by us?  What would be the scenarios we would want to offer this option and support it if so?

docs/mux-logging-service.md
```diff
@@ -41,30 +41,27 @@ There are environment variables for the logging-mux DC config:
...
+  If `full_no_k8s_meta`, then fluentd will perform *all* of the data processing
```
portante: Perhaps, "then the fluentd collectors will perform as much of the log transformations and enrichment as possible that scale in the collectors.  Currently, the transformation to enhance logs with Kubernetes metadata annotations does not scale when handled by fluentd collectors as the number of collectors grows, while using a smaller number of mux pods to handle those transformations does scale."

docs/mux-logging-service.md
```diff
@@ -275,13 +278,14 @@ The next thing is this match:
 This uses the `relabel` plugin to redirect the standard OpenShift logging tags
 to their usual destination via the `@INGRESS` label.
 
-This is all of the processing done if `USE_MUX true` and
-`MUX_ALLOW_EXTERNAL false`.
+The next thing is this match:
```
portante: instead of `thing` perhaps `transformation` or `filter`?

docs/mux-logging-service.md
```diff
@@ -275,13 +278,14 @@ The next thing is this match:
...
+
+    <match journal.container** journal.system>
 
-If `MUX_ALLOW_EXTERNAL true`, there is some additional processing for tags
-matching `project.**`, and for tags which do not match any of the above.
+Records with these tags are operations logs and have already been processed and
+formatted by an openshift fluentd, so we can skip any additional processing.
 
-The first thing is this filter:
+The next thing is this filter:
```
portante: also replace `thing` with something else (pun intended)?

deployer/templates/fluentd.yaml
richm: @portante Give me a short, no more than 2 or 3 or 4 very short words, that I can use, that will explain that this is the mode to use for fluentd if you want to do perform all of the processing except the kubernetes metadata enrichment, because 'enrich', 'normalize', etc. do not mean that.

docs/mux-logging-service.md
richm: @smarterclayton Do you want to be able to run fluentd in a mode where it ships the logs off of the node as fast as possible, doing minimal processing?  Is there ever a case where you would want to run fluentd in such a mode?  Note that we have not even begun to try to optimize the actual processing pipeline code in fluentd/mux, so we don't really know how well it will actually scale.  @portante is correct that currently, fluentd uses way, way too much cpu because the processing pipeline is incredibly poorly designed, so the only way we can scale the logging pipeline today is to push all of that processing onto each node.

deployer/templates/fluentd.yaml
portante: If we replace the fluentd collector in the future with something else, would this option still be applicable?  Do you believe this name is general enough that it would apply in other situations?  If you do, then let's keep it.  If you don't, would not a more general name be appropriate for use where we can keep the name but define the behavior appropriately via documentation, so that user's don't have to change their inventory files in ansible, or scripts?

richm: > If we replace the fluentd collector in the future with something else, would this option still be applicable?

yes, because `full_no_k8s_meta` is not specific to fluentd.

> Do you believe this name is general enough that it would apply in other situations? 

No.  It applies only in the specific situation where you want to offload k8s processing.

So how about `maximal` instead, which implies doing the maximum possible amount of processing?

docs/mux-logging-service.md
richm: I've added a note in the docs about `full_no_k8s_meta` and `minimal` usage.

richm:
> do external logs require viaq formatting?

They should already be in the viaq format with the _default_ namespace fields and any context specific namespaced metadata fields.

The only reason we still have viaq in the pipeline on the mux side for those records is to ensure that they have no empty fields and have the `@timestamp` field.

I have tried in the latest version of the docs to answer your other questions.

deployer/templates/fluentd.yaml
portante: Sure, that might work.

richm: > Sure, that might work

Shall I go ahead then and rework both the ansible PR and this PR to use `maximal` instead of `full_no_k8s_meta`?

richm:
Also, with respect to `deployer/templates/fluentd.yaml` - any changes here are temporary until we can get rid of that file - we unfortunately are still using it some of our tests.

deployer/templates/fluentd.yaml
portante: Yes, thanks.

richm:
Another comment - mux only runs in 1 mode - it will accept records sent from any sort of client - so you can have fluentd maximal clients, fluentd minimal clients, and fluentd external clients, all sending to the same mux, with no further mux configuration required.

richm:
Any more changes?  Is this ready to merge?

docs/mux-logging-service.md
```diff
@@ -41,30 +41,34 @@ There are environment variables for the logging-mux DC config:
...
-  `full_no_k8s_meta`, then fluentd will perform *all* of the data processing
-  and filtering *except* the Kubernetes metadata annotation.  The records will
-  be sent to mux for Kubernetes metadata annotation before sending the logs to
-  Elasticsearch.  In this case, it is assumed you want to deploy mux to be as
-  lightweight as possible, and move the processing burden to the individual
-  nodes.
-
-logging-mux must be deployed with `MUX_ALLOW_EXTERNAL true` in order to receive
-records sent from outside the cluster.
+* `MUX_CLIENT_MODE` - `minimal`/`maximal` - default is unset If `maximal`, then
```
portante: Is there a missing dash character between "`unset`" and "`If`"?

docs/mux-logging-service.md
```diff
@@ -73,12 +77,17 @@ There are several new Ansible parameters which can be used with the
...
 * `openshift_logging_use_mux` - `True`/`False` - default `False` - if
   `True`, create the mux Service, DeploymentConfig, Secrets, ConfigMap, etc.
-* `openshift_logging_use_mux_client` - `True`/`False` - default `False` - if
-  `True`, configure the Fluentd collectors running on each node to send the raw
-  logs to mux instead of directly to Elasticsearch
-* `openshift_logging_use_mux_service` - `True`/`False` - default `False` - if
-  `True`, expose the mux service external to the cluster, and configure the mux
-  fluentd to accept logs from outside the cluster
+- `openshift_logging_mux_client_mode`: Values - `minimal`, `maximal`.
+  Default is unset.  Setting this value will cause the Fluentd node agent to
```
jcantrill: Can we clarify what the behavior is when default?  If unset it operations in max mode? min mode?

richm:
[test]

richm:
[merge]

PR #544, opened 2017-07-24
Impl fluentd file buffer.

jcantrill:
This PR removes memory buffer for fluentd and supercedes #525

PR #556, opened 2017-07-27
This PR succeeds #544.

nhosoi:
@jcantrill, @richm and @portante, could you please review this PR once more?  Thanks.

### DIFF
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -61,11 +61,7 @@ IPADDR4=`/usr/sbin/ip -4 addr show dev eth0 | grep inet | sed -e "s/[ \t]*inet \
 IPADDR6=`/usr/sbin/ip -6 addr show dev eth0 | grep inet6 | sed "s/[ \t]*inet6 \([a-f0-9:]*\).*/\1/"`
 export IPADDR4 IPADDR6
 
-BUFFER_SIZE_LIMIT=${BUFFER_SIZE_LIMIT:-1048576}
-MUX_CPU_LIMIT=${MUX_CPU_LIMIT:-500m}
-MUX_MEMORY_LIMIT=${MUX_MEMORY_LIMIT:-2Gi}
-FLUENTD_CPU_LIMIT=${FLUENTD_CPU_LIMIT:-100m}
-FLUENTD_MEMORY_LIMIT=${FLUENTD_MEMORY_LIMIT:-512Mi}
+BUFFER_SIZE_LIMIT=${BUFFER_SIZE_LIMIT:-16777216}
 
 CFG_DIR=/etc/fluent/configs.d
 if [ "${USE_MUX:-}" = "true" ] ; then
@@ -129,30 +125,55 @@ else
 fi
 export K8S_FILTER_REMOVE_KEYS
 
-if [ "${USE_MUX:-}" = "true" ] ; then
-    TOTAL_MEMORY_LIMIT=`echo $MUX_MEMORY_LIMIT |  sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc`
-else
-    TOTAL_MEMORY_LIMIT=`echo $FLUENTD_MEMORY_LIMIT |  sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc`
-fi
-BUFFER_SIZE_LIMIT=`echo $BUFFER_SIZE_LIMIT |  sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc`
-if [ $BUFFER_SIZE_LIMIT -eq 0 ]; then
-    BUFFER_SIZE_LIMIT=1048576
-fi
-
-DIV=1
-if [ "$ES_HOST" != "$OPS_HOST" ] || [ "$ES_PORT" != "$OPS_PORT" ] ; then
-    # using ops cluster
-    DIV=`expr $DIV \* 2`
-fi
+# How many outputs?
 if [ -n "${MUX_CLIENT_MODE:-}" -o "${USE_MUX_CLIENT:-}" = "true" ] ; then
-    DIV=`expr $DIV \* 2`
+    # A fluentd collector configured as a mux client has just one output: sending to a mux.
+    NUM_OUTPUTS=1
+else
+    # fluentd usually has 2 outputs.
+    NUM_OUTPUTS=2
+    if [ "$ES_COPY" = "true" ]; then
+        NUM_OUTPUTS=`expr $NUM_OUTPUTS \* 2`
+    fi
 fi
 
-# MEMORY_LIMIT per buffer
-MEMORY_LIMIT=`expr $TOTAL_MEMORY_LIMIT / $DIV`
-BUFFER_QUEUE_LIMIT=`expr $MEMORY_LIMIT / $BUFFER_SIZE_LIMIT`
-if [ $BUFFER_QUEUE_LIMIT -eq 0 ]; then
-    BUFFER_QUEUE_LIMIT=1024
+# If FILE_BUFFER_PATH exists and it is not a directory, mkdir fails with the error.
+FILE_BUFFER_PATH=/var/lib/fluentd
+mkdir -p $FILE_BUFFER_PATH
+
+# Get the available disk size; use 1/4 of it
+DF_LIMIT=$(df -B1 $FILE_BUFFER_PATH | grep -v Filesystem | awk '{print $2}')
+DF_LIMIT=${DF_LIMIT:-0}
+DF_LIMIT=$(expr $DF_LIMIT / 4) || :
+if [ $DF_LIMIT -eq 0 ]; then
+    echo "ERROR: No disk space is available for file buffer in $FILE_BUFFER_PATH."
+    exit 1
+fi
+# Determine final total given the number of outputs we have.
+TOTAL_LIMIT=$(echo ${FILE_BUFFER_LIMIT:-2Gi} | sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc) || :
+if [ $TOTAL_LIMIT -le 0 ]; then
+    echo "ERROR: Invalid file buffer limit ($FILE_BUFFER_LIMIT) is given.  Failed to convert to bytes."
+    exit 1
+fi
+TOTAL_LIMIT=$(expr $TOTAL_LIMIT \* $NUM_OUTPUTS) || :
+if [ $DF_LIMIT -lt $TOTAL_LIMIT ]; then
+    echo "WARNING: Available disk space ($DF_LIMIT bytes) is less than the user specified file buffer limit ($FILE_BUFFER_LIMIT times $NUM_OUTPUTS)."
+    TOTAL_LIMIT=$DF_LIMIT
+fi
+
+BUFFER_SIZE_LIMIT=$(echo $BUFFER_SIZE_LIMIT |  sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc)
+BUFFER_SIZE_LIMIT=${BUFFER_SIZE_LIMIT:-16777216}
+
+# TOTAL_BUFFER_SIZE_LIMIT per buffer
+TOTAL_BUFFER_SIZE_LIMIT=$(expr $TOTAL_LIMIT / $NUM_OUTPUTS) || :
+if [ -z $TOTAL_BUFFER_SIZE_LIMIT -o $TOTAL_BUFFER_SIZE_LIMIT -eq 0 ]; then
+    echo "ERROR: Calculated TOTAL_BUFFER_SIZE_LIMIT is 0. TOTAL_LIMIT $TOTAL_LIMIT is too small compared to NUM_OUTPUTS $NUM_OUTPUTS. Please increase FILE_BUFFER_LIMIT $FILE_BUFFER_LIMIT and/or the volume size of $FILE_BUFFER_PATH."
+    exit 1
+fi
+BUFFER_QUEUE_LIMIT=$(expr $TOTAL_BUFFER_SIZE_LIMIT / $BUFFER_SIZE_LIMIT) || :
+if [ -z $BUFFER_QUEUE_LIMIT -o $BUFFER_QUEUE_LIMIT -eq 0 ]; then
+    echo "ERROR: Calculated BUFFER_QUEUE_LIMIT is 0. TOTAL_BUFFER_SIZE_LIMIT $TOTAL_BUFFER_SIZE_LIMIT is too small compared to BUFFER_SIZE_LIMIT $BUFFER_SIZE_LIMIT. Please increase FILE_BUFFER_LIMIT $FILE_BUFFER_LIMIT and/or the volume size of $FILE_BUFFER_PATH."
+    exit 1
 fi
 export BUFFER_QUEUE_LIMIT BUFFER_SIZE_LIMIT
 
@@ -167,12 +188,15 @@ OPS_COPY_PASSWORD="${OPS_COPY_PASSWORD:-$ES_COPY_PASSWORD}"
 export OPS_COPY_HOST OPS_COPY_PORT OPS_COPY_SCHEME OPS_COPY_CLIENT_CERT \
        OPS_COPY_CLIENT_KEY OPS_COPY_CA OPS_COPY_USERNAME OPS_COPY_PASSWORD
 
-if [ "$ES_COPY" = "true" ] ; then
+if [ "$ES_COPY" = "true" -a "${USE_MUX_CLIENT:-}" != "true" ] ; then
     # user wants to split the output of fluentd into two different elasticsearch
     # user will provide the necessary COPY environment variables as above
     cp $CFG_DIR/{openshift,dynamic}/es-copy-config.conf
     cp $CFG_DIR/{openshift,dynamic}/es-ops-copy-config.conf
 else
+    if [ "$ES_COPY" = "true" ] ; then
+        echo "WARNING: When USE_MUX_CLIENT is true, logs are forwarded to MUX; COPY won't work with it."
+    fi
     # create empty files for the ES copy config
     echo > $CFG_DIR/dynamic/es-copy-config.conf
     echo > $CFG_DIR/dynamic/es-ops-copy-config.conf
--- a/fluentd/configs.d/openshift/es-copy-config.conf
+++ b/fluentd/configs.d/openshift/es-copy-config.conf
@@ -20,6 +20,8 @@
       flush_interval "#{ENV['ES_FLUSH_INTERVAL'] || '5s'}"
       max_retry_wait "#{ENV['ES_RETRY_WAIT'] || '300'}"
       disable_retry_limit true
+      buffer_type file
+      buffer_path '/var/lib/fluentd/buffer-es-copy-config'
       buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
       buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT'] || '1m' }"
       # the systemd journald 0.0.8 input plugin will just throw away records if the buffer
--- a/fluentd/configs.d/openshift/es-ops-copy-config.conf
+++ b/fluentd/configs.d/openshift/es-ops-copy-config.conf
@@ -20,6 +20,8 @@
       flush_interval "#{ENV['OPS_FLUSH_INTERVAL'] || ENV['ES_FLUSH_INTERVAL'] || '5s'}"
       max_retry_wait "#{ENV['OPS_RETRY_WAIT'] || ENV['ES_RETRY_WAIT'] || '300'}"
       disable_retry_limit true
+      buffer_type file
+      buffer_path '/var/lib/fluentd/buffer-es-ops-copy-config'
       buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
       buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT'] || '1m' }"
       # the systemd journald 0.0.8 input plugin will just throw away records if the buffer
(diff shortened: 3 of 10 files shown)


fluentd/run.sh
```diff
@@ -129,32 +125,58 @@ else
...
+if [ "${USE_MUX_CLIENT:-}" = "true" ] ; then
+    # A fluentd collector configured as mux client has just one output: sending to a mux.
+    DIV=1
 else
-    TOTAL_MEMORY_LIMIT=`echo $FLUENTD_MEMORY_LIMIT |  sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc`
-fi
-BUFFER_SIZE_LIMIT=`echo $BUFFER_SIZE_LIMIT |  sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc`
-if [ $BUFFER_SIZE_LIMIT -eq 0 ]; then
-    BUFFER_SIZE_LIMIT=1048576
+    # fluend usually has 2 outputs.
```
portante: s/fluend/fluentd/

fluentd/run.sh
```diff
@@ -129,32 +125,58 @@ else
...
+        DIV=`expr $DIV \* 2`
+    fi
 fi
 
-DIV=1
-if [ "$ES_HOST" != "$OPS_HOST" ] || [ "$ES_PORT" != "$OPS_PORT" ] ; then
-    # using ops cluster
-    DIV=`expr $DIV \* 2`
+FILE_BUFFER_PATH=/var/lib/fluentd
+if [ ! -d $FILE_BUFFER_PATH ]; then
```
portante: If the `$FILE_BUFFER_PATH` is a file, this check will fail.  But since the `mkdir` below will also fail if `$FILE_BUFFER_PATH` is a file, but will ignore an already created directory, and create the directory when it does not exist, why not only perform the `mkdir -p $FILE_BUFFER_PATH`?

fluentd/run.sh
```diff
@@ -129,32 +125,58 @@ else
...
+    mkdir -p $FILE_BUFFER_PATH
 fi
+
 if [ -n "${MUX_CLIENT_MODE:-}" -o "${USE_MUX_CLIENT:-}" = "true" ] ; then
```
portante: Can you add a comment as to why we multiply `DIV` by 2 here?  This conditional appears to say, "if we don't have `MUX_CLIENT_MODE` defined, or `USE_MUX_CLIENT` is `true`, then we need to double the number of outputs we have."

But didn't we already consider this at lines 128-138?

fluentd/run.sh
```diff
@@ -129,32 +125,58 @@ else
 fi
 export K8S_FILTER_REMOVE_KEYS
 
-if [ "${USE_MUX:-}" = "true" ] ; then
-    TOTAL_MEMORY_LIMIT=`echo $MUX_MEMORY_LIMIT |  sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc`
+# How many outputs?
+if [ "${USE_MUX_CLIENT:-}" = "true" ] ; then
+    # A fluentd collector configured as mux client has just one output: sending to a mux.
+    DIV=1
```
portante: Instead of using `DIV`, can we name this `NUM_OUTPUTS`?

fluentd/run.sh
```diff
@@ -129,32 +125,58 @@ else
...
+DF_LIMIT=$(df -B1 $FILE_BUFFER_PATH | grep -v Filesystem | awk '{print $2}')
+DF_LIMIT=${DF_LIMIT:-0}
+DF_LIMIT=$(expr $DF_LIMIT / 4) || :
+if [ $DF_LIMIT -eq 0 ]; then
+    echo "ERROR: No disk space is available for file buffer in $FILE_BUFFER_PATH."
+    exit 1
+fi
+# Given buffer limit per output
+TOTAL_LIMIT=$(echo ${FILE_BUFFER_LIMIT:-2Gi} | sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc) || :
+# In total
```
portante: Perhaps reword to "Determine final total given the number of outputs we have."

fluentd/run.sh
```diff
@@ -129,32 +125,58 @@ else
...
+TOTAL_LIMIT=$(expr $TOTAL_LIMIT \* $DIV) || :
```
portante: Perhaps rename `DIV` to `NUM_OUTPUTS`?

fluentd/run.sh
```diff
@@ -129,32 +125,58 @@ else
...
+if [ $TOTAL_LIMIT -le 0 ]; then
```
portante: Can you consider moving this check to just after line 157?  It seems this check is not dependent on the intervening adjustment.

fluentd/run.sh
```diff
@@ -129,32 +125,58 @@ else
...
+    echo "ERROR: Invalid file buffer limit ($FILE_BUFFER_LIMIT) is given.  Failed to convert to bytes."
+    exit 1
+fi
+# Available disk space is less than the specified size.
+if [ $DF_LIMIT -lt $TOTAL_LIMIT ]; then
+    echo "WARNING: Available disk space ($DF_LIMIT bytes) is less than the user specified file buffer limit ($FILE_BUFFER_LIMIT times $DIV)."
```
portante: Perhaps rename `DIV` to `NUM_OUTPUTS`?

fluentd/run.sh
```diff
@@ -129,32 +125,58 @@ else
...
+fi
+# Given buffer limit per output
+TOTAL_LIMIT=$(echo ${FILE_BUFFER_LIMIT:-2Gi} | sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc) || :
+# In total
+TOTAL_LIMIT=$(expr $TOTAL_LIMIT \* $DIV) || :
+if [ $TOTAL_LIMIT -le 0 ]; then
+    echo "ERROR: Invalid file buffer limit ($FILE_BUFFER_LIMIT) is given.  Failed to convert to bytes."
+    exit 1
+fi
+# Available disk space is less than the specified size.
```
portante: Does this comment add anything not said in the error message at line 166?

fluentd/run.sh
```diff
@@ -129,32 +125,58 @@ else
...
+BUFFER_SIZE_LIMIT=$(echo $BUFFER_SIZE_LIMIT |  sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc)
+BUFFER_SIZE_LIMIT=${BUFFER_SIZE_LIMIT:-1048576}
 
-# MEMORY_LIMIT per buffer
-MEMORY_LIMIT=`expr $TOTAL_MEMORY_LIMIT / $DIV`
-BUFFER_QUEUE_LIMIT=`expr $MEMORY_LIMIT / $BUFFER_SIZE_LIMIT`
-if [ $BUFFER_QUEUE_LIMIT -eq 0 ]; then
-    BUFFER_QUEUE_LIMIT=1024
+# TOTAL_BUFFER_SIZE_LIMIT per buffer
+TOTAL_BUFFER_SIZE_LIMIT=$(expr $TOTAL_LIMIT / $DIV) || :
```
portante: Perhaps rename `DIV` to `NUM_OUTPUTS`?

fluentd/run.sh
```diff
@@ -129,32 +125,58 @@ else
...
+BUFFER_QUEUE_LIMIT=$(expr $TOTAL_BUFFER_SIZE_LIMIT / $BUFFER_SIZE_LIMIT) || :
+if [ -z $BUFFER_QUEUE_LIMIT -o $BUFFER_QUEUE_LIMIT -eq 0 ]; then
+    BUFFER_QUEUE_LIMIT=256
```
portante: Can we really just pick an arbitrary number here?  Isn't this the case that either the environmental constraints (size of the partition on which the file buffer will live) is too small or that the user specified `FILE_BUFFER_LIMIT` is too small?  If so, arbitrarily picking a queue limit that would result in having more queue entries than would fit in the file buffer seems like it would present a problem to diagnose in the field.

fluentd/run.sh
nhosoi: How about adding rm?
```
 if [ ! -d $FILE_BUFFER_PATH ]; then
+    rm -rf $FILE_BUFFER_PATH || :
     mkdir -p $FILE_BUFFER_PATH
 fi
```

fluentd/run.sh
nhosoi: Yes, you are right.  I removed it in #525...  I'm removing this part.
```
-if [ -n "${MUX_CLIENT_MODE:-}" -o "${USE_MUX_CLIENT:-}" = "true" ] ; then
-    DIV=`expr $DIV \* 2`
-fi
```

fluentd/run.sh
nhosoi: @portante, @richm, fluentd/run.sh still sets BUFFER_SIZE_LIMIT to 1M by default.  Is it ok?  
` 64 BUFFER_SIZE_LIMIT=${BUFFER_SIZE_LIMIT:-1048576}`
Shall I change the default value to 8M or 16M or ...?

fluentd/run.sh
```diff
@@ -129,32 +125,58 @@ else
 fi
 export K8S_FILTER_REMOVE_KEYS
 
-if [ "${USE_MUX:-}" = "true" ] ; then
-    TOTAL_MEMORY_LIMIT=`echo $MUX_MEMORY_LIMIT |  sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc`
+# How many outputs?
+if [ "${USE_MUX_CLIENT:-}" = "true" ] ; then
+    # A fluentd collector configured as mux client has just one output: sending to a mux.
```
nhosoi: Done.

fluentd/run.sh
nhosoi: Done. (replaced every DIV with NUM_OUTPUTS)

fluentd/run.sh
nhosoi: Done.

fluentd/run.sh
portante: @nhosoi, changing the default to 16 MB would be fine.

fluentd/run.sh
portante: Will the use of `rm` mask an error?  Why not just use a one line `mkdir`?

fluentd/run.sh
```diff
@@ -167,7 +186,8 @@ OPS_COPY_PASSWORD="${OPS_COPY_PASSWORD:-$ES_COPY_PASSWORD}"
 export OPS_COPY_HOST OPS_COPY_PORT OPS_COPY_SCHEME OPS_COPY_CLIENT_CERT \
        OPS_COPY_CLIENT_KEY OPS_COPY_CA OPS_COPY_USERNAME OPS_COPY_PASSWORD
 
-if [ "$ES_COPY" = "true" ] ; then
+# When USE_MUX_CLIENT is true, logs are forwarded to MUX; COPY won't work with it.
```
portante: Should we warn the user instead of having this only in a comment?

fluentd/run.sh
```diff
@@ -129,32 +125,55 @@ else
...
-if [ $BUFFER_QUEUE_LIMIT -eq 0 ]; then
-    BUFFER_QUEUE_LIMIT=1024
+BUFFER_SIZE_LIMIT=$(echo $BUFFER_SIZE_LIMIT |  sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc)
+BUFFER_SIZE_LIMIT=${BUFFER_SIZE_LIMIT:-1048576}
+
+# TOTAL_BUFFER_SIZE_LIMIT per buffer
+TOTAL_BUFFER_SIZE_LIMIT=$(expr $TOTAL_LIMIT / $NUM_OUTPUTS) || :
+BUFFER_QUEUE_LIMIT=$(expr $TOTAL_BUFFER_SIZE_LIMIT / $BUFFER_SIZE_LIMIT) || :
+if [ -z $BUFFER_QUEUE_LIMIT -o $BUFFER_QUEUE_LIMIT -eq 0 ]; then
+    echo "ERROR: Failed to calculate BUFFER_QUEUE_LIMIT from (TOTAL_BUFFER_SIZE_LIMIT $TOTAL_BUFFER_SIZE_LIMIT) / (BUFFER_SIZE_LIMIT $BUFFER_SIZE_LIMIT)."
```
portante: Does this error only result when `BUFFER_SIZE_LIMIT` is greater than `TOTAL_BUFFER_SIZE_LIMIT`?  If so, should we just tell the users they need to adjust one or the other?

fluentd/run.sh
```diff
@@ -129,32 +125,55 @@ else
...
+    fi
 fi
-BUFFER_SIZE_LIMIT=`echo $BUFFER_SIZE_LIMIT |  sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc`
-if [ $BUFFER_SIZE_LIMIT -eq 0 ]; then
-    BUFFER_SIZE_LIMIT=1048576
+
+FILE_BUFFER_PATH=/var/lib/fluentd
+if [ ! -d $FILE_BUFFER_PATH ]; then
+    rm -rf $FILE_BUFFER_PATH || :
+    mkdir -p $FILE_BUFFER_PATH
```
portante: Would using `rm` mask a problem where something other than a directory is being provided in the image?  Wouldn't a simple `mkdir` be sufficient here?

nhosoi: I think I haven't understood your concern yet...

I thought if a non-directory /var/lib/fluentd was found, it'd be better to be replaced with /var/lib/fluentd directory.  I guess this is what you disagree?

If we remove the line 141 check and always call mkdir, it'd fail as you commented in your original comment (if non-directory /var/libfluentd is there).  That is, fluentd/run.sh is supposed to fail there instead of going forward?  If so, do we want to add an error message about it?  Or the error message from the mkdir command is good enough?

fluentd/run.sh
nhosoi: How about this?
```
diff --git a/fluentd/run.sh b/fluentd/run.sh
index 5523244..555006b 100644
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -168,9 +168,13 @@ BUFFER_SIZE_LIMIT=${BUFFER_SIZE_LIMIT:-1048576}
 
 # TOTAL_BUFFER_SIZE_LIMIT per buffer
 TOTAL_BUFFER_SIZE_LIMIT=$(expr $TOTAL_LIMIT / $NUM_OUTPUTS) || :
+if [ -z $TOTAL_BUFFER_SIZE_LIMIT -o $TOTAL_BUFFER_SIZE_LIMIT -eq 0 ]; then
+    echo "ERROR: Calculated TOTAL_BUFFER_SIZE_LIMIT is 0. TOTAL_LIMIT $TOTAL_LIMIT is too small compared to NUM_OUTPUTS $NUM_OUTPUTS. Please increase FILE_BUFFER_LIMIT $FILE_BUFFER_LIMIT and/or the volume size of $FILE_BUFFER_PATH."
+    exit 1
+fi
 BUFFER_QUEUE_LIMIT=$(expr $TOTAL_BUFFER_SIZE_LIMIT / $BUFFER_SIZE_LIMIT) || :
 if [ -z $BUFFER_QUEUE_LIMIT -o $BUFFER_QUEUE_LIMIT -eq 0 ]; then
-    echo "ERROR: Failed to calculate BUFFER_QUEUE_LIMIT from (TOTAL_BUFFER_SIZE_LIMIT $TOTAL_BUFFER_SIZE_LIMIT) / (BUFFER_SIZE_LIMIT $BUFFER_SIZE_LIMIT)."
+    echo "ERROR: Calculated BUFFER_QUEUE_LIMIT is 0. TOTAL_BUFFER_SIZE_LIMIT $TOTAL_BUFFER_SIZE_LIMIT is too small compared to BUFFER_SIZE_LIMIT $BUFFER_SIZE_LIMIT. Please increase FILE_BUFFER_LIMIT $FILE_BUFFER_LIMIT and/or the volume size of $FILE_BUFFER_PATH."
     exit 1
 fi
 export BUFFER_QUEUE_LIMIT BUFFER_SIZE_LIMIT BUFFER_TYPE

```

fluentd/run.sh
nhosoi: Do you think this is ok?
```
@@ -186,8 +190,8 @@ OPS_COPY_PASSWORD="${OPS_COPY_PASSWORD:-$ES_COPY_PASSWORD}"
 export OPS_COPY_HOST OPS_COPY_PORT OPS_COPY_SCHEME OPS_COPY_CLIENT_CERT \
        OPS_COPY_CLIENT_KEY OPS_COPY_CA OPS_COPY_USERNAME OPS_COPY_PASSWORD
 
-# When USE_MUX_CLIENT is true, logs are forwarded to MUX; COPY won't work with it.
 if [ "$ES_COPY" = "true" -a "${USE_MUX_CLIENT:-}" != "true" ] ; then
+    echo "WARNING: When USE_MUX_CLIENT is true, logs are forwarded to MUX; COPY won't work with it."
     # user wants to split the output of fluentd into two different elasticsearch
     # user will provide the necessary COPY environment variables as above
     cp $CFG_DIR/{openshift,dynamic}/es-copy-config.conf
```

fluentd/run.sh
richm: > I thought if a non-directory /var/lib/fluentd was found, it'd be better to be replaced with /var/lib/fluentd directory. I guess this is what you disagree?

I think we should just print an ERROR message and exit.  If /var/lib/fluentd is something that isn't a directory, there is something very wrong that should be solved by the operator (or us via bug fix) immediately.

nhosoi:
@jcantrill , @richm , I've pushed the fixes for the issues pointed by @portante.  Since @portante is not available until the late afternoon, could you please take a look at them?  Thanks!!

fluentd/run.sh
nhosoi: Thanks, @richm.  As @portante suggested, I got rid of the directory check.  This is the current code:
```
# If FILE_BUFFER_PATH exists and it is not a directory, mkdir fails with the error.
FILE_BUFFER_PATH=/var/lib/fluentd
mkdir -p $FILE_BUFFER_PATH
```
And if non-directory /var/lib/fluentd exists, run.sh fails as follows.  Do we want to have a clearer error message?
```
+ FILE_BUFFER_PATH=/var/lib/fluentd
+ mkdir -p /var/lib/fluentd
mkdir: cannot create directory '/var/lib/fluentd': File exists
```

nhosoi: Note: if /var/lib/fluentd exists and it is a directory, mkdir -p does nothing and just returns 0.

fluentd/run.sh
```diff
@@ -137,11 +137,9 @@ else
     fi
 fi
 
+# If FILE_BUFFER_PATH exists and it is not a directory, mkdir fails with the error.
```
richm: correct, because the script is run with `set -e`

richm:
[test]

fluentd/run.sh
nhosoi: If it were a symlink pointing to some volume, would it be acceptable?

richm:
lgtm

fluentd/run.sh
nhosoi: Thanks.  Done.

fluentd/run.sh
```diff
@@ -129,32 +125,57 @@ else
...
+    echo "ERROR: Calculated TOTAL_BUFFER_SIZE_LIMIT is 0. TOTAL_LIMIT $TOTAL_LIMIT is too small compared to NUM_OUTPUTS $NUM_OUTPUTS. Please increase FILE_BUFFER_LIMIT $FILE_BUFFER_LIMIT and/or the volume size of $FILE_BUFFER_PATH."
+    exit 1
+fi
+BUFFER_QUEUE_LIMIT=$(expr $TOTAL_BUFFER_SIZE_LIMIT / $BUFFER_SIZE_LIMIT) || :
+if [ -z $BUFFER_QUEUE_LIMIT -o $BUFFER_QUEUE_LIMIT -eq 0 ]; then
+    echo "ERROR: Calculated BUFFER_QUEUE_LIMIT is 0. TOTAL_BUFFER_SIZE_LIMIT $TOTAL_BUFFER_SIZE_LIMIT is too small compared to BUFFER_SIZE_LIMIT $BUFFER_SIZE_LIMIT. Please increase FILE_BUFFER_LIMIT $FILE_BUFFER_LIMIT and/or the volume size of $FILE_BUFFER_PATH."
+    exit 1
 fi
-export BUFFER_QUEUE_LIMIT BUFFER_SIZE_LIMIT
+export BUFFER_QUEUE_LIMIT BUFFER_SIZE_LIMIT BUFFER_TYPE
```
portante: `BUFFER_TYPE` does not seem to appear anywhere, is it needed?

fluentd/run.sh
```diff
@@ -167,12 +188,13 @@ OPS_COPY_PASSWORD="${OPS_COPY_PASSWORD:-$ES_COPY_PASSWORD}"
...
        OPS_COPY_CLIENT_KEY OPS_COPY_CA OPS_COPY_USERNAME OPS_COPY_PASSWORD
 
-if [ "$ES_COPY" = "true" ] ; then
+if [ "$ES_COPY" = "true" -a "${USE_MUX_CLIENT:-}" != "true" ] ; then
     # user wants to split the output of fluentd into two different elasticsearch
     # user will provide the necessary COPY environment variables as above
     cp $CFG_DIR/{openshift,dynamic}/es-copy-config.conf
     cp $CFG_DIR/{openshift,dynamic}/es-ops-copy-config.conf
 else
+    echo "WARNING: When USE_MUX_CLIENT is true, logs are forwarded to MUX; COPY won't work with it."
```
portante: Is this because the COPY operation happens in MUX instead?

fluentd/run.sh
nhosoi: Right.  Since memory buffer is no longer supported, there is no need to have it in run.sh. 
Thanks for finding it out, @portante !

fluentd/run.sh
nhosoi: Ahhh, thanks again, @portante .
The logic should be fine.  (MUX_CLIENT_MODE and USE_MUX_CLIENT are not passed to mux.  So, if ES_COPY is true, the copy config files are set to the expected location /etc/fluent/configs.d/dynamic on the mux pod.)  B*U*T, the WARNING should be issued only when ES_COPY is true...  The fix has been pushed.

richm:
[merge]

nhosoi:
[(5) Use `file` buffer instead of `memory` buffer for fluentd[logging][epic-ois-agl-perf]](https://trello.com/c/XpreI533/509-5-use-file-buffer-instead-of-memory-buffer-for-fluentdloggingepic-ois-agl-perf)

richm:
[merge]

PR #564, opened 2017-08-02
base kibana on SCL node6 image

jcantrill:
This change:

* makes the downstream image dependent upon SCL node maintained image node6
* Aligns the node runtime with the version provided with the version of Kibana we are using

This change does not modify upstream to use the comparable image since we install kibana from a download that includes the node runtime.

Should we consider backporting this to v3.4 and v3.5?  During my research it seemed there were potential memory issues associated with node4 which may be the root of our OOM

### DIFF
--- a/kibana/Dockerfile
+++ b/kibana/Dockerfile
@@ -1,4 +1,4 @@
-FROM rhel7.3:7.3-released
+FROM rhscl/nodejs-6-rhel7
 
 MAINTAINER OpenShift Development <dev@lists.openshift.redhat.com>
 
@@ -13,7 +13,7 @@ ENV AOP_KIBANA_PLUGIN_VER=4.5.1 \
     CONFIG_PATH=/etc/kibana \
     NODE_ENV=production \
     KIBANA_VER=4.6.4 \
-    NODE_BIN=/usr/bin/node \
+    NODE_BIN=/opt/rh/rh-nodejs6/root/usr/bin/node \
     RELEASE_STREAM=prod
 
 LABEL io.k8s.description="Kibana container for querying Elasticsearch for aggregated logs" \
@@ -26,7 +26,11 @@ LABEL io.k8s.description="Kibana container for querying Elasticsearch for aggreg
       version="3.6.0" \
       release="1"
 
-RUN yum-config-manager --enable rhel-7-server-ose-3.6-rpms && \
+USER 0
+
+RUN yum repolist > /dev/null && \
+    yum-config-manager --enable rhel-7-server-ose-3.6-rpms && \
+    yum-config-manager --disable epel >/dev/null || : && \
     INSTALLED_PKGS="kibana-${KIBANA_VER} \
                     origin-kibana-${AOP_KIBANA_PLUGIN_VER}" && \
     yum install -y --setopt=tsflags=nodocs  ${INSTALLED_PKGS} && \


jcantrill:
[test]

kibana/Dockerfile
```diff
@@ -26,7 +26,13 @@ LABEL io.k8s.description="Kibana container for querying Elasticsearch for aggreg
       version="3.6.0" \
       release="1"
 
-RUN yum-config-manager --enable rhel-7-server-ose-3.6-rpms && \
+USER 0
+
+RUN yum repolist > /dev/null && \
+    yum-config-manager --enable rhel-server-rhscl-7-rpms && \
+    yum-config-manager --enable rhel-7-server-optional-rpms && \
```
richm: What packages are we still pulling from these two repos, that are not already baked into the base image?

jcantrill: I'll check... copied from the jenkins image

wozniakjan:
Even though it doesn't show any signs of seg faults as the kibana-proxy, would we perhaps like to wait until this bz is addressed by the image maintainer? 

https://bugzilla.redhat.com/show_bug.cgi?id=1473330

jcantrill:
@wozniakjan My preference would be to use the version for which kibana is released to run.  We have been out of sync with it for some time.

jcantrill:
[test] flake?

richm:
@jcantrill not a test flake:
```
To install a containerized Origin release, you must set openshift_release or
openshift_image_tag in your inventory to specify which version of the OpenShift
component images to use. You may want the latest (usually alpha) releases or
a more stable release. (Suggestion: add openshift_release="x.y" to inventory.)
```
Note that we have this in our init-log-stack: https://github.com/openshift/origin-aggregated-logging/blob/master/hack/testing/init-log-stack#L39
```
containerized=true
```
Do we need that?  If so, looks like we will need to add `openshift_release` to the inventory in init-log-stack

kibana/Dockerfile
```diff
@@ -13,7 +13,7 @@ ENV AOP_KIBANA_PLUGIN_VER=4.5.1 \
     CONFIG_PATH=/etc/kibana \
     NODE_ENV=production \
     KIBANA_VER=4.6.4 \
-    NODE_BIN=/usr/bin/node \
+    NODE_BIN=/opt/rh/rh-nodejs6/root/usr/bin/node \
```
wozniakjan: It is possible that following issue is only my local, as I was informed that mileage may vary when building rhel based images on Fedora, but is it possible that we may need to setup more ENV variables? 

Looking at what [scl enable](https://gist.githubusercontent.com/wozniakjan/321faa0f3163ef95f341867237c64369/raw/8064abfc7fea705b9719b0d6161033ce3c1aa3df/enable) does in this image, it may be insufficient to only change the path to nodejs binary. An [error message](https://gist.githubusercontent.com/wozniakjan/321faa0f3163ef95f341867237c64369/raw/8064abfc7fea705b9719b0d6161033ce3c1aa3df/log-install%2520failure) I see when building the image locally is a weak support but still may be worth addressing, both seem to relate to shared library path.

And it looks like kibana is still pulling own nodejs as a dependency in my case - https://gist.githubusercontent.com/wozniakjan/321faa0f3163ef95f341867237c64369/raw/8064abfc7fea705b9719b0d6161033ce3c1aa3df/log-dependencies

jcantrill: You are correct it is still pulling in the nodejs specified in the spec file.  We need to changes that as well.  I just need to figure out how to do that.  I tested the build of the image by creating a scratch build through OSBS, basically what is described in the 'building images' section.  This produced the image found here [2].

[1] https://mojo.redhat.com/docs/DOC-1137707
[2] https://brewweb.engineering.redhat.com/brew/taskinfo?taskID=13768194

jcantrill:
[test] again after we decided to remove the original logging job

jcantrill:
[merge]

Issue #443, opened 2017-05-29
Port all test scripts to work on Ansible-installed cluster


stevekuznetsov:
| Script | OK | Notes |
| ----- | ---- | ---- |
| `test-curator.sh`         | ✔️ | Needs `"${ENABLE_OPS_CLUSTER}"` as first arg |
| `test-datetime-future.sh` | ❌ | Needs edit to `get_test_user_token` for project, needs journal logging |
| `test-es-copy.sh`         | ❌ | #565  |
| `test-fluentd-forward.sh` | ❌ | #566  |
| `test-json-parsing.sh`    | ✔️ | #567 |
| `test-mux.sh`             | ❓ | #570|
| `test-upgrade.sh`         | ✔️  | get rid of this - doesn't currently work - need a new ansible based test |
| `test-viaq-data-model.sh` | ❓  | #568  |
| `check-EFK-running.sh` | ✔️ | |
| `check-logs.sh` | ✔️ | |

Proposal:

 - `hack/testing/logging.sh` takes parameters `$KUBECONFIG` and `$TEST_ONLY`, to trigger whether it calls into `hack/testing/setup.sh` before running component tests
 - `hack/testing/setup.sh` does cluster setup that normally would be done by a job (install OpenShift-Ansible, start cluster, etc)
 - `hack/testing/test-X.sh` (???) does set-up on cluster for specific test, then calls `test/X.sh` to run it
 - `test/X.sh` runs a test

We would use `hack/testing/test-X.sh` for instance to translate "is there an ops cluster" into simple input to the _actual_ test -- adding more items to the list of `DeploymentConfig`s or changing the `--selector` field, so the test script under `test/` can be super simple.

stevekuznetsov:
Also a bunch of these expect to be called from `hack/testing` dir and are not well formed when called anywhere else.

ewolinetz:
We should be able to create one-off plays to run to configure for the different tests now that the components are split up into different roles...

richm:
I think we should get rid of the need to set `ENABLE_OPS_CLUSTER` in the individual tests.  Instead, the code can just figure out if ops clusters are in use or not.  For the CI tests, this is mostly to know whether or not es-ops is being used, which can easily be determined.  For example, look at https://github.com/openshift/origin-aggregated-logging/blob/master/deployer/scripts/util.sh#L506:
```
    local es_pod=`get_running_pod es`
    local es_ops_pod=`get_running_pod es-ops`
[...]

PR #572, opened 2017-08-04
Add readme guidelines for writing tests

jcantrill:
I'm trying to capture the guidelines and assumptions from #443 and https://trello.com/c/UBias2xZ for writing tests using the framework provided.


test/README.md
```diff
@@ -0,0 +1,60 @@
+# Guidelines for Writing Tests
+
+## Assumptions & Requirements
+
+1. Directory paths should be well formed such that they can be run from any directory.
+1. Test scripts in the `$HOME/test` directory assume they are run against a functional Openshift cluster that was setup prior to the execution of the test script.
```
stevekuznetsov: `$HOME` here could be misleading

test/README.md
```diff
@@ -0,0 +1,60 @@
...
+1. Each script in `$HOME/test` should has a companion script in `$HOME/hack/testing`
```
stevekuznetsov: ... which can be used to launch the test and do requisite setup steps...

test/README.md
```diff
@@ -0,0 +1,60 @@
...
+1. Each script in `$HOME/test` performs cleanup logic to reset the state of the cluster and logging stack
+
+```
+#!/bin/bash
+
+source "$(dirname "${BASH_SOURCE[0]}" )/../hack/lib/init.sh"
+
+os::util::environment::use_sudo
```
stevekuznetsov: let people figure out they need this, don't suggest it outright

test/README.md
```diff
@@ -0,0 +1,60 @@
+# Guidelines for Writing Tests
+
+## Assumptions & Requirements
+
+1. Directory paths should be well formed such that they can be run from any directory.
+1. Test scripts in the `$HOME/test` directory assume they are run against a functional Openshift cluster that was setup prior to the execution of the test script.
+1. Each script in `$HOME/test` should has a companion script in `$HOME/hack/testing`
+1. Each script in `$HOME/test` performs cleanup logic to reset the state of the cluster and logging stack
+
+```
```
stevekuznetsov: use 

    ```sh

for highlighting

test/README.md
```diff
@@ -0,0 +1,60 @@
...
+#!/bin/bash
+
+source "$(dirname "${BASH_SOURCE[0]}" )/../hack/lib/init.sh"
+
+os::util::environment::use_sudo
+
+os::test::junit::declare_suite_start "test/<TESTNAME>"
```
stevekuznetsov: declare end

test/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+if [ -n "${DEBUG:-}" ] ; then
```
stevekuznetsov: don't suggest this -- `os::cmd` + `set -x` = thousands of lines of output nobody needs

stevekuznetsov: there is a `$OS_DEBUG` var that turns on printing from `os::log::debug` functions, use instead

test/README.md
jcantrill: What do you mean as the change here?

stevekuznetsov: If you declare start, declare end too in this example

test/README.md
```diff
@@ -3,23 +3,18 @@
 ## Assumptions & Requirements
 
 1. Directory paths should be well formed such that they can be run from any directory.
-1. Test scripts in the `$HOME/test` directory assume they are run against a functional Openshift cluster that was setup prior to the execution of the test script.
-1. Each script in `$HOME/test` should has a companion script in `$HOME/hack/testing`
-1. Each script in `$HOME/test` performs cleanup logic to reset the state of the cluster and logging stack
+1. Test scripts in the `origin-aggregated-logging/test` directory assume they are run against a functional Openshift cluster that was setup prior to the execution of the test script.
+1. Each script in `origin-aggregated-logging/test` should has a companion script in `origin-aggregated-logging/hack/testing` which can be used to launch the test and do requisite setup steps.
```
richm: I don't think this is necessary.  I thought the `/hack/testing` scripts were only to allow us to run the `/test` scripts using `logging.sh`.

test/README.md
richm: In practice, `declare_suite_end` will be called from `os::test::junit::reconcile_output` which should be the last thing in the `cleanup` function.

richm:
What's the rationale for calling `os::cleanup::all` from an individual test?  e.g. in `test/curator.sh` do we need to `dump_events`, `dump_etcd`, `dump_container_logs`, `dump_pprof_output`, etc.?  Or is that something we would generally do after `entrypoint.sh` is completed? or perhaps only upon failure of an individual test?

stevekuznetsov:
@richm +1 to cleanup in the entrypoint. The tests themselves should only clean up if they make disruptive changes to the cluster

test/README.md
jcantrill: I defer to you and @stevekuznetsov regarding this statement.  I thought it was need for any custom setup to run the script in /test.  If that is not the case then I will remove it; you tell me.

test/README.md
```diff
@@ -21,7 +21,7 @@ os::test::junit::declare_suite_start "test/<TESTNAME>"
 
 ## Role of Companion scripts in hack/testing
 
-The wrapper or companion script for each test is responsible for 
+The wrapper or companion script for each test is responsible for
```
richm: @stevekuznetsov is this true?  I thought the purpose of the `hack/testing/test-*` scripts was to bridge us until we could get rid of `hack/testing`

stevekuznetsov: If the only setup, etc that is necessary lives in `entrypoint.sh` then yes, these are not necessary. I think some of these scripts right now call into the test 2x, one for ops and one for not

stevekuznetsov: That's kind of a question of how you want to structure your test cases TBH and I don't really have a strong opinion -- but it does seem cleaner to say "we expect these components to work this way" then call that test twice with two different sets of components

richm: > I think some of these scripts right now call into the test 2x, one for ops and one for not

I don't think so.  Some of the tests, like curator.sh, will run with and without ops _inside_ the test script itself.

stevekuznetsov: Yeah, looks like drift has happened -- whatever you feel is appropriate here should be enforced so it's a little more clear

PR #575, opened 2017-08-07
monitor fluentd perf; use hack/lib functions in util.sh

richm:
Add some performance monitoring for fluentd:

- top
- how far behind fluentd is reading the journal
- what is the rate of logs being written to the journal

if there was a problem in wait_for_fluentd_to_catch_up, the function
would potentially spew a lot of journal messages - do not print
these unless running in debug mode - figured I might as well convert
this function and a couple of others to use the hack/lib functions
[test]

### DIFF
--- a/deployer/scripts/util.sh
+++ b/deployer/scripts/util.sh
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+if ! type -t os::log::info > /dev/null ; then
+    source "${OS_O_A_L_DIR:-..}/hack/lib/init.sh"
+fi
+
 function generate_JKS_chain() {
     dir=${SCRATCH_DIR:-_output}
     ADD_OID=$1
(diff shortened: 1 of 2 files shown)


richm:
It's tough with journal to get % stats.  There isn't a cheap way, afaik, to say "journalctl give me the number of records in the journal" - you have to do `journalctl | wc -l`.

I could introduce another stat - number of records that fluentd has read from the journal so far - that would essentially be `journalctl -r -c $( cat /var/log/journal.pos ) | wc -l`.  But the problem there is that we don't know exactly where fluentd began reading - that is, fluentd reads from the tail by default, but we don't record the position of the tail - I suppose the stat gathering could save a copy of the first `/var/log/journal.pos` that it finds, and use `journalctl -c $(cat first.pos) | wc -l` to find out how many records total it is possible for fluentd to have read so far.

richm:
[merge]

PR #583, opened 2017-08-11
Test base 3.6.0 fluentd perf but use new viaq filter plugin

richm:
Test base 3.6.0 fluentd perf but use new viaq filter plugin
[test]


richm:
[test]

richm:
[test]

PR #601, opened 2017-08-17
port test-mux.sh to hack/lib mux.sh

richm:
(cherry picked from commit f5af94804e01bee7464e55a1f0723f412978915f)


jcantrill:
[test]

richm:
[merge]

test/mux.sh
```diff
@@ -0,0 +1,353 @@
...
+fi
+
+source "$(dirname "${BASH_SOURCE[0]}" )/../hack/lib/init.sh"
+source "${OS_O_A_L_DIR}/deployer/scripts/util.sh"
+os::util::environment::use_sudo
+
+os::test::junit::declare_suite_start "test/mux"
+
+if [ -n "${DEBUG:-}" ] ; then
+    set -x
```
stevekuznetsov: Still not sure anyone wants this, but ok

test/mux.sh
```diff
@@ -0,0 +1,353 @@
...
+fi
+
+reset_fluentd_daemonset() {
+  # this test only works with MUX_CLIENT_MODE=minimal for now
+  os::log::debug "$( oc set env daemonset/logging-fluentd MUX_CLIENT_MODE=minimal )"
```
stevekuznetsov: Do you care if the stuff in `$()` fails here?

test/mux.sh
```diff
@@ -0,0 +1,353 @@
...
+        </record>\n\
+      </filter>"}]'
+  else
+      os::log::info "Enabling secure forward"
+  fi
+
+  reset_fluentd_daemonset
+
+  os::log::debug "$( oc label node --all logging-infra-fluentd=true )"
+  os::cmd::try_until_text "oc get pods -l component=fluentd" "^logging-fluentd-.* Running "
```
stevekuznetsov: What's driving this? Can you use `oc rollout status`?

test/mux.sh
```diff
@@ -0,0 +1,353 @@
...
+            os::log::debug "$( oc exec $muxpod -- ls /etc/fluent/configs.d/openshift )"
+            os::log::debug "$( oc exec $muxpod -- ls /etc/fluent/configs.d/user )"
+        fi
+    fi
+    $mycmd mux test finished at $( date )
+    # dump the pod before we restart it
+    if [ -n "${fpod:-}" ] ; then
+        oc logs $fpod > $ARTIFACT_DIR/$fpod.log 2>&1
+    fi
+    os::log::debug "$( oc label node --all logging-infra-fluentd- 2>&1 || : )"
```
stevekuznetsov: Why is it ok if this fails?

test/mux.sh
richm: yes, but if it does fail, it is wholly unexpected, and the `set -e` will make the error propagate up

test/mux.sh
richm: not with a daemonset, afaik?

test/mux.sh
richm: This is in the cleanup code - it may be that something failed when the node was unlabeled, so if the node is already unlabeled that's ok

test/mux.sh
stevekuznetsov: I would suggest you use `os::cmd::expect_success` then -- it will collapse the output into one or two lines and give nice output on failures. You'll get jUnit too, which makes programmatic identification of issues easier.

test/mux.sh
stevekuznetsov: Hm right -- I had to do this: https://github.com/openshift/origin-aggregated-logging/blob/master/test/cluster/rollout.sh#L47-L48

Might be more robust than this

PR #609, opened 2017-08-25
elasticsearch: readiness probe extension

wozniakjan:
Interim readiness probe until multi-role Elasticsearch topology from https://github.com/openshift/openshift-ansible/pull/5001 is introduced.

Proposed changes are based on this document: https://docs.google.com/document/d/1m9-LXr-UThs9WhRfkFRbwf1Lppxa5Opw_PYRY1wuDbY/

**List of implemented readiness probe checks:**
1) `localhost:9200/` returns `OK 200`
2) `localhost:9200/.searchguard.${DC_NAME}` returns `OK 200`
3) `localhost:9200/_template/${template}` returns `OK 200` for each `$template` in `$ES_HOME/index_templates/*.json`
4) `localhost:9200/.searchguard.$DC_NAME/_count` returns count of 5 documents in the response body

**Depends on this PR**
- [X] https://github.com/fabric8io/elasticsearch-cloud-kubernetes/pull/98

### DIFF
--- a/elasticsearch/probe/readiness.sh
+++ b/elasticsearch/probe/readiness.sh
@@ -19,20 +19,45 @@
 # TODO: try re-use code from ./run.sh
 ES_REST_BASEURL=https://localhost:9200
 EXPECTED_RESPONSE_CODE=200
+EXPECTED_SG_DOCUMENT_COUNT=5
 secret_dir=/etc/elasticsearch/secret
 max_time=${READINESS_PROBE_TIMEOUT:-30}
 
-response_code=$(curl -s -X HEAD \
-    --cacert $secret_dir/admin-ca \
-    --cert $secret_dir/admin-cert \
-    --key  $secret_dir/admin-key \
-    --max-time $max_time \
-    -w '%{response_code}' \
-    "${ES_REST_BASEURL}/")
 
-if [ ${response_code} == ${EXPECTED_RESPONSE_CODE} ]; then
-  exit 0
-else
-  echo "Elasticsearch node is not ready to accept HTTP requests yet [response code: ${response_code}]"
+function check_if_ready() {
+  path="$1"
+  err_msg="$2"
+  response_code=$(curl -s --head \
+      --cacert $secret_dir/admin-ca \
+      --cert $secret_dir/admin-cert \
+      --key  $secret_dir/admin-key \
+      --max-time $max_time \
+      -o /dev/null \
+      -w '%{response_code}' \
+      "${ES_REST_BASEURL}${path}")
+
+  if [ "${response_code}" != ${EXPECTED_RESPONSE_CODE} ]; then
+    echo "${err_msg} [response code: ${response_code}]"
+    exit 1
+  fi
+}
+
+check_if_ready "/" "Elasticsearch node is not ready to accept HTTP requests yet"
+check_if_ready "/.searchguard.$DC_NAME" "Searchguard index '.searchguard.$DC_NAME' is missing in ES cluster"
+sg_doc_count=$(curl -s --get \
+  --cacert $secret_dir/admin-ca \
+  --cert $secret_dir/admin-cert \
+  --key  $secret_dir/admin-key \
+  --max-time $max_time \
+  "${ES_REST_BASEURL}/.searchguard.$DC_NAME/_search?size=0" \
+  | python -c "import sys, json; print json.load(sys.stdin)['hits']['total']")
+
+if [ "$sg_doc_count" != $EXPECTED_SG_DOCUMENT_COUNT ]; then
+  echo "Incorrect SG document count, expected $EXPECTED_SG_DOCUMENT_COUNT [received doc count: ${sg_doc_count}]"
   exit 1
 fi
+
+for template_file in ${ES_HOME}/index_templates/*.json; do
+  template=`basename $template_file`
+  check_if_ready "/_template/$template" "Index template '$template' is missing in ES cluster"
+done
--- a/elasticsearch/Dockerfile
+++ b/elasticsearch/Dockerfile
@@ -6,7 +6,7 @@ EXPOSE 9200
 EXPOSE 9300
 USER 0
 
-ENV ES_CLOUD_K8S_VER=2.4.4 \
+ENV ES_CLOUD_K8S_VER=2.4.4_01 \
     ES_CONF=/usr/share/java/elasticsearch/config \
     ES_JAVA_OPTS="-Dmapper.allow_dots_in_name=true" \
     ES_HOME=/usr/share/java/elasticsearch \
@@ -22,7 +22,7 @@ ENV ES_CLOUD_K8S_VER=2.4.4 \
     RECOVER_AFTER_TIME=5m \
     RELEASE_STREAM=prod
 
-ARG ES_CLOUD_K8S_VER=2.4.4
+ARG ES_CLOUD_K8S_VER=2.4.4_01
 ARG OSE_ES_VER=2.4.4.13
 ARG ES_CLOUD_K8S_URL
 ARG OSE_ES_URL
--- a/elasticsearch/Dockerfile.centos7
+++ b/elasticsearch/Dockerfile.centos7
@@ -6,7 +6,7 @@ EXPOSE 9200
 EXPOSE 9300
 USER 0
 
-ENV ES_CLOUD_K8S_VER=2.4.4 \
+ENV ES_CLOUD_K8S_VER=2.4.4_01 \
     ES_CONF=/usr/share/java/elasticsearch/config \
     ES_JAVA_OPTS="-Dmapper.allow_dots_in_name=true" \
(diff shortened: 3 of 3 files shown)


elasticsearch/probe/readiness.sh
```diff
@@ -22,17 +22,27 @@ EXPECTED_RESPONSE_CODE=200
...
-if [ ${response_code} == ${EXPECTED_RESPONSE_CODE} ]; then
-  exit 0
-else
-  echo "Elasticsearch node is not ready to accept HTTP requests yet [response code: ${response_code}]"
-  exit 1
-fi
+function check_if_ready() {
+  path="$1"
+  err_msg="$2"
+  response_code=$(curl -s -X HEAD \
```
t0ffel: please use `--head` instead. ES5 doesn't like `HEAD`

wozniakjan:
[(3) Elasticsearch label discovery [epic-stability][logging]](https://trello.com/c/IxWno3R8/537-3-elasticsearch-label-discovery-epic-stabilitylogging)

elasticsearch/probe/readiness.sh
```diff
@@ -22,17 +22,27 @@ EXPECTED_RESPONSE_CODE=200
...
+      "${ES_REST_BASEURL}${path}")
+
+  if [ "${response_code}" != ${EXPECTED_RESPONSE_CODE} ]; then
+    echo "${err_msg} [response code: ${response_code}]"
+    exit 1
+  fi
+}
+
+check_if_ready "/" "Elasticsearch node is not ready to accept HTTP requests yet"
+check_if_ready "/.searchguard.$DC_NAME" "Searchguard index '.searchguard.$DC_NAME' is missing in ES cluster"
```
t0ffel: I thought @jcantrill wanted to check individual documents in that index.

wozniakjan: Yes, you are correct, thanks! I misread that part in the original document

lukas-vlcek: Minor: you might want to update the commit comment respectively (mention this test as well).

jcantrill:
@wozniakjan Regarding 'We need to ensure pod is still accessible by admin or by using admin certs' this is not intended to implemented in the probe but more a note we need to ensure that when a pod is in a status such that it is not ready, that we can still access the member of the node to obtain information (e.g. _cat/indices).  I believe the ACL backend registry will allow you to do this assuming you have access to the admin certs

elasticsearch/probe/readiness.sh
```diff
@@ -19,20 +19,45 @@
...
+sg_doc_count=$(curl -s --get \
+  --cacert $secret_dir/admin-ca \
+  --cert $secret_dir/admin-cert \
+  --key  $secret_dir/admin-key \
+  --max-time $max_time \
+  "${ES_REST_BASEURL}/.searchguard.$DC_NAME/_count" \
+  | python -c "import sys, json; print json.load(sys.stdin)['count']")
```
richm: `_count` is not supported in ES 5 afaik.  Instead, do a `_search?size=0` and look at the `total` field.

elasticsearch/probe/readiness.sh
```diff
@@ -19,20 +19,45 @@
...
+      --cacert $secret_dir/admin-ca \
+      --cert $secret_dir/admin-cert \
+      --key  $secret_dir/admin-key \
+      --max-time $max_time \
+      -o /dev/null \
+      -w '%{response_code}' \
+      "${ES_REST_BASEURL}${path}")
+
+  if [ "${response_code}" != ${EXPECTED_RESPONSE_CODE} ]; then
+    echo "${err_msg} [response code: ${response_code}]"
```
richm: - not sure what are the guidelines about readiness probes printing messages - I don't know what will happen, but I don't want to spew hundreds of error messages at the user about es not being available if it is a normal condition during startup
- if we do want these error messages, they should be more descriptive - the above message is meaningless to a user, and they don't know what else to check, or what actions to take to correct the situation

elasticsearch/probe/readiness.sh
```diff
@@ -19,20 +19,45 @@
...
+sg_doc_count=$(curl -s --get \
+  --cacert $secret_dir/admin-ca \
+  --cert $secret_dir/admin-cert \
+  --key  $secret_dir/admin-key \
+  --max-time $max_time \
+  "${ES_REST_BASEURL}/.searchguard.$DC_NAME/_count" \
+  | python -c "import sys, json; print json.load(sys.stdin)['count']")
+
+if [ "$sg_doc_count" != $EXPECTED_SG_DOCUMENT_COUNT ]; then
+  echo "Incorrect SG document count, expected $EXPECTED_SG_DOCUMENT_COUNT [received doc count: ${sg_doc_count}]"
```
richm: see above comment about errors

elasticsearch/probe/readiness.sh
wozniakjan: > not sure what are the guidelines about readiness probes printing messages - I don't know what will happen, but I don't want to spew hundreds of error messages at the user about es not being available if it is a normal condition during startup

maximal number of error messages this could produce is 9 (but it is usually around 0-3, given the probe initial delay is 10s and period is 5s)
- ES not available at all
- search guard index not available
- search guard document count not as expected (5x)
- index template missing (2x)

And wouldn't it be useful to know what went awry when the pod doesn't start due to the readiness probe? Or should it rather be logged elsewhere?

elasticsearch/probe/readiness.sh
wozniakjan: appears to be supported: https://www.elastic.co/guide/en/elasticsearch/reference/5.6/search-count.html

lukas-vlcek: It seems `_count` [has been removed](https://github.com/elastic/elasticsearch/pull/14166) for Java API only in the end.

wozniakjan: some information I was able to find on this topic is https://github.com/elastic/elasticsearch/pull/9296. Given that internally, they should be equivalent but the Java API has removed `count` already, I will go with what Rich suggested and make a fix

jcantrill:
[test]

elasticsearch/probe/readiness.sh
richm: Yes, it would be useful to know.  Ideally, the user would only see an error message for the _last_ failed probe, instead of for _all_ of them.

wozniakjan: That is exactly what it does per readiness probe instance

wozniakjan:
@jcantrill I updated versions of elasticsearch cloud kubernetes plugin and tested again. All seems to be ready for merge and QE tests.

wozniakjan:
[test]

wozniakjan:
@jcantrill tests are passing, my local tests also succeeded, do you think we can merge it?

jcantrill:
[merge]

portante:
I have some reservations about this probe.  This relies on the cluster being operational for the probes to work.  If the searchguard index is sharded across all nodes of the cluster, and one of the nodes is not available, then will this probe work?

Fundamentally, readiness probes should not relied on the cluster-wide state of Elasticsearch to determine if one pod is ready to join the service.

wozniakjan:
> I have some reservations about this probe. This relies on the cluster being operational for the probes to work. If the searchguard index is sharded across all nodes of the cluster, and one of the nodes is not available, then will this probe work?

Each probe verifies only its own searchguard index. If fewer than half of ES pods aren't available, rest of the service remains operational. Those malfunctioning ES pods will be removed from ES service and the service will loadbalance only across working pods forming a cluster.

> Fundamentally, readiness probes should not relied on the cluster-wide state of Elasticsearch to determine if one pod is ready to join the service.

I am afraid there is no way around it. It is ES internal mechanism and a single pod is not ready until it negotiates with a subset of his peers to form an operational cluster according to its configuration.

richm:
How can we test this on a real, live, multi-node cluster, so we can try to reproduce some of these scenarios @portante is describing?

portante:
> Each probe verifies only its own searchguard index.

Are not indices in Elasticsearch maintained cluster wide?  Don't they have state that is managed across all nodes in the cluster?  If that is not the case, share where that is documented to be otherwise.

Assuming it is cluster-wide state, making it part of a pod's readiness probe does not make sense.

There is a way to drop per-ES pod searchguard indices which would make this particular issue go away.

But it still holds a that a readiness probe is per pod, and as such, should not rely on service or ES cluster state to determine its readiness to join the service.

wozniakjan:
@richm locally I tested the searchguard index miss by creating an image that had this commented out and deployed cluster with 3 nodes. https://github.com/openshift/origin-aggregated-logging/blob/ca917ff85ea4df05ee9cb7a2e12179c8b434af29/elasticsearch/run.sh#L116 
Then another image with this line back and redeploy 2 out of 3 images. Two images formed an operational cluster, one was waiting for the probe to succeed and was, therefore, dropped out of the service

PR #622, opened 2017-08-29
Use UTF-8 instead of ASCII-8BIT

### LINKED ISSUE #621: Issues with syslog character encoding in fluentd
As context, instead of using the built-in Elasticsearch instance, I'm pushing messages to a RabbitMQ instance and having something else pick them up. This involves a conversion to JSON. Unfortunately, fluentd's `in_tail` plugin will, by default, [force the encoding of messages it reads to `ASCII-8BIT`][ASCII-8BIT]. This is mostly fine, except that the origin-node service occasionally logs messages like this:
```
Aug 29 17:19:10 ip-192-168-32-82 origin-node: I0829 17:19:10.500708   14033 docker_manager.go:1373] Container "e8a6b32b1e4d101ff6fd7950de8eb425e16253005cf53924b9d3480c1b972e2a dev/config-1-deploy" exited after 883.317µs
```
`µ` is encoding in UTF-8 as the two-byte sequence `0xC2B5`. When fluentd attempts to convert a record containing this character to JSON, it fails with the following error:
```
Encoding::UndefinedConversionError: "\xC2" from ASCII-8BIT to UTF-8
```
While this is surfacing due to my nonstandard use of the container, it's possible that it might surface in sanctioned use later. Additionally, since we're logging UTF-8 into syslog (and, as a side note, containers might be as well), we should probably have fluentd treat these files as UTF-8.

The good news is that this is pretty easy to fix. By setting `encoding` to `UTF-8` in the various `in_tail` configuration sections, fluentd will treat the files as `UTF-8`, and the problem goes away entirely.

[ASCII-8BIT]: https://docs.fluentd.org/v0.12/articles/in_tail#encoding-fromencoding

elyscape:
The various origin services log messages including the character `µ`, which is encoded in UTF-8 as the 2-byte sequence `0xC2B5`. Both the [`in_tail`][in_tail] and [`in_systemd`][in_systemd] inputs treat messages as ASCII-8BIT by default, with only the `in_tail` plugin providing a way to override this. As a result, if one of these messages (or any other message with UTF-8 characters) is read in and later needs to be converted to UTF-8 (e.g. as part of being marshaled to JSON), the conversion will fail with an `UndefinedConversionError`. By forcing the message to be treated as UTF-8, we avoid this issue and other potential issues arising from the use of ASCII-8BIT.

Fixes #621.

[in_tail]: https://docs.fluentd.org/v0.12/articles/in_tail#encoding-fromencoding
[in_systemd]: https://github.com/openshift/origin-aggregated-logging/issues/621#issuecomment-325780592

### DIFF
--- a/test/utf8-characters.sh
+++ b/test/utf8-characters.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+
+# This logging test ensures that a log message containing
+# extended UTF-8 characters will be correctly processed by
+# Fluentd and appear in Elasticsearch.
+
+source "$(dirname "${BASH_SOURCE[0]}" )/../hack/lib/init.sh"
+source "${OS_O_A_L_DIR}/hack/testing/util.sh"
+os::util::environment::use_sudo
+
+os::test::junit::declare_suite_start "test/utf8-characters"
+
+if [ -n "${DEBUG:-}" ] ; then
+    set -x
+fi
+
+cleanup() {
+    local return_code="$?"
+    set +e
+    if [ $return_code = 0 ] ; then
+        mycmd=os::log::info
+    else
+        mycmd=os::log::error
+    fi
+    $mycmd "utf8-characters test finished at $( date )"
+    # this will call declare_test_end, suite_end, etc.
+    os::test::junit::reconcile_output
+    exit $return_code
+}
+trap "cleanup" EXIT
+
+os::log::info "Starting utf8-characters test at $( date )"
+
+message_uuid="$( uuidgen )"
+message="$(printf '%s-\xC2\xB5' "$message_uuid" )"
+logger -p local6.info -t "$message_uuid" "$message"
+
+wait_for_fluentd_ready
+wait_for_fluentd_to_catch_up
+
+es_pod="$( get_running_pod es )"
+
+os::log::info "Checking that message was successfully processed..."
+os::cmd::expect_success "curl_es $es_pod /.operations.*/_search?q=systemd.u.SYSLOG_IDENTIFIER:$message_uuid | \
+                         python $OS_O_A_L_DIR/hack/testing/test-utf8-characters.py $message $message_uuid"
--- a/fluentd/Dockerfile.centos7
+++ b/fluentd/Dockerfile.centos7
@@ -35,6 +35,7 @@ RUN mkdir -p ${HOME} && mkdir -p /etc/fluent/plugin && \
      'elasticsearch:<5' \
       fluentd:${FLUENTD_VERSION} \
       fluent-plugin-kubernetes_metadata_filter \
+      'fluent-plugin-record-modifier:<1.0.0' \
       fluent-plugin-rewrite-tag-filter \
       fluent-plugin-secure-forward \
      'fluent-plugin-systemd:<0.1.0' \
--- a/fluentd/configs.d/openshift/filter-pre-force-utf8.conf
+++ b/fluentd/configs.d/openshift/filter-pre-force-utf8.conf
@@ -1,0 +1,4 @@
+<filter **>
+  @type record_modifier
+  char_encoding utf-8
+</filter>
--- a/hack/testing/test-utf8-characters.py
+++ b/hack/testing/test-utf8-characters.py
@@ -1,0 +1,31 @@
+import sys
+import json
+
+obj = json.loads(sys.stdin.read())
+
+message_uuid = sys.argv[1]
+message = sys.argv[2]
+
+for dd in obj['hits']['hits']:
+    if dd['_score'] < 1.0:
+        print "ignoring spurious hit"
+        continue
+    if not dd['_source']['message'].startswith(message):
+        print 'Error: message field does not start with [%s]: [%s]' % (message, dd['_source']['message'])
+        sys.exit(1)
+    if not 'systemd' in dd['_source']:
+        print 'Error: systemd field not in record: [%s]' % json.dumps(dd['_source'])
+        sys.exit(1)
+    if not 'u' in dd['_source']['systemd']:
+        print 'Error: systemd.u field not in record: [%s]' % json.dumps(dd['_source'])
+        sys.exit(1)
+    if not 'SYSLOG_IDENTIFIER' in dd['_source']['systemd']['u']:
+        print 'Error: systemd.u.SYSLOG_IDENTIFIER field not in record: [%s]' % json.dumps(dd['_source'])
+        sys.exit(1)
+    syslog_identifier = dd['_source']['systemd']['u']['SYSLOG_IDENTIFIER']
+    if syslog_identifier == message_uuid:
+        print 'Error: field systemd.u.SYSLOG_IDENTIFIER does not have expected value [%s]: [%s]' % (message_uuid, str(syslog_identifier))
+        sys.exit(1)
+
+print 'Success: record contains all of the expected fields/values'
+sys.exit(0)
--- a/hack/testing/test-utf8-characters.sh
+++ b/hack/testing/test-utf8-characters.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+source "$(dirname "${BASH_SOURCE[0]}" )/../lib/init.sh"
+
+exec ${OS_O_A_L_DIR}/test/utf8-characters.sh


jcantrill:
@richm does this maybe need to happen elsewhere as well.  We had file conversion issue with filebuffering did we not?

[test]

elyscape:
On further investigation, it looks like the journald input plugin doesn't force things to UTF-8 either. From an IRB session using the underlying library:
```
irb(main):028:0> puts j.current_entry.message
[1.02304307s] [8.72µs] Self-link added
=> nil
irb(main):029:0> j.current_entry.message
=> "[1.02304307s] [8.72\xC2\xB5s] Self-link added"
irb(main):030:0> j.current_entry.message.encoding
=> #<Encoding:ASCII-8BIT>
```
So we should probably use a filter instead.

fluentd/generate_throttle_configs.rb
```diff
@@ -108,6 +109,7 @@ def create_default_docker(excluded)
   @label @INGRESS
   path /var/log/containers/*.log
   pos_file /var/log/es-containers.log.pos
+  encoding UTF-8
```
richm: We have the same utf-8 problem with container log files as well as /var/log/messages?

elyscape: If they log UTF-8 messages, then yes.

richm:
Please add a test to `origin-aggregated-logging/test` that will test utf-8 input - test should write utf-8 to the various input sources, and verify that the correct value can be read from Elasticsearch

elyscape:
@richm I have pushed a test. Let me know what you think.

fluentd/configs.d/user/fluent.conf
```diff
@@ -17,6 +17,7 @@
 <label @INGRESS>
 ## filters
   @include configs.d/openshift/filter-pre-*.conf
+  @include configs.d/openshift/filter-force-utf8.conf
```
richm: This fluent.conf isn't used.  The source is in openshift-ansible: https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging_fluentd/templates/fluent.conf.j2

However, you could make this work by creating this filter as a new file e.g. `configs.d/openshift/filter-pre-force-utf8.conf` which will be picked up by the wildcard match `@include configs.d/openshift/filter-pre-*.conf`

elyscape: Done.

richm:
[test]

elyscape:
The testing appears to be failing in an unrelated test ([one of these checks](https://github.com/elyscape/origin-aggregated-logging/blob/349cd7a36ec58fd09ba8e449242581018a581956/test/viaq-data-model.sh#L135-L139)) but, for good measure, I went ahead and rebased onto master.

jcantrill:
[test]

richm:
yeah - test fails in es-copy - not sure if that is related to the recent work by @nhosoi

nhosoi:
@richm 
> yeah - test fails in es-copy - not sure if that is related to the recent work by @nhosoi
The test/es-copy.sh in the master branch is supposed to work as is.  My further changes on the test script is not merged yet.  Actually, it's in pr#577...  (I'm going to run the CI test with this pr applied in my local test env.)

richm:
Should Elasticsearch store UTF-8 or UC2?
How does it work without this patch?

nhosoi:
I think Elasticsearch currently expects UTF-8 (other encodings are rejected).

With or without this pr#622, I believe the stored logs in Elasticsearch look the same -- in the \u format.

Without pr#622
```
"Però non è la città di Barnabù" 
==>
                    "message": "Per\u00f2 non \u00e8 la citt\u00e0 di Barnab\u00f9", 
```

elyscape:
>I was hoping this pr#622 may preserve log messages in multi-byte UTF-8 in the Elasticsearch as is, but I could not see the expected side-effect. :)

That's ultimately a result of whatever library Elasticsearch is using to serialize the document to JSON.

>I think Elasticsearch currently expects UTF-8 (other encodings are rejected).

JSON is by definition only valid when encoded in UTF-8.

elyscape:
@portante The benefit to doing it as a separate step is that it means that, in non-standard use cases that use a different output plugin (e.g. the amqp plugin), the encoding is handled and won't cause problems like in #621.

elyscape:
Earlier it was failing in the viaq-data-model.sh tests. Now it's failing in the es-copy.sh tests. I can't figure out what the issue is, and would welcome any ideas others might have.

nhosoi:
Definitely, it's a test problem.. :(  Could you please rebase with the latest master and force to push to your repo again?

elyscape:
@nhosoi Done.

richm:
[test]

nhosoi:
Thanks, @elyscape !!  Test suites passed!!

richm:
Sorry this is taking so long to merge.  It is a seemingly small change, but it affects every log record, coming in from every source, so I want to be very careful.
For example, in the mux case, where we have a fluentd acting as a forwarder for secure_forward messages, which may come from inside the cluster as well as outside the cluster, is it safe to re-process messages to utf-8 that may already be in utf-8 format?  Or do we have to have some logic to detect if the log record is already properly encoded in utf-8?

elyscape:
>For example, in the mux case, where we have a fluentd acting as a forwarder for secure_forward messages, which may come from inside the cluster as well as outside the cluster, is it safe to re-process messages to utf-8 that may already be in utf-8 format?

This change doesn't actually reprocess them in UTF-8. It basically just tells Ruby, "hey, interpret this line as a UTF-8 bytestream". In theory, this could cause issues if the logs contain data that isn't valid UTF-8. In practice, that was probably already an unsupported scenario.

Also, based on a cursory test, it seems to work fine. When using the JSON log driver, the Docker daemon logs in UTF-8, and converts invalid UTF-8 characters to the Unicode replacement character (i.e. they are replaced with `\ufffd` in the JSON log file). When using the journald log driver, it appears to pass it unmodified. That being said, it also looks like `JSON#dump` will just kinda trust the string if the encoding is declared to be UTF-8 and pass things on without crashing, which seems like basically what we want anyway:
```
irb(main):041:0> j.current_entry.message
=> "\xC2\xB5"
irb(main):042:0> j.current_entry.message.force_encoding(Encoding::UTF_8)
=> "\u00B5"
irb(main):043:0> puts JSON.dump(j.current_entry.message.force_encoding(Encoding::UTF_8))
"µ"
=> nil
irb(main):044:0> puts JSON.dump(j.current_entry.message)
Encoding::UndefinedConversionError: "\xC2" from ASCII-8BIT to UTF-8
	from /usr/share/gems/gems/json-1.7.7/lib/json/common.rb:224:in `encode'
	from /usr/share/gems/gems/json-1.7.7/lib/json/common.rb:224:in `generate'
	from /usr/share/gems/gems/json-1.7.7/lib/json/common.rb:224:in `generate'
	from /usr/share/gems/gems/json-1.7.7/lib/json/common.rb:395:in `dump'
	from (irb):44
	from /usr/bin/irb:12:in `<main>'
irb(main):045:0> j.move_next
=> true
irb(main):046:0> j.current_entry.message
=> "\xC5"
irb(main):047:0> j.current_entry.message.force_encoding(Encoding::UTF_8)
=> "\xC5"
irb(main):048:0> puts JSON.dump(j.current_entry.message.force_encoding(Encoding::UTF_8))
"?"
=> nil
irb(main):049:0> puts JSON.dump(j.current_entry.message)
Encoding::UndefinedConversionError: "\xC5" from ASCII-8BIT to UTF-8
	from /usr/share/gems/gems/json-1.7.7/lib/json/common.rb:224:in `encode'
	from /usr/share/gems/gems/json-1.7.7/lib/json/common.rb:224:in `generate'
	from /usr/share/gems/gems/json-1.7.7/lib/json/common.rb:224:in `generate'
	from /usr/share/gems/gems/json-1.7.7/lib/json/common.rb:395:in `dump'
	from (irb):49
	from /usr/bin/irb:12:in `<main>'
```
So this should be fine.

>Or do we have to have some logic to detect if the log record is already properly encoded in utf-8?

As described above, this shouldn't be necessary, though some further testing is probably necessary.

richm:
How can we reproduce https://github.com/openshift/origin-aggregated-logging/issues/621 ?  I don't think I've ever seen this error before.

richm:
That is, do you see an exception and stack trace in the fluentd pod log with `Encoding::UndefinedConversionError`?

elyscape:
It doesn't occur under normal operation[1], but it can occur if you add different outputs. I'm using [the `fluent-plugin-amqp` output][amqp], which [uses the built-in `json` gem][json].

[1] Fluentd [normally uses][fluentd-json] either the `yajl-ruby` gem or the `oj` gem to serialize to JSON, neither of which choke on invalid UTF-8 byte sequences.

[fluentd-json]: https://github.com/fluent/fluentd/blob/v0.12.40/lib/fluent/formatter.rb#L143-L150
[amqp]: https://github.com/giraffi/fluent-plugin-amqp/tree/v0.9.3
[json]: https://github.com/giraffi/fluent-plugin-amqp/blob/v0.9.3/lib/fluent/plugin/out_amqp.rb#L93

nhosoi:
Thanks for your patience, @elyscape. I think your pr is almost ready to merge.  

Some time back, pull/664 was merged, which switched the path of util.sh.  Could you please run "git rebase master" one more time and update the path to "${OS_O_A_L_DIR}/hack/testing/util.sh"?  Thanks!

elyscape:
I guess I can't retest. In any event, I have rebased this.

test/utf8-characters.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+
+# This logging test ensures that a log message containing
+# extended UTF-8 characters will be correctly processed by
+# Fluentd and appear in Elasticsearch.
+
+source "$(dirname "${BASH_SOURCE[0]}" )/../hack/lib/init.sh"
+source "${OS_O_A_L_DIR}/deployer/scripts/util.sh"
```
nhosoi: util.sh has been moved to ${OS_O_A_L_DIR}/hack/testing in the latest master and release-3.6 branch.

elyscape: Good catch. Fixed.

richm:
@elyscape This is severely hurting performance.  It makes the processing pipeline take twice as long.  I'm going to propose making this an optional feature with the default off.  I think the only way we can do this for all messages, and not negatively impact performance, is to do the processing _at the source only_.  That is, have `in_tail` do the checking and utf8 conversion, and implement something similar for `in_systemd`.  For `secure_forward` or other inputs, we will just assume that utf8 conversion has already been done.

What do you think?

@portante @mffiedler @jcantrill

elyscape:
@richm That makes sense to me. Sorry I didn't catch that earlier.

portante:
@richm, seems reasonable to have the default to be `off` for this feature.  It cases where performance is not an issue, one can enable it easily enough.

Fixing this in the `in_tail` and `in_systemd` input plugins also seems reasonable, though I think the ultimate solution is to replace fluentd as a collector with a much lighter-weight technology.  Not sure how much effort should be spent on fixing those plugins to handle this.

PR #623, opened 2017-08-29
move formatting to viaq; use viaq for index names

richm:
Move systemd and other expensive processing from record_transformer
embedded ruby code to the viaq plugin.
Create index names in the viaq plugin and switch to the use of
static elasticsearch plugin rather than dynamic.
@jcantrill @portante @nhosoi ptal
[test]


richm:
[merge]

nhosoi:
Looks good to me, too.

richm:
[merge] flake?

fluentd/configs.d/openshift/filter-viaq-data-model.conf
```diff
@@ -9,4 +9,32 @@
...
+    tag "kubernetes.var.log.containers**"
+    type k8s_json_file
+    remove_keys log,stream
+  </formatter>
+  <elasticsearch_index_name>
+    tag "journal.system** system.var.log** **_default_** **_openshift_** **_openshift-infra_** mux.ops"
+    name_type operations_full
+  </elasticsearch_index_name>
+  <elasticsearch_index_name>
+    tag "**"
```
portante: Will this also match the same tags as line 33?  In other words, how to records with tags match on line 33 avoid be processed again with this filter due to also matching on line 37?

richm: no - the way it works is that once it matches, there is no further processing - so once the match is hit on line 33, it will not go to line 37

richm:
@nhosoi updated - ptal

nhosoi:
Thanks, Rich!  I'm running the test with setting ENABLE_ES_INDEX_NAME in fluentd/run.sh.
Note: you already have my approval. :)

PR #629, opened 2017-09-01
Replacing deprecated excludeN with <exclude> section.

nhosoi:
This patch eliminates the following warning from the fluentd log:

` [warn]: 'exclude1' parameter is deprecated: Use <exclude> section`


richm:
[merge]

PR #617, opened 2017-08-29
refactor es_util to pass query as arg. introduce es_acl

jcantrill:
This PR:

* Makes it so you can use es_util from 'oc exec' by allowing you to pass the query as an arg
* Introduces es_acl to facilitate retrieving the ACL documents.

PR #632, opened 2017-09-01
fix kibana_dashboard test from es_util change

jcantrill:
This PR fixes es_util to accept additional args for cases like kibana_dashboard test.  Broken by #617

### DIFF
--- a/elasticsearch/utils/es_util
+++ b/elasticsearch/utils/es_util
@@ -20,6 +20,9 @@
 set -euo pipefail
 source es_util_env
 
+extra_args=("")
+index=0
+
 while (($#))
 do
 case $1 in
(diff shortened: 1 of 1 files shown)


jcantrill:
[test]

elasticsearch/utils/es_util
```diff
@@ -20,14 +20,18 @@
...
+index=0
+
 while (($#))
 do
 case $1 in
     --query=*)
       QUERY=${1#*=}
       ;;
     *)
-      echo Ignoring unknown argument $1
```
richm: how about just `break` here, so the remaining args are left in `"$@"`?

richm: or - have a case `--*) echo Error: unknown argument $1 ; exit 1 ;;`
then `*) break ;;`

jcantrill: What if we decide to add another switch?

richm: @jcantrill just add it below `--query` and above `--*)`.  This assumes that all of the `--name` arguments will come before the other positional arguments.

If we think we are going to need more/better processing, then we should use `getopt` or something like that, rather than "roll-your-own" argument processing.

PR #636, opened 2017-09-05
Proposal to gather docker events

jcantrill:
This PR introduces a proposal to gather docker events in support of https://trello.com/c/qxoFsKz9/ 

cc @mpbarrett @smarterclayton @pweil-

### DIFF
--- a/docs/proposals/docker-event-collection.md
+++ b/docs/proposals/docker-event-collection.md
@@ -1,15 +1,15 @@
 # Docker Event Collection
 ## Overview
-Docker events are not collected or stored by the Openshift cluster.  Certain customers are [required](https://trello.com/c/qxoFsKz9/) to audit these events (e.g. security).  This document is the architecture and design proposal for gathering Docker events for persistence.
+Docker events are not collected or stored by the Openshift cluster.  Certain customers are [required](https://trello.com/c/qxoFsKz9/) to audit these events (e.g. security).  This document is the architecture and design proposal for gathering Docker events for persistence as log entries interleaved with other operational logs.
 
 ## Requirements
 ### Security 
-Docker events shall only be visible by users in a cluster admin role.
+Docker event log entries shall have the same visibility as other operational log entries.
 ### Storage and Retention
 Docker events shall be stored for the same retention period as other log event messages.  The default curation period is 31 days.  The default indices are the operations indices
 
 # Proposal
-Openshift is deployed with [ProjectAtomic Docker](https://github.com/projectatomic/docker/blob/docker-1.12.6/api/server/middleware/audit_linux.go) that enhances audit logging of container events.  These events are logged to the audit file on the node.  The log collector for the logging stack already has access to the file system and will collect docker events from this log.  
+Openshift is deployed with [ProjectAtomic Docker](https://github.com/projectatomic/docker/blob/docker-1.12.6/api/server/middleware/audit_linux.go) that enhances audit logging of container events.  These events are logged to the audit file on the node.  The log collector for the logging stack already has access to the file system and will collect docker events from this file.  
 
 **NOTE:** Only Docker Events will be parsed from the audit log; all others will be dropped.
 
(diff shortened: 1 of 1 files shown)


jcantrill:
[(8) [SCCFSI] docker events and API calls need to be logged[logging][epic-ois-agl-exp]](https://trello.com/c/qxoFsKz9/497-8-sccfsi-docker-events-and-api-calls-need-to-be-loggedloggingepic-ois-agl-exp)

docs/proposals/docker-event-collection.md
```diff
@@ -0,0 +1,86 @@
+# Docker Event Collection
+## Overview
+Docker events are not collected or stored by the Openshift cluster.  Certain customers are [required](https://trello.com/c/qxoFsKz9/) to audit these events (e.g. security).  This document is the architecture and design proposal for gathering Docker events for persistence.
```
portante: Consider changing last few words to, "for gathering Docker events for persistence as log entries interleaved with other operational logs."

docs/proposals/docker-event-collection.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Requirements
+### Security 
+Docker events shall only be visible by users in a cluster admin role.
```
portante: s/Docker events/Docker event log entries/, and s/cluster admin role/just as other operational log entries are visible/.

docs/proposals/docker-event-collection.md
```diff
@@ -0,0 +1,86 @@
...
+### Storage and Retention
+Docker events shall be stored for the same retention period as other log event messages.  The default curation period is 31 days.  The default indices are the operations indices
+
+# Proposal
+Openshift is deployed with [ProjectAtomic Docker](https://github.com/projectatomic/docker/blob/docker-1.12.6/api/server/middleware/audit_linux.go) that enhances audit logging of container events.  These events are logged to the audit file on the node.  The log collector for the logging stack already has access to the file system and will collect docker events from this log.
```
portante: s/from this log/from this file/.

docs/proposals/docker-event-collection.md
```diff
@@ -0,0 +1,86 @@
...
+    exe: container-entrypoint
+    hostname: logging-kibana-1-xxmr3
+```
+## Fluentd Modifications
+* Modify fluentd configuration to conditionally include transformation plugin from the CDM
+
+## Dependencies
+ * [ProjectAtomic Docker](https://github.com/projectatomic/docker) version 1.12.6 - Docker fork with patch for enhanced auditing
+
+# Alternative Considerations
```
portante: Can you provide a description of why these paths were not chosen?

jcantrill: added reasons

docs/proposals/docker-event-collection.md
```diff
@@ -0,0 +1,90 @@
...
+addr=? terminal=? res=success'
+
+```
+
+The additional fields:
+
+|Field|Type|Definition| Example |
+|----|-----|-|------|
+|op| string|The operation or action being performed|exec
+|reason | string|The constant value 'api'|api
```
wozniakjan: given the `reason` is constant, would we maybe want this to be implicit? Is it possible that this will carry other values in future or is there another reason to have this explicitly stated?

jcantrill: I don't actually know if this value will ever change.  I presume you could have some event generated that is not related to an API call and maybe this is the source of this field.  Do you think maybe it would be better to simply not identify this field in the proposal but to persist it to storage?  Or maybe for the sake of disk savings we should exclude it.  I don't know what value it really provides.

wozniakjan: > I presume you could have some event generated that is not related to an API call and maybe this is the source of this field

Then I think you were correct to include it here if it could potentially change in future and provide some unique information.

tkatarki:
LGTM.

jcantrill:
[merge]

PR #651, opened 2017-09-12
try to fix test flakes by using more precise searches

richm:
Use more precise searches to avoid test flakes caused by Elasticsearch
searches that return too many matches.
For project searches that look for a Kibana generated log message, use
a query with "match_phrase" to look for a message field containing
the phrase "GET /$uuid 404 " as opposed to just looking for $uuid in
the message field which can have multiple hits.
For operations searches, use a "term" query in order to exactly match
the $uuid_ops value.
@jcantrill @nhosoi ptal
[test]


richm:
[test]

richm:
[merge]

PR #659, opened 2017-09-13
update to viaq-data-model 0.0.11 and add script to grab latest data model

richm:
@jcantrill @jwozniak ptal
[test]


richm:
[merge]

PR #660, opened 2017-09-13
port es-copy to the new hack/lib tests

richm:
does not currently work with mux client mode, so skip in that case

(cherry picked from commit f35878ba83e547bc5d8e0badd00ab97ae527d642)
@nhosoi @jcantrill ptal
[test]


richm:
[merge]

PR #666, opened 2017-09-18
increase rsyslog rate limits for json-file; add debugging for json-file

richm:
When using json-file on 3.6, the number of messages exceeds the rsyslog
imjournal limit, so test messages are dropped.  This increases the
limits for rsyslog.
This also adds some debugging when using json-file.
@jcantrill @nhosoi ptal
[test]


richm:
[test]

richm:
[merge]

PR #669, opened 2017-09-19
Backporting fluentd file buffer to release 1.4

nhosoi:
Bug 1477513 - [3.4] Data loss of logs can occur if fluentd pod is terminated/restarted when Elasticsearch is unavailable

Backporting "Impl fluentd file buffer [1]" with dependent patches [2] and [3]

[1] - commit 6252ba1d1ae2695a51795339478b2f0a876bced3
[2] - allow flush_interval to be tunable; name tunable depending on output
       (commit a9a14f5f79d05cf4a14ef188a5c9d937063b1760)
[3] - Introducing following environment variables with the initial value for fluentd.
       (commit 3818101b713d92f11a1ff33fe44f6d735fda2eda)

### DIFF
--- a/deployer/templates/fluentd.yaml
+++ b/deployer/templates/fluentd.yaml
@@ -308,3 +308,11 @@ parameters:
   description: 'Read the journal starting from the very beginning (false|true).'
   name: JOURNAL_READ_FROM_HEAD
   value: "false"
+-
+  description: 'Fluentd buffer queue limit (0.12 only)'
+  name: BUFFER_QUEUE_LIMIT
+  value: "1024"
+-
+  description: 'Fluentd buffer size limit'
+  name: BUFFER_SIZE_LIMIT
+  value: "1048576"
--- a/fluentd/Dockerfile
+++ b/fluentd/Dockerfile
@@ -22,6 +22,7 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
       ruby-devel \
       libcurl-devel \
       make \
+      bc \
       iproute && \
     yum clean all
 RUN mkdir -p ${HOME} && \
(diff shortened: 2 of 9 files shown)


nhosoi:
@richm 
> does this require any ansible changes to create the persistent volume? 

I don't think so.  The persistent volume settings are required just for MUX.  The fluentd piggybacks on hostmount. 

> If not, then we still need to document what to do in order to get a persistent log buffer

Is that for upgrading?  We can build and provide a new fluentd image.  Once it's deployed, I'd think the file buffer would be enabled.

richm:
@nhosoi ok - then there is nothing to do

nhosoi:
@richm 
I manually tested the patch on 3.4 where I did not see any problem with the file buffer enabled.  I've just noticed the fluentd version for 3.4 are set as follows, respectively.
`FLUENTD_VERSION=0.12.29`

What confuses me is we removed fluentd-0.12.29-fix-UTF-error.patch from fluentd-0.12.39-2.el7.  But it looks there was no problem with fluentd 0.12.29...

My question is should we set FLUENTD_VERSION to 0.12.39 as done in release-3.6 Dockerfile?
`FLUENTD_VERSION=0.12.39`

richm:
@nhosoi I am surprised that it worked downstream if it were using 0.12.29 without the utf-8 fix - can you investigate?  and yes, we'll need to use the latest fluentd 0.12.39-2

nhosoi:
> I am surprised that it worked downstream if it were using 0.12.29 without the utf-8 fix - can you investigate? and yes, we'll need to use the latest fluentd 0.12.39-2

It turned out I was testing with the upstream fluentd which did not have the UTF-8 patch.  That's why I did not see the file buffer problem.

I also tested the (manual) downstream builds with fluentd-0.12.39-2 on 3.4 as well as 3.5 which did not cause any problem with the file buffer enabled.

PR #670, opened 2017-09-19
Backporting fluentd file buffer to release 1.5

nhosoi:
Bug 1477515 - [3.5] Data loss of logs can occur if fluentd pod is terminated/restarted when Elasticsearch is unavailable

Backporting "Impl fluentd file buffer [1]" with dependent patches [2] and [3]

[1] - commit 6252ba1d1ae2695a51795339478b2f0a876bced3
[2] - allow flush_interval to be tunable; name tunable depending on output
(commit a9a14f5f79d05cf4a14ef188a5c9d937063b1760)
[3] - Introducing following environment variables with the initial value for fluentd.
(commit 3818101b713d92f11a1ff33fe44f6d735fda2eda)

### DIFF
--- a/fluentd/Dockerfile
+++ b/fluentd/Dockerfile
@@ -22,6 +22,7 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
       ruby-devel \
       libcurl-devel \
       make \
+      bc \
       iproute && \
     yum clean all
 RUN mkdir -p ${HOME} && \
--- a/fluentd/Dockerfile
+++ b/fluentd/Dockerfile
@@ -5,7 +5,7 @@ MAINTAINER OpenShift Development <dev@lists.openshift.redhat.com>
 ENV HOME=/opt/app-root/src \
   PATH=/opt/app-root/src/bin:/opt/app-root/bin:$PATH \
   RUBY_VERSION=2.0 \
-  FLUENTD_VERSION=0.12.31 \
+  FLUENTD_VERSION=0.12.39 \
   GEM_HOME=/opt/app-root/src
 
 LABEL io.k8s.description="Fluentd container for collecting of docker container logs" \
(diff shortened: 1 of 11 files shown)


nhosoi:
@richm 
> does this require any ansible changes to create the persistent volume? 

I don't think so.  The persistent volume settings are required just for MUX.  The fluentd piggybacks on hostmount. 

> If not, then we still need to document what to do in order to get a persistent log buffer

Is that for upgrading?  We can build and provide a new fluentd image.  Once it's deployed, I'd think the file buffer would be enabled.

richm:
@nhosoi ok - then there is nothing to do

nhosoi:
@richm 
I manually tested the patch on 3.5 where I did not see any problem with the file buffer enabled.  I've just noticed the fluentd version for 3.5 are set as follows.
` FLUENTD_VERSION=0.12.31`

What confuses me is we removed fluentd-0.12.29-fix-UTF-error.patch from fluentd-0.12.39-2.el7.  But it looks there was no problem with fluentd 0.12.31...

My question is should we set FLUENTD_VERSION to 0.12.39 as done in release-3.6 Dockerfile?
    `FLUENTD_VERSION=0.12.39`

richm:
@nhosoi I am surprised that it worked downstream if it were using 0.12.31 without the utf-8 fix - can you investigate?  and yes, we'll need to use the latest fluentd 0.12.39-2

nhosoi:
@richm, I'm also puzzled...

I'm on flexy where I deployed, e.g., 3.5 from the 3.5 image.  I applied the file buffer patch and built logging-fluentd image and restarted fluentd.

In the fluentd pod, it shows the expected version.
```
sh-4.2# /opt/app-root/src/bin/fluentd --version
fluentd 0.12.31
```
But the file in the fluentd pod has no patch applied as follows (if fluentd-0.12.29-fix-UTF-error.patch is applied, we'd see UTF-8 instead of ASCII8-BIT).  No wonder I did not run into the problem in my file buffer tests...
"gems/fluentd-0.12.31/lib/fluent/engine.rb"
```
     68       BasicSocket.do_not_reverse_lookup = true
     69       Plugin.load_plugins
     70       if defined?(Encoding)
     71         Encoding.default_internal = 'ASCII-8BIT' if Encoding.respond_to?(:default_internal)
     72         Encoding.default_external = 'ASCII-8BIT' if Encoding.respond_to?(:default_external)
     73       end
```
But why I did not get the fluentd-0.12.31 with the patch???

And yes, I'm going to update the fluentd version and retest logging-fluentd.

fluentd/Dockerfile
```diff
@@ -5,7 +5,7 @@ MAINTAINER OpenShift Development <dev@lists.openshift.redhat.com>
 ENV HOME=/opt/app-root/src \
   PATH=/opt/app-root/src/bin:/opt/app-root/bin:$PATH \
   RUBY_VERSION=2.0 \
-  FLUENTD_VERSION=0.12.31 \
+  FLUENTD_VERSION=0.12.39-2 \
```
richm: Does this work?  I don't think we've ever tried a `version-release`

PR #686, opened 2017-09-28
Fluentd - one output tag per plugin

nhosoi:
Removing "WARNING: There is no @OUTPUT label declared in /etc/fluent/fluent.conf."


richm:
[merge]

richm:
@nhosoi does this need to be backported to 3.6?

PR #695, opened 2017-10-03
bz#1497937: run oc deploy when installedVersion is >= 1

nak3:
This patch changes to run `oc deploy $object --latest` when
`installedVersion` >= 1. Without this patch, `oc deploy --latest` will
not be executed and deployer pod will not complete.


nak3:
Sorry, i pushed wrong branch. Updated here https://github.com/openshift/origin-aggregated-logging/pull/696

PR #699, opened 2017-10-04
preserve ops namespace tags in mux mode

richm:
When mux receives logs from fluentd, look for the `@timestamp` field
to mean that the logs have already been processed in the viaq data
model format, so they need no further processing, just k8s meta if
necessary, and construct the index name.  Records that have already
been processed by fluentd get a ".mux" tag suffix, and "raw"
unprocessed records get a ".raw" tag suffix.
Also, remove the `docker` and `kubernetes` fields from records sent
externally.  These fields are added by the k8s meta to when getting
the id for the namespace but are not needed for external records
once the index name is created.

When using the json-file container logs, the mux test will add
CONTAINER_NAME in the same places where it is implicitly added by
the journald log driver logs.
/test

### DIFF
--- a/test/mux-client-mode.sh
+++ b/test/mux-client-mode.sh
@@ -6,26 +6,16 @@ source "$(dirname "${BASH_SOURCE[0]}" )/../hack/lib/init.sh"
 source "${OS_O_A_L_DIR}/hack/testing/util.sh"
 os::util::environment::use_sudo
 
-os::test::junit::declare_suite_start "test/curator"
-
-if [ -n "${DEBUG:-}" ] ; then
-    set -x
-    curl_output() {
-        python -mjson.tool
-    }
+# only works if there is a mux dc
+if oc get dc/logging-mux > /dev/null 2>&1 ; then
+    os::log::debug "$( oc get dc/logging-mux )"
 else
-    curl_output() {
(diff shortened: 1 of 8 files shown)


test/mux-client-mode.sh
```diff
@@ -62,28 +46,16 @@ cleanup() {
...
-        - mountPath: /etc/fluent/muxkeys\
-          name: muxcerts\
-          readOnly: true\
-' | oc replace -f -
+      os::log::debug "$( oc set volumes daemonset/logging-fluentd --add --overwrite \
+               --name=muxcerts --default-mode=0400 -t secret -m /etc/fluent/muxkeys --secret-name logging-mux 2>&1 )"
   fi
 }
 
-os::log::info Starting mux-client-mode test at $( date )
```
nhosoi: We could keep this "Starting mux-client-mode" line?

richm: It is redundant when running this test in CI, since entrypoint.sh also prints a similar message

nhosoi: Ok.  Thanks.

richm:
Tests passing now - can I get a lgtm?

richm:
@nhosoi thanks!

PR #701, opened 2017-10-04
use record['fieldname'] instead of 'fieldname' - was throwing exception

richm:
Was seeing an exception being thrown in some cases, relating to an
undefined variable `log`.  Instead of relying on the `record_transformer`
behavior of adding local variables with the same name as the field,
use the explicit behavior of using the field from `record`.  This also
clears up any ambiguity between the `log` logging object and a field
named `log`.
/test


richm:
@nhosoi with the new openshift-jenkins bot, if you approve of a PR, after the tests have passed, add a comment with /lgtm

PR #702, opened 2017-10-05
preserve ops namespace tags in mux mode

richm:
When mux receives logs from fluentd, look for the `@timestamp` field
to mean that the logs have already been processed in the viaq data
model format, so they need no further processing, just k8s meta if
necessary, and construct the index name.  Records that have already
been processed by fluentd get a ".mux" tag suffix, and "raw"
unprocessed records get a ".raw" tag suffix.
Also, remove the `docker` and `kubernetes` fields from records sent
externally.  These fields are added by the k8s meta to when getting
the id for the namespace but are not needed for external records
once the index name is created.

When using the json-file container logs, the mux test will add
CONTAINER_NAME in the same places where it is implicitly added by
the journald log driver logs.

The test creates indices like `project.testproj.*` and `project.default.*`
which are not normally present, so make sure these are removed.

(cherry picked from commit 0c0ddce486380819a5ad12d10efa12a248be573c)
/test

### DIFF
--- a/docs/mux-logging-service.md
+++ b/docs/mux-logging-service.md
@@ -321,6 +321,16 @@ TLS server `cert` and `key`, as well as the `ca` cert and the `shared_key`.
 These records are sent to the `@MUX` label.  The first thing here is this
 filter:
 
+    <match journal>
+      @type relabel
+      @label @INGRESS
+    </match>
+
+Records tagged with `journal` are "raw" records from a Fluentd running in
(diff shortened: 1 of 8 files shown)


richm:
cherry pick of https://github.com/openshift/origin-aggregated-logging/pull/699

richm:
@jcantrill openshift-bot says no lgtm but @nhosoi gave it - does she have rights to give /lgtm?

PR #703, opened 2017-10-05
use record['fieldname'] instead of 'fieldname' - was throwing exception

richm:
(cherry picked from commit 84107183f466d4ea336f022697c76ec0149ee955)
/test


richm:
cherry pick of https://github.com/openshift/origin-aggregated-logging/pull/701

PR #705, opened 2017-10-06
Bug 1490647 - logging-fluentd deployed with openshift_logging_use_mux=false fails to start due to missing

nhosoi:
If openshift_logging_use_mux=False and openshift_logging_mux_allow_external=False,
then all other mux related parameters should be set to False (if boolean) or
removed (e.g. openshift_logging_mux_client_mode should be undefined).

To determine if mux is configured in run.sh, check whether the directory
/etc/fluent/muxkeys exits or not.  If it does not exist, MUX_CLIENT_MODE
is unset.

### DIFF
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -42,10 +42,15 @@ if [ -z "${USE_MUX:-}" -o "${USE_MUX:-}" = "false" ] ; then
     else
         export USE_JOURNAL=false
     fi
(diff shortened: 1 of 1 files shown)


richm:
@nhosoi this should be backported to release-3.6

nhosoi:
@richm 
> this should be backported to release-3.6

Sorry...  Done.
https://github.com/openshift/origin-aggregated-logging/pull/718

PR #711, opened 2017-10-09
security test

FlorinPeter:
removed authorization header for security test

I checked the ability to access the elasticsearch cluster from a different namespace and I found out that it is not necessary to provide an authorization header if an X-Forwarded-For header is provided.

That allow an non ops user to access ops logs if the non ops user is knowing the admin login, for example:
```
curl -vk \
-H "X-Proxy-Remote-User: admin" \
-H "X-Forwarded-For: 127.0.0.1" \
"https://logging-es.logging.svc:9200/.operations.*/_search?q=*"
```
  
please do not merge


richm:
@FlorinPeter I have some questions about this that I would like to discuss privately.  Please contact me offline - rmeggins@redhat.com

FlorinPeter:
@richm why did you close this PR?
It would be interesting if this hack works against the upstream and if yes it would also a good idea to write a test case to avoid this situation in the future.

PR #713, opened 2017-10-09
Get more data from ES

ruromero:
Some enhancements to the logging-dump.sh
* Sort environment variables
* Get internal ES logs
* Get latest documents stored (if cluster is GREEN)
* Get indices size in bytes


hack/logging-dump.sh
```diff
@@ -239,10 +240,30 @@ get_elasticsearch_status() {
...
+  local pod=$1
+  local logs_folder=$2/logs
+  echo -- POD $1 Elasticsearch Logs
+  if [ ! -d "$logs_folder" ]
+  then
+    mkdir $logs_folder
+  fi
+  if [[ $pod == logging-es-ops* ]]
+  then
+    path=/elasticsearch/logging-es-ops/logs
```
portante: Might want to consider deriving these from the final elasticsearch.yaml file that is rendered in the pod.

PR #716, opened 2017-10-09
CI test improvement

nhosoi:
```
    entrypoint.sh - Allowing SUITE to have multiple tests. E.g.,
                    SUITE="test-mux.sh test-viaq-data-model.sh"
    
    viaq-data-model.sh - Checking MUX_CLIENT_MODE=maximal as a string.
    
    mux.sh - In the file buffer's persistency test, replacing "sleep ##"
             with the code checking the file buffers' existency.
             When ops_cluster is enabled, ES_HOST and OPS_HOST should not
             have the same temporary test name.
```

### DIFF
--- a/hack/testing/entrypoint.sh
+++ b/hack/testing/entrypoint.sh
@@ -128,13 +128,16 @@ function run_suite() {
 	fi
 }
 
-suite_selector="${SUITE:-".*"}"
-for test in $( find "${OS_O_A_L_DIR}/hack/testing" -type f -name 'check-*.sh' | grep -E "${suite_selector}" | sort ); do
+for suite_selector in ${SUITE:-".*"} ; do
+  for test in $( find "${OS_O_A_L_DIR}/hack/testing" -type f -name 'check-*.sh' | grep -E "${suite_selector}" | sort ); do
 	run_suite "${test}"
+  done
 done
 
-for test in $( find "${OS_O_A_L_DIR}/hack/testing" -type f -name 'test-*.sh' | grep -E "${suite_selector}" | sort ); do
+for suite_selector in ${SUITE:-".*"} ; do
+  for test in $( find "${OS_O_A_L_DIR}/hack/testing" -type f -name 'test-*.sh' | grep -E "${suite_selector}" | sort ); do
 	run_suite "${test}"
+  done
 done
 
 if [[ -n "${failed:-}" ]]; then
--- a/test/fluentd-forward.sh
+++ b/test/fluentd-forward.sh
@@ -6,7 +6,7 @@ source "$(dirname "${BASH_SOURCE[0]}" )/../hack/lib/init.sh"
 source "${OS_O_A_L_DIR}/hack/testing/util.sh"
 os::util::environment::use_sudo
 
-FLUENTD_WAIT_TIME=$(( 2 * minute ))
+FLUENTD_WAIT_TIME=${FLUENTD_WAIT_TIME:-$(( 2 * minute ))}
 
 os::test::junit::declare_suite_start "test/fluentd-forward"
 
@@ -43,6 +43,7 @@ update_current_fluentd() {
     # redeploy fluentd
     os::cmd::expect_success flush_fluentd_pos_files
     os::log::debug "$( oc label node --all logging-infra-fluentd=true )"
+    os::cmd::try_until_success "oc rollout status -w ds/logging-fluentd"
     os::cmd::try_until_text "oc get pods -l component=fluentd" "^logging-fluentd-.* Running "
     fpod=$( get_running_pod fluentd )
 }
--- a/test/mux.sh
+++ b/test/mux.sh
@@ -16,9 +16,22 @@ else
     exit 0
 fi
 
-FLUENTD_WAIT_TIME=$(( 2 * minute ))
+FLUENTD_WAIT_TIME=${FLUENTD_WAIT_TIME:-$(( 2 * minute ))}
 
 os::test::junit::declare_suite_start "test/mux"
+
+ARTIFACT_DIR=${ARTIFACT_DIR:-${TMPDIR:-/tmp}/origin-aggregated-logging}
+
+os::log::debug ARTIFACT_DIR is $ARTIFACT_DIR
+
+es_pod=$( get_es_pod es )
+es_ops_pod=$( get_es_pod es-ops )
+if [ -n "$es_ops_pod" ] ; then
+  ops_cluster=true
+else
+  ops_cluster=false
+  es_ops_pod=$es_pod
+fi
 
 reset_fluentd_daemonset() {
   # this test only works with MUX_CLIENT_MODE=minimal for now
@@ -246,8 +259,11 @@ write_and_verify_logs() {
 }
 
 reset_ES_HOST() {
-    oc set env dc logging-mux $1 $2
+    muxpod=$( get_running_pod mux )
+    os::cmd::expect_success "oc set env dc logging-mux $1 $2"
     os::cmd::try_until_failure "oc get pod $muxpod"
+    os::log::debug $( oc get pods -l component=mux )
+    oc rollout status -w dc/logging-mux # wait for mux to be redeployed
     os::cmd::try_until_text "oc get pods -l component=mux" "^logging-mux-.* Running "
     muxpod=$( get_running_pod mux )
 }
@@ -255,6 +271,11 @@ reset_ES_HOST() {
 cleanup() {
     local return_code="$?"
     set +e
+
+    # In case test failed in Test case FILE_BUFFER_STORAGE_TYPE: $MUX_FILE_BUFFER_STORAGE_TYPE 
+    # reset ES_HOST and OPS_HOST
+    reset_ES_HOST $ES_HOST_BAK $OPS_HOST_BAK
+
     if [ $return_code = 0 ] ; then
         mycmd=os::log::info
     else
@@ -336,6 +357,9 @@ if [ "$MUX_FILE_BUFFER_STORAGE_TYPE" = "pvc" ]; then
     os::log::debug "$( oc get pvc )"
 fi
 
+ES_HOST_BAK=$( oc set env --list dc/logging-mux | grep \^ES_HOST= )
+OPS_HOST_BAK=$( oc set env --list dc/logging-mux | grep \^OPS_HOST= )
+
 # make sure fluentd is working normally
 fpod=$( get_running_pod fluentd )
 wait_for_fluentd_ready
@@ -346,11 +370,12 @@ if [ "$MUX_FILE_BUFFER_STORAGE_TYPE" = "pvc" -o "$MUX_FILE_BUFFER_STORAGE_TYPE" 
 
     update_current_fluentd $ENABLE_SECURE_FORWARD
 
-    ES_HOST_BAK=$( oc set env --list dc/logging-mux | grep \^ES_HOST= )
-    OPS_HOST_BAK=$( oc set env --list dc/logging-mux | grep \^OPS_HOST= )
-
-    # set ES_HOST and OPS_HOST to bogus
-    reset_ES_HOST ES_HOST=bogus OPS_HOST=bogus
+    # set ES_HOST and OPS_HOST to non-existing hostname
+    if [ "$ops_cluster" = "true" ]; then
+        reset_ES_HOST ES_HOST=bogus OPS_HOST=bogus_ops
+    else
+        reset_ES_HOST ES_HOST=bogus OPS_HOST=bogus
+    fi
 
     uuid_es=$( uuidgen )
     uuid_es_ops=$( uuidgen )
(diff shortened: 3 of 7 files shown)


hack/testing/entrypoint.sh
```diff
@@ -127,13 +127,16 @@ function run_suite() {
 	fi
 }
 
-suite_selector="${SUITE:-".*"}"
-for test in $( find "${OS_O_A_L_DIR}/hack/testing" -type f -name 'check-*.sh' | grep -E "${suite_selector}" | sort ); do
+for suite_selector in ${SUITE:-".*"} ; do
```
richm: SUITE is set like this?
```
SUITE="viaq-data-model.sh mux-client-mode.sh" make test
```
?

test/viaq-data-model.sh
```diff
@@ -115,6 +115,7 @@ get_logmessage2() {
 }
 
 # TEST 1
+TIMEOUT=600
```
richm: it really shouldn't take this long - that would indicate that there is some other problem

Is it the first `wait_for_fluentd_to_catch_up` that fails without the timeout?  Is it a specific line in the test that is failing?

One thing I have noticed is that some of the tests do not clean up correctly - specifically, they do not wait until mux and fluentd are up and running in some cases.  So, for example, if you see that it is the first test that fails, maybe it is because fluentd is not yet up and running?

hack/testing/entrypoint.sh
nhosoi: Yes, exactly.

test/viaq-data-model.sh
nhosoi: > Is it the first wait_for_fluentd_to_catch_up that fails without the timeout? Is it a specific line in the test that is failing?

Sometimes, yes.  But more often, {"message":"GET /$UUID 404 "} fails.

> One thing I have noticed is that some of the tests do not clean up correctly

Ok.  Let me try just running test-viaq-data-model.sh a couple of times...  (I think I ran into it with the single test case, but not 100% sure...)

test/mux.sh
```diff
@@ -237,15 +243,26 @@ write_and_verify_logs() {
...
     local return_code="$?"
     set +e
+
+    # In case test failed in Test case FILE_BUFFER_STORAGE_TYPE: $MUX_FILE_BUFFER_STORAGE_TYPE 
+    # reset ES_HOST and OPS_HOST
+    reset_ES_HOST $ES_HOST_BAK $OPS_HOST_BAK
+
+    # wait for the file buffer disappears once
+    os::cmd::try_until_failure 'oc get pods | egrep "logging-mux-.*-deploy"'
+    os::cmd::try_until_failure 'oc get pods | egrep "logging-mux-.*Terminating"'
```
richm: These two lines do not seem right to me - what are you trying to do?  Follow a deployment which you have deployed either explicitly using `oc rollout` or implicitly by editing the mux dc, `oc set env dc/logging-mux`, or etc.?  If so, you should use `oc rollout status -w dc/logging-mux` which will follow the deployment and either return success when it is deployed, or failure if the deployment failed or timed out.

test/mux.sh
```diff
@@ -282,8 +299,8 @@ cleanup() {
     fi
     # delete indices created by this test
     curl_es $es_pod /project.testproj.* -XDELETE
-    curl_es $es_pod /project.default.* -XDELETE
-    curl_es $es_pod /project.mux-undefined.* -XDELETE
+    os::log::debug $( curl_es $es_pod /project.default.* -XDELETE ) || :
```
richm: This will always fail now - there should never be "project.default.*" indices

test/mux.sh
```diff
@@ -18,6 +18,12 @@ fi
 
 os::test::junit::declare_suite_start "test/mux"
 
+ARTIFACT_DIR=${ARTIFACT_DIR:-${TMPDIR:-/tmp}/origin-aggregated-logging}
+
+os::log::debug ARTIFACT_DIR is $ARTIFACT_DIR
+
+ops_cluster=${1:-"true"}
```
richm: The test should not rely on this. The test should do something like 
```
espod=$( get_es_pod es )
esopspod=$( get_es_pod es-ops )
if [ -n "$esopspod" ] ; then
  ops_cluster=true
else
  ops_cluster=false
  esopspod=$espod
fi
```

test/mux.sh
```diff
@@ -237,15 +243,26 @@ write_and_verify_logs() {
 }
 
 reset_ES_HOST() {
-    oc set env dc logging-mux $1 $2
+    muxpod=$( get_running_pod mux )
+    os::cmd::try_until_success "oc set env dc logging-mux $1 $2"
     os::cmd::try_until_failure "oc get pod $muxpod"
+    os::log::debug $( oc get pods -l component=mux )
     os::cmd::try_until_text "oc get pods -l component=mux" "^logging-mux-.* Running "
-    muxpod=$( get_running_pod mux )
```
richm: Why get rid of setting `muxpod` here?

test/mux.sh
richm: Or
If `$muxpod` is set, can't you just do 
```
```
?

test/mux.sh
```diff
@@ -331,40 +348,78 @@ fpod=$( get_running_pod fluentd )
...
-    # set ES_HOST and OPS_HOST to bogus
-    reset_ES_HOST ES_HOST=bogus OPS_HOST=bogus
+    # set ES_HOST and OPS_HOST to non-existing hostname
+    if [ "$ops_cluster" = "true" ]; then
+        reset_ES_HOST ES_HOST=bogus OPS_HOST=bogus_ops
+    else
+        reset_ES_HOST ES_HOST=bogus OPS_HOST=bogus
+    fi
+    os::cmd::try_until_failure 'oc get pods | egrep "logging-mux-.*-deploy"'
+    os::cmd::try_until_failure 'oc get pods | egrep "logging-mux-.*Terminating"'
```
richm: see above for commentary about this

test/mux.sh
```diff
@@ -331,40 +348,78 @@ fpod=$( get_running_pod fluentd )
...
+    os::cmd::try_until_text "oc get pods -l component=mux" "Running"
 
     uuid_es=$( uuidgen )
     uuid_es_ops=$( uuidgen )
 
-    logger -i -p local6.info -t $uuid_es_ops $uuid_es_ops
     add_test_message $uuid_es
+    logger -i -p local6.info -t $uuid_es_ops $uuid_es_ops
```
richm: Does the ordering matter here?

test/mux.sh
```diff
@@ -331,40 +348,78 @@ fpod=$( get_running_pod fluentd )
...
+    muxpod=$( get_running_pod mux )
+    if [ "$ops_cluster" = "true" ]; then
+       ops_filename="output-es-ops-config.output_ops_tag"
+    else
+       ops_filename="output-es-config.output_tag"
+    fi
+    retry="true"
+    while [ "$retry" = "true" ]; do
+        ops_logs=$( oc exec $muxpod -- ls /var/lib/fluentd/ | egrep $ops_filename ) || :
+        if [ -z "$ops_logs" ]; then
```
richm: do you need a `sleep 1` here to avoid a too tight loop?

test/mux.sh
nhosoi: Ok.  Being replaced.

test/mux.sh
nhosoi: Well, the issue is after resetting ES_HOST and OPS_HOST, if the test logs are issued without making sure the new mux pod is 100% up and running by [1], the fluentd client fails to send the data and they are lost (since the current fluentd client are not using the file buffer [2].)  muxpod have to be retrieved after the code [1] is done.  If the way how to confirm the new mux pod's full availability [1] is not correct, I'm open to your suggestion.
```
[1]
    # wait for the file buffer disappears once
    os::cmd::try_until_failure 'oc get pods | egrep "logging-mux-.*-deploy"'
    os::cmd::try_until_failure 'oc get pods | egrep "logging-mux-.*Terminating"'
    os::cmd::try_until_text "oc get pods -l component=mux" "Running"
```
[2] https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/configs.d/filter-pre-mux-client.conf

test/mux.sh
nhosoi: Well, I followed the test steps manually and saw these lines in the transition state.  I had to make sure these two lines disappear before going forward to make the test successfully run. 
```
logging-mux-.*-deploy    Running
logging-mux-PREVIOUS_MUX_POD    Terminating
```

test/mux.sh
nhosoi: Ok. Removing the line.

test/mux.sh
nhosoi: No, it was a part of my investigation. :)  Reverting this change.

test/mux.sh
nhosoi: A good idea.  Adding it.

test/mux.sh
richm: ok - I think the following should be sufficient:

    # do something to cause mux to be redeployed
    oc rollout status -w dc/logging-mux # wait for mux to be redeployed
    os::cmd::try_until_text "oc get pods -l component=mux" "Running"

After this, there still may be a mux pod in "Terminating" for a short period of time while the new one is in "Running", but that should be ok

richm:
Test failure: https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_origin-aggregated-logging/716/test_pull_request_origin_aggregated_logging_ansible/732/
```
test/mux.sh:
[INFO] ------- Test case 4 -------
Running test/mux.sh:254: executing 'oc set env dc logging-mux ES_HOST=logging-es OPS_HOST=logging-es-ops' expecting success; re-trying every 0.2s until completion or 60.000s...
SUCCESS after 0.263s: test/mux.sh:254: executing 'oc set env dc logging-mux ES_HOST=logging-es OPS_HOST=logging-es-ops' expecting success; re-trying every 0.2s until completion or 60.000s
Running test/mux.sh:255: executing 'oc get pod logging-mux-1-tt9sb' expecting failure; re-trying every 0.2s until completion or 60.000s...
FAILURE after 59.187s: test/mux.sh:255: executing 'oc get pod logging-mux-1-tt9sb' expecting failure; re-trying every 0.2s until completion or 60.000s: the command timed out
NAME                  READY     STATUS    RESTARTS   AGE
logging-mux-1-tt9sb   1/1       Running   0          34m
```
I'm guessing that doing the `oc set env dc logging-mux ES_HOST=logging-es OPS_HOST=logging-es-ops` didn't actually change anything, so mux did not redeploy, therefore the pod never went away.

nhosoi:
> I'm guessing that doing the oc set env dc logging-mux ES_HOST=logging-es OPS_HOST=logging-es-ops didn't actually change anything, so mux did not redeploy, therefore the pod never went away.

The following reset_ES_HOST is called from cleanup and its purpose is to reset the ES_HOST and OPS_HOST in case they still have the test hosts bogus and bogus_ops.  So, I'd think if they don't have to be reset, it'd be ok...
`Running test/mux.sh:254: executing 'oc set env dc logging-mux ES_HOST=logging-es OPS_HOST=logging-es-ops' expecting success; re-trying every 0.2s until completion or 60.000s...`

But indeed there is a regression.  Keep investigating...

test/mux.sh
nhosoi: @richm, it seems if we don't delete project.default.* here, it makes zzz-correct-index-names.sh fail later...
```
Running test/zzz-correct-index-names.sh:29: executing 'curl_es logging-es-data-master-jp217vjp-1-983bd /_cat/indices' expecting success and not text 'project.default.'...
FAILURE after 0.484s: test/zzz-correct-index-names.sh:29: executing 'curl_es logging-es-data-master-jp217vjp-1-983bd /_cat/indices' expecting success and not text 'project.default.': the output content test failed
```

test/fluentd-forward.sh
```diff
@@ -129,4 +129,5 @@ create_forwarding_fluentd
 update_current_fluentd
 
 wait_for_fluentd_ready
+TIMEOUT=3000
```
richm: This is just for testing purposes, right?

test/mux.sh
richm: That means there is yet another bug somewhere - possibly we need to change something about the fluentd forward configuration so that default namespace messages go to .operations?  Otherwise, it is a bug.

nhosoi:
Squashed and pushed.

test/mux.sh
```diff
@@ -237,15 +250,25 @@ write_and_verify_logs() {
 }
 
 reset_ES_HOST() {
-    oc set env dc logging-mux $1 $2
+    muxpod=$( get_running_pod mux )
+    os::cmd::try_until_success "oc set env dc logging-mux $1 $2"
```
richm: I think you can just use `expect_success` here instead of `try_until_success` - that is, `oc set env` should be run once, and only once.

test/mux.sh
```diff
@@ -237,15 +250,25 @@ write_and_verify_logs() {
...
     os::cmd::try_until_failure "oc get pod $muxpod"
+    os::log::debug $( oc get pods -l component=mux )
     os::cmd::try_until_text "oc get pods -l component=mux" "^logging-mux-.* Running "
```
richm: After this line returns, then the old mux has been removed and the new one is running.  Which means mux has been redeployed.  Which means you do not need to check for rollout status, and you do not need to check again that mux is running below.

nhosoi:
@richm, could you please review this pr one more time?  Thanks!

richm:
@nhosoi please backport this to 3.6

nhosoi:
@richm, this patch depends upon pr/728.
`bumping up timeout waiting for fluentd to terminate from 60s to 120s …`
Is pr/728 going to be backported?

richm:
> Is pr/728 going to be backported?

yes

nhosoi:
@richm, done. 
https://github.com/openshift/origin-aggregated-logging/pull/755
[3.6] CI test improvements/stabilization (backport #716)

PR #718, opened 2017-10-10
Bug 1490647 - logging-fluentd deployed with openshift_logging_use_mux=false fails to start due to missing

nhosoi:
(3.6 backport of pr#705)

If openshift_logging_use_mux=False and openshift_logging_mux_allow_external=False,
then all other mux related parameters should be set to False (if boolean) or
removed (e.g. openshift_logging_mux_client_mode should be undefined).

To determine if mux is configured in run.sh, check whether the directory
/etc/fluent/muxkeys exits or not.  If it does not exist, MUX_CLIENT_MODE
is unset.

(cherry picked from commit 9954e962e62c8ef512ce6c7a0af7974f2afdbaf9)

### DIFF
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -42,10 +42,15 @@ if [ -z "${USE_MUX:-}" -o "${USE_MUX:-}" = "false" ] ; then
     else
         export USE_JOURNAL=false
     fi
+    unset MUX_FILE_BUFFER_STORAGE_TYPE
 else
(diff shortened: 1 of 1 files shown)


richm:
@stevekuznetsov not sure what's wrong here - Submit Queue says "Required Github CI test is not green: ci/openshift-jenkins/logging" but that test is green?

stevekuznetsov:
Unclear, but it looks like the bot figured it out in the end :)

PR #728, opened 2017-10-17
bumping up timeout waiting for fluentd to terminate from 60s to 120s …

ewolinetz:
…for tests

Hopefully waiting a little bit longer will cut down on the number of flakes. I noticed in the controller logs that our test failed before the controller was able to completely flush the terminating fluentd pod.

@stevekuznetsov 
@richm 
@jcantrill

### DIFF
--- a/hack/lib/init.sh
+++ b/hack/lib/init.sh
@@ -11,6 +11,8 @@ set -o pipefail
 
 OS_SCRIPT_START_TIME="$( date +%s )"; export OS_SCRIPT_START_TIME
 
+FLUENTD_WAIT_TIME=120
+
 # os::util::absolute_path returns the absolute path to the directory provided
 function os::util::absolute_path() {
 	local relative_path="$1"
--- a/hack/lib/init.sh
+++ b/hack/lib/init.sh
@@ -11,8 +11,6 @@ set -o pipefail
 
 OS_SCRIPT_START_TIME="$( date +%s )"; export OS_SCRIPT_START_TIME
 
-FLUENTD_WAIT_TIME=120
-
 # os::util::absolute_path returns the absolute path to the directory provided
 function os::util::absolute_path() {
 	local relative_path="$1"
--- a/hack/testing/util.sh
+++ b/hack/testing/util.sh
@@ -177,7 +177,7 @@ function wait_for_fluentd_to_catch_up() {
     local uuid_es=$( uuidgen )
     local uuid_es_ops=$( uuidgen )
     local expected=${3:-1}
-    local timeout=${TIMEOUT:-300}
+    local timeout=${TIMEOUT:-600}
     local project=${4:-logging}
 
     add_test_message $uuid_es
--- a/test/fluentd-forward.sh
+++ b/test/fluentd-forward.sh
@@ -16,7 +16,7 @@ update_current_fluentd() {
 
     # undeploy fluentd
     os::log::debug "$( oc label node --all logging-infra-fluentd- )"
-    os::cmd::try_until_failure "oc get pod $fpod" $FLUENTD_WAIT_TIME
+    os::cmd::try_until_text "oc get daemonset logging-fluentd -o jsonpath='{ .status.numberReady }'" "0" $FLUENTD_WAIT_TIME
 
     # edit so we don't send to ES
     oc get configmap/logging-fluentd -o yaml | sed '/## matches/ a\
@@ -105,7 +105,7 @@ cleanup() {
         oc logs $fpod > $ARTIFACT_DIR/$fpod.log 2>&1
     fi
     os::log::debug "$( oc label node --all logging-infra-fluentd- 2>&1 || : )"
-    os::cmd::try_until_failure "oc get pod $fpod" $FLUENTD_WAIT_TIME
+    os::cmd::try_until_text "oc get daemonset logging-fluentd -o jsonpath='{ .status.numberReady }'" "0" $FLUENTD_WAIT_TIME
     if [ -n "${savecm:-}" -a -f "${savecm:-}" ] ; then
         os::log::debug "$( oc replace --force -f $savecm )"
     fi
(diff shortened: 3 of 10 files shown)


richm:
@nhosoi ptal

richm:
There's also this test failure which has become common lately:
```
Running test/cluster/rollout.sh:48: executing 'oc get daemonset logging-fluentd -o jsonpath='{ .status.numberReady }'' expecting any result and text '1'; re-trying every 0.2s until completion or 60.000s...
FAILURE after 60.217s: test/cluster/rollout.sh:48: executing 'oc get daemonset logging-fluentd -o jsonpath='{ .status.numberReady }'' expecting any result and text '1'; re-trying every 0.2s until completion or 60.000s: the command timed out
```
That is, we time out waiting for fluentd to start up too.  Perhaps that code should also have a 120 second timeout instead of the default 60.

nhosoi:
@ewolinetz, @richm, I'm having the similar timeout problem when I run the Rich's script oshift-oct-new-dev.sh.  So, I vote "yes" to increase the timeout value.

And forgive me to piggyback...  I think these are to be fixed, too...
```
diff --git a/hack/lib/cmd.sh b/hack/lib/cmd.sh
index eec4e60..e858cce 100644
--- a/hack/lib/cmd.sh
+++ b/hack/lib/cmd.sh
@@ -108,7 +108,7 @@ minute=$(( 60 * second ))
 function os::cmd::try_until_success() {
        if [[ $# -lt 1 ]]; then echo "os::cmd::try_until_success expects at least one arguments, got $#"; return 1; fi
        local cmd=$1
-       local duration=${2:-minute}
+       local duration=${2:-$minute}
        local interval=${3:-0.2}
 
        os::cmd::internal::run_until_exit_code "${cmd}" "os::cmd::internal::success_func" "${duration}" "${interval}"
@@ -133,7 +133,7 @@ function os::cmd::try_until_text() {
        if [[ $# -lt 2 ]]; then echo "os::cmd::try_until_text expects at least two arguments, got $#"; return 1; fi
        local cmd=$1
        local text=$2
-       local duration=${3:-minute}
+       local duration=${3:-$minute}
        local interval=${4:-0.2}
 
        os::cmd::internal::run_until_text "${cmd}" "${text}" "os::cmd::internal::success_func" "${duration}" "${interval}"
@@ -146,7 +146,7 @@ function os::cmd::try_until_not_text() {
        if [[ $# -lt 2 ]]; then echo "os::cmd::try_until_not_text expects at least two arguments, got $#"; return 1; fi
        local cmd=$1
        local text=$2
-       local duration=${3:-minute}
+       local duration=${3:-$minute}
        local interval=${4:-0.2}
 
        os::cmd::internal::run_until_text "${cmd}" "${text}" "os::cmd::internal::failure_func" "${duration}" "${interval}"
```

nhosoi:
Sorry, I clicked a wrong button... :(

hack/lib/init.sh
```diff
@@ -11,6 +11,8 @@ set -o pipefail
 
 OS_SCRIPT_START_TIME="$( date +%s )"; export OS_SCRIPT_START_TIME
 
+FLUENTD_WAIT_TIME=120
```
stevekuznetsov: The bar is usually _really_ high to add anything to this file. Especially since it's coming from Origin and this will become a long-term carry commit on top. I think it's fine to just use 120 in the areas you need it

test/fluentd-forward.sh
```diff
@@ -14,7 +14,7 @@ update_current_fluentd() {
 
     # undeploy fluentd
     os::log::debug "$( oc label node --all logging-infra-fluentd- )"
-    os::cmd::try_until_failure "oc get pod $fpod"
+    os::cmd::try_until_failure "oc get pod $fpod" $FLUENTD_WAIT_TIME
```
stevekuznetsov: You should take the time to change all of  these to use the comparison I did in `rollout.sh` -- just because a pod exists doesn't mean it's been scheduled, the init containers have fired, it hasn't died, etc...

nhosoi:
Looking into hack/lib/cmd.sh, we need to set (120 * 60 * 1000) to FLUENTD_WAIT_TIME?
```
millisecond=1
second=$(( 1000 * millisecond ))
minute=$(( 60 * second ))

function os::cmd::try_until_failure() {
        local duration=${2:-$minute}
```
But, I think we could just replace "minute=$(( 60 * second ))" with
`minute=$(( 120 * second ))`
in hack/lib/cmd.sh?

richm:
fluentd still hasn't started after 120 seconds - methinks this is a deeper problem as in  https://bugzilla.redhat.com/show_bug.cgi?id=1503351

ewolinetz:
@richm I think based on @nhosoi 's comments we actually are using 120 ms... updating to set the time to `$((2 * minute))`

May need to circle back after and try to implement @stevekuznetsov 's suggestions.

ewolinetz:
/hold

I have another change to make after this one passes

stevekuznetsov:
> But, I think we could just replace `minute=$(( 60 * second ))` with `minute=$(( 120 * second ))` in hack/lib/cmd.sh?

Please don't do that.  A minute is 60 seconds. The calls all allow for you to pass a time-out if you are not happy with the default. Pass one.

ewolinetz:
@stevekuznetsov I didn't plan on doing that... I instead reused the `minute` variable when setting the time to be `$(( 2 * minute ))`

ewolinetz:
Not sure why the bot wasn't updating its status... but on jenkins both tests passed again. If these tests pass then im good with merging

ewolinetz:
/kind bug

the merge gate has gone up...

hack/testing/util.sh
```diff
@@ -177,7 +177,7 @@ function wait_for_fluentd_to_catch_up() {
     local uuid_es=$( uuidgen )
     local uuid_es_ops=$( uuidgen )
     local expected=${3:-1}
-    local timeout=${TIMEOUT:-300}
+    local timeout=${TIMEOUT:-600}
```
richm: ok - but I can't imagine what's going on that makes fluentd take 10 minutes to read a record from journal/json-file and store it in Elasticsearch :-P

richm:
in the interest of getting CI working as quickly as possible
/lgtm
then we can figure out why `project.default.*` is cropping up

richm:
@ewolinetz please backport this to 3.6

PR #730, opened 2017-10-18
dump indices before and after test

richm:
/test


richm:
@ewolinetz @jcantrill This PR is based off the latest master.  It just extends the test to dump the contents of /_cat/indices from es and es-ops before and after each test.  As you can see, there is no `project.default.*`.  So I'm not sure how/where in the failing PRs something is being introduced which causes `project.default.*` to be created.

PR #731, opened 2017-10-19
dump indices before and after test

richm:
So we can find out of project.default is being created by some test
/test

### DIFF
--- a/hack/testing/test-zzz-correct-index-names.sh
+++ b/hack/testing/test-zzz-correct-index-names.sh
@@ -1,6 +1,4 @@
 #!/bin/bash
-# Currently not checking since it is causing CI tests to fail
-# TODO: Add back test and fix cause.
-# source "$(dirname "${BASH_SOURCE[0]}" )/../lib/init.sh"
+source "$(dirname "${BASH_SOURCE[0]}" )/../lib/init.sh"
 
-#exec ${OS_O_A_L_DIR}/test/zzz-correct-index-names.sh
+exec ${OS_O_A_L_DIR}/test/zzz-correct-index-names.sh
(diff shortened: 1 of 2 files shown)


richm:
bingo - must be a test flake in the read-throttling test:
```
before /data/src/github.com/openshift/origin-aggregated-logging/hack/testing/test-remote-syslog.sh logging-es-data-master-d6ks5ubv-1-8vnxg indices
green open .searchguard.logging-es-data-master-d6ks5ubv                          1 0    5 0  40.9kb  40.9kb 
green open project.default.044be73a-b4fe-11e7-8927-0ec06334315e.2017.10.19       1 0    0 0    130b    130b 
```
although it doesn't show up after that test - it must take a few seconds for es to create the index

richm:
I changed the ansible to explicitly pass in the value on the command line:
```
                         -e openshift_logging_fluentd_audit_container_engine=True                 \
                          /usr/share/ansible/openshift-ansible/playbooks/byo/openshift-cluster/openshift-logging.yml \
```
We'll see if that makes a difference
/retest

hack/testing/entrypoint.sh
```diff
@@ -113,6 +113,13 @@ function run_suite() {
 	os::cmd::expect_success "oc project logging"
 	os::test::junit::declare_suite_end
 
+    local es_pod=$( get_es_pod es )
```
ewolinetz: im guessing once the uncommented out tests all pass we'll remove these sections?

richm: They are passing now - you want me to remove the checking for the indices now?

ewolinetz: Sure, other than that this seems good to go to me.

PR #732, opened 2017-10-20
remove cpu limit if it exists

jcantrill:
This essentially reverts #723 by only removing cpu limit if one exists and leaves request cpu


richm:
curator flake
/retest

portante:
@jcantrill, can we preserve this test run so we can find out more about the environment in which it fails?

PR #733, opened 2017-10-20
[WIP] Logging rearch proposal doc

ewolinetz:
Not sure who else we should have on this discussion. I'd like to make sure we capture and address all general concerns, we may need to have a follow up to flesh out the specifics for how we will re architect the logging stack

### DIFF
--- a/docs/proposals/logging_rearch.md
+++ b/docs/proposals/logging_rearch.md
@@ -10,29 +10,48 @@ We should also take the time to evaluate how we deploy/configure the logging
 stack and incorporate newer API objects where able, such as using Stateful Sets
 to deploy Elasticsearch instead of Deployment Configs.
 
+The installation restructure is based on the following proposal in the openshift-ansible
+repository: https://github.com/openshift/openshift-ansible/pull/5826.
+
 ## How?
 
 Taking from our experiences with supporting operations and customers, we should
 be sure to incorporate those gotchas while minimizing the number of hacky
 approaches we've included over releases.
 
-We will need to create templates that allow granular deployment options (e.g.
+We will want to create APB ("Ansible Playbook Bundles" [see references]) so we
+can deliver a versioned container that will contain configurations for a specific
+release and bundle that with a specific image version of our components.
+
+An APB can best be thought of as a containerized playbook execution. Within the
+container we would still maintain our roles and a playbook to install and uninstall
+(called "provision" and "deprovision" within the APB naming scheme) our components.
+
+We can create templates that allow granular deployment options (e.g.
 how do we deploy Elasticsearch with stateful sets so that we can have different
 zones and some be masters) while still maintaining the fewest number of template
-files to keep the role simple.
+files to keep our role simple.
 
 Once the template files are created the logging role(s) will need to be updated
 to consume and process these template files in a specific order and provide
 parameters when appropriate.
 
-This would mirror work that was done for the Template Service Broker role [see
-references].
+I think we could also use this time to remove the openshift_logging role, instead
+making it a playbook that would call the different component roles, and leverage
+the configuration structure that Anton has been working on to have a very
+customizable aggregated logging installation.
+
+The added benefit of using the APB model is that we could potentially install
+the logging stack using the Service Catalog on an existing cluster instead of
+relying solely on the Ansible installer.
 
 ## Proposal
 
-Update the deployment pattern for the logging stack to use OCP templates and
-update our openshift-ansible role to use these templates for deploying
-and configuring a complete logging solution using newer API objects.
+Update the deployment pattern for the logging stack to use APBs and OCP templates,
+update our openshift-ansible role to kick off our APB container install/uninstall
+while using these templates for deploying and configuring a complete logging
+solution using newer API objects. We would maintain our APB within this repository,
+similar to what we had done previously with the deployer pod.
 
 # Limitations
 
@@ -48,8 +67,8 @@ components to simplify updating images and ensuring we have the latest version
 without needing to `docker pull` from each node where a logging component is
 deployed.
 
-This would also be a good time to update the openshift-ansible role to ensure
-that we have the latest release of image versions pulled down to the registry.
-
 ## References
 * https://github.com/openshift/openshift-ansible/blob/master/roles/template_service_broker/
+* https://github.com/ansibleplaybookbundle/manageiq-apb
+* https://github.com/ansibleplaybookbundle/ansible-playbook-bundle
+* https://github.com/openshift/openshift-ansible/pull/5826
--- a/docs/proposals/logging_rearch.md
+++ b/docs/proposals/logging_rearch.md
@@ -1,0 +1,55 @@
+# Re architect origin-aggregated-logging to use updated API objects and Templates to install
+
+## Why?
+
+Currently when we develop features or make changes that require configuration
+updates as well as updates to our images we must coordinate those changes in two
+repositories which can lead to things being out of sync or missed.
+
+We should also take the time to evaluate how we deploy/configure the logging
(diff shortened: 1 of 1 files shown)


docs/proposals/logging_rearch.md
```diff
@@ -0,0 +1,55 @@
+# Re architect origin-aggregated-logging to use updated API objects and Templates to install
+
+## Why?
+
+Currently when we develop features or make changes that require configuration
+updates as well as updates to our images we must coordinate those changes in two
+repositories which can lead to things being out of sync or missed.
```
t0ffel: 2 repositories is a big problem, but how will the usage of templates fix that?

richm: Because we will keep the templates in origin-aggregated-logging?  Then all of the configuration/images will be in one place.

ewolinetz: Correct, then the role itself will just `oc process -f | oc apply -f -` the templates we provide. We'd keep the majority of the logic in this repo with our images.

Then we can get a Jenkins job set up to automatically sync our template install files over to openshift-ansible for us.

docs/proposals/logging_rearch.md
```diff
@@ -0,0 +1,55 @@
...
+## How?
+
+Taking from our experiences with supporting operations and customers, we should
+be sure to incorporate those gotchas while minimizing the number of hacky
+approaches we've included over releases.
+
+We will need to create templates that allow granular deployment options (e.g.
+how do we deploy Elasticsearch with stateful sets so that we can have different
+zones and some be masters) while still maintaining the fewest number of template
+files to keep the role simple.
```
t0ffel: in my PR https://github.com/openshift/openshift-ansible/pull/5001 I have logic of deployment of certain node type in `tasks/node_provision.yaml`, so it should be relatively  easy to abstract to use OCP template for "similarly-purposed" nodes (i.e. for masters, or for clients or for data nodes).

docs/proposals/logging_rearch.md
t0ffel: So configmaps will fully be in the templates now, not like it is done here? https://github.com/openshift/openshift-ansible/blob/master/files/origin-components/apiserver-template.yaml#L79

t0ffel:
One of the major concerns with logging deployment somewhat related to having code in 2 repos is the fact that in order to deploy logging user must have root access to the master + preferably the inventory that was used to deploy the cluster.

If we're trying to move logic to origin-aggregated-logging repo, why don't we move all of it here?
So that logging is deployable via playbook that is run in ASB-like way.

ewolinetz:
@t0ffel We want to still keep our role/playbook within the openshift-ansible repository because ultimately that's where all the ansible logic should reside. We would be moving the object definitions from that repository to this one where it actually makes the most sense to keep them so we can have image changes and their necessary config changes be merged in the same PR.

docs/proposals/logging_rearch.md
ewolinetz: Correct

t0ffel:
@ewolinetz why should the logic to deploy logging reside in openshift-ansible though?
elasticsearch/kibana are applications that can potentially be deployed by end-users and shouldn't require admin's credentials. logging requires some extra privileges, and that can be isolated in openshift-ansible, but most of the deployment doesn't.

Moving resource definitions into origin-aggregated-logging will somewhat mitigate the issue with project spanning 2 repos, but it won't fix the issue because logging will not be deployable from origin-aggregated-logging repo. Specifically ansible is needed when: you need to see the entire cluster (all the node types), generate certs, config settings like various quorums, upgrade.

This is not an objection to the proposal, I just think that more can be done.

ewolinetz:
@t0ffel I'm not sure I'm following your concern about "logging will not be deployable from origin-aggregated-logging repo". Is this in relation to the CI tests? Currently we leverage the logging playbook within openshift-ansible to set up our logging stack.

I was unaware that there currently is a concern about needing ansible to install the logging stack so my intent with this proposal was to maintain the installation pattern of using openshift-ansible and the logging role but transfer the responsibility of knowing object definition back to this repository.

Could you expand on that a bit more?

docs/proposals/logging_rearch.md
```diff
@@ -0,0 +1,55 @@
...
+to deploy Elasticsearch instead of Deployment Configs.
+
+## How?
+
+Taking from our experiences with supporting operations and customers, we should
+be sure to incorporate those gotchas while minimizing the number of hacky
+approaches we've included over releases.
+
+We will need to create templates that allow granular deployment options (e.g.
+how do we deploy Elasticsearch with stateful sets so that we can have different
```
jcantrill: I like the idea of stateful sets assuming the way in which the platform provides the storage is something that is compatible with the way ES functions.  I know @portante has had some issues stating that SS are the wrong solution; they allow for the stranding of data if the cluster size is inappropriately reduced in size.  I believe that is a risk but given the complexity of wiring DC's to storage, I believe it is worth taking on.

docs/proposals/logging_rearch.md
ewolinetz: I need to update this, after having a discussion with Fabian to try to address concerns that Peter had, we'll likely move to an Ansible Playbook Bundle approach. We can choose to use OCP templates or continue to use roles. Whichever we are more comfortable with. I think the intent would be to still use templates as we should leverage `oc` being idempotent.

docs/proposals/logging_rearch.md
ewolinetz: I wonder if that's something we can attempt to control. I'd like to document those concerns and see if we can come up with a solution to address them

ewolinetz:
@t0ffel I think if we move to this APB model we could still be able to independently test and install the logging components as it would be a container that installs the logging parts.

jcantrill:
At request of @ewolinetz adding in my comment that we should move to using stateful sets or use an ordinal number for DCs instead of generating the DC suffix.  Working in the OPS rotations demonstrates its difficult to track dcs and associated pods.

portante:
@jcantrill, what BZs or issues have been filed that would be solved by a different method of DC suffix generation?  Or how would stateful sets solve a particular BZ that can't be solved another way?

Do stateful sets provide a way for ES to know what the expected number of master nodes is vs what is the quorum number given the state of a stateful set?  Wouldn't those values have to be provided separately by a modification to an environment variable (or via confimap change) for ALL ES pods?  And given that Elasticsearch does not read those parameters dynamically during the lifetime of the Java process, wouldn't that require a restart of all the ES pods in order to effect that change?

If a user has defined a quorum of 3 master nodes, but scales the stateful set down to 1, won't that cause ES to stop all operations in the cluster until the master nodes are restored?  How is that any different from using DCs mapped to PVCs (1-to-1) today?

Additionally, if a stateful set has to restart from the beginning (fresh bringup of all pods) and a given pod in the series cannot be started properly, won't that prevent the rest of the pods from coming up?

If somebody can demonstrate how Stateful sets can be using to effectively run an Elasticsearch cluster in operation, day-to-day, in the face of PV failures, pod failures, cluster restarts, etc., that is, "day 2 operational considerations solved", then we should consider their use.

As it stands, it seems Anton's work on breaking out master, data, and client nodes is our best bet (overly simplified description below):

- Use DCs for data-less master nodes, configmap defines master node configuration, admin sets replica count based on configmap target number of master nodes
- Use DCs mapped one-to-one to PVCs for data nodes (not running as masters), where any number of data nodes can be created above the configured minimum required (for ensuring data durability and failure-zone representation)
- Use DCs for client nodes (not masters, data-less), admin sets replica count to accommodate traffic

jcantrill:
@ewolinetz consider closing since we are moving to operators for 4.0

PR #734, opened 2017-10-20
Flushing fluentd pos files prior to starting up

ewolinetz:
Attempt to resolve flakes where we are waiting for fluentd to catch up

### DIFF
--- a/hack/testing/util.sh
+++ b/hack/testing/util.sh
@@ -165,6 +165,15 @@ function add_test_message() {
        http://localhost:5601/$1 > /dev/null 2>&1
 }
 
+function flush_fluentd_pos_files() {
+    os::cmd::expect_success "sudo rm -f /var/log/journal.pos"
+    if docker_uses_journal ; then
+        : # done
+    else
+        os::cmd::expect_success "sudo rm -f /var/log/es-containers.log.pos"
+    fi
+}
+
 # $1 - command to call to pass the uuid_es
 # $2 - command to call to pass the uuid_es_ops
 # $3 - expected number of matches
--- a/test/fluentd-forward.sh
+++ b/test/fluentd-forward.sh
@@ -39,6 +39,7 @@ update_current_fluentd() {
   </server>"}]'
 
     # redeploy fluentd
+    os::cmd::expect_success flush_fluentd_pos_files
     os::log::debug "$( oc label node --all logging-infra-fluentd=true )"
     os::cmd::try_until_text "oc get pods -l component=fluentd" "^logging-fluentd-.* Running "
     fpod=$( get_running_pod fluentd )
@@ -69,6 +70,7 @@ create_forwarding_fluentd() {
        --name=filebufferstorage --type=hostPath \
        --path=/var/lib/fluentd/forward --mount-path=/var/lib/fluentd
 
+    os::cmd::expect_success flush_fluentd_pos_files
     os::log::debug "$( oc label node --all logging-infra-forward-fluentd=true )"
 
     # wait for forward-fluentd to start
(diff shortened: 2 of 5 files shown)


hack/testing/util.sh
```diff
@@ -165,6 +165,17 @@ function add_test_message() {
        http://localhost:5601/$1 > /dev/null 2>&1
 }
 
+# $1 the timeout in seconds
+function flush_fluentd_pos_files() {
+    local timeout=${1:-60}
+    os::cmd::try_until_success "sudo rm -f /var/log/journal.pos" $(( timeout * second ))
```
richm: Why is this a `try_until_success`?  Is it because if fluentd is still running, it will keep trying to recreate the file until it is completely shutdown?

richm:
does this require any modification to the code in entrypoint.sh that tries to measure the fluentd journal reading performance by seeing how far fluentd is behind the journal?

hack/testing/util.sh
ewolinetz: I guess it doesn't need to be try_until_success... I made it so we only call this function when fluentd is stopped and before we start it again

ewolinetz:
@richm I haven't looked in that file.... I only searched for where `wait_for_fluentd_to_catch_up` is called and call this function prior to that one being called when starting fluentd up...

ewolinetz:
I've seen this twice now... but both were from the same time on this PR so not sure if its related
```
Running test/mux.sh:239: executing 'curl_es logging-es-data-master-019tn3gq-1-fqflg /project.testproj.*/_count?q=SYSLOG_IDENTIFIER:96ca3a8b-1d27-4c5a-ace0-6ee24899f5ed | get_count_from_json' expecting any result and text '^1$'; re-trying every 0.2s until completion or 600.000s...
FAILURE after 600.150s: test/mux.sh:239: executing 'curl_es logging-es-data-master-019tn3gq-1-fqflg /project.testproj.*/_count?q=SYSLOG_IDENTIFIER:96ca3a8b-1d27-4c5a-ace0-6ee24899f5ed | get_count_from_json' expecting any result and text '^1$'; re-trying every 0.2s until completion or 600.000s: the command timed out
0
... repeated 768 times
```
/test json-file

Trying to see if we still hit the previous flake... 
/test logging

ewolinetz:
I see this happening on some other PRs in openshift-ansible:
```
FAILURE after 600.108s: test/mux.sh:221: executing 'curl_es logging-es-data-master-d8xj6hei-1-21ct8 /project.testproj.*/_count -XPOST -d '{"query":{"bool":{"filter":{"match_phrase":{"_all":"GET /82870ac9-ca5e-49de-951f-ef0386656f1e 404 "}},"must_not":[{"exists":{"field":"docker"}},{"exists":{"field":"kubernetes"}},{"exists":{"field":"CONTAINER_NAME"}},{"exists":{"field":"CONTAINER_ID_FULL"}},{"exists":{"field":"mux_namespace_name"}},{"exists":{"field":"mux_need_k8s_meta"}},{"exists":{"field":"namespace_name"}},{"exists":{"field":"namespace_uuid"}}]}}}' | get_count_from_json' expecting any result and text '^1$'; re-trying every 0.2s until completion or 600.000s: the command timed out
0
```

ewolinetz:
@richm I haven't seen any test failures from waiting for fluentd to catch up at this point. However I only see failures like the above. Do you think it's related to this change? I'm not super familiar with what the mux test is doing

nhosoi:
@ewolinetz, I have another pr to stabilize CI test including mux.
https://github.com/openshift/origin-aggregated-logging/pull/716

I ran your patch with mine and did not see the failure you mentioned.
https://github.com/openshift/origin-aggregated-logging/pull/734#issuecomment-338768963

ewolinetz:
That reminds me... I should update the command used to remove the position files. we don't need to `wait until`

richm:
@ewolinetz please backport this to release-3.6 branch

PR #737, opened 2017-10-21
Fixing a typo in hack/lib/cmd.sh 

nhosoi:
The default value of duration in try_until_success, try_until_text and try_until_not_text is minute, which is replaced with $minute.

### DIFF
--- a/hack/lib/cmd.sh
+++ b/hack/lib/cmd.sh
@@ -108,7 +108,7 @@ minute=$(( 60 * second ))
 function os::cmd::try_until_success() {
 	if [[ $# -lt 1 ]]; then echo "os::cmd::try_until_success expects at least one arguments, got $#"; return 1; fi
 	local cmd=$1
-	local duration=${2:-minute}
+	local duration=${2:-$minute}
 	local interval=${3:-0.2}
 
 	os::cmd::internal::run_until_exit_code "${cmd}" "os::cmd::internal::success_func" "${duration}" "${interval}"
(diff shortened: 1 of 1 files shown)


richm:
What problem are you trying to solve?
```bash
> minute=60
> duration=minute
> echo $(( $duration * 60 ))
3600
> echo "scale=3; $(( duration )) / 1000" | bc
.060
> echo "scale=3; $(( 60 )) / 1000" | bc
.060
```

nhosoi:
@richm, I just thought this line is valid, where minute is a variable with '$', while others do not have '$'.
https://github.com/openshift/origin-aggregated-logging/blob/master/hack/lib/cmd.sh#L123

Do you think L123 is rather invalid?

With or without this pr, the CI tests pass anyway. :)

richm:
> Do you think L123 is rather invalid?

I don't know - the original seems to work as intended - @stevekuznetsov ?

stevekuznetsov:
@nhosoi this is a good patch -- but can you make it in Origin, please, and we can re-vendor to pick it up?

```
$ cat /tmp/nhosoi.sh 
#!/bin/bash

variable=1

function call() {
  local argument="${1:-variable}"
  echo "${argument}"
}

call
call 2

$ /tmp/nhosoi.sh 
variable
2
```

nhosoi:
@stevekuznetsov, sure! 

Before submitting a PR in Origin, let me double check one more time not to make a mistake... :)
```
(1) local duration=${2:-minute}
(2) local duration=${2:-$minute}
```
As @richm pointed out, between (1) and (2), (1) is correct, right?

stevekuznetsov:
No, here is a simple reproducer: 

```sh
#!/bin/bash

variable=1

function current() {
  local argument="${1:-variable}"
  echo "${argument}"
}

current
current 2

function updated() {
  local argument="${1:-"${variable}"}"
  echo "${argument}"
}

updated
updated 2
```

```
$ nohosoi.sh
variable
2
1
2
```

stevekuznetsov:
That is to say -- you are making the correct change in this PR

nhosoi:
@stevekuznetsov, ok!  Thanks for the clarification.  Closing this PR and submitting a new one on Origin.

PR #679, opened 2017-09-25
Add fluent-plugin-remote-syslog gem

sqtran:
This is the work for adding the fluent-plugin-remote-syslog gem to openshift aggregated logging.  

There's one workaround here (7939ef3) that creates symlink to get around an issue https://github.com/openshift/origin-aggregated-logging/issues/604 where the gems cannot be found on the GEM_PATH after installation.

Issue #604, opened 2017-08-22
fluent-plugin-** gems not found


sqtran:
I'm trying to install a new fluentd plugin (fluent-plugin-remote-syslog) into the fluentd openshift image.  I updated the Dockerfile.centos7 file, and the gem appears to install correctly.  The docker build output looks like it was happy, so I start up a container with the newly built image.  When I'm inside the container, I tried doing a fluentd --dry-run, but it errors out.

`2017-08-22 05:02:06 +0000 [error]: dry run failed: Unknown output plugin 'syslog_buffered'. Run 'gem search -rd fluent-plugin' to find plugins`

Upon investigation, I executed a "gem list" to see what was installed.  I see all the gems, so I execute a "gem which <gem name>" on a bunch of gems to figure out if they're loaded correctly.  Most of the gems look like they're working, but all of the fluent-plugin-** ones can't be found.

- fluent-mixin-config-placeholders (0.4.0)
- fluent-plugin-elasticsearch (1.9.5)
- fluent-plugin-kubernetes_metadata_filter (0.29.0)
- fluent-plugin-remote-syslog (1.1)
- fluent-plugin-rewrite-tag-filter (1.5.6)
- fluent-plugin-secure-forward (0.4.5)
- fluent-plugin-systemd (0.0.8)
- fluent-plugin-viaq_data_model (0.0.5) 

Here's the specific error.

`bash-4.2# gem which fluentd-plugin-remote-syslog
ERROR:  Can't find ruby library file or shared library fluentd-plugin-remote-syslog
`

I thought it was something funky with the Docker install of the gems, so I installed the hola gem, and it  returned the expected output when I executed "gem which hola".  So it's not something with the Docker build.

What am I missing here?

richm:
Can you provide your updated Dockerfile.centos7?

sqtran:
Sure, here it is.  

https://github.com/sqtran/origin-aggregated-logging/blob/test/fluentd/Dockerfile.centos7

richm:
I'm at a loss to explain the `gem which` behavior.  fluentd is somehow able to find the plugins at runtime, at least for the ones in the base image.  I've tried using `gem which --debug -v` but it doesn't give any additional information.  I've tried using strace inside the container, but it fails with permission problems.  Is there a ruby/gem expert available?  My next step would be to trace through the gem source code to find out exactly what `which` is doing.

sqtran:
[...]

PR #742, opened 2017-10-24
[WIP] Bug 1505683 - fluentd downstream vendor record_modifier

wozniakjan:
Vendoring the missing record_modifier to downstream image. The #722 PR introduced record_modifier to upstream Dockerfile but not to the downstream.

TODO - needs the hack like remote_syslog has where copies/links the remote_syslog ruby files under /etc/fluent/plugin/
https://github.com/openshift/origin-aggregated-logging/issues/604
https://github.com/openshift/origin-aggregated-logging/pull/679


richm:
@wozniakjan please cherry pick this to master so we can get it into 3.7

PR #744, opened 2017-10-24
Bug 1505273 fluentd failed to load plugin when remote_syslog was enabled

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1505273
The syslog plugin is installed in a different location when using
RPM packaging.
/test

### DIFF
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -235,7 +235,21 @@ fi
 
 if [[ "${USE_REMOTE_SYSLOG:-}" = "true" ]] ; then
     # The symlink is a workaround for https://github.com/openshift/origin-aggregated-logging/issues/604
-    ln -s /opt/app-root/src/gems/fluent-plugin-remote-syslog*/lib/fluentd/plugin/*.rb /etc/fluent/plugin/
+    found=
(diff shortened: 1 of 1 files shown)


fluentd/run.sh
```diff
@@ -235,7 +235,21 @@ fi
...
+        fi
+    done
+    if [ -z "${found:-}" ] ; then
+        # not found in rpm location - look in alternate location
+        for file in /opt/app-root/src/gems/fluent-plugin-remote-syslog*/lib/fluentd/plugin/*.rb ; do
+            if [ -f "$file" ] ; then
+                ln -s $file /etc/fluent/plugin/
+            fi
+        done
+    fi
```
nhosoi: Do we have some error report if the plugin does not exist in the both places?  

If it is guaranteed to be found at the either location, please ignore this comment. :)

richm: There are no guarantees in software :-)

The error report will be "no such plugin - remote_syslog" from fluentd

nhosoi: I thought it could be guaranteed by the gem/rpm dependency...  But the error message from fluentd looks clear enough. :)

/lgtm

PR #745, opened 2017-10-25
Bug 1505683 - fluentd downstream vendor record_modifier

wozniakjan:
Vendoring the missing record_modifier to downstream image. The #622 PR
introduced record_modifier to upstream Dockerfile but not to the downstream.

This is porting forward the https://github.com/openshift/origin-aggregated-logging/pull/742


richm:
/test json-file
flake

wozniakjan:
looks like the failing test is always the same: 
- https://github.com/wozniakjan/origin-aggregated-logging/blob/bz_1505683/fluentd/record-modifier/test/mux.sh#L407

I would like to take some time and try to find you why, because this PR doesn't change anything in upstream images or tests, only in downstream image and that is not used or built for CI tests AFAIK.

PR #747, opened 2017-10-26
Bug - forward docker build image exit status

wozniakjan:
The '!' is a reserved word and if it precedes a pipeline, it reverses the exit status of the pipeline. In this case, 'return "$?"' was always 0 inside of the 'if' statement.

https://linux.die.net/man/1/bash - section about pipelines
http://tldp.org/LDP/abs/html/exit-status.html#EXITSTATUSREF

### DIFF
--- a/hack/lib/build/images.sh
+++ b/hack/lib/build/images.sh
@@ -127,8 +127,10 @@ function os::build::image::internal::docker() {
 	local extra_tag="${3-}"
 	local options=()
 
-	if ! docker build ${OS_BUILD_IMAGE_ARGS:-} -t "${tag}" "${directory}"; then
-		return "$?"
+	docker build ${OS_BUILD_IMAGE_ARGS:-} -t "${tag}" "${directory}"
+	local ret="$?"
+	if [[ $ret -ne 0 ]]; then
+		return "$ret"
 	fi
 
 	if [[ -n "${extra_tag}" ]]; then


jcantrill:
@stevekuznetsov can this go here or does it need to go upstream?

hack/lib/build/images.sh
```diff
@@ -127,12 +127,14 @@ function os::build::image::internal::docker() {
 	local extra_tag="${3-}"
 	local options=()
 
-	if ! docker build ${OS_BUILD_IMAGE_ARGS:-} -t "${tag}" "${directory}"; then
-		return "$?"
+	docker build ${OS_BUILD_IMAGE_ARGS:-} -t "${tag}" "${directory}"
```
stevekuznetsov: Because we're running with `set -o errexit` on this will change the control flow. The bug is real, though -- correct fix IMO is just to `return 1` in the `if`.

stevekuznetsov:
Should go into Orign

hack/lib/build/images.sh
wozniakjan: yeah, that should work as well. I thought the actual return code was useful so I wanted to forward it as was probably originally intended.

stevekuznetsov: Sorry, I mis-remembered. The `set -o errexit` is _not_ on, so we need `if` to force the return. The return code not important, so `return 1` should be fine. See commit f661ac9d1cac752d19c6c251cfa02a7b630b7e98 in Origin.

wozniakjan:
@stevekuznetsov I created a PR to Origin - https://github.com/openshift/origin/pull/17066. Should this one be closed without merge or update to be exactly same as the one in Origin and then merged?

stevekuznetsov:
Let's close this one and re-vendor Origin when it merges.

/close

wozniakjan:
@stevekuznetsov this fix probably never got propagated to this repo (or re-vendored)

the build jobs still succeed even if a build of an image fails
https://ci.openshift.redhat.com/jenkins/job/push_origin_aggregated_logging_release/65/consoleFull
https://ci.openshift.redhat.com/jenkins/job/build-and-release-latest-origin-aggregated-logging/170/consoleFull

PR #748, opened 2017-10-26
Stream output of docker to prevent travis timeout

wozniakjan:
Originally, the output of `docker build` was stored in a variable and only after the image was built, sent to STDOUT. Travis has 10 minute kill period when nothing goes to STDOUT which lead to premature test failures if building of the image took longer (occasionally happened with Elasticsearch image).

Also adds a timestamp to each output line to better evaluate time requirements of image builds.

### DIFF
--- a/hack/lib/build/images.sh
+++ b/hack/lib/build/images.sh
@@ -32,21 +32,17 @@ function os::build::image() {
 	fi
 
 	local result=1
-	local image_build_log
-	image_build_log="$( mktemp "${BASETMPDIR}/imagelogs.XXXXX" )"
 	for (( i = 0; i < "${OS_BUILD_IMAGE_NUM_RETRIES:-2}"; i++ )); do
 		if [[ "${i}" -gt 0 ]]; then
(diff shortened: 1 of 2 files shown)


jcantrill:
@stevekuznetsov Does this change go here or in an upstream repo that gets vendored in?

stevekuznetsov:
Is Travis used for merge gating in this repo? The change should go into Origin first but before we tweak that code I just want to make sure we need the change.

wozniakjan:
we don't need it, but recently I had a few travis builds fail on a PR because `docker build` for elasticsearch was taking over 10 minutes. Travis didn't see anything on STDOUT and therefore killed the job. It is always possible to rerun travis but having this could help provide better feedback from our CI I think.

stevekuznetsov:
We should just turn off Travis for this repo if it is not being used, we do not gain much from this and it would make parallel builds have poor output. The current design was to support parallel image builds.

richm:
Here is the travis config: https://github.com/openshift/origin-aggregated-logging/blob/master/.travis.yml

If we feel this isn't useful, then someone submit a PR to turn this off.

richm:
either close this PR, or change it to disable the travis CI

PR #749, opened 2017-10-26
do not remove es-containers.log.pos; wait for journal.pos

richm:
Do not remove es-containers.log.pos - by default, fluentd will
read from the head of this file.  If the pos file is removed,
all of the container json-file log files will be read in their
entirety again.
https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/generate_throttle_configs.rb#L109

After journal.pos has been removed, wait for fluentd to create it
again before writing test data to be traced.  Otherwise, fluentd will
never see it, and it will never show up in Elasticsearch.

Call wait_for_fluentd_ready from wait_for_fluentd_to_catch_up, and
do not call wait_for_fluentd_ready when using wait_for_fluentd_to_catch_up

Clean up the utf8 test.

Add additional debug information to the mux test which is dumped to
an artifact file.
/test

### DIFF
--- a/hack/testing/util.sh
+++ b/hack/testing/util.sh
@@ -167,11 +167,6 @@ function add_test_message() {
 
 function flush_fluentd_pos_files() {
     os::cmd::expect_success "sudo rm -f /var/log/journal.pos"
-    if docker_uses_journal ; then
-        : # done
-    else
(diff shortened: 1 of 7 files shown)


richm:
@jcantrill this should help with your mux test flakes

PR #751, opened 2017-10-27
bug 1503404. Turn of api watch to reduce fluentd memory

jcantrill:
This PR:

* Turns off watch completely of pod and namespaces
* Allows configuration of watch enablement via env var
* Allows configuration of cache size via env var


jcantrill:
@portante done

PR #755, opened 2017-10-27
[3.6] CI test improvements/stabilization (backport #716)

nhosoi:
entrypoint.sh - Allowing SUITE to have multiple tests. E.g.,
                SUITE="test-mux.sh test-viaq-data-model.sh"

viaq-data-model.sh - Checking MUX_CLIENT_MODE=maximal as a string.

mux.sh - In the file buffer's persistency test, replacing "sleep ##"
         with the code checking the file buffers' existency.
         When ops_cluster is enabled, ES_HOST and OPS_HOST should not
         have the same temporary test name.
         Plus making sure the mux pod is up after restarted.

Making FLUENTD_WAIT_TIME configurable.

(cherry picked from commit 1383952db2df4e0f51efc7f02863932479bcf159)

### DIFF
--- a/test/fluentd-forward.sh
+++ b/test/fluentd-forward.sh
@@ -6,7 +6,7 @@ source "$(dirname "${BASH_SOURCE[0]}" )/../hack/lib/init.sh"
 source "${OS_O_A_L_DIR}/hack/testing/util.sh"
 os::util::environment::use_sudo
 
-FLUENTD_WAIT_TIME=$(( 2 * minute ))
+FLUENTD_WAIT_TIME=${FLUENTD_WAIT_TIME:-$(( 2 * minute ))}
 
 os::test::junit::declare_suite_start "test/fluentd-forward"
 
--- a/test/mux.sh
+++ b/test/mux.sh
@@ -16,7 +16,7 @@ else
     exit 0
 fi
 
-FLUENTD_WAIT_TIME=$(( 2 * minute ))
+FLUENTD_WAIT_TIME=${FLUENTD_WAIT_TIME:-$(( 2 * minute ))}
 
 os::test::junit::declare_suite_start "test/mux"
 
@@ -312,7 +312,9 @@ write_and_verify_logs() {
 }
 
 reset_ES_HOST() {
+    muxpod=$( get_running_pod mux )
     os::cmd::expect_success "oc set env dc logging-mux $1 $2"
+    os::cmd::try_until_failure "oc get pod $muxpod"
     os::log::debug $( oc get pods -l component=mux )
     oc rollout status -w dc/logging-mux # wait for mux to be redeployed
     os::cmd::try_until_text "oc get pods -l component=mux" "^logging-mux-.* Running "
@@ -322,6 +324,11 @@ reset_ES_HOST() {
 cleanup() {
     local return_code="$?"
     set +e
+
+    # In case test failed in Test case FILE_BUFFER_STORAGE_TYPE: $MUX_FILE_BUFFER_STORAGE_TYPE 
+    # reset ES_HOST and OPS_HOST
+    reset_ES_HOST $ES_HOST_BAK $OPS_HOST_BAK
+
     if [ $return_code = 0 ] ; then
         mycmd=os::log::info
     else
@@ -399,6 +406,9 @@ if [ "$MUX_FILE_BUFFER_STORAGE_TYPE" = "pvc" ]; then
     os::log::debug "$( oc get pvc )"
 fi
 
+ES_HOST_BAK=$( oc set env --list dc/logging-mux | grep \^ES_HOST= )
+OPS_HOST_BAK=$( oc set env --list dc/logging-mux | grep \^OPS_HOST= )
+
 # make sure fluentd is working normally
 fpod=$( get_running_pod fluentd )
 wait_for_fluentd_to_catch_up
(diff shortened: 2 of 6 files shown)


test/fluentd-forward.sh
```diff
@@ -42,6 +42,7 @@ update_current_fluentd() {
 
     # redeploy fluentd
     os::log::debug "$( oc label node --all logging-infra-fluentd=true )"
+    os::cmd::try_until_success "oc rollout status -w ds/logging-fluentd"
```
richm: I don't think you can use rollout with a daemonset:
```bash
$ oc label node --overwrite --all logging-infra-fluentd-
node "ip-172-18-1-3.ec2.internal" labeled
$ oc get pods | grep fluentd
[origin@ip-172-18-1-3 artifacts]$ oc rollout status ds/logging-fluentd && echo success
daemon set "logging-fluentd" successfully rolled out
success
```
I think you can just omit this line - I think the following line is sufficient to know that fluentd is running.

test/mux.sh
```diff
@@ -245,15 +258,23 @@ write_and_verify_logs() {
 }
 
 reset_ES_HOST() {
-    oc set env dc logging-mux $1 $2
+    muxpod=$( get_running_pod mux )
+    os::cmd::expect_success "oc set env dc logging-mux $1 $2"
     os::cmd::try_until_failure "oc get pod $muxpod"
+    os::log::debug $( oc get pods -l component=mux )
+    oc rollout status -w dc/logging-mux # wait for mux to be redeployed
     os::cmd::try_until_text "oc get pods -l component=mux" "^logging-mux-.* Running "
```
richm: This conflicts with https://github.com/openshift/origin-aggregated-logging/pull/754

test/fluentd-forward.sh
nhosoi: Hmmm, the master branch already has it.  What shall we do?
https://github.com/openshift/origin-aggregated-logging/blob/master/test/fluentd-forward.sh#L49

Let me double check that mux is ok since mux uses dc?
https://github.com/openshift/origin-aggregated-logging/blob/master/test/mux.sh#L318

richm: it should be fine to use `oc rollout` with `dc/logging-mux`.

We should fix master branch then because we cannot use `oc rollout` with `ds/logging-fluentd`

nhosoi: @richm,
> We should fix master branch then because we cannot use oc rollout with ds/logging-fluentd

I've submitted one line pr for the master branch...
https://github.com/openshift/origin-aggregated-logging/pull/756

test/mux.sh
```diff
@@ -16,10 +16,14 @@ else
...
 fi
 
-FLUENTD_WAIT_TIME=$(( 2 * minute ))
+FLUENTD_WAIT_TIME=${FLUENTD_WAIT_TIME:-$(( 2 * minute ))}
 
 os::test::junit::declare_suite_start "test/mux"
 
+ARTIFACT_DIR=${ARTIFACT_DIR:-${TMPDIR:-/tmp}/origin-aggregated-logging}
+
+os::log::debug ARTIFACT_DIR is $ARTIFACT_DIR
```
richm: remove these - just use the default ARTIFACT_DIR set by init.sh

richm: Actually, just rebase on top of https://github.com/openshift/origin-aggregated-logging/pull/754 now that it has merged

nhosoi: Yep, I revived it... :p  Sorry, removing it again.

test/mux.sh
```diff
@@ -408,35 +418,64 @@ if [ "$MUX_FILE_BUFFER_STORAGE_TYPE" = "pvc" -o "$MUX_FILE_BUFFER_STORAGE_TYPE"
...
+    os::log::debug "$( oc logs $muxpod )"
+
     # kibana logs with kibana container/pod values
     myproject=project.logging
-    mymessage=$uuid_es
-    os::cmd::try_until_success "curl_es $es_pod /${myproject}.*/_count?q=message:$mymessage | get_count_from_json | grep -q 1" "$(( 10*minute ))"
+    mymessage="GET /${uuid_es} 404 "
+    qs='{"query":{"match_phrase":{"message":"'"${mymessage}"'"}}}'
+    os::log::debug "Check kibana log - message \"${mymessage}\""
+    os::cmd::try_until_success "curl_es $es_pod /${myproject}.*/_count -XPOST -d '$qs' | get_count_from_json | grep -q 1" "$(( 10*minute ))"
```
richm: do you want to test if the output from count contains `1` (e.g. `100`, `21` all match `grep -q 1`) or do you want to match exactly `1` e.g. `grep -q '^1$'`?

test/mux.sh
```diff
@@ -408,35 +418,64 @@ if [ "$MUX_FILE_BUFFER_STORAGE_TYPE" = "pvc" -o "$MUX_FILE_BUFFER_STORAGE_TYPE"
...
 
     myproject=.operations
     mymessage=$uuid_es_ops
-    os::cmd::try_until_success "curl_es $es_ops_pod /${myproject}.*/_count?q=message:$mymessage | get_count_from_json | grep -q 1" "$(( 10*minute ))"
+    qs='{"query":{"term":{"systemd.u.SYSLOG_IDENTIFIER":"'"${mymessage}"'"}}}'
+    os::log::debug "Check system log - SYSLOG_IDENTIFIER \"${mymessage}\""
+    os::cmd::try_until_success "curl_es $es_ops_pod /${myproject}.*/_count -XPOST -d '$qs' | get_count_from_json | grep -q 1" "$(( 10*minute ))"
```
richm: see above

test/mux.sh
nhosoi: Good catch!  Pushed the fix.  (Adding the fix to pr#756 for master, too)

PR #758, opened 2017-11-01
proposal: updating upstream images

wozniakjan:
The proposal for image tags based on released branch merged PRs.

New CI job prepared for PR after this proposal is accepted: 
https://github.com/wozniakjan/aos-cd-jobs/blob/aggregated_logging/push_images/sjb/config/test_cases/push_origin_aggregated_logging_release.yml

### DIFF
--- a/docs/proposals/updating_upstream_images.md
+++ b/docs/proposals/updating_upstream_images.md
@@ -1,0 +1,62 @@
+Updating upstream image tags for supported branches
+---------------------------------------------------
+Our logging team received a few issue and bug reports for a stable upstream 
+branch, where the proper code was already merged in github, but propagating
+the code to upstream images was out of sync with openshift ansible releases.
+
+Upstream image tags
+-------------------
+For upstream images, we have the `latest` image tag and then image tag based
+on the `git tag` (for example `v3.6.1`). The `latest` is frequently updated
+when we merge a PR to the `master` branch, the `v3.6.1` tag is created once and
+does not get updated. For downstream images, we also have updating tags based
+on `git branch` (for example branch `rhaos-3.6-rhel-7` to image tag `v3.6`) 
+and these get updated by our release engineers.
+
+I would like to propose similar updating tags for the upstream images based on 
+the `git branch`.
+
+Proposal
+--------
+We would like to keep the original workflow for `master` and `git tag` for
+upstream image tags and add a separate workflow for when we merge a PR to
+any release branch. The latest updates in our supported branches will be
+tracked by floating image tags. These tags will get updated each time a PR
+is merged to a certain branch, for example if we merge a PR to `release-3.6`
+the image tag `v3.6` will also get updated and point to the latest image build.
+
+This will work together with our current workflow. If we merge a PR to `master`,
+we will still update the `latest` image tag. If we merge a PR to `v3.6` and
+tag the commit `v3.6.1`, we will still get a new image tag `v3.6.1` and at
+the same time, update the `v3.6` tag.
+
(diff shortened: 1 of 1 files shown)


docs/proposals/updating_upstream_images.md
```diff
@@ -0,0 +1,54 @@
...
+We would like to keep the original workflow for `master` and `git branch` for
+upstream image tags and add a separate workflow for when we merge a PR to
+any release branch.
+
+Mapping from git to the image tag:
+```
+git                   | image tag
+----------------------|--------------------
+branch - master       | updateable - latest
+branch - release-vX.Y | updateable - vX.Y         <--- proposed new image tag
```
richm: the branch name is just `release-X.Y` e.g. `release-3.6`, `release-1.5`, etc.  But the image tag has the `v`

docs/proposals/updating_upstream_images.md
```diff
@@ -0,0 +1,54 @@
...
+We would like to keep the original workflow for `master` and `git branch` for
+upstream image tags and add a separate workflow for when we merge a PR to
+any release branch.
+
+Mapping from git to the image tag:
+```
+git                   | image tag
+----------------------|--------------------
+branch - master       | updateable - latest
+branch - release-X.Y  | updateable - vX.Y         <--- proposed new image tag
```
ewolinetz: So the thought here is that `vX.Y` is a floating tag -- it can point to either the latest tagged release of `vX.Y.Z` or the last merge from `release-X.Y` ?

richm: yes, that's my understanding - that's the way downstream images work as well, so it would be consistent

wozniakjan: exactly, that is what I was trying to describe. Do you have any suggestions how to improve the proposal and make the wording clearer so it is easier to understand?

jcantrill: Consider explicitly noting this is a floating tag

docs/proposals/updating_upstream_images.md
```diff
@@ -0,0 +1,54 @@
...
+For upstream images, we have the `latest` image tag and then image tag based
+on the `git tag` (for example `v3.6.1`). The `latest` is frequently updated
+when we merge a PR to the `master` branch, the `v3.6.1` tag is created once and 
+does not get updated. For downstream images, we also have updating tags based
+on `git branch` (for example branch `rhaos-3.6-rhel-7` to image tag `v3.6`) 
+and these get updated by our release engineers.
+
+I would like to propose similar updating tags for the upstream images based on 
+the `git branch`.
+
```
ewolinetz: To clarify the table below you could add something like the following here:
"We will create a floating tag that matches the pattern `vX.Y` that will be updated whenever we push to its corresponding branch or when a `vX.Y.Z` tag is created. For example, if we push an update to `release-3.6` then we would also rebuild and pushing a update to the tag `v3.6`. If we create a github tag `v3.6.5` then we would push a new image tagged `v3.6.5` and also point `v.3.6` to this newly created image tag since it should contain the latest 3.6 change."

docs/proposals/updating_upstream_images.md
wozniakjan: I tried to explicitly state that with "updateable", will swap that for "floating"

docs/proposals/updating_upstream_images.md
wozniakjan: Thanks for the review, added very similar text to what you recommended and modified the table so its easier to read.

Issue #769, opened 2017-11-07
[warn]: rewrite_tag_filter: [DEPRECATED] Use <rule> section instead of rewriterule1


nhosoi:
Even though Dockerfile.centos7 sets up the correct version for fluent-plugin-rewrite-tag-filter as follows
` 'fluent-plugin-rewrite-tag-filter:<2.0.0' \`
I started getting the warning If I test the master branch.
` [warn]: rewrite_tag_filter: [DEPRECATED] Use <rule> section instead of rewriterule1`

This diff cleans up the warnings.
https://github.com/openshift/origin-aggregated-logging/compare/master...nhosoi:rewrite_tag_filter
But according to the fluent-plugin-rewrite-tag-filter README, the change is required for fluentd >= v0.14.2.
https://github.com/fluent/fluent-plugin-rewrite-tag-filter

Are we migrating to Fluentd v0.14.x?  Or is this accidental?

richm:
We are not migrating yet to fluent 0.14
I'm not sure why we are getting the warning, but the code seems to work
Is it possible that there is a new version of rewrite_tag_filter 1.x.y _and_ a new 2.0.0 version, which both have the `<rule>` feature?

nhosoi:
@richm, yes, that was it!
```
Date:   Mon Nov 6 09:52:13 2017 +0900
    Release v1.6.0
```
And v1.6.0 code has the warning:
` log.warn "rewrite_tag_filter: [DEPRECATED] Use <rule> section instead of #{k}"`
Now, should we switch to <rule> now or wait for the downstream to catch up?

josefkarasek:
Did not find any useful info what's in 1.6 release. Please let me know if you find something.
In the meantime I'm in favor for locking to 1.5

jcantrill:
@richm Can we close this because of #773

richm:
closed - fixed by https://github.com/openshift/origin-aggregated-logging/pull/773

PR #771, opened 2017-11-08
escaping "." for regex

ewolinetz:
our logging test isn't escaping regex characters for `os::cmd::expect_success_and_not_text` which is causing it to fail as below:

```
FAILURE after 0.436s: test/zzz-correct-index-names.sh:29: executing 'curl_es logging-es-data-master-1mn7xrgj-1-2lvwp /_cat/indices' expecting success and not text 'project.openshift.': the output content test failed
Standard output from the command:
green open .kibana.95c4a1f66e0e565b219bbd5ba5bad200116351c3                                          1 0    3 0 35.3kb 35.3kb 
green open .kibana                                                                                   1 0    1 0  3.1kb  3.1kb 
green open project.logging.44059e9a-c4c0-11e7-9de3-0e79b72cd796.2017.11.08                           1 0 2179 0  1.6mb  1.6mb 
green open project.mux-undefined.8d3953d4-c4c1-11e7-9de3-0e79b72cd796.2017.11.08                     1 0   34 0 62.8kb 62.8kb 
green open project.openshift-template-service-broker.cdba30b1-c4c0-11e7-9de3-0e79b72cd796.2017.11.08 1 0  168 0  216kb  216kb 
green open .kibana.d033e22ae348aeb5660fc2140aec35850c4da997                                          1 0    6 1 59.7kb 59.7kb 
green open project.kube-service-catalog.a57bff2e-c4c0-11e7-9de3-0e79b72cd796.2017.11.08              1 0 6614 0  6.7mb  6.7mb 
green open project.openshift-ansible-service-broker.c2d1db49-c4c0-11e7-9de3-0e79b72cd796.2017.11.08  1 0   13 0 80.1kb 80.1kb 
green open .kibana.d5675a53ada6b361fb95c821bf35696b926d5acf                                          1 0    3 0 35.3kb 35.3kb 
green open .searchguard.logging-es-data-master-1mn7xrgj  
```

PR #773, opened 2017-11-09
To avoid DEPRECATED warning, set 'fluent-plugin-rewrite-tag-filter:<1.6.0'

nhosoi:
Sample warning:
[warn]: rewrite_tag_filter: [DEPRECATED] Use <rule> section instead of rewriterule1

Discussed on:
https://github.com/openshift/origin-aggregated-logging/issues/769


nhosoi:
The test failures should be solved once #771 'escaping "." for regex' is merged.

PR #777, opened 2017-11-10
if only 1 output, just remove output-operations.conf so it will not match ops logs

richm:
If there is no @OUTPUT tag in fluent.conf, and
if there is only one output (i.e. not using ops cluster), then we need
all of the output logs to go to the `match **` in output-applications.conf
instead of the matches for the various types of operations logs in
output-operations.conf.  So, just remove output-operations.conf in that
case.

NOTE WELL: This means that if you want to send operations logs and
applications logs to different destinations, out of the box, it will
not work.  There will be some customization of fluent.conf in the
configmap required.

(cherry picked from commit 938f8f5bc7c6384a57c44c72834ffb718c4c5900)
/test


richm:
cherry pick of https://github.com/openshift/origin-aggregated-logging/pull/774
not sure why automated cherry pick failed, so had to do this one manually

PR #784, opened 2017-11-13
Bug 1511925 Can not login Kibana, Kibana error:[security_exception] no permissions for indices:data/read/mget

richm:
Need these backports to fix https://bugzilla.redhat.com/show_bug.cgi?id=1511925

bug 1476062. Allow write to kibana index
bug 1491227. Modify request if user has back slash
bug 1490719. Operations missing from .all alias
bug 1463436. Allow defining cluster perms to allow export of ui objects


richm:
We don't have CI for 1.5, so if you approve, just go ahead and merge

PR #787, opened 2017-11-16
Bug 1513254 - logging-fluentd performance regression between 3.7.0-0.158.0 and 3.7.7 + message loss.

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1513254
The utf8 filter causes the peformance to be cut in half.  Disable
it by default.  If you need to use it, and can pay the performance
penalty,

    oc set env ds/logging-fluentd ENABLE_UTF8_FILTER=true

see also https://github.com/openshift/origin-aggregated-logging/pull/622

/test

### DIFF
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -268,6 +268,11 @@ else
     touch $CFG_DIR/openshift/filter-pre-a-audit-exclude.conf
 fi
 
+if [ "${ENABLE_UTF8_FILTER:-}" != true ] ; then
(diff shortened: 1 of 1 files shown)


ewolinetz:
Should we also create documentation for this?

richm:
@ewolinetz This PR is just a temporary work around.  What I would like to do is to move the utf8 processing to the source plugin e.g. make sure when in_tail or in_systemd reads the data, it converts it to utf8 there.

ewolinetz:
ack
/lgtm

PR #788, opened 2017-11-16
jemalloc support - vendored in - source built in Dockerfile

richm:
jemalloc support
docker build will pull down the released tarball, configure, and
make install
If you want to use jemalloc:

    oc set env ds/logging-fluentd USE_JEMALLOC=true

This will cause fluentd to run with jemalloc.  It is off by default.

This also disables fluentd supervisor mode so that we
are only running the one fluentd process with jemalloc.

You can set jemalloc options by

    oc set env ds/logging-fluentd MALLOC_CONF="stats_print:true,prof:true,...."

see https://github.com/jemalloc/jemalloc/wiki/Getting-Started for more
information
/test

### DIFF
--- a/fluentd/Dockerfile.centos7
+++ b/fluentd/Dockerfile.centos7
@@ -20,12 +20,15 @@ LABEL io.k8s.description="Fluentd container for collecting of docker container l
       io.openshift.tags="logging,elk,fluentd"
 
 USER 0
+
 # activesupport version 5.x requires ruby 2.2
 # iproute needed for ip command to get ip addresses
 RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
     yum install -y --setopt=tsflags=nodocs \
-      bc iproute && \
+      bc iproute \
+      autoconf redhat-rpm-config && \
     yum clean all
+
 RUN mkdir -p ${HOME} && mkdir -p /etc/fluent/plugin && \
     scl enable ${SCL_VERSION} -- \
      gem install -N --conservative --minimal-deps --no-ri --no-doc \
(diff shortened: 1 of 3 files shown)


fluentd/Dockerfile.centos7
```diff
@@ -25,8 +28,15 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
       libcurl-devel \
       make \
       bc \
+      bzip2 \
       iproute && \
     yum clean all
+
+RUN mkdir -p ${HOME} && \
```
jcantrill: In order for us to use downstream, why not actually vendor in the dependency instead of pulling it in via curl?

richm: @jcantrill I don't know if we can do downstream without an rpm package . . . but sure, I can vendor it in - I just didn't want to check in a big tarball into git.  I suppose I could check in the uncompressed source tree, but if I'm going to do that, I should just make it a git submodule

jcantrill: I would prefer not to vendor the source but if that is our only recourse we should do it.  Alternatively, maybe its worth us setting up the necessary repos to properly build it into an rpm.  Is that something you are willing to take on?

richm: @jcantrill for downstream. separate rpm for the rhlog channel, then we just tag those builds in to build the fluentd image

richm:
Are we ok with this?  If so, I'll remove the do-not-merge label, and someone can give it a /lgtm

fluentd/run.sh
```diff
@@ -1,14 +1,14 @@
 #!/bin/bash
 
+fluentdargs="--no-supervisor"
```
portante: Is this necessary to use jemalloc?

jcantrill:
@richm I am find merging this PR as long as we have a story for downstream

richm:
> the purpose of the fluentd arg --no-supervisor

it is problematic for monitoring and metrics to have two separate fluentd processes running in each pod.  It is even more problematic if we want to use jemalloc stats/tracing features, because we will get two sets of stats.  When running in a container, it is much better to let the container runtime (i.e. openshift) manage the process lifecycle rather than an external supervisor process running inside the same pod/container.

richm:
fixes bz 1502764 slow memory leak in logging-mux during steady-state logging.
https://bugzilla.redhat.com/show_bug.cgi?id=1502764

PR #796, opened 2017-11-24
proposal: optional application logging

wozniakjan:


wozniakjan:
[(3) Opt-in Application Logging [epic-rnd]](https://trello.com/c/PXl5qKLM/619-3-opt-in-application-logging-epic-rnd)

docs/proposals/optional_application_loging.md
```diff
@@ -0,0 +1,106 @@
+Optional Application Logging
+============================
+
+The EFK aggregated logging stack by default collects application logs with
+fluentd from STDOUT of each container managed by OpenShift. Currently only way
+of opting out is based on a node label and ansible variable
```
jcantrill: We should expand this to clarify:
* opting out the entire node for logging
* explicitly controlled by operator

docs/proposals/optional_application_loging.md
```diff
@@ -0,0 +1,106 @@
...
+Latest development shifted recommended docker driver from journald to json-file,
+therefore, the proposed implementation has focused on usage of json-file as
+well. Rationale being, journald currently doesn't match performance expectations
+and desired throttling control for larger clusters.
+
+Each container on the host has its own json structured log file
+- `/var/log/containers/[pod]_[namespace]_[container]-[docker_id].log`
+
+Every line is treated as a separate log message and processed in the cascade
+of fluentd filters with currently only supported sink Elasticsearch.
```
jcantrill: `.... of fluentd filters.`
I believe we can remove ref to Elasticsearch as the final sync is immaterial of opt-in logging

docs/proposals/optional_application_loging.md
```diff
@@ -0,0 +1,106 @@
...
+
+#### Dynamic Control of Optional Logging
+
+By 'dynamic control' we understand the ability of an operator or a user to
+opt-in or opt-out certain OpenShift object from aggregated logging without
+having to redeploy fluentd.
+
+This approach has immediate effect, offers very granular level of control,
+is flexible, and doesn't bring any measurable latency for other containers logs.
+It requires additional resources and brings an overhead to the API server, where
```
jcantrill: additional resources?  cpu or mem?

docs/proposals/optional_application_loging.md
```diff
@@ -0,0 +1,106 @@
...
+in the fluentd `k8_metadata_filter` plugin.
+
+#### Dynamic Control of Optional Logging
+
+By 'dynamic control' we understand the ability of an operator or a user to
+opt-in or opt-out certain OpenShift object from aggregated logging without
+having to redeploy fluentd.
+
+This approach has immediate effect, offers very granular level of control,
+is flexible, and doesn't bring any measurable latency for other containers logs.
```
jcantrill: ...doesn't bring any measurable latency...    do we know this enough to be able to state it here?

docs/proposals/optional_application_loging.md
```diff
@@ -0,0 +1,106 @@
...
+It requires additional resources and brings an overhead to the API server, where
+the fluentd input plugin has to `watch` certain API endpoints.
+
+Following schema describes how this can be implemented as an opt-out kind:
```
jcantrill: Following solution describes how this can be implemented as an opt-out strategy

docs/proposals/optional_application_loging.md
```diff
@@ -0,0 +1,106 @@
...
+- a user will label a namespace or a pod with `aggregated_logging_opt_out=true`
```
jcantrill: Write in the present:
a user labels a namespace or a pod with `aggregated_logging_opt_out=true`

jcantrill: How do they label.... 'oc label namespace` ?

docs/proposals/optional_application_loging.md
```diff
@@ -0,0 +1,106 @@
...
+- the fluentd will exclude log files based on this information
+![fluentd optin pipeline](https://raw.githubusercontent.com/wozniakjan/origin-aggregated-logging/proposal/optional_logging/docs/proposals/optin_pipeline.png)
+
+One way how to achieve this could be creating additional layer of symlinks to
+the log files. Currently the log files in `/var/log/containers/` are symlinks
+to files in `/var/log/pods/`, those are symlinks to files in `/var/lib/docker/containers/`,
+and finally those are UTF-8 formatted text files. We could create symlinks to
+`/var/log/containers/*` in a different directory, `/var/log/containers/optin/*`
+for instance. Then a daemon watching the OpenShift APIs and keep the symlinks
```
jcantrill: Where does this daemon live and how is it deployed?  @pweil- Is this a case for a custom controller?

docs/proposals/optional_application_loging.md
```diff
@@ -0,0 +1,106 @@
...
+  - otherwise would have to wait for a [`refresh_interval`](https://docs.fluentd.org/v0.12/articles/in_tail#refresh_interval)
+- in `expand_paths` filter out files matching cached notices
+
+#### Explicit Optional Logging
+
+Users and cluster operators would have to create an agreed precondition upon
+which logging will be gathered. Any change to the existing setup will require
+an action from a cluster operator and possibly redeployment of fluentd.
+
+This approach could be achieved using [`exclude_path`](https://docs.fluentd.org/v0.12/articles/in_tail#exclude_path)
```
jcantrill: Are you saying this can be achieved now?  And it requires a config change and fluentd restart?

docs/proposals/optional_application_loging.md
```diff
@@ -0,0 +1,106 @@
...
+## Comparison
+
+There are multiple criteria worth addressing, following discusses some
+advantages and disadvantages.
+
+#### Opt-in vs. Opt-out
+
+|         | requires changes                  | collect logs by default | upgrade scenario        |
+|---------|-----------------------------------|-------------------------|-------------------------|
+|Opt-in   | openshift-ansible, fluentd image  | no                      | needs preliminary steps |
```
jcantrill: Maybe clarify  what 'needs prelimanary steps' means

docs/proposals/optional_application_loging.md
wozniakjan: I should probably elaborate what I meant by that statement. I wanted to use it in comparison with static control and emphasize the impact on all containers during any change of settings.

the way I imagine static control is:
- set [`exclude_path`](https://docs.fluentd.org/v0.12/articles/in_tail#exclude_path) to something like `/var/log/containers/mypod_test_test-*.log`
- only place, where this takes effect is here
  - https://github.com/fluent/fluentd/blob/v0.12.41/lib/fluent/plugin/in_tail.rb#L167
- we need to redeploy fluentd to acknowledge new settings
- during redeployment, there is a brief period of time when fluentd doesn't collect logs even from  pods not excluded => a change of settings introduces a spike in latency for other containers

to put it in contrast with dynamic control:
- set label `aggregated_logging_opt_out=true` to a container
- the modified fluentd already queries OpenShift API regardless
- it doesn't need to be redeployed => a change of settings doesn't introduce a spike in latency for other containers

I will try to rephrase the original statement to explain this more clearly

docs/proposals/optional_application_loging.md
wozniakjan: The way I tried to implement a prototype was simply having a bash script running in the background inside each fluentd container. It did work but our fluentd json-file pipeline currently tags container logs using their full path and depends on this for further processing on multiple places. To hack around this, I used the `rewrite_tag_filter` back to to the old know container log path just to see it works. It does, but also at that point, I thought this is probably too hackerish and decided to try another approach, modifying the tail plugin to see if that leads to something cleaner.

Custom controller could be very elegant way, it didn't occur to me and I haven't explored that path yet. Another viable approach to the solution could be a separate container in fluentd pod mounting mentioned hostPaths, watching the OpenShift API, and keeping the symlinks up to date.

docs/proposals/optional_application_loging.md
wozniakjan: Something along those lines. Given the `in_tail` plugin config is generated here:
https://github.com/openshift/origin-aggregated-logging/blob/829725d65b7139e79373b49d84c3333597287b3b/fluentd/generate_throttle_configs.rb#L100-L111

and is not included in the fluentd ConfigMap AFAIK, the process I tested is following:
- deploy logging
- modify `generate_throttle_configs.rb` to exclude desired pod with `exclude_path` in the `in_tail` plugin config
- rebuild and redeploy fluentd image

Not entirely trivial but trivial enough so I could do it. Also, I would imagine if we figure aout an elegant way of adding the the `in_tail` config to ConfigMap, that would remove the "rebuild the fluentd image" requirement.

wozniakjan:
@jcantrill thanks for the review! I tried to address all of the issues, rephrased awkward wording where appropriate, additional info where lacked. Let me please know if you find something still missing.

docs/proposals/optional_application_loging.md
richm: As you have discovered, the file path is tied to the fluentd tags which is tied to the workflow.  It will be risky to change that.  Why isn't using `exclude_path` sufficient?  Because it isn't dynamic?

wozniakjan: I was trying to only document multiple options and compare them. But I personally would prefer a dynamic control, as it appears a bit more elegant for the end user (either the `in_tail` plugin changes or additional layer of symlinks). On the other hand, it will require more maintenance from us in the future.

docs/proposals/optional_application_loging.md
```diff
@@ -0,0 +1,144 @@
+Optional Application Logging
+============================
+
+The EFK aggregated logging stack collects application logs from STDOUT of each 
+container managed by OpenShift using fluentd. Currently, only way of opting out
```
ewolinetz: I think this is a little misleading talking about 'opting out' here. Would we want to say something about how there is no granular control currently... logs are all or nothing based on what nodes fluentd is deployed to (which flows into the next sentence).

docs/proposals/optional_application_loging.md
```diff
@@ -0,0 +1,144 @@
...
+container logs are processed and stored using the EFK stack.
+
+## Fluentd pipeline
+
+The container logs can be obtained from one of these supported logging drivers
+- json-file - logs read from `/var/log/containers/*` mounted in fluentd
+              container
+- journald
+
+Latest development shifted recommended docker driver from journald to json-file,
```
ewolinetz: I'd add backticks (\`) around journald and json-file here.

docs/proposals/optional_application_loging.md
```diff
@@ -0,0 +1,144 @@
...
+- `/var/log/containers/[pod]_[namespace]_[container]-[docker_id].log`
+
+Every line is treated as a separate log message and processed in the cascade
+of fluentd filters. A log line doesn't contain OpenShift metadata, this is 
+added to the message in the fluentd `k8_metadata_filter` plugin.
+
+#### Dynamic Control of Optional Logging
+
+By 'dynamic control' we understand the ability of an operator or a user to
+opt-in or opt-out certain OpenShift object from aggregated logging without
```
ewolinetz: example objects we would mark for opting in/out?

docs/proposals/optional_application_loging.md
```diff
@@ -0,0 +1,144 @@
...
+
+It requires additional resources:
+- memory - cache `O(count(namespaces)+count(pods))` + process overhead
+- CPU - parsing notices + process overhead
+
+and also brings overhead to the API server, where the fluentd input plugin has 
+to `watch` `namespaces` and `pods` API endpoints.
+
+Following solution describes how this can be implemented as an opt-out strategy:
+- a user labels a namespace or a pod with `aggregated_logging_opt_out=true`
```
ewolinetz: I actually would suggest we annotate rather than label since labels can be used for flow/deployment control as well, where as I feel that annotations are more appropriate for when a process is watching for them (e.g. annotating a service to have OCP generate a key pair for it, or annotating a route to change its default timeouts).

There is an `oc annotate` command that can be used

t0ffel: shouldn't the label (or annotation) be something more like `io.openshift.logging=true`

docs/proposals/optional_application_loging.md
t0ffel: The daemon must be deployed on each node then?

Have you considered instead specifying "destinations" instead of "source"? i.e.
if namespace has label/annotation directing it to "exclude" logs, then logs go to `/dev/null`.
if namespace has label to include logs, then logs go to default service `logging-es.logging.svc.service.cluster.local`.
if namespace is infrastructure ns, then logs go to ops service `logging-ops-es`.
if user wants to send logs to external service, then logs go to this service specified in the annotation.

i.e. all logs are collected, decision whether to forward them somewhere or not and where to forward to is made within the collector.

wozniakjan: > The daemon must be deployed on each node then?

yes, every mentioned solution affects fluentd only, therefore, must be deployed on each node (the prototype is a part of fluentd container)

> Have you considered instead specifying "destinations" instead of "source"

That originally wasn't the scope of this effort, please read [here](https://docs.google.com/document/d/1m9-LXr-UThs9WhRfkFRbwf1Lppxa5Opw_PYRY1wuDbY/edit#heading=h.s6k4plkzrdji). From my understanding, we wanted a way of dealing with pods that went awry and tried to log excessive amounts of data (pod exclusion) and a feature for openshift.io (namespace exclusion). I was not aware we would like to change the routing logic for `ops` and `non-ops`. But in case we do, I can go back to drawing board and play around with it a bit more. Or maybe it can have its own proposal as a distinct feature.

docs/proposals/optional_application_loging.md
wozniakjan: Yes, you are correct! This is an exact example where annotations should be used also according to kubernetes page about [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/), that I have previously missed.

wozniakjan: of course, that is open for discussion. I selected a string that fell unique and difficult to get into a collision with any other component, but I like your suggestion better. Do we have a document with label/annotation naming convention so I comply with it in my future proposals?

docs/proposals/optional_application_loging.md
t0ffel: on changing the routing loging for `ops` and non `ops` I was maybe going too far.
The key thing I meant was that if destinations were being considered instead of source then 1 daemon per cluster might suffice. That operator would reconfigure the fluentd pods to tweak the log destination

ewolinetz:
Looks good so far, I'd prefer the dynamic approach as it would be the cleanest and quickest approach. I think that goes beyond the scope of this immediate PR though.

docs/proposals/optional_application_loging.md
wozniakjan: Sounds like an interesting concept, but I am not sure I can fully imagine how will I go about implementing it. That cluster-wide manager daemon will set the `exclude_path` in fluentd or will somewhere in the middle of the pipeline throw away certain logs? What will be the discriminating factor for throwing away unwanted logs? A pod annotation or something else?

Also worth mentioning, one of the requirements was to be able to exclude excessive logs as early as possible so an awry pod does not clog the logging pipeline (prevent DoS of aggregated logging)

t0ffel: The exclusion may come on the step where the k8s metadata is already present based on the annotation.
That won't prevent  clogging on the individual nodes (or wherever k8s metadata is attached), but it might prevent problems on elasticsearch ingest.

docs/proposals/optional_application_loging.md
```diff
@@ -0,0 +1,141 @@
...
+    pos_file "#{ENV['JSON_FILE_POS_FILE'] || '/var/log/es-containers.log.pos'}" 
+    time_format %Y-%m-%dT%H:%M:%S.%N%Z 
+    tag kubernetes.* 
+    format json 
+    keep_time_key true 
+    read_from_head "#{ENV['JSON_FILE_READ_FROM_HEAD'] || 'true'}" 
+    exclude_path #{excluded} 
+  </source> 
+  ```
+- rebuild and redeploy fluentd image
```
richm: The image does not have to be rebuilt, just redeployed.

jcantrill:
closing without prejudice

PR #798, opened 2017-11-27
Elasticsearch 5.5.2 added

t0ffel:
Elasticsearch and Kibana version 5.5.2 added.
Search-guard version 5.5.2-16.
Prometheus exporter version 5.5.2.0 is installed from the web.

Norms are removed from `ipaddr` field in templates as required by ES 5.x.
OpenShift Elasticsearch plugin is temporarily removed.
Search-guard permissions temporarily adjusted to allow Kibana to view everything.
Search-guard admin uses client certificates, instead of server certificates.
Disabled Kibana startup optimization.
Various unit test fixes.


elasticsearch/run.sh
```diff
@@ -68,11 +68,9 @@ if [[ "${INSTANCE_RAM:-}" =~ $regex ]]; then
         error "A minimum of $(($MIN_ES_MEMORY_BYTES/$BYTES_PER_MEG))m is required but only $(($num/$BYTES_PER_MEG))m is available or was specified"
         exit 1
     fi
-
-    # Set JVM HEAP size to half of available space
     num=$(($num/2/BYTES_PER_MEG))
-    export ES_HEAP_SIZE="${num}m"
-    info "ES_HEAP_SIZE: '${ES_HEAP_SIZE}'"
+    export ES_JAVA_OPTS="${ES_JAVA_OPTS:-} -Xms${num}m -Xmx${num}m"
```
jcantrill: Did this env var change in 5.x?  In 2.4 settting ES_HEAP_SIZE accomplished the same thing

t0ffel: yes, ES_HEAP_SIZE is gone

elasticsearch/utils/es_seed_acl
```diff
@@ -12,17 +12,18 @@ index=$(python -c "import yaml; print yaml.load(open('${config}'))['searchguard'
...
     -kst JKS \
     -kspass kspass \
-    -ts /etc/elasticsearch/secret/searchguard.truststore \
+    -ts /etc/elasticsearch/secret/truststore \
     -tst JKS \
     -tspass tspass \
     -nhnv \
-    -icl ${DEBUG:+-dg} && [ ${now} -lt ${MAX_WAIT_SEED} ]
+    -cn ${CLUSTER_NAME} \
+    ${DEBUG:+-dg} && [ ${now} -lt ${MAX_WAIT_SEED} ]
```
t0ffel: replica count  = min (3, number of data nodes ) - 1
shard 1

test/fluentd-forward.sh
```diff
@@ -16,7 +16,7 @@ update_current_fluentd() {
 
     # undeploy fluentd
     os::log::debug "$( oc label node --all logging-infra-fluentd- )"
-    os::cmd::try_until_text "oc get daemonset logging-fluentd -o jsonpath='{ .status.numberReady }'" "0" $FLUENTD_WAIT_TIME
+    os::cmd::try_until_text "oc get daemonset -n logging logging-fluentd -o jsonpath='{ .status.numberReady }'" "0" $FLUENTD_WAIT_TIME
```
nhosoi: There are more "oc get daemonset logging-fluentd ..." command lines (20 of them in total) in the test directory.  Do we want to add "-n logging" to all of them?

richm: it shouldn't be necessary - all of the commands should happen in the `logging` namespace

richm: It should not be needed

t0ffel: hm, it didn't work for me w/o '-n logging' there. Might be that manual execution is different from execution via Jenkins?

richm: @t0ffel one of the first things entrypoint.sh does is to `oc project logging` - almost all of the tests use the `oc` command without `-n logging` - if there is a particular place where `oc` requires `-n logging` that is a bug that needs to be fixed asap

nhosoi:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_origin-aggregated-logging/798/test_pull_request_origin_aggregated_logging_ansible/1342/
The above error is caused by elasticsearch-### is already installed in Dockerfile.centos7.  We could remove this line in install.sh or adding || : to the line as follows.
-yum -y install https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-${ES_VER}.rpm
+yum -y install https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-${ES_VER}.rpm || :

nhosoi:
@wozniakjan, 
I'm afraid it may not be a help for your work, but here's what I did.

This is my working branch, which is based on the current latest master branch with 'Revert "bug 1519705. bump openshift-elasticsearch-plugin-2.4.4.20 and…', then applied this PR798. 
https://github.com/openshift/origin-aggregated-logging/compare/master...nhosoi:pr_798-noprep
 (Please note that the rh-maven33 part in your error is in the "bug 1519705" code to be reverted, which is not ready to work with ES5 yet.)

The ansible side is more questionable...  I tried openshift/openshift-ansible#6628 to find out the ES pod not starting for me too, due to this IllegalArgumentException error.
```
[WARN ][o.e.b.ElasticsearchUncaughtExceptionHandler] [logging-es-data-master-2zxhhpys] uncaught exception in thread [main]
org.elasticsearch.bootstrap.StartupException: java.lang.IllegalArgumentException: node settings must not contain any index level settings
```
@jcantrill  gave me this advice on IRC:
jcantril: edit your logging-elasticsearch configmap to remove all the settings shown in the exception
jcantril: ES now will only allow settings in the file it is expecting

So, I got rid of all the arguments which caused IllegalArgumentException.  This is my ansible branch (sorry, it's a bit old.)
https://github.com/openshift/openshift-ansible/compare/master...nhosoi:es5

Using these branches, ES5 successfully builds and starts.  But I'd think I removed too much config params or it's lacking some necessary data(?)/schema(?), fluentd fails to forward the logs to the ES5 as follows: 
```
[warn]: Elasticsearch rejected 729 documents due to invalid field arguments
[error]: Could not push log to Elasticsearch: ... "error"=>{"type"=>"mapper_parsing_exception", "reason"=>"failed to parse", "caused_by"=>{"type"=>"illegal_argument_exception", "reason"=>"object field starting or ending with a [.] makes object resolution ambiguous: []"}}}},
```
Currently, I'm waiting for the valid roles/openshift_logging_elasticsearch/templates/elasticsearch.yml.j2 ready...

PR #803, opened 2017-11-30
Throttle different pos

ewolinetz:
When we generate our throttle config we should not be reusing the same `.pos` file for each config.
Adds test to confirm the files are being generated correctly.

To address https://bugzilla.redhat.com/show_bug.cgi?id=1496261

### DIFF
--- a/fluentd/generate_throttle_configs.rb
+++ b/fluentd/generate_throttle_configs.rb
@@ -26,6 +26,10 @@ def json_pos_file
   ENV['JSON_FILE_POS_FILE'] || '/var/log/es-containers.log.pos'
 end
 
+def get_json_pos_file_name(name)
+  return "/var/log/es-container-#{name}.log.pos"
+end
+
 def get_file_name(name)
   ## file_name follows pattern: gen-#{name}-YYYYMMDD.conf ##
 
@@ -38,24 +42,75 @@ def get_file_name(name)
   return file_name
 end
 
+## Returns the names of all throttle configs for parsing through when reverting
+def get_all_throttle_files()
+  return Dir.glob('/var/log/es-container-*.log.pos')
+end
+
+def move_pos_file_project_entry(source_file, dest_file, project)
+
+  if File.file?(source_file)
+    project_pattern = ".*_#{project}_.*\.log"
+
+    matches = Array.new
+    File.open(source_file) { |file|
+      file.grep(/#{project_pattern}/) { |match| matches << match }
+    }
+
+    update = ""
+    update = File.read(dest_file) if File.file?(dest_file)
+    matches.each { |match|
+      log_file_pattern = Regexp.escape(match.split(" ")[0])
+      update.gsub!(/#{log_file_pattern}.*$\n/, "")
+      ## We are clearing out the line while if it exists and then appending the source_file line
+      update << match
+    }
+    File.open(dest_file, "w") { |file| file.write update }
+
+    ## Remove matches from the old file now that they're safely in the new file
+    update = File.read(source_file)
+    matches.each { |match|
+      log_file_pattern = Regexp.escape(match.split(" ")[0])
+      update.gsub!(/#{log_file_pattern}.*$\n/, "")
+    }
+    File.open(source_file, "w") { |file| file.write update }
+  end
+
+end
+
+## This will copy the pos entries from the throttle configs back to the default pos file
+def revert_throttle()
+  get_all_throttle_files.each { |file_name| move_pos_file_project_entry(file_name, json_pos_file, '.*') }
+end
+
 def seed_file(file_name, project)
 
   if project.eql?('.operations')
     path = DEFAULT_OPS_PROJECTS.map{|p| get_project_pattern(p)}.join(',')
+
+    ## openshift-* is a protected project prefix -- so users would not be able to create
+    ## a project that starts with this. Guard taken to prevent possible collision with a
+    ## user created project 'operations'
+    pos_file = get_json_pos_file_name('openshift-operations')
   else
     path = get_project_pattern(project)
+    pos_file = get_json_pos_file_name(project)
   end
 
   File.open(file_name, 'w') { |file|
-    @log.debug "Seeding #{file_name} with path: '#{path}' and pos_file: '#{json_pos_file}'"
+    @log.debug "Seeding #{file_name} with path: '#{path}' and pos_file: '#{pos_file}'"
     file.write(<<-CONF)
 <source>
   @type tail
   @label @INGRESS
   path #{path}
-  pos_file #{json_pos_file}
+  pos_file #{pos_file}
     CONF
   }
+
+  # Set up the initial pos file in case we had already read from container files
+  move_pos_file_project_entry(json_pos_file, pos_file, project) unless project.eql?('.operations')
+  DEFAULT_OPS_PROJECTS.each{ |ops_project| move_pos_file_project_entry(json_pos_file, pos_file, ops_project) } if project.eql?('.operations')
 end
 
 def write_to_file(project, key, value)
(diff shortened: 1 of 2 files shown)


test/read-throttling.sh
```diff
@@ -25,6 +25,36 @@ oc get daemonset logging-fluentd -o yaml > $saveds
...
+
+start_fluentd() {
+  os::log::debug "$( oc label node --all logging-infra-fluentd=true 2>&1 || : )"
+  os::cmd::try_until_text "oc get pods -l component=fluentd" "^logging-fluentd-.* Running "
+}
+
+check_fluentd_logs_for() {
+  local entries=$@
+
+  fpod=$( get_running_pod fluentd )
```
richm: does fpod need to be local?

ewolinetz: probably wouldn't hurt

richm:
One of the problem is this:
- user has existing es-containers.log.pos file with existing positions from existing pods
- configure throttling for a namespace - will create a brand new empty pos file - since we read from head, there will be a lot of duplicate records for that namespace

I think we need to do this:
- shutdown fluentd
- configure throttling
- before fluentd starts up, parse out entries in es-containers.log.pos for the namespaces to throttle, and "move" those entries from es-containers.log.pos to the new es-containters-namespace-log.pos file.

Then, when "un-throttling", the reverse will need to happen - move the entries from es-containters-namespace-log.pos to es-containers.log.pos

ewolinetz:
Need to figure out logic for throttle vs unthrottle and call the method to move pos entries from one file to another...

fluentd/generate_throttle_configs.rb
```diff
@@ -47,6 +47,36 @@ def get_file_name(name)
...
+    ## Remove matches from the old file now that they're safely in the new file
+    update = File.read(source_file)
+    matches.each { |match|
+      log_file_pattern = Regexp.escape(match.split(" ")[0])
+      update.gsub!(/#{log_file_pattern}.*$\n/, "")
+    }
+    File.open(source_file, "w") { |file| file.puts update }
+  end
+
+end
```
richm: nice!

ewolinetz:
/test logging

Saw this, not sure if its a flake or something more serious...
```
TASK [openshift_logging : include_role] ****************************************
task path: /usr/share/ansible/openshift-ansible/roles/openshift_logging/tasks/install_logging.yaml:308
  File "/usr/lib64/python2.7/multiprocessing/queues.py", line 266, in _feed
    send(obj)
RuntimeError: maximum recursion depth exceeded in cmp
```

richm:
/retest
make sure it works with correct jenkins tests

ewolinetz:
looks like it passed

fluentd/generate_throttle_configs.rb
```diff
@@ -38,24 +47,75 @@ def get_file_name(name)
...
   if project.eql?('.operations')
     path = DEFAULT_OPS_PROJECTS.map{|p| get_project_pattern(p)}.join(',')
+
+    ## openshift-* is a protected project prefix -- so users would not be able to create
+    ## a project that starts with this. Guard taken to prevent possible collision with a
+    ## user created project 'operations'
+    pos_file = get_json_pos_file_name('openshift-operations')
   else
     path = get_project_pattern(project)
+    pos_file = get_json_pos_file_name(project)
```
richm: Is it possible that `project` could contain characters which are not valid in a filename?

ewolinetz: I don't believe so. It could be if a user enters it incorrectly in their config, otherwise the project names themselves are [:alnum:-_] which should be valid for file names

test/read-throttling.sh
```diff
@@ -25,6 +25,35 @@ oc get daemonset logging-fluentd -o yaml > $saveds
...
+  local fpod=$( get_running_pod fluentd )
+
+  for file in ${files}; do
+    os::cmd::expect_success "oc exec $fpod -- test -f $file"
+  done
+}
+
+check_fluentd_pod_file_content_for() {
+  local file=$1
+  local content=$2
```
richm: I think this should be quoted `local content="$2"`

test/read-throttling.sh
```diff
@@ -51,28 +80,35 @@ trap "cleanup" EXIT
...
 os::cmd::expect_success_and_text "oc logs $fpod" "Could not parse YAML file"
 
+# generate a throttle config that properly generates different pos files
+stop_fluentd
+os::log::debug "$( oc patch configmap/logging-fluentd --type=json \
+   --patch '[{ "op": "replace", "path": "/data/throttle-config.yaml", "value": "\
+    test-proj:\n  read_lines_limit: 5\n.operations:\n  read_lines_limit: 5"}]' )"
+start_fluentd
+check_fluentd_pod_for_files '/var/log/es-container-test-proj.log.pos' '/var/log/es-container-openshift-operations.log.pos'
+check_fluentd_pod_file_content_for '/var/log/es-container-openshift-operations.log.pos' '[^ ]*'
```
richm: You're trying to see if the file contains zero or more non-space characters?  Won't that always succeed?  Is there something more concrete that you could search the file for to see if it correctly constructed a pos file?

ewolinetz: Likely... I had thought that the file should contain something if its actually reading in log files. I could specifically check for specific project entries like `_default_`

richm:
looks good - squash into a single commit and i'll lgtm

fluentd/generate_throttle_configs.rb
```diff
@@ -26,6 +26,15 @@ def json_pos_file
   ENV['JSON_FILE_POS_FILE'] || '/var/log/es-containers.log.pos'
 end
 
+def get_json_pos_file_name(name)
+  ## file_name follows pattern /var/log/es-container-#{name}.log.pos
```
jcantrill: why not just 
```
def get_json_pos_file_name(name)
  "/var/log/es-container-#{name}.log.pos"
end

```

fluentd/generate_throttle_configs.rb
```diff
@@ -38,24 +47,75 @@ def get_file_name(name)
...
+def get_all_throttle_files()
+  return Dir.glob('/var/log/es-container-*.log.pos')
+end
+
+def move_pos_file_project_entry(source_file, dest_file, project)
+
+  if File.file?(source_file)
+    project_pattern = ".*_#{project}_.*\.log"
+
+    matches = Array.new()
```
jcantrill: No parens required... or just `[]`

fluentd/generate_throttle_configs.rb
```diff
@@ -38,24 +47,75 @@ def get_file_name(name)
...
+    File.open(source_file) { |file|
```
jcantrill: when using a multiline block, prefer
```
do  |file|
end
```

fluentd/generate_throttle_configs.rb
```diff
@@ -38,24 +47,75 @@ def get_file_name(name)
...
+
+    update = ""
+    update = File.read(dest_file) if File.file?(dest_file)
+    matches.each { |match|
+      log_file_pattern = Regexp.escape(match.split(" ")[0])
+      update.gsub!(/#{log_file_pattern}.*$\n/, "")
+      ## We are clearing out the line while if it exists and then appending the source_file line
+      update << match
+    }
+    File.open(dest_file, "w") { |file| file.puts update }
```
jcantrill: Should this be 'write' instead of 'puts'... might make it more obvious what is happening

ewolinetz: we can probably use `write` over `puts` since we don't necessarily need an extra carriage return at the end of `update`...

ewolinetz:
@richm squashed @jcantrill addressed your formatting comments

richm:
not sure if we need this for 1.5 and 1.4 - if so, you'll have to port the tests to work with the old framework

richm:
https://github.com/openshift/origin-aggregated-logging/pull/841

PR #811, opened 2017-12-05
Include DEBUG log level messages when collecting from journald

lukas-vlcek:
Closes BZ-1505602

### DIFF
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -279,6 +279,13 @@ if type -p jemalloc-config > /dev/null 2>&1 && [ "${USE_JEMALLOC:-true}" = true 
     export LD_PRELOAD=$( jemalloc-config --libdir )/libjemalloc.so.$( jemalloc-config --revision )
 fi
 
+# Include DEBUG log level messages when collecting from journald
+# https://bugzilla.redhat.com/show_bug.cgi?id=1505602
+if [ "${COLLECT_JOURNAL_DEBUG_LOGS:-true}" = true ] ; then
+  rm -f $CFG_DIR/openshift/filter-exclude-journal-debug.conf
+  touch $CFG_DIR/openshift/filter-exclude-journal-debug.conf
+fi
+
 if [[ $DEBUG ]] ; then
     exec fluentd $fluentdargs > /var/log/fluentd.log 2>&1
 else
--- a/test/debug_level_logs.sh
+++ b/test/debug_level_logs.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+# This is a test suite for the debug_level_logs
+
+source "$(dirname "${BASH_SOURCE[0]}" )/../hack/lib/init.sh"
+source "${OS_O_A_L_DIR}/hack/testing/util.sh"
+os::test::junit::declare_suite_start "test/debug_level_logs"
+
+get_logmessage() {
+    logmessage="$1"
+}
+get_logmessage2() {
+    logmessage2="$1"
+}
+
+es_pod=$( get_es_pod es )
+es_ops_pod=$( get_es_pod es-ops )
+es_ops_pod=${es_ops_pod:-$es_pod}
+
+TEST_REC_PRIORITY=debug wait_for_fluentd_to_catch_up get_logmessage get_logmessage2
+qs='{"query":{"term":{"systemd.u.SYSLOG_IDENTIFIER":"'"${logmessage2}"'"}}}'
+os::cmd::expect_success "curl_es ${es_ops_pod} /.operations.*/_count -X POST -d '$qs' | jq '.count > 0'"
--- a/test/viaq-data-model.sh
+++ b/test/viaq-data-model.sh
@@ -215,3 +215,4 @@ if [ -n "$is_maximal" ] ; then
     os::cmd::try_until_failure "oc get pod $muxpod"
     os::cmd::try_until_text "oc get pods -l component=mux" "^logging-mux-.* Running "
 fi
+
--- a/hack/testing/entrypoint.sh
+++ b/hack/testing/entrypoint.sh
@@ -35,6 +35,9 @@ if [[ -z "${USE_DEFAULT_FLUENTD_CPU_LIMIT:-}" && -n "$(oc get ds logging-fluentd
           {"op":"remove","path":"/spec/template/spec/containers/0/resources/limits/cpu"}]'
 fi
 
+# Make CI run with enabled debug logs for journald (BZ 1505602)
+oc set -n logging env ds/logging-fluentd COLLECT_JOURNAL_DEBUG_LOGS=true
+
 # start a fluentd performance monitor
 monitor_fluentd_top() {
     # assumes running in a subshell
--- a/hack/testing/test-debug_level_logs.sh
+++ b/hack/testing/test-debug_level_logs.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+source "$(dirname "${BASH_SOURCE[0]}" )/../lib/init.sh"
+
+exec ${OS_O_A_L_DIR}/test/debug_level_logs.sh
--- a/hack/testing/util.sh
+++ b/hack/testing/util.sh
@@ -194,11 +194,12 @@ function wait_for_fluentd_to_catch_up() {
     local expected=${3:-1}
     local timeout=${TIMEOUT:-600}
     local project=${4:-logging}
+    local priority=${TEST_REC_PRIORITY:-info}
 
     wait_for_fluentd_ready
     add_test_message $uuid_es
     os::log::debug added es message $uuid_es
-    logger -i -p local6.info -t $uuid_es_ops $uuid_es_ops
+    logger -i -p local6.${priority} -t $uuid_es_ops $uuid_es_ops
     os::log::debug added es-ops message $uuid_es_ops
 
     local rc=0


lukas-vlcek:
Documentation is still missing. There is a new opt-in option `COLLECT_JOURNAL_DEBUG_LOGS` to enable this.

lukas-vlcek:
@richm My question would be _if_ and _where_ it would make sense to set default value (`false`?) to `COLLECT_JOURNAL_DEBUG_LOGS`.

richm:
please squash the commits and repush

richm:
> @richm My question would be if and where it would make sense to set default value (false?) to COLLECT_JOURNAL_DEBUG_LOGS.

I think we want to make sure it works well with `true`.  So the next step will be to add some code to the test in `hack/testing/entrypoint.sh` to set this env. var. so we can run the CI tests with this enabled:

```
oc set env ds/logging-fluentd COLLECT_JOURNAL_DEBUG_LOGS=true
```

If it is working well with CI, we'll just make the default be `true` at some point in the future e.g. so that it makes it into 3.9.
@portante

jcantrill:
@lukas-vlcek If you modify the commit comment to be something like:

```
bug 1505602. Include DEBUG log level messages when collecting from journald 
```
The BZ will be updated with the commit hash

lukas-vlcek:
@jcantrill And you've just done that right so I do not have to now?

jcantrill:
@lukas-vlcek we should squash commits and edit your merge comment.  I dont think the bot looks at PR comments but commit history

richm:
/test logging
see if test flake

lukas-vlcek:
@richm @jcantrill I pushed updated code.

lukas-vlcek:
@richm Is there any reason why this command is failing? For example the pod does not exist in these test scenarios?

```
Stack Trace: 
[INFO] 1: hack/testing/entrypoint.sh:39: `oc set env ds/logging-fluentd COLLECT_JOURNAL_DEBUG_LOGS=true`
[INFO]   Exiting with code 1.
```

richm:
> Error from server (NotFound): daemonsets.extensions "logging-fluentd" not found

Ah, sorry, use this:

```
oc set -n logging env ds/logging-fluentd COLLECT_JOURNAL_DEBUG_LOGS=true
```

lukas-vlcek:
@richm Do we want to implement more specific test for this? Like for example run ES query to check that there are any records at the `debug` level for journald use case?

richm:
> Do we want to implement more specific test for this?

Yes, good idea.

There is a function in our test library which adds test records and verifies they are in es - it uses `logger` with `info`: https://github.com/openshift/origin-aggregated-logging/blob/master/hack/testing/util.sh#L190

You can add the ability to set that with an env. var.
```bash
  local priority=${TEST_REC_PRIORITY:-info}
  ...
  logger -i -p local6.${TEST_REC_PRIORITY} -t $uuid_es_ops $uuid_es_ops
```

Then, to verify, something like this: https://github.com/openshift/origin-aggregated-logging/blob/master/test/viaq-data-model.sh#L126

```bash
  TEST_REC_PRIORITY=debug wait_for_fluentd_to_catch_up get_logmessage get_logmessage2
  qs='{"query":{"term":{"systemd.u.SYSLOG_IDENTIFIER":"'"${logmessage2}"'"}}}'
  curl_es ${es_ops_pod} /.operations.*/_search -X POST -d '$qs' | something to parse the record JSON and verify it has the "level" field with a value of "debug"
```

lukas-vlcek:
@richm It would be great to have `jq` available so that specific test checks can be implemented without a need to introduce a new python scripts (AFAIK Python is not good for one-liners). For example in my case I could do something like this:

```bash
TEST_REC_PRIORITY=debug wait_for_fluentd_to_catch_up get_logmessage get_logmessage2
qs='{"query":{"term":{"systemd.u.SYSLOG_IDENTIFIER":"'"${logmessage2}"'"}}}'
os::cmd::expect_success "curl_es ${es_ops_pod} /.operations.*/_count -X POST -d '$qs' | \
                        jq '.count > 0'"
```
Simple and elegant.

fluentd/run.sh
```diff
@@ -279,6 +279,15 @@ if type -p jemalloc-config > /dev/null 2>&1 && [ "${USE_JEMALLOC:-}" = true ] ;
     export LD_PRELOAD=$( jemalloc-config --libdir )/libjemalloc.so.$( jemalloc-config --revision )
 fi
 
+if [ "${USE_JOURNAL:-}" = "true" ] ; then
+  # Include DEBUG log level messages when collecting from journald
+  # https://bugzilla.redhat.com/show_bug.cgi?id=1505602
+  if [ "${COLLECT_JOURNAL_DEBUG_LOGS:-false}" = true ] ; then
```
portante: Shouldn't this default to `true`?  Don't we want to collect everything by default, so that if an admin does not want debug messages collected, they can set `COLLECT_JOURNAL_DEBUG_LOGS` to `false`?

lukas-vlcek:
Do I understand it correctly that CI failed tests failed before they reached the test that was added in this PR?

richm:
/retest
test flake in viaq test?

Sure - we can use jq for a lot of things.

Yes, please make the default to be `true` so we can get a lot of CI testing for that setting - if it proves to be problematic, we can always switch it back to false

test/viaq-data-model.sh
```diff
@@ -215,3 +215,12 @@ if [ -n "$is_maximal" ] ; then
     os::cmd::try_until_failure "oc get pod $muxpod"
     os::cmd::try_until_text "oc get pods -l component=mux" "^logging-mux-.* Running "
 fi
+
+# TEST 6
+# catch logs at the debug level
+TEST_REC_PRIORITY=debug wait_for_fluentd_to_catch_up get_logmessage get_logmessage2
```
richm: This test is consistently failing.  I think this is probably not a good place to add this test.  The viaq test leaves fluentd and mux in a "bad" state, which is reverted in the cleanup method.

I'm not sure which test you should add this to.  It isn't really related to the viaq data model.  I suppose if you want to add this test to this file, it should be the first test in the file, before the test modifies fluentd and mux, but again, this isn't related.

test/viaq-data-model.sh
```diff
@@ -215,3 +215,12 @@ if [ -n "$is_maximal" ] ; then
...
+qs='{"query":{"term":{"systemd.u.SYSLOG_IDENTIFIER":"'"${logmessage2}"'"}}}'
+os::cmd::expect_success "curl_es ${es_ops_pod} /.operations.*/_count -X POST -d '$qs' | \
+                         jq '.count > 0'"
+unset TEST_REC_PRIORITY
```
richm: You do not need to unset it.  When you do
```bash
VAR=value command arg ...
```
the `VAR` is set as an environment variable for the command *only*, and nowhere else.

fluentd/run.sh
lukas-vlcek: Changed to `true`.

test/viaq-data-model.sh
lukas-vlcek: I moved the test to he beginning for now (I made it `TEST 0`) to see if it will pass. Let's find a better place for it before we merge this in...

test/viaq-data-model.sh
lukas-vlcek: ack

richm:
/retest
curator flake?

richm:
It just isn't working.  Not sure if it is because it is in the viaq test.  I suggest moving the test elsewhere.  I strongly encourage you to add this as a separate test, rather than adding it to an existing unrelated test.  When I suggested to take a look at the viaq test, I meant as an example, not as the place to put the new test.

lukas-vlcek:
I moved the test to extra (new) script.

test/debug_level_logs.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# This is a test suite for the debug_level_logs
+
+source "$(dirname "${BASH_SOURCE[0]}" )/../hack/lib/init.sh"
+source "${OS_O_A_L_DIR}/hack/testing/util.sh"
+os::test::junit::declare_suite_start "test/debug_level_logs"
+
```
richm:
```bash
get_logmessage() {
    logmessage="$1"
}
get_logmessage2() {
    logmessage2="$1"
}

TEST_REC_PRIORITY=debug wait_for_fluentd_to_catch_up get_logmessage get_logmessage2
qs='{"query":{"term":{"systemd.u.SYSLOG_IDENTIFIER":"'"${logmessage2}"'"}}}'
```

test/debug_level_logs.sh
```diff
@@ -0,0 +1,10 @@
...
+qs='{"query":{"term":{"systemd.u.SYSLOG_IDENTIFIER":"'"debug"'"}}}'
```
richm: remove this `qs` and use the one above

richm:
```
hack/testing/entrypoint.sh: line 119: /data/src/github.com/openshift/origin-aggregated-logging/hack/testing/test-debug_level_logs.sh: Permission denied
```
both hack/testing/test-debug_level_logs.sh and test/debug_level_logs.sh must be executable - chmod +x.

lukas-vlcek:
Ah... sorry, my bad.

lukas-vlcek:
It is failing due to missing ES plugin in maven repository. See <https://repo.maven.apache.org/maven2/io/fabric8/elasticsearch/openshift-elasticsearch-plugin/> for all available versions. Version `2.4.4.20` is not listed. Does it mean we need to wait until specific version of plugin is officially released?

```
Downloading: https://repo.maven.apache.org/maven2/io/fabric8/elasticsearch/openshift-elasticsearch-plugin/2.4.4.20/openshift-elasticsearch-plugin-2.4.4.20.jar
...
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:2.8:get (default-cli) on project standalone-pom: Couldn't download artifact: Missing:
...
```

richm:
I think we're waiting for the PR from @jcantrill that vendors in 2.4.4.20

lukas-vlcek:
@richm I am discussing this with @jcantrill over IRC now

test/debug_level_logs.sh
```diff
@@ -13,6 +13,10 @@ get_logmessage2() {
     logmessage2="$1"
 }
 
+es_pod=$( get_es_pod es )
+es_ops_pod=$( get_es_pod es-ops )
+es_ops_pod=${es_ops_pod:-$es_pod}
```
lukas-vlcek: Ahhh.... good catch.

richm:
squash these commits into one and I will approve

PR #815, opened 2017-12-06
peg metadatafilter to 0.33 until we fix cache update

jcantrill:


jcantrill:
Close as the #760 already merged

PR #818, opened 2017-12-06
logging CI test flake fixes - mux, multi_tenancy

richm:
get internal log file from mux
have to poll when creating or deleting project/namespace
/test


richm:
This is finally working.  Please review and lgtm if ok

PR #819, opened 2017-12-06
disable mux test until test flakes are fixed

richm:
disable mux test until test flakes are fixed
/test


richm:
This should unblock CI until we can sort out why mux testing has so many problems.

PR #825, opened 2017-12-07
bug 1520629. Use 'orphan' namespace when namespace or namespace_id is…

jcantrill:
… missing

peg metadata filter to less then 1.0.0

### DIFF
--- a/fluentd/Dockerfile
+++ b/fluentd/Dockerfile
@@ -33,7 +33,7 @@ RUN mkdir -p ${HOME} && \
       'elasticsearch-transport:<5' \
       'elasticsearch-api:<5' \
       'elasticsearch:<5' \
-      fluent-plugin-kubernetes_metadata_filter \
+      'fluent-plugin-kubernetes_metadata_filter:<1.0.0' \
       'fluent-plugin-elasticsearch:<2.0.0' \
       'fluent-plugin-systemd:<0.1.0' \
       systemd-journal \
--- a/fluentd/configs.d/openshift/es-copy-config.conf
+++ b/fluentd/configs.d/openshift/es-copy-config.conf
@@ -3,7 +3,7 @@
       host "#{ENV['ES_COPY_HOST']}"
       port "#{ENV['ES_COPY_PORT']}"
       scheme "#{ENV['ES_COPY_SCHEME']}"
-      index_name project.${record['kubernetes']['namespace_name']}.${record['kubernetes']['namespace_id']}.${Time.parse(record['@timestamp']).getutc.strftime(@logstash_dateformat)}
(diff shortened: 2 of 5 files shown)


fluentd/configs.d/openshift/es-copy-config.conf
```diff
@@ -3,7 +3,7 @@
       host "#{ENV['ES_COPY_HOST']}"
       port "#{ENV['ES_COPY_PORT']}"
       scheme "#{ENV['ES_COPY_SCHEME']}"
-      index_name project.${record['kubernetes']['namespace_name']}.${record['kubernetes']['namespace_id']}.${Time.parse(record['@timestamp']).getutc.strftime(@logstash_dateformat)}
+      index_name project.${begin record['kubernetes']['namespace_name']+'.'+record['kubernetes']['namespace_id']; rescue; $log.error('record is missing kubernetes or kubernetes.namespace_id:' + record.to_s); '.orphan.orphan'; end}.${t = !record['@timestamp'].nil? ? Time.parse(record['@timestamp']) : ($log.error('record is missing @timestamp' + record.to_s);Time.at(time)) ; t.getutc.strftime(@logstash_dateformat)}"
```
portante: If we are missing the `@timestamp` field, then the record will not be found at all.  I think we need a filter to ensure every record has a timestamp.

fluentd/Dockerfile
```diff
@@ -33,7 +33,7 @@ RUN mkdir -p ${HOME} && \
       'elasticsearch-transport:<5' \
       'elasticsearch-api:<5' \
       'elasticsearch:<5' \
-      fluent-plugin-kubernetes_metadata_filter \
+      'fluent-plugin-kubernetes_metadata_filter:<1.0.0' \
```
portante: Is there a reason for this version restriction?

fluentd/configs.d/openshift/es-copy-config.conf
richm: We should add that to fluent-plugin-viaq_data_model

fluentd/Dockerfile
richm: yes, until we sort out why the new proposed cache code isn't working

fluentd/configs.d/openshift/es-copy-config.conf
jcantrill: @portante I'm missing why the record wont be found.  Are you saying it wont be found in the expected index?

portante: If a record does not have an @timestamp field, it won't know how to consider that record in the time range it is looking for.

richm: We need some way to generate the `yyyy.mm.dd` suffix for the index name for curation, so we need some sort of time field or `@timestamp` to do that.

jcantrill: What about if we can not use the `@timstamp` because its missing that we insert it right here?  Does that seem reasonable? What is our alternative?

richm: We cannot insert it here - we would need to insert it in a record-transformer filter, or in the viaq filter plugin

jcantrill:
Merging over @portante comment given we are not ready to move the referenced 1.0.0 into the 1.5 code base

PR #826, opened 2017-12-07
Bug 1500086 - no authentication required to access elasticsearch

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1500086
Test access control in many different ways, both against
Elasticsearch directly and via Kibana.
- proper token and username has access to correct projects
- proper token and username is denied access to disallowed projects
- admin has access to all indices
- username with no token is an error
- username with bogus token is an error
- username with wrong token is an error
- no username and no token is an error
- admin with no token is an error
- admin with bogus token is an error
- client cert auth with kibana cert issued by unknown CA is an error
/test

### DIFF
--- a/hack/testing/signing.conf
+++ b/hack/testing/signing.conf
@@ -1,0 +1,103 @@
+# Simple Signing CA
+
+# The [default] section contains global constants that can be referred to from
+# the entire configuration file. It may also hold settings pertaining to more
+# than one openssl command.
+
+#[ default ]
+#dir                     = _output               # Top dir
+
+# The next part of the configuration file is used by the openssl req command.
+# It defines the CA's key pair, its DN, and the desired extensions for the CA
+# certificate.
+
+[ req ]
+default_bits            = 2048                  # RSA key size
+encrypt_key             = yes                   # Protect private key
+default_md              = sha1                  # MD to use
+utf8                    = yes                   # Input is UTF-8
+string_mask             = utf8only              # Emit UTF-8 strings
+prompt                  = no                    # Don't prompt for DN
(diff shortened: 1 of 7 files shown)


richm:
/retest
curator flake

portante:
So there are no changes to the code, just test environment.  Perhaps worth stating that explicitly, since the subject refers to a BZ that one might consider is fixed by this change?

richm:
This PR is only to add a test - it does not fix the BZ referred to, but it does add a test to ensure the behavior fixed by the BZ is working.

richm:
@ewolinetz We will need to make sure that the openshift-elasticsearch-plugin work you are doing doesn't cause a regression - I never could get the SAR request to work properly when I was using the new es api java code.

ewolinetz:
@richm the work I'm doing is just to reenable the integration tests, no actual code change

richm:
> @richm the work I'm doing is just to reenable the integration tests, no actual code change

Are you upgrading the `kubernetes-client.version` ?  If so, be aware that that may change the way the api behaves.  That's what I ran into - version 3 SAR request did not behave the same way as 1.3.99.

richm:
can I get a lgtm?

ewolinetz:
@richm ack, i did upgrade that... is there an integration test I can write to verify it?

richm:
> @richm ack, i did upgrade that... is there an integration test I can write to verify it?

This PR will verify it - not sure if there is an openshift-elasticsearch-plugin integration test for doing token auth for a SAR request.

richm:
can I get a lgtm?

PR #849, opened 2017-12-13
need to include jemalloc license in image

richm:
need to include jemalloc license in image


richm:
/test logging
curator flake

PR #858, opened 2017-12-14
vendor in the source files instead of the gem

richm:
get source files from
https://github.com/ViaQ/fluent-plugin-viaq_docker_audit_log_parser/commit/c38cb33a88a00832274d53655e2e174b7e6d6705
Also gets rid of the record-modifier gem
/test


richm:
does @nhosoi have /lgtm rights?  If so, then something is wrong with @openshift-bot

nhosoi:
@richm, I see lgtm label is set after my "/lgtm" about an hour ago...

PR #862, opened 2017-12-15
Modularized run_cron.py and added support for using

josefkarasek:
regular expressions in curator config:

https://bugzilla.redhat.com/show_bug.cgi?id=1517875


richm:
test fails here:
```
+ exec python -u src/run_cron.py
python: can't open file 'src/run_cron.py': [Errno 2] No such file or directory
```
Did the `COPY src/ ${HOME}/src/` fail?

richm:
otherwise, looking good
need to add some tests for regex project patterns

curator/README.md
```diff
@@ -46,12 +61,19 @@ For example, using::
       runhour: 0
       runminute: 0
       timezone: America/New_York
+
+    ^project\..+\-dev.*$:
```
richm: Are all regular expressions also valid YAML keys?  If not, then we will need to change the structure (and maybe we should change it anyway, just to be safe):
```yaml
myapp-dev:
  delete:
...
.regex:
  - pattern: "^project\..+\-dev.*$"
    delete:
      days: 1
```
In general: the `.regex` element is a List - each list element has a `pattern` key - the value of this key is a YAML string properly escaped e.g. you can't have a string value like `{..something`.  See http://www.yaml.org/spec/1.2/spec.html#id2786942 for a definition of a yaml string scalar - or a block scalar e.g.
```yaml
.regex:
  - pattern: '^project\..+\-dev'
    delete:
      days: 3
```
or I suppose a block scalar would work too
```yaml
.regex:
  - pattern: |
      ^project\..+\-dev
    delete:
      days: 3
```

josefkarasek:
@jcantrill @richm Please review

richm:
prereq failed - maybe fixed now in aos-cd-jobs or openshift-ansible?
/retest

richm:
@josefkarasek 
```
+ exec python -u src/run_cron.py
python: can't open file 'src/run_cron.py': [Errno 2] No such file or directory
```

josefkarasek:
@richm I did not see the above error on my laptop nor in latest origin code running on devenv.
However, I did changes to the image based on our dockerfile guide:
https://docs.openshift.org/latest/creating_images/guidelines.html#use-uid

josefkarasek:
@richm obviously my dev environment is different from test env
Any idea how to replicate it?

I am not running into the above issue. 

This works for me in both, devenv and my laptop (pun not intended):
```
python -u ./src/run_cron.py
```
With the above changes I made, I understand that the process doesn't have permissions to list `/opt`. Which is probably why it failed on `python -u /opt/app-root/src/run_cron.py`. I removed the `$HOME` part of the path, it's just `python -u ./src/run_cron.py` now. Hopefully it helps.

jcantrill:
@josefkarasek Consider using ADD instead of COPY.

josefkarasek:
@richm @jcantrill Finally figured out what the issue was. Docker build handled the `COPY` instruction differently on my fedora26, centos7 AMI and the CI machine.
```
COPY dir target/dir
```
resulted in two different FS structures, depending on the version of docker:
```
target/dir/dir
vs.
target/dir
```

I squashed and rebased. Waiting for last test runs.

curator/Dockerfile
```diff
@@ -4,7 +4,7 @@ MAINTAINER OpenShift Development <dev@lists.openshift.redhat.com>
 
 # by default, run the curator cron job at midnight UTC every night, and delete
 # indices older than $CURATOR_DEFAULT_DAYS days old
-ENV HOME=/opt/app-root/src \
+ENV HOME=/opt/app-root \
```
ewolinetz: `/opt/app-root/src` is common for all images we're producing (even outside of logging IIRC). any particular reason why we're changing that?

richm:
please squash into a single commit

josefkarasek:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1517875

PR #786, opened 2017-11-16
use centos ruby-24-centos7

richm:
use the centos ruby-24 base image, and use scl to install gems

PR #865, opened 2017-12-19
revert to ruby 2.0 support

richm:
revert to ruby 2.0 support

### DIFF
--- a/fluentd/Dockerfile.centos7
+++ b/fluentd/Dockerfile.centos7
@@ -1,35 +1,39 @@
-FROM centos/ruby-24-centos7:latest
+FROM centos:centos7
 
 MAINTAINER OpenShift Development <dev@lists.openshift.redhat.com>
 
 ENV DATA_VERSION=1.6.0 \
-    FLUENTD_VERSION=0.12.41 \
-    FLUENTD_AUDIT_LOG_PARSER_VERSION=0.0.5 \
+    FLUENTD_VERSION=0.12.42 \
     GEM_HOME=/opt/app-root/src \
     HOME=/opt/app-root/src \
     PATH=/opt/app-root/src/bin:/opt/app-root/bin:$PATH \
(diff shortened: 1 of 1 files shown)


nhosoi:
/lgtm
How about the Dockerfile side in pr #786?
And this means we have to give up to upgrade to fluentd 1.0, right? :(

fluentd/Dockerfile.centos7
```diff
@@ -1,35 +1,39 @@
-FROM centos/ruby-24-centos7:latest
+FROM centos:centos7
 
 MAINTAINER OpenShift Development <dev@lists.openshift.redhat.com>
 
 ENV DATA_VERSION=1.6.0 \
-    FLUENTD_VERSION=0.12.41 \
-    FLUENTD_AUDIT_LOG_PARSER_VERSION=0.0.5 \
```
portante: Why do we also drop the `FLUENTD_AUDIT_LOG_PARSER_VERSION` variable?

fluentd/Dockerfile.centos7
```diff
@@ -1,35 +1,39 @@
...
+    FLUENTD_VERSION=0.12.42 \
```
portante: Can you also post a comment as to why we bump to `FLUENTD_VERSION` to `0.12.42`?

richm:
> How about the Dockerfile side in pr #786?

I will submit another PR for that.

> And this means we have to give up to upgrade to fluentd 1.0, right? :(

Right, at least for the short term

fluentd/Dockerfile.centos7
richm: It is now unused

fluentd/Dockerfile.centos7
richm: It is the latest available 0.12.x version - I suppose I could have bumped that version in a separate commit

PR #868, opened 2017-12-20
fluentd minor fixes

nhosoi:
fluent-plugin-remote-syslog: class Time in out_syslog_buffer.rb should
    inherit the standard class Time as done in out_syslog.rb.

fluent-plugin-remote-syslog: out_syslog.rb, out_syslog_buffered.rb
    Adding '\s' to record[@tag_key][0..31].gsub(/[\[\]]/,'') to get rid
    of spaces to avoid an argument error "Tag may not contain spaces"
    raised in syslog_protocol. (bz 1519213)

fluentd config files: removed trailing spaces

### DIFF
--- a/fluentd/configs.d/filter-pre-mux-client.conf
+++ b/fluentd/configs.d/filter-pre-mux-client.conf
@@ -1,7 +1,7 @@
 <match **>
   @type secure_forward
   log_level "#{ENV['FORWARD_INPUT_LOG_LEVEL'] || ENV['LOG_LEVEL'] || 'warn'}"
-  self_hostname ${HOSTNAME}                   
+  self_hostname ${HOSTNAME}
   shared_key "#{File.open('/etc/fluent/muxkeys/shared_key') do |f| f.readline end.rstrip}"
   secure yes
   ca_cert_path /etc/fluent/muxkeys/ca
--- a/fluentd/configs.d/openshift/filter-kibana-transform.conf
+++ b/fluentd/configs.d/openshift/filter-kibana-transform.conf
@@ -2,7 +2,7 @@
   @type record_transformer
   enable_ruby
   <record>
-    log ${record['err'] || record['msg'] || record['MESSAGE'] || record['log']}   
+    log ${record['err'] || record['msg'] || record['MESSAGE'] || record['log']}
   </record>
   remove_keys req,res,msg,name,level,v,pid,err
 </filter>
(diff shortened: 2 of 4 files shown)


richm:
does not work
```
2017-12-20 20:31:56 +0000 [info]: reading config file path="/etc/fluent/fluent.conf"
2017-12-20 20:31:56 +0000 [error]: unexpected error error="superclass mismatch for class Time"
  2017-12-20 20:31:56 +0000 [error]: /etc/fluent/plugin/out_syslog_buffered.rb:142:in `<module:Fluent>'
  2017-12-20 20:31:56 +0000 [error]: /etc/fluent/plugin/out_syslog_buffered.rb:5:in `<top (required)>'
  2017-12-20 20:31:56 +0000 [error]: /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:55:in `require'
```
This doesn't make any sense to me that it would work:
```ruby
  class Time < Time
     def timezone(timezone = 'UTC') 
```

nhosoi:
Hmmm.  What's the difference between these 2 files?  And I got an error "Time.now" is not defined (or something like that...)  But I might have been using ruby 2.4 then.  Let me double check.
```
==> out_syslog_buffered.rb <==
  class Time
    def timezone(timezone = 'UTC')
      old = ENV['TZ']
      utc = self.dup.utc
      ENV['TZ'] = timezone
      output = utc.localtime
      ENV['TZ'] = old
      output
    end
  end
end

==> out_syslog.rb <==
  class Time < Time
    def timezone(timezone = 'UTC')
      old = ENV['TZ']
      utc = self.dup.utc
      ENV['TZ'] = timezone
      output = utc.localtime
      ENV['TZ'] = old
      output
    end
  end
end
```

nhosoi:
@richm, could you please take a look one more time?

I was confused by both out_syslog.rb and out_syslog_buffered.rb having class Time define class Time, in which one with inheritance ( < Time) and the other without.  Class Time in out_syslog.rb derives from the standard Time, while the one in out_syslog_buffered.rb is just referring the class in out_syslog.rb, isn't it?

If both has "< Time", the out_syslog_buffered fails with "superclass mismatch for class Time".  If both does not have "< Time", then it does with "undefined method `now' for Fluent::Time:Class (NoMethodError)".

That's being said "class Time" in out_syslog_buffered.rb is not necessary be defined there.

PR #876, opened 2018-01-02
bug 1524644. Curator configuration produces wrong regular expressions

josefkarasek:
For example `^.operations.` instead of `^\.operations\..*$`
Both regexes match, but the first one only accidentaly. Example index name:
`.operations.1827ac8a-ef8f-11e7-92de-0ebdd5807cb2.2018.01.02`

Curator code: https://github.com/elastic/curator/blob/3.5/curator/api/filter.py#L90-L138

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1524644


jcantrill:
Do we have any test to verify what we are seeing here?

josefkarasek:
@jcantrill  https://github.com/openshift/origin-aggregated-logging/blob/master/test/curator.sh

ewolinetz:
@josefkarasek should we also be cherry picking this?

PR #881, opened 2018-01-03
Kibana55

jcantrill:
This PR:

* Updates Kibana to v5.5.2 in support of Elastic 5.5.x
* Updates the origin-kibana plugin to include Openshift skinning of Kibana
* Allows config overrides by setting 

Depends on vendored code from https://github.com/openshift/origin-kibana/pull/27


jcantrill:
[(5) Update kibana to 5.x [logging][epic-ois-es5]](https://trello.com/c/kgyvA8uz/544-5-update-kibana-to-5x-loggingepic-ois-es5)

ewolinetz:
@jcantrill why are we able to remove all those files for kibana as part of this?

jcantrill:
@ewolinetz The previous installation of the origin-kibana plugin was vendored in and install was copying all the files into the proper location.  They would then be recompiled by us to improve startup.  This change vendors in those files as a plugin archive by which we can use the kibana-plugin tool to install which compiles as part of its install process.

kibana/utils
```diff
@@ -0,0 +1,112 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+#override config by environment variables
+#inspired by https://github.com/elastic/kibana-docker/blob/master/build/kibana/bin/kibana-docker
+
+update_config_from_env_vars() {
+  conf_dir=$1
+  kibana_vars=(
```
ewolinetz: do we want to let all of these be configured with env vars? e.g. `server.host` may be alright to keep as `0.0.0.0` in the config... is there a request that we be able to customize this via env vars?

jcantrill: This is open to debate; I sourced what is here from the script that is used by the Kibana docker image.  I know we have an outstanding issue where the request timeout needs to be increased but the work around is complicated.  This change would make that easy.  I am open to whitelisting the config items to a smaller set.

PR #885, opened 2018-01-05
remove dashes in ident values used for term searches to avoid test flakes

richm:
Searches like this are failing randomly and often:

    message_uuid=$( uuidgen )
    qs='{"query":{"term":{"systemd.u.SYSLOG_IDENTIFIER":"'"${message_uuid}"'"}}}'
    curl_es $espod /.operations.*/_search -X POST -d $qs

it seems as if Elasticsearch is analyzing the systemd.u.SYSLOG_IDENTIFIER
field which causes exact matches to fail, even though it is defined as
a non_analyzed field.
See https://www.elastic.co/guide/en/elasticsearch/reference/2.4/query-dsl-term-query.html

Until we can figure this out, remove the dashes from the uuidgen output
so that there will be no analysis of the string.

### DIFF
--- a/hack/testing/util.sh
+++ b/hack/testing/util.sh
@@ -189,8 +189,8 @@ function wait_for_fluentd_to_catch_up() {
     local es_pod=$( get_es_pod es )
     local es_ops_pod=$( get_es_pod es-ops )
     es_ops_pod=${es_ops_pod:-$es_pod}
-    local uuid_es=$( uuidgen )
-    local uuid_es_ops=$( uuidgen )
+    local uuid_es=$( uuidgen | sed 's/[-]//g' )
+    local uuid_es_ops=$( uuidgen | sed 's/[-]//g' )
     local expected=${3:-1}
     local timeout=${TIMEOUT:-600}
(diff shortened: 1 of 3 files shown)


ewolinetz:
makes sense to me /lgtm

richm:
wth?
```
    "reason": "You cannot use 'static' on an include_role with a loop\n\nThe error appears to have been in '/usr/share/ansible/openshift-ansible/roles/openshift_logging/tasks/install_logging.yaml': line 78, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n# We don't allow scaling down of ES nodes currently\n- import_role:\n  ^ here\n"
```
/retest

ewolinetz:
@richm `import_*` are static calls vs `include_*` which are dynamic. I think we need to switch it for that line?

ewolinetz:
https://github.com/openshift/openshift-ansible/pull/6660

richm:
/test json-file
appears to be a new failure, but test again just in case

PR #888, opened 2018-01-10
Bug 1502744 - [3.5] Fluentd strips away timestamp milliseconds

nhosoi:
Adding "record['@timestamp'] || record['time']" to the time directive in the
record transformer filter for json-file (filter-k8s-record-transform.conf).
The patch was proposed and tested by Steven Walter <stwalter@redhat.com>.
https://bugzilla.redhat.com/show_bug.cgi?id=1502744#c12


rahnarsson:
New docker image will be released for this?

jcantrill:
Working on building this now.  It should be available within the hour.

PR #892, opened 2018-01-12
1126 es5 dev

ewolinetz:


ewolinetz:
@jcantrill I couldn't get this to build locally again for some reason... not sure if you are able to do so? I incorporated your changes from https://github.com/openshift/origin-aggregated-logging/pull/893/files

richm:
```
[openshift/origin-logging-elasticsearch] /opt/app-root/src/install.sh: line 13: /usr/share/java/elasticsearch/bin/elasticsearch-plugin: No such file or directory
```
location is different now?

jcantrill:
@ewolinetz @richm   I modified the home env vars for the installation of Elasticsearch to be the locations as defined using the RPM installation [1] .

[1] https://www.elastic.co/guide/en/elasticsearch/reference/5.4/rpm.html#rpm-layout

ewolinetz:
@jcantrill currently I'm not seeing searchguard install during the building here so we fail when trying to chmod the sgadmin script. Otherwise ES is building again for origin

elasticsearch/install.sh
```diff
@@ -20,7 +23,7 @@ mkdir /elasticsearch
 mkdir -p $ES_CONF
 chmod -R og+w $ES_CONF ${ES_HOME} ${HOME} /elasticsearch
 chmod -R o+rx /etc/elasticsearch
-chmod +x ${ES_HOME}/plugins/openshift-elasticsearch/sgadmin.sh
+chmod +x /usr/share/elasticsearch/plugins/search-guard-5/tools/sgadmin.sh
```
jcantrill: This should be reverted

elasticsearch/Dockerfile.centos7
```diff
@@ -6,28 +6,30 @@ EXPOSE 9200
 EXPOSE 9300
 USER 0
 
-ENV ES_CLOUD_K8S_VER=2.4.4_01 \
-    ES_CONF=/usr/share/java/elasticsearch/config \
+ENV ES_CLOUD_K8S_VER=5.5.2 \
+    ES_CONF=/etc/elasticsearch/ \
     ES_JAVA_OPTS="-Dmapper.allow_dots_in_name=true" \
     ES_HOME=/usr/share/java/elasticsearch \
```
jcantrill: Should be `/usr/share/elasticsearch`

elasticsearch/Dockerfile.centos7
```diff
@@ -60,7 +60,8 @@ ADD run.sh prep-install.${RELEASE_STREAM} install.sh ${HOME}/
 COPY utils/** /usr/local/bin/
 RUN ln -s /usr/local/bin/logging ${HOME}/logging && \
     ${HOME}/install.sh && \
-    rm -rf /tmp/lib
+    rm -rf /tmp/lib && \
+    echo "vm.max_map_count=262144" > /etc/sysctl.conf
```
richm: Do we need a similar change in elasticsearch/Dockerfile?

ewolinetz: I need to remove this line.. I was testing to see if that would make a difference and it doesn't. The way I was able to resolve this was doing that on my actual OCP nodes... so we need to investigate the best course of action when it comes to ensuring that is correctly set. If it is not at least that value ES 5 will not start up.

ewolinetz: Continuing the discussion...
So I asked around and it seems there is a way to modify the sysctl settings of pods using annotations [1] however updating a `vm` setting is not allowed by default and would actually make it so the entire cluster could use unsafe annotations.

The `openshift_node` has a task that updates the `sysctl`, so that may be the best course of action if one is going to deploy logging with es5.

[1] https://kubernetes.io/docs/concepts/cluster-administration/sysctl-cluster/

ewolinetz: https://github.com/openshift/openshift-ansible/pull/6628/commits/283e1902d2ec95ea3c6b369e9da234d5e13b02f7

ewolinetz:
Currently the only error I'm seeing in the logs when ES is starting (I think this is preventing the pod from being deemed `ready`) is this:
```
[2018-01-22 18:44:53,707][INFO ][container.run            ] Updating node index settings
{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Can't update non dynamic settings [[index.number_of_shards]] for open indices [[.searchguard.logging-es-data-master-l83eq4ay/OM8xzjmpRxmLbQoCUnr2aw]]"}],"type":"illegal_argument_exception","reason":"Can't update non dynamic settings [[index.number_of_shards]] for open indices [[.searchguard.logging-es-data-master-l83eq4ay/OM8xzjmpRxmLbQoCUnr2aw]]"},"status":400}[2018-01-22 18:44:54,405][INFO ][container.run            ] Finished updating node index settings
```

Currently investigating. This may require changes to the index templates only

elasticsearch/index_templates/com.redhat.viaq-openshift-operations.template.json
```diff
@@ -943,7 +934,11 @@
   },
   "order": 10,
   "settings": {
-    "index.refresh_interval": "5s"
+    "index.refresh_interval": "5s",
+    "index.number_of_replicas": $REPLICA_SHARDS,
```
ewolinetz: FYI @lukas-vlcek - added this in here since we could no longer specify `index` settings in the elasticsearch config.
We do substitution of these values in `run.sh` before we push them into ES

lukas-vlcek: @ewolinetz I moved all the `settings.*` into extra json file. See my PR https://github.com/openshift/origin-aggregated-logging/pull/922

Going forward we could maintain the `settings.*` json files in this repo and keep the pure mappings in common data model repo. Pure separation... WDYT?
//cc @richm @jcantrill

richm: Yes, that sounds good to me.

ewolinetz:
should see ci test this correctly now with https://github.com/openshift/aos-cd-jobs/pull/1078 merged and pushed

ewolinetz:
kibana didn't deploy for some reason....
/retest

ewolinetz:
This is currently failing due to a change that happened in Kibana
https://gist.github.com/ewolinetz/da9e9077c6cc5ff84c394df32192c278

I'm alright with manually merging this seeing as how ES builds and starts up correctly (per ci pod logs)

PR #883, opened 2018-01-05
Bug 1531157 logging-fluentd v3.9.0-0.16.0.2 immediately starts flooding " missing namespace" errors on startup

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1531157
We're still using k8s filter version 0.33 which still requires
include_namespace_metadata true
/test

PR #898, opened 2018-01-16
Metadata cpu cache fix

jcantrill:
This PR bums the fluent-kubernetes_metadata_filter to:
* handle case for missing namespace and project metadata
* apiserver cpu fix where metadata will never be found

This depends on https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/107


jcantrill:
@richm This probably makes #883 obsolete

jcantrill:
@richm or @nhosoi can you please reapprove ... i would like to merge assuming the tests pass.  previous failure was a build failure.

jcantrill:
@richm if/when this merges can you move the gem downstream.  I don't have the capability to do so

PR #899, opened 2018-01-16
host openshift-elasticsearch-plugin on github instead of building

jcantrill:
This PR removes the vendoring of the multi-tenant plugin in favor of hosting on github release

@ewolinetz @richm Do you think we should backport?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,6 +7,3 @@
 [submodule "fluentd/jemalloc"]
 	path = fluentd/jemalloc
 	url = https://github.com/jemalloc/jemalloc
-[submodule "elasticsearch/lib/openshift-elasticsearch-plugin"]
-	path = elasticsearch/lib/openshift-elasticsearch-plugin
-	url = https://github.com/fabric8io/openshift-elasticsearch-plugin
--- a/elasticsearch/Dockerfile.centos7
+++ b/elasticsearch/Dockerfile.centos7
@@ -39,10 +39,8 @@ LABEL io.k8s.description="Elasticsearch container for EFK aggregated logging sto
 ADD elasticsearch.repo /etc/yum.repos.d/elasticsearch.repo
 # install the RPMs in a separate step so it can be cached
 RUN rpm --import https://packages.elastic.co/GPG-KEY-elasticsearch && \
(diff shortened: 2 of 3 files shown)


richm:
yes, I think we should backport - too difficult to keep track of otherwise

richm:
@jcantrill @nhosoi I think we need to change the tests to allow project.openshift-web-console ASAP

jcantrill:
@richm @nhosoi What is specific in our tests to that namespace?  This should  become a an operations index or maybe we should consider adding an 'infra' prefix to represent other infra project (e.g. metrics, openshift-template-broker)

richm:
> @richm @nhosoi What is specific in our tests to that namespace? 

All `openshift*` namespace records should go into `.operations` - there was a BZ related to this - I added a test to check for regressions for this BZ - then openshift added a few new namespaces which broke our tests - so yes, we need to change our fluentd config to put these records in the `.operations` index, but in the meantime, in order to unblock CI and be able to merge commits, we should change the test that checks for `project.openshift*` to allow these records.

nhosoi:
@richm @jcantrill 
> in order to unblock CI and be able to merge commits, we should change the test that checks for project.openshift* to allow these records.

I was thinking to add NOT "kubernetes.namespace_name":"openshift-web-console" to the query in zzz-correct-index-names.sh, something like this ... (sorry, i'm about to test it...)
```
-    qs='{"query":{"term":{"kubernetes.namespace_name":"'"${project}"'"}}}'
+    qs='{"query":{"bool":{"must":{"term":{"kubernetes.namespace_name":"'"${project}"'"}},"must_not":{"match":{"kubernetes.namespace_name":"openshift-web-console"}}}}}'
```
but you mean we want to just skip openshift and openshift-infra checking for now?  (it is super easy... :)
```
-for project in default openshift openshift-infra ; do
+for project in default ; do
```

richm:
> but you mean we want to just skip openshift and openshift-infra checking for now? (it is super easy... :)

Yes.  The problem is that the query term searches are supposed to be exact match searches, but because of the index template problem, the searches are really doing "fuzzy" match searches so that a search for `openshift` returns `openshift-web-console`, etc.

PR #900, opened 2018-01-17
bug 1534878. fix sending openshift-web-console logs to operations index

jcantrill:
This PR:
* Makes the operations projects configurable

### DIFF
--- a/fluentd/README.md
+++ b/fluentd/README.md
@@ -1,0 +1,9 @@
+# Fluentd
+[Fluentd](https://www.fluentd.org/) is the log collector that resides on each Openshift node to gather application and node logs
+
+## Configuration
+Following are the environment variables that can be modified to adjust the configuration:
+
+| Environment Variable | Description |Example|
+|----------------------|-------------|---|
+| `OCP_OPERATIONS_PROJECTS`| The list of project or patterns for which messages will be sent to the operations indices|`OCP_OPERATIONS_PROJECTS="default openshift openshift-"`
--- a/fluentd/generate_throttle_configs.rb
+++ b/fluentd/generate_throttle_configs.rb
@@ -17,7 +17,7 @@ ENV.sort.each do |entry|
   @log.debug entry
 end if @log.debug?
 
-DEFAULT_OPS_PROJECTS = ['default', 'openshift', 'openshift-infra']
+DEFAULT_OPS_PROJECTS = !ENV['OCP_OPERATIONS_PROJECTS'].nil? ? ENV['OCP_OPERATIONS_PROJECTS'].split(' ') : ['default', 'openshift', 'openshift-infra']
 DEFAULT_FILENAME = "/etc/fluent/configs.d/user/throttle-config.yaml"
 
 VALID_SETTINGS = {"read_lines_limit" => "number"}
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -1,5 +1,21 @@
 #!/bin/bash
 
+CFG_DIR=/etc/fluent/configs.d
+OCP_OPERATIONS_PROJECTS=${OCP_OPERATIONS_PROJECTS:-"default openshift openshift-"}
+OCP_FLUENTD_TAGS="journal.system.** audit.log**"
+for p in ${OCP_OPERATIONS_PROJECTS}; do
+    if [[ "${p}" == *- ]] ; then
+      p="${p}*"
+    fi
+    OCP_FLUENTD_TAGS+=" **_${p}_**"
+done
+ocp_fluentd_files=$( grep -l %OCP_FLUENTD_TAGS% ${CFG_DIR}/* ${CFG_DIR}/*/* 2> /dev/null || : )
+tmpfile=$( mktemp )
+for file in ${ocp_fluentd_files} ; do
+    sed -e "s/%OCP_FLUENTD_TAGS%/${OCP_FLUENTD_TAGS}/" $file > $tmpfile
+    mv $tmpfile $file
+done
+
 fluentdargs="--no-supervisor"
 if [[ $VERBOSE ]]; then
   set -ex
@@ -59,7 +75,6 @@ export IPADDR4 IPADDR6
 
 BUFFER_SIZE_LIMIT=${BUFFER_SIZE_LIMIT:-16777216}
 
-CFG_DIR=/etc/fluent/configs.d
 if [ "${USE_MUX:-}" = "true" ] ; then
     # copy our standard mux configs to the openshift dir
     cp $CFG_DIR/input-*-mux.conf $CFG_DIR/filter-*-mux.conf $CFG_DIR/openshift
--- a/test/zzz-correct-index-names.sh
+++ b/test/zzz-correct-index-names.sh
@@ -24,6 +24,12 @@ es_pod=$( get_es_pod es )
 es_ops_pod=$( get_es_pod es-ops )
 es_ops_pod=${es_ops_pod:-$es_pod}
 
+# INFRA_PROJECTS=$(oc set env ds/logging-fluentd --list | grep OCP_OPERATIONS_PROJECTS | sed "s/.*=//")
+# for project in ${INFRA_PROJECTS} ; do
+
+# Testing to revert the change in pr/900
+# Since logs under the tag openshift-* (e.g., openshift-web-console) are indexed in .operations,
+# this test should not fail by those logs.
 for project in default openshift openshift-infra ; do
     qs='{"query":{"term":{"kubernetes.namespace_name":"'"${project}"'"}}}'
     os::cmd::expect_success_and_not_text "curl_es $es_pod /_cat/indices" "project\.${project}\."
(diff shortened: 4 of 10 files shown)


fluentd/README.md
```diff
@@ -0,0 +1,9 @@
+# Fluentd
+[Fluentd](https://www.fluentd.org/) is the log collector that resides on each Openshift node to gather application and node logs
+
+## Configuration
+Following are the environment variables that can be modified to adjust the configuration:
+
+| Environment Variable | Description |Example|
+|----------------------|-------------|---|
+| `OCP_OPERATIONS_PROJECTS`| The list of project for which messages will be sent to the operations indices|`OCP_OPERATIONS_PROJECTS="journal.system** system.var.log** **_default_** **_openshift_** **_openshift-** audit.log**"`
```
richm: It isn't really a list of projects - it is a list of fluentd tag patterns which match records tagged as operations records.

fluentd/generate_throttle_configs.rb
```diff
@@ -17,7 +17,7 @@
   @log.debug entry
 end if @log.debug?
 
-DEFAULT_OPS_PROJECTS = ['default', 'openshift', 'openshift-infra']
+DEFAULT_OPS_PROJECTS = !ENV['OCP_OPERATIONS_PROJECTS'].nil? ? ENV['OCP_OPERATIONS_PROJECTS'].split(' ') : ['default', 'openshift', 'openshift-infra']
```
richm: I'm not sure if this will work since these are fluentd tag patterns, not project names.

richm:
You'll need to make changes to these files too:
https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/configs.d/openshift/filter-viaq-data-model.conf
https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/configs.d/openshift/filter-retag-journal.conf
https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/configs.d/input-post-forward-mux.conf#L47

jcantrill:
@richm I made an attempt to address your questions but I'm not exactly sure how to modify.  Can it take a regex pattern to say 'substitute capture group from source into the target'?
[1] https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/configs.d/openshift/filter-retag-journal.conf

richm:
@jcantrill not sure I understand your question, but probably, yes - https://github.com/fluent/fluent-plugin-rewrite-tag-filter/tree/v1.5.6#usage

fluentd/run.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/bash
 
+OCP_OPERATIONS_PROJECTS=${OCP_OPERATIONS_PROJECTS:-"default openshift openshift-"}
+OCP_FLUENTD_TAGS=('journal.system.** system.var.log** audit.log**')
+for p in ${OCP_OPERATIONS_PROJECTS}; do
+    OCP_FLUENTD_TAGS+=("_**${p}_**")
```
richm: should be
```
 +    OCP_FLUENTD_TAGS+=("**_${p}_**") 
```
and - one of the tags will be `**_openshift-_**` which will match nothing - I think you need it to be `**_openshift-*_**`

This is going to be tricky . . .

test/zzz-correct-index-names.sh
```diff
@@ -24,7 +24,15 @@ es_pod=$( get_es_pod es )
...
 es_ops_pod=${es_ops_pod:-$es_pod}
 
-for project in default openshift openshift-infra ; do
+all=$(oc get project -o jsonpath={.items[*].metadata.name})
+read -r -d '' SCRIPT<< EOF
+projects = [p for p in "$all".split(' ') if p.startswith('openshift-') or p == 'default']
+print " ".join(projects)
+EOF
+
+INFRA_PROJECTS=$(python -x "${SCRIPT}")
```
richm: I don't think you want to get the list of projects from the system like this.  I think you want to use `OCP_OPERATIONS_PROJECTS` as the source of project names and patterns to check.

The trick is - how to get this list in this context?  If the env var is set in the fluentd daemonset, you could get it like this:
```
oc set env ds/logging-fluentd --list | grep and sed for OCP_OPERATIONS_PROJECTS
```
if it isn't set in the ds, then you could find the pid of the fluentd process, and do something like this:
```
sudo cat /proc/$pid/environ | tr '\0' '\n' | grep and sed for OCP_OPERATIONS_PROJECTS
```

or ....

jcantrill: @richm  I image going forward the pattern will be anything prefixed with 'openshift' should go to the operations index.  Do we not want to insulate ourselves from other changes?  For instance the openshift-template-broker is also a namespace I am seeing.  Like logging, shouldnt that really go to  some infra index?  In the long term we should probably redefine IMO '.operations' to be an alias and infra projects should maybe go into an 'infra' prefixed index.  Short term, maybe your solution is better; i defer to you.

richm: > anything prefixed with 'openshift' should go to the operations index.

and `kube` too?  Yes, openshift-template-broker should go into .operations.

My "solution" is that the script that checks to see if the data in es is correct should use the logging configuration of which projects and patterns to use, rather than just use the list of projects returned by `oc get projects`.

richm:
The problem is that no records are going to `.operations`
```
FAILURE after 599.575s: test/cluster/functionality.sh:80: executing 'curl_es 'logging-es-ops-data-master-ku06fax8-1-6n96n' '/_cat/indices?h=index'' expecting any result and text '^(project|\.operations)\.'; re-trying every 0.2s until completion or 600.000s: the command timed out
.kibana                                          
.searchguard.logging-es-ops-data-master-ku06fax8
... repeated 914 times
```
That is, the es ops instance has no `.operations.*` indices, which it should have.  This means the fluentd configuration is wrong.

fluentd/run.sh
```diff
@@ -3,7 +3,10 @@
 OCP_OPERATIONS_PROJECTS=${OCP_OPERATIONS_PROJECTS:-"default openshift openshift-"}
 OCP_FLUENTD_TAGS=('journal.system.** system.var.log** audit.log**')
 for p in ${OCP_OPERATIONS_PROJECTS}; do
-    OCP_FLUENTD_TAGS+=("_**${p}_**")
+    if [[ "${p}" == *- ]] ; then
```
richm: I think you need to use the match operator, not `==`

fluentd/run.sh
```diff
@@ -1,5 +1,14 @@
 #!/bin/bash
 
+OCP_OPERATIONS_PROJECTS=${OCP_OPERATIONS_PROJECTS:-"default openshift openshift-"}
+OCP_FLUENTD_TAGS=('journal.system.** system.var.log** audit.log**')
+for p in ${OCP_OPERATIONS_PROJECTS}; do
+    if [[ "${p}" == *- ]] ; then
+      p="${p}*"
+    fi
+    OCP_FLUENTD_TAGS+=("**_${p}_**")
+done
```
nhosoi: Not sure this is causing any error, but these 2 are different.
```
echo "${OCP_FLUENTD_TAGS[@]}"
journal.system.** system.var.log** audit.log** **_default_** **_openshift_** **_openshift-*_**
```
```
echo "${OCP_FLUENTD_TAGS}"
journal.system.** system.var.log** audit.log**
```
And we may need "export OCP_FLUENTD_TAGS"?

nhosoi: Changing the type of OCP_FLUENTD_TAGS from an array to a string separated by a space, OCP_FLUENTD_TAGS is likely passed to ruby ENV successfully.  But still .operations is not created...  I'm suspecting a ruby code may not be evaluated in <match ...>?
```
diff --git a/fluentd/run.sh b/fluentd/run.sh
index b904a68..cb8a716 100644
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -1,12 +1,12 @@
 #!/bin/bash
 
 OCP_OPERATIONS_PROJECTS=${OCP_OPERATIONS_PROJECTS:-"default openshift openshift-"}
-OCP_FLUENTD_TAGS=('journal.system.** system.var.log** audit.log**')
+OCP_FLUENTD_TAGS="journal.system.** system.var.log** audit.log**"
 for p in ${OCP_OPERATIONS_PROJECTS}; do
     if [[ "${p}" == *- ]] ; then
       p="${p}*"
     fi
-    OCP_FLUENTD_TAGS+=("**_${p}_**")
+    OCP_FLUENTD_TAGS+=" **_${p}_**"
 done
 export OCP_FLUENTD_TAGS
```

jcantrill:
@richm @nhosoi replaced the original PR with #908 and squashed

jcantrill:
@richm @ewolinetz @nhosoi Can I get an LGTM for this one?

richm:
should we go ahead and backport to 3.8 and 3.7, or wait until there is a problem?

jcantrill:
@richm I don't believe the other namespaces are relevant until 3.9.

PR #903, opened 2018-01-17
Update to reference only the container_name field

portante:
We no longer need to reference the `.raw` subfield since the mapping for the `container_name` now uses `not_analyzed`.


nhosoi:
Still failing?
```
FAILURE after 0.505s: test/zzz-correct-index-names.sh:31: executing 'curl_es logging-es-data-master-xa6wxvm8-1-kx8nq /project.*/_count -X POST -d '{"query":{"term":{"kubernetes.namespace_name":"openshift"}}}' | get_count_from_json' expecting success and text '^0$': the output content test failed
5
```

richm:
/test json-file
curator flake

PR #909, opened 2018-01-19
bz1515715 -- remote syslog hostname

nhosoi:
Extending "use_record" to include hostname, which is going be put in the log in rsyslog.

Other changes.

- remote-syslog.sh would test mux, as well.
- remote-syslog.sh has another test case -- use rsyslogd on the node,
 in which it restarting rsyslogd enabling:
    $ModLoad imtcp
    $InputTCPServerRun 514

### DIFF
--- a/fluentd/out_syslog.rb
+++ b/fluentd/out_syslog.rb
@@ -67,11 +67,12 @@ module Fluent
       tag = tag.sub(@remove_tag_prefix, '') if @remove_tag_prefix
       chain.next
       es.each {|time,record|
-        @packet.hostname = hostname
         if @use_record
+          @packet.hostname = record['hostname'] || hostname
           @packet.facility = record['facility'] || @facilty
           @packet.severity = record['severity'] || @severity
         else
+          @packet.hostname = hostname
           @packet.facility = @facilty
           @packet.severity = @severity
         end
@@ -91,7 +92,14 @@ module Fluent
                           tag[0..31] # tag is trimmed to 32 chars for syslog_protocol gem compatibility
                       end
         packet = @packet.dup
-        packet.content = record[@payload_key]
+        if @use_record && (record.key?('kubernetes'))
+            packet.content = @payload_key + "=" + record[@payload_key] + \
+                             (((record["kubernetes"]).key?('namespace_name')) ? ", namespace_name=" + record["kubernetes"]["namespace_name"] : "" ) + \
+                             (((record["kubernetes"]).key?('container_name')) ? ", container_name=" + record["kubernetes"]["container_name"] : "" ) + \
+                             (((record["kubernetes"]).key?('pod_name')) ? ", pod_name=" + record["kubernetes"]["pod_name"] : "" )
+        else
+            packet.content = record[@payload_key]
+        end
         @socket.send(packet.assemble, 0, @remote_syslog, @port)
     }
     end
--- a/fluentd/out_syslog.rb
+++ b/fluentd/out_syslog.rb
@@ -92,11 +92,11 @@ module Fluent
                           tag[0..31] # tag is trimmed to 32 chars for syslog_protocol gem compatibility
                       end
         packet = @packet.dup
-        if @use_record && (record.key?('kubernetes'))
-            packet.content = @payload_key + "=" + record[@payload_key] + \
-                             (((record["kubernetes"]).key?('namespace_name')) ? ", namespace_name=" + record["kubernetes"]["namespace_name"] : "" ) + \
-                             (((record["kubernetes"]).key?('container_name')) ? ", container_name=" + record["kubernetes"]["container_name"] : "" ) + \
-                             (((record["kubernetes"]).key?('pod_name')) ? ", pod_name=" + record["kubernetes"]["pod_name"] : "" )
+        if @use_record && (record.key?('kubernetes')) && @payload_key && record[@payload_key]
+            packet.content = (((record['kubernetes']).key?('namespace_name')) ? 'namespace_name=' + record['kubernetes']['namespace_name'] + ', ' : '' ) + \
+                             (((record['kubernetes']).key?('container_name')) ? 'container_name=' + record['kubernetes']['container_name'] + ', ' : '' ) + \
+                             (((record['kubernetes']).key?('pod_name')) ? 'pod_name=' + record['kubernetes']['pod_name'] + ', ' : '' ) + \
+                             @payload_key + '=' + record[@payload_key]
         else
             packet.content = record[@payload_key]
         end
--- a/test/remote-syslog.sh
+++ b/test/remote-syslog.sh
@@ -9,8 +9,17 @@ source "${OS_O_A_L_DIR}/hack/testing/util.sh"
 os::util::environment::use_sudo
 
 FLUENTD_WAIT_TIME=${FLUENTD_WAIT_TIME:-$(( 2 * minute ))}
+MUX_WAIT_TIME=$(( 5 * minute ))
+ALTPORT=601
 
 os::test::junit::declare_suite_start "Remote Syslog Configuration Tests"
+
+# switch pods type depending on the mux configuration
+fluentdtype="fluentd"
+if oc get dc/logging-mux > /dev/null 2>&1 ; then
+    os::log::debug "$( oc get dc/logging-mux )"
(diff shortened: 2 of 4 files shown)


anpingli:
@nhosoi Could we include "container_name", "namespace_name", and  "pod_name" in the output records?    I guess only the hostname couldn't know where the record come from.

nhosoi:
@anpingli @richm @jcantrill 
Per request from @anpingli, I've added the second patch to this pr.
https://github.com/openshift/origin-aggregated-logging/pull/909/commits/a816c5b5277fbccfca0ac5895b252f2b8a53e228

If use_record is set to true via env variable REMOTE_SYSLOG_USE_RECORD, additional "container_name", "namespace_name", and "pod_name" are added to the payload (message, by default) if any of them is available.

Format:
`payload_name=PAYLOAD_STRING[, namespace_name=NAMESPACE_NAME][, container_name=CONTAINER_NAME][, pod_name=POD_NAME]`

My only concern is we borrow remote syslog plugin code from https://github.com/docebo/fluent-plugin-remote-syslog and I wonder how relevant this change would be to feedback to the original repo.

test/remote-syslog.sh
```diff
@@ -9,189 +9,301 @@ source "${OS_O_A_L_DIR}/hack/testing/util.sh"
...
-  savemuxdc=$( mktemp )
-  oc export dc/logging-mux -o yaml > $savemuxdc
-fi
-
-function reset_mux_deployconfig() {
-  if [ "$savemuxdc" != "" ]; then
-    os::log::info Restoring original mux deployconfig environment variable
-    os::log::debug "$( oc replace -f $savemuxdc )"
-  fi
+function artifact_log () {
```
richm: There is already an `artifact_log` method and an `artifact_out` method e.g. https://github.com/openshift/origin-aggregated-logging/blob/master/hack/testing/util.sh#L295 and https://github.com/openshift/origin-aggregated-logging/blob/master/hack/testing/util.sh#L298

Here is an example: https://github.com/openshift/origin-aggregated-logging/blob/master/test/mux.sh#L51

test/remote-syslog.sh
```diff
@@ -9,189 +9,301 @@ source "${OS_O_A_L_DIR}/hack/testing/util.sh"
...
-  os::log::debug "$( oc set env ds/logging-fluentd USE_REMOTE_SYSLOG=true )"
-  os::log::debug "$( oc label node --all logging-infra-fluentd=true --overwrite=true )"
-  os::cmd::try_until_text "oc get pods -l component=fluentd" "^logging-fluentd-.* Running "
-  fpod=$( get_running_pod fluentd )
-  os::cmd::try_until_success "oc exec $fpod find /etc/fluent/configs.d/dynamic/output-remote-syslog.conf"
-  os::cmd::expect_success_and_text "oc exec $fpod grep 'remote_syslog' /etc/fluent/configs.d/dynamic/output-remote-syslog.conf" "remote_syslog ${my_remote_syslog_host}"
-fi
-
-os::log::info Test 1, expecting generate_syslog_config.rb to have created configuration file
+extra_rsyslog_artifacts=$ARTIFACT_DIR/rsyslog-artifacts.txt
```
richm: see above

fluentd/out_syslog.rb
```diff
@@ -92,7 +92,14 @@ def emit(tag, es, chain)
                           tag[0..31] # tag is trimmed to 32 chars for syslog_protocol gem compatibility
                       end
         packet = @packet.dup
-        packet.content = record[@payload_key]
+        if @use_record && (defined? record["kubernetes"]) && (!record["kubernetes"].nil?)
```
richm: I think you should use `key?` instead of `defined?`
`record.key?('kubernetes')`
http://ruby-doc.org/core-2.0.0/Hash.html#method-i-key-3F

test/remote-syslog.sh
```diff
@@ -284,10 +284,16 @@ artifact_log $title $mypod
 
 mymessage="rsyslogTestMessage-"$( date +%Y%m%d-%H%M%S )
 logger -i -p local6.info -t rsyslogTestTag $mymessage
-os::cmd::try_until_text "sudo grep $mymessage /var/log/messages | awk -F':' '{print \$4}'" " $mymessage" $MUX_WAIT_TIME
+os::cmd::try_until_text "sudo grep $mymessage /var/log/messages" ".*${mymessage}.*" $MUX_WAIT_TIME
```
richm: it would be better to use `journalctl` - but it would be very expensive to call `journalctl | grep something` every 0.2 seconds, which is what `try_until_text` does.   Maybe something like this?
```bash
tmpjournal=$( mktemp )
journalctl -f > $tmpjournal & jcpid=$!
logger -i -p local6.info -t rsyslogTestTag $mymessage
os::cmd::try_until_text "sudo grep $mymessage $tmpjournal" ".*${mymessage}.*" $MUX_WAIT_TIME
kill $jcpid # also need to do this in cleanup()
```
If you know that the message is going to be logged to a specific unit (e.g. rsyslog) then you can use `journalctl -f -u rsyslog`

test/remote-syslog.sh
```diff
@@ -284,10 +284,16 @@ artifact_log $title $mypod
...
 echo "Log test message by logger: $mymessage" >> $extra_rsyslog_artifacts
 sudo grep $mymessage /var/log/messages >> $extra_rsyslog_artifacts
 
+mymessage="testKibanaMessage-"$( date +%Y%m%d-%H%M%S )
+kpod=$( get_running_pod kibana )
+oc exec $kpod -c kibana -- curl --connect-timeout 1 -s http://localhost:5601/$mymessage || :
+os::cmd::try_until_text "sudo grep $mymessage /var/log/messages" ".*${mymessage}.*" $MUX_WAIT_TIME
```
richm: see above

test/remote-syslog.sh
```diff
@@ -284,10 +284,16 @@ artifact_log $title $mypod
...
 mymessage="rsyslogTestMessage-"$( date +%Y%m%d-%H%M%S )
 logger -i -p local6.info -t rsyslogTestTag $mymessage
-os::cmd::try_until_text "sudo grep $mymessage /var/log/messages | awk -F':' '{print \$4}'" " $mymessage" $MUX_WAIT_TIME
+os::cmd::try_until_text "sudo grep $mymessage /var/log/messages" ".*${mymessage}.*" $MUX_WAIT_TIME
 echo "Log test message by logger: $mymessage" >> $extra_rsyslog_artifacts
 sudo grep $mymessage /var/log/messages >> $extra_rsyslog_artifacts
 
+mymessage="testKibanaMessage-"$( date +%Y%m%d-%H%M%S )
+kpod=$( get_running_pod kibana )
+oc exec $kpod -c kibana -- curl --connect-timeout 1 -s http://localhost:5601/$mymessage || :
```
richm: Use `add_test_message`

test/remote-syslog.sh
nhosoi: In this test, i've restarted rsyslog with tcp input:
https://github.com/nhosoi/origin-aggregated-logging/blob/bz1515715/test/remote-syslog.sh#L228-L235
and let fluentd output-remote-syslog.conf point it.  That is, I'm expecting to see the forwarded message is stored in /var/log/messages by rsyslogd (instead of capturing the message from journald as in this flow: logger -> journald -> fluentd -> rsyslogd).  Do you think the test case is not written as I described?

richm: Ok, I see.  Then it is ok to use /var/log/messages

nhosoi:
accidentally reopened pr.  closing it.

nhosoi:
As described in the bug 1543761 - "openshift_logging_fluentd_remote_syslog_use_record=true is dysfunctional", severity and facility are not respected.  I'm suspecting we do not store them in the record hash?  (I.e., record['severity'] and record['facility'])  I noticed we set record['level'] in viaq plugin.  But replacing record['severity'] in the remote-syslog plugin did not work...  @richm, do you have any idea how to get them in the remote-syslog plugin?  Also, the original tag is lost at this phase since we had replaced each tag with output_tag.  Can we keep the original tag somewhere to pass to remote-syslog plugin?

nhosoi:
It looks I cannot keep using this pr. :(  Opening another one...

PR #916, opened 2018-01-24
reduce test flakes with false hits for ops projects

richm:
Because of the problem where index templates can be loaded after the
data, we cannot rely on exact match searches (query term).  Instead,
do a possibly "fuzzy" search for namespace_name, then use jq to further
parse and query the output.
In order to make sure the code which processes operational and
"openshift-*" projects is working, create log messages for those
projects.
/test

### DIFF
--- a/test/zzz-correct-index-names.sh
+++ b/test/zzz-correct-index-names.sh
@@ -24,25 +24,59 @@ es_pod=$( get_es_pod es )
 es_ops_pod=$( get_es_pod es-ops )
 es_ops_pod=${es_ops_pod:-$es_pod}
 
-# INFRA_PROJECTS=$(oc set env ds/logging-fluentd --list | grep OCP_OPERATIONS_PROJECTS | sed "s/.*=//")
-# for project in ${INFRA_PROJECTS} ; do
+OPS_NAMESPACES="default openshift openshift-infra openshift-this-is-a-test"
 
-# Testing to revert the change in pr/900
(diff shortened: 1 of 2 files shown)


richm:
/test logging
flake - Submit Queue reports test is not green

test/zzz-correct-index-names.sh
```diff
@@ -24,25 +24,59 @@ es_pod=$( get_es_pod es )
...
+    if [ -n "$delete_project" ] ; then
+        oc delete project $delete_project 2>&1 | artifact_out
+        os::cmd::try_until_failure "oc get project $delete_project" 2>&1 | artifact_out
+    fi
+done
+
+for project in $OPS_NAMESPACES ; do
     qs='{"query":{"term":{"kubernetes.namespace_name":"'"${project}"'"}}}'
-    os::cmd::expect_success_and_not_text "curl_es $es_pod /_cat/indices" "project\.${project}\."
+    os::cmd::expect_success_and_not_text "curl_es $es_pod /_cat/indices" "project.${project}."
```
ewolinetz: do we not want to escape the regex here and make sure its an actual '.' vs a regex '.' ?

ewolinetz:
other than question about escaping, lgtm

richm:
@ewolinetz good catch - I had messed up the rebase - fixed

richm:
/retest
curator flake

PR #922, opened 2018-01-29
1126 es5 dev

lukas-vlcek:
Testing updated index mappings and index template.


ewolinetz:
@lukas-vlcek the diffs aren't formatting correctly when I'm looking at them... I see the extra index templates... Does this just work with the idea that more than one index template will be matched? Once with the common settings and once with the prior templates?

elasticsearch/run.sh
```diff
@@ -133,6 +134,8 @@ verify_or_add_index_templates() {
     shopt -s failglob
     for template_file in ${ES_HOME}/index_templates/*.json
     do
+        sed -i "s,\$REPLICA_SHARDS,$REPLICA_SHARDS," $template_file
```
ewolinetz: @lukas-vlcek we should consider trying to make this more efficient if we only need to update 2 of the 5 templates provided. Maybe we can match on the fact that the new ones are "common."

richm: What about when you want to upgrade index templates or settings - that is, replace what's already there?

lukas-vlcek:
@ewolinetz yes, I added two new "common" index templates and they contain just the `settings.*` section. That means I assume `sed` will be used mostly on these two files. And yes, we can make `sed` more efficient by including "common" or "settings" into matching pattern (does it really make any difference from perf POW?).

All index template files matching the `template` pattern in it are used when creating a new index. The order is driven by `order` value (lower goes first). Unfortunately, the `template` pattern does not support full regex syntax so we need two common setting files (one for `project` second for `operation` pattern). We could use just `*` to match all index names but that would include kibana, sg, ... indices as well, and that is not what we want I think.

elasticsearch/run.sh
lukas-vlcek: @richm If we need upgrade then we need to make sure particular index templates are also pushed into ES cluster (I think they are not pushed today if they are already present). Then, as soon as new index is created in ES it will follow updated index template(s).

lukas-vlcek:
@richm I updated `hack/update-viaq-data-model.sh` to download and process elasticsearch templates for specific ES version only (ie the script requires an argument now). PTAL.

lukas-vlcek:
Closing in favour of https://github.com/openshift/origin-aggregated-logging/pull/961

PR #923, opened 2018-01-30
fix default projects so they better match the previous values

jcantrill:
This PR should make the changes more compatible to what existed prior to #900


jcantrill:
I am unable to validate this against a job triggered from ansible.  Assuming all tests pass here, I think we should merge and try the ansible ones again to see if the changes are related to the variances this fixes

PR #924, opened 2018-01-30
Using sed -i instead of doing a mv for file security

ewolinetz:
If we don't end up getting output from the sed at all we could potentially be `mv`ing a blank file onto `$file`.
When debugging I saw that my `/etc/fluent/conf.d/openshift/output-operations.conf` file was empty, which shouldn't happen since we provide it at image build time.


ewolinetz:
This may or may not resolve the issues we're seeing where the intended namespaces aren't ending up in .operations since nothing gets routed through our operations based output plugin

nhosoi:
> If we don't end up getting output from the sed at all we could potentially be mving a blank file onto $file.
When debugging I saw that my /etc/fluent/conf.d/openshift/output-operations.conf file was empty, which shouldn't happen since we provide it at image build time.

I agree with you.  Thanks for the fix, @ewolinetz

richm:
do you want to remove the do-not-merge label?

PR #926, opened 2018-01-31
Bug 1534588 - Size of the emitted data exceeds buffer_chunk_limit

nhosoi:
If secure-forward outputs are additionally configured, the output
count is also to be added to NUM_OUTPUTS, which is a divider to
calculate the buffer size per output.

Adding tests to cover the cases - having 1 then 2 intermediate
forwarding fluentds.


nhosoi:
There could be some cleanup issues.  Turning into WIP.

nhosoi:
adding debug prints covered up the problem... :(

nhosoi:
> @richm: #926 failed to apply on top of branch "release-3.6":

I'm going to apply the patch to release-3.6 manually.

PR #933, opened 2018-02-08
add bulk index rejection test

richm:
This tests the ability of fluentd to handle bulk index rejections.
The test starts up several curl processes which hammer Elasticsearch
with bulk index requests which causes fluentd to receive bulk rejections.
The methodology is to add some messages for both projects and operations
indices which should be rejected and queued and eventually resent
successfully.
The test will ensure that Elasticsearch is rejecting bulk requests and
fluentd is receiving bulk rejections.

### DIFF
--- a/test/zzzz-bulk_rejection.sh
+++ b/test/zzzz-bulk_rejection.sh
@@ -1,0 +1,265 @@
+#!/bin/bash
+
+# test bulk rejection handling
+source "$(dirname "${BASH_SOURCE[0]}" )/../hack/lib/init.sh"
+source "${OS_O_A_L_DIR}/hack/testing/util.sh"
+trap os::test::junit::reconcile_output EXIT
+os::util::environment::use_sudo
+
+os::test::junit::declare_suite_start "test/bulk_rejection"
+
+espod=$( get_es_pod es )
+esopspod=$( get_es_pod es-ops )
+esopspod=${esopspod:-$espod}
+
+function cleanup() {
+    local result_code="$?"
+    set +e
+    if [ -n "${bulkdonefile:-}" ] ; then
+        echo done > $bulkdonefile
+    fi
+    if [ -n "${bulktestjson:-}" -a -f "${bulktestjson:-}" ] ; then
+        rm -f $bulktestjson
+    fi
+    if [ -f "${es_settings:-}" ] ; then
+        echo restore settings | artifact_out
+        cat $es_settings | artifact_out
+        cat $es_settings | curl_es_input $espod /_cluster/settings -XPUT --data-binary @- | jq . | artifact_out
+        rm -f $es_settings
+    fi
+    curl_es $espod /bulkindextest -XDELETE | jq . | artifact_out
+    fpod=$( get_running_pod fluentd )
+    if [ -f /var/log/fluentd.log ] ; then
+        cp /var/log/fluentd.log $ARTIFACT_DIR/fluentd-with-bulk-index-rejections.log
+    else
+        oc logs $fpod > $ARTIFACT_DIR/fluentd-with-bulk-index-rejections.log
+    fi
+    if [ -n "${f_cm:-}" -a -f "${f_cm:-}" ] ; then
+        oc replace --force -f $f_cm
+    fi
+    if [ -n "${f_ds:-}" -a -f "${f_ds:-}" ] ; then
+        oc replace --force -f $f_ds
+    fi
+    os::cmd::try_until_failure "oc describe pod $fpod"
+    sleep 1
+    os::cmd::try_until_text "oc get pods -l component=fluentd" "^logging-fluentd-.* Running "
+    if [ -n "${bulkdonefile:-}" -a -f "${bulkdonefile:-}" ] ; then
+        rm -f $bulkdonefile
+    fi
+    # this will call declare_test_end, suite_end, etc.
+    os::test::junit::reconcile_output
+    exit $result_code
+}
+
+trap cleanup EXIT
+
+# save current fluentd settings
+f_cm=$( mktemp )
+f_ds=$( mktemp )
+
+oc get cm/logging-fluentd -o yaml > $f_cm
+oc get ds/logging-fluentd -o yaml > $f_ds
+
+# turn on debug output
+cat $f_cm | \
+    sed -e 's,@include configs.d/openshift/system.conf,<system>\
+      log_level debug\
+    </system>,' | oc replace --force -f -
+
+oc set env ds/logging-fluentd DEBUG=true
+
+# save current es settings
+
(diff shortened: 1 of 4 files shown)


richm:
https://trello.com/c/DwR5URpr/332-5-create-integration-test-for-bulk-indexing-rejectionsepic-ois-agl-perfcda

richm:
The motivation for doing this now is https://bugzilla.redhat.com/show_bug.cgi?id=1541429

richm:
hmm - looks like bulk index rejection test is interfering with mux test?

richm:
/retest
flakes in forward and mux tests???

The json-file test flake in mux is very strange.  It looks like an extra `n` character is being added to the messages?
```
"message": "67108864 is less then the default 128m.  Setting to default.n",
"message": "2018-02-09 23:12:39 +0000 [warn]: 'block' action stops input process until the buffer full is resolved. Check your pipeline this action is fit or notn",
```
here is an example of a raw log message from a file:
```
/var/log/containers/logging-kibana-1-tvhbm_logging_kibana-962e091dcc81d53f2231df604c6c3469db2af890781a977031fbd21b3de03b7e.log:{"log":"{"type":"response","@timestamp":"2018-02-09T23:13:13Z","tags":[],"pid":66,"method":"get","statusCode":404,"req":{"url":"/bd0cf33bdd464f9f924a47c80bc877f5","method":"get","headers":{"user-agent":"curl/7.29.0","host":"localhost:5601","accept":"*/*"},"remoteAddress":"127.0.0.1","userAgent":"127.0.0.1"},"res":{"statusCode":404,"responseTime":2,"contentLength":9},"message":"GET /bd0cf33bdd464f9f924a47c80bc877f5 404 2ms - 9.0B"}n","stream":"stdout","time":"2018-02-09T23:13:13.559006504Z"}
```
notice the extra `n` after the closing `}`
wtf? `docker --log-driver=json-file` bug?

portante:
Could that be a \n that was process as “escape the n” leaving it as a ‘n’?

richm:
> Could that be a \n that was process as “escape the n” leaving it as a ‘n’?

Yes, looks like that's it.  So, probably false alarm.  Looks like there is a bug in `artifact_out` - it strips off the `\`

richm:
see https://github.com/openshift/origin-aggregated-logging/pull/974

richm:
ok - everything is finally passing - please review

test/zzzz-bulk_rejection.sh
```diff
@@ -0,0 +1,259 @@
...
+bulktestjson=$( mktemp )
+python -c 'import sys
+for ii in xrange(0,int(sys.argv[1])):
+    print """
+{{"index":{{"_index":"bulkindextest","_type":"bulkindextest"}}}}
+{{"field0":"value value value {0}"}}
+""".format(ii)
+' 100000 > $bulktestjson
+wc $bulktestjson | artifact_out
+ls -al $bulktestjson | artifact_out
```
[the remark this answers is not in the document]
richm: yeah, just for a sanity check

test/zzzz-bulk_rejection.sh
```diff
@@ -0,0 +1,259 @@
...
+count=40
+countops=500
+os::log::info Adding $count project log records and $countops operations log records . . .
+starttime=$( date +%s )
+# not sure why, but it seems the operations messages get sent
+# to es much faster - so add more of them to see if we can get
+# them to be rejected
+for jj in $( seq 1 $count ) ; do
+    add_test_message "$uuid_es-$jj"
+    os::log::debug added es message $uuid_es-$jj
```
lukas-vlcek: Do I get it correctly that calling 40x times the `add_test_message` should result into BulkRequestRejection? Is it because we changed bulk thread configuration in ES prior that? In other words is it correct to assume that fluentd will produce more than one bulk request here?

lukas-vlcek:
Left one question.
Otherwise LGTM. Thanks!

test/zzzz-bulk_rejection.sh
richm: > Do I get it correctly that calling 40x times the add_test_message should result into BulkRequestRejection?

The intention is that fluentd gets several bulk index rejection exceptions when trying to send these test messages.  I picked 40 because it seemed like a good number based on repeated trial and error.

>  Is it because we changed bulk thread configuration in ES prior that?

That, and the fact that there are multiple bulk index requests from curl happening at the same time.  That causes fluentd to fail its bulk index requests.

> In other words is it correct to assume that fluentd will produce more than one bulk request here?

There is no way to guarantee that, but in my testing, there were always at least 4-5 bulk index rejections reported by fluentd.

lukas-vlcek: >> Is it because we changed bulk thread configuration in ES prior that?

> That, and the fact that there are multiple bulk index requests from curl happening at the same time. That causes fluentd to fail its bulk index requests.

That means the `add_test_message` is async? Or other curl requests interfere with this?

richm: > the add_test_message is async?

Not sure what you mean by `async` in this context.  `add_test_message` uses `curl` to hit kibana with an unknown URL which causes it to emit a `404 not found` to its log, which is then picked up by docker, then fluentd, which tries to submit a bulk index request to Elasticsearch with this record and usually many more.  `add_test_message` returns immediately - it does not run as a background job.

This function starts 6 curl processes running in parallel in the background, all hitting Elasticsearch with as many bulk index requests as possible.  https://github.com/openshift/origin-aggregated-logging/pull/933/files#diff-1e94ccb79adaf3b43bc5ad62f61efd0bR96  All of these curl processes cause Elasticsearch to become overloaded and start rejecting bulk index requests.

richm:
Thanks @lukas-vlcek - Can I get a lgtm from someone else?  Please review.

richm:
```
FAILURE after 59.976s: test/zzz-correct-index-names.sh:54: executing 'curl_es logging-es-ops-data-master-asow3bq3-1-58f9s /.operations.*/_search?q=message:e192bc7e7884400d85e3dc5b160c35fc |         jq '.hits.hits | map(select(._source.kubernetes.namespace_name == "default")) | length | . > 0'' expecting any result and text '^true$'; re-trying every 0.2s until completion or 60.000s: the command timed out
```
Might need to increase the timeout for this test - the records are logged - https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_origin-aggregated-logging/933/test_pull_request_origin_aggregated_logging_ansible_json_file/1239/artifacts/scripts/entrypoint/logs/containers/k8s_test_test-pod_default_26f342e3-1bdc-11e8-a471-0ef28cadd622_0.log

/test json-file

PR #937, opened 2018-02-08
parse log field for mux minimal mode for json-file logs

richm:
When using MUX_CLIENT_MODE=minimal and json-file logs, we have to
parse the `log` field - mux only runs in journald mode so expects
the `MESSAGE` field to contain the JSON value to parse.
In the mux tests, make sure the `MESSAGE` field is populated with
the contents of the `log` field so that it will be parsed.
There appears to be a new behavior with the internal openshift
logging which is causing test flakes in the mux test - query the
`message` field only instead of `_all` (all fields) for the test
data.

### DIFF
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -118,8 +118,12 @@ if [ -n "${MUX_CLIENT_MODE:-}" ] ; then
         cp -f $CFG_DIR/user/filter-pre-mux-client.conf $CFG_DIR/openshift/$mux_client_filename
     fi
     # rm k8s meta plugin - do not hit the API server - just do json parsing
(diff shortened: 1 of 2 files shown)


richm:
/retest
curator flake

richm:
/test logging
I can't seem to reproduce the failure - flake?

richm:
have not seen this failure before . . .
```
FAILURE after 394.661s: test/cluster/rollout.sh:28: executing 'oc rollout status deploymentconfig/logging-curator' expecting success: the command returned the wrong error code
Waiting for rollout to finish: 0 of 1 updated replicas are available...

error: replication controller "logging-curator-1" has failed progressing
```

PR #954, opened 2018-02-12
impl use of openssl by SearchGuard

jcantrill:
The PR makes ES and SearchGuard utilize openssl per SG recommendations


jcantrill:
@richm @ewolinetz @simo5  This PR uses the version specified by SG documentation which is supposed to EOL this month.  Not sure if this is an issue:
```
2018-02-12 22:24:28 INFO  DefaultSearchGuardKeyStore:420 - Open SSL OpenSSL 1.0.1e-fips 11 Feb 2013 available

```

simo5:
@jcantrill unclear what is EOLed ?

jcantrill:
@simo5 ES2.x is shortly in EOL but also the security plugin we use, SearchGuard is EOL this version.  I'm more interested in knowing if the Openssl version listed in the logs is an issue or if we should leave well enough alone.  The version listed in the comment doesnt seem to match the rpm that was installed which was 1.0.2

simo5:
RHEL7 has only 1.0.2, is that plugin pulling down its own openssl library ?

jcantrill:
```
<jcantril> simo, docs say it is dynamically linked from here http://docs.search-guard.com/v5/openssl which is one of these jars...http://repo1.maven.org/maven2/io/netty/netty-tcnative/1.1.33.Fork17/  i installed on centos 'yum install openssl' which pulled in 1.0.2
<jcantril> simo, is it possible for a consumer of openssl to downgrade the version functionality somehow
<jcantril> simo, and this is the rpm installed in the image:
<jcantril> $ rpm -qva | grep openssl
<jcantril> openssl-libs-1.0.2k-8.el7.x86_64
<jcantril> openssl-1.0.2k-8.el7.x86_64
```

jcantrill:
depends on https://github.com/fabric8io/openshift-elasticsearch-plugin/pull/121

richm:
re: openssl version difference - can you verify that there are no other openssl libraries on the system?  Can you verify that elasticsearch or sg or some jar that they pull in does not include bundled openssl libraries?

PR #955, opened 2018-02-13
bz1515715 -- remote syslog hostname

nhosoi:
Changes in the CI test scrip remote-syslog.sh
-  use artifact_log, artifact_out and add_test_message.
- testing fluentd -> mux -> rsyslogd, as well.
- added cleanup function
- use_record for hostname, too.
  use_record true, by default.
  Including "container_name", "namespace_name", and "pod_name" in the
  output content, when use_record is true.
- adding test case to cover use_record (Test 6)
  use rsyslogd on the node
  modified /etc/rsyslog.conf so that rsyslogd only takes the input
  from fluentd remote-syslog plugin and outputs more log fields such
  as priority, facility, tag, etc. in the Test 6.

Migrated from https://github.com/openshift/origin-aggregated-logging/pull/909

### DIFF
--- a/fluentd/configs.d/openshift/output-operations.conf
+++ b/fluentd/configs.d/openshift/output-operations.conf
@@ -1,6 +1,7 @@
 <match output_ops_tag journal.** system.var.log** mux.ops audit.log** %OCP_FLUENTD_TAGS%>
   @type copy
   @include ../dynamic/output-es-ops-config.conf
+  @include ../dynamic/output-remote-syslog.conf
   @include ../user/output-ops-extra-*.conf
   @include ../user/secure-forward.conf
 </match>
--- a/fluentd/out_syslog.rb
+++ b/fluentd/out_syslog.rb
@@ -3,7 +3,6 @@
 
 require 'fluent/mixin/config_placeholders'
 module Fluent
-
   class SyslogOutput < Fluent::Output
     # First, register the plugin. NAME is the name of this plugin
     # and identifies the plugin in the configuration file.
@@ -11,7 +10,7 @@ module Fluent
 
     # This method is called before starting.
 
-    config_param :remote_syslog, :string, :default => nil
+    config_param :remote_syslog, :string, :default => ""
     config_param :port, :integer, :default => 25
     config_param :hostname, :string, :default => ""
     config_param :remove_tag_prefix, :string, :default => nil
@@ -26,6 +25,7 @@ module Fluent
       super
       require 'socket'
       require 'syslog_protocol'
+      require 'securerandom'
     end
 
     def configure(conf)
@@ -45,6 +45,7 @@ module Fluent
       if not @payload_key
         @payload_key = "message"
       end
+      @random_string = SecureRandom.hex
     end
 
 
@@ -73,13 +74,13 @@ module Fluent
           if @use_record && record.key?('systemd') && (record['systemd']).key?('u') && (record['systemd']['u']).key?('SYSLOG_FACILITY')
             fval = record['systemd']['u']['SYSLOG_FACILITY'].to_i
             if (1..23).include?(fval)
-                @packet.facility = fval
-            elsif record['systemd']['u']['SYSLOG_FACILITY'].eql? '0'
-                @packet.facility = 0
-            elsif record['systemd']['u']['SYSLOG_FACILITY'].eql? 'kern'
-                @packet.facility = 'kern'
+              @packet.facility = fval
             else
+              if record['systemd']['u']['SYSLOG_FACILITY'].eql? '0'
+                @packet.facility = 0
+              else
                 @packet.facility = record['systemd']['u']['SYSLOG_FACILITY'] || @facility
+              end
             end
           elsif record.key?('_KERNEL_DEVICE')
             @packet.facility = 'kern'
@@ -91,36 +92,48 @@ module Fluent
           @packet.facility = @facilty
           @packet.severity = @severity
         end
-        if record['time']
-          time = Time.parse(record['time'])
+        time = if record['time']
+          Time.parse(record['time'])
         else
-          time = Time.now
+          Time.now
         end
         @packet.time = time
-        if @tag_key.any?
+        @packet.tag = @random_string
+        if @tag_key.any? && record.any?
+          # tag_key is an array type
+          # E.g., tag_key ident,systemd.u.SYSLOG_IDENTIFIER,ident
+          #       tkey = ident
+          #       tkey = systemd.u.SYSLOG_IDENTIFIER
           @tag_key.each { |tkey|
+            # ident => record[ident]
+            # systemd.u.SYSLOG_IDENTIFIER => record[systemd][u][SYSLOG_IDENTIFIER]
             mytag = record
+            # check if tkey is '.' separated.
             tkey.split('.').each { |p|
-              break if ! mytag.key?(p)
+              if ! mytag.key?(p)
+                if ! p.eql? tkey
+                  log.debug "out:syslog: #{p} from #{tkey} in tag_key #{@tag_key} is not a key of record."
+                end
+                break
+              end
               mytag = mytag[p]
             }
             next if ! mytag.is_a? String
-            next if mytag.empty?
             @packet.tag = mytag[0..31].gsub(/[\[\]\s]/,'') # tag is trimmed to 32 chars for syslog_protocol gem compatibility
             break
           }
         end
-        if @packet.tag.nil? || @packet.tag.empty?
+        if @packet.tag.eql? random_string
           @packet.tag = tag[0..31] # tag is trimmed to 32 chars for syslog_protocol gem compatibility
         end
         packet = @packet.dup
-        if @use_record && (record.key?('kubernetes')) && @record[@payload_key]
-            packet.content = (((record['kubernetes']).key?('namespace_name')) ? 'namespace_name=' + record['kubernetes']['namespace_name'] + ', ' : '' ) + \
-                             (((record['kubernetes']).key?('container_name')) ? 'container_name=' + record['kubernetes']['container_name'] + ', ' : '' ) + \
-                             (((record['kubernetes']).key?('pod_name')) ? 'pod_name=' + record['kubernetes']['pod_name'] + ', ' : '' ) + \
-                             @payload_key + '=' + record[@payload_key]
+        packet.content = if @use_record && (record.key?('kubernetes')) && record[@payload_key]
+          (((record['kubernetes']).key?('namespace_name')) ? 'namespace_name=' + record['kubernetes']['namespace_name'] + ', ' : '' ) + \
+          (((record['kubernetes']).key?('container_name')) ? 'container_name=' + record['kubernetes']['container_name'] + ', ' : '' ) + \
+          (((record['kubernetes']).key?('pod_name')) ? 'pod_name=' + record['kubernetes']['pod_name'] + ', ' : '' ) + \
+          @payload_key + '=' + record[@payload_key]
         else
-            packet.content = record[@payload_key]
+          record[@payload_key]
         end
         @socket.send(packet.assemble, 0, @remote_syslog, @port)
     }
--- a/fluentd/out_syslog.rb
+++ b/fluentd/out_syslog.rb
@@ -15,7 +15,7 @@ module Fluent
     config_param :port, :integer, :default => 25
     config_param :hostname, :string, :default => ""
     config_param :remove_tag_prefix, :string, :default => nil
-    config_param :tag_key, :string, :default => nil
+    config_param :tag_key, :array, default: nil
     config_param :facility, :string, :default => 'user'
     config_param :severity, :string, :default => 'debug'
     config_param :use_record, :string, :default => nil
@@ -67,11 +67,27 @@ module Fluent
       tag = tag.sub(@remove_tag_prefix, '') if @remove_tag_prefix
       chain.next
       es.each {|time,record|
-        @packet.hostname = hostname
         if @use_record
-          @packet.facility = record['facility'] || @facilty
-          @packet.severity = record['severity'] || @severity
+          @packet.hostname = record['hostname'] || hostname
+          @packet.severity = (record['level'].eql? 'warning')? 'warn' : record['level'] || @severity
+          if @use_record && record.key?('systemd') && (record['systemd']).key?('u') && (record['systemd']['u']).key?('SYSLOG_FACILITY')
+            fval = record['systemd']['u']['SYSLOG_FACILITY'].to_i
+            if (1..23).include?(fval)
+                @packet.facility = fval
+            elsif record['systemd']['u']['SYSLOG_FACILITY'].eql? '0'
+                @packet.facility = 0
+            elsif record['systemd']['u']['SYSLOG_FACILITY'].eql? 'kern'
+                @packet.facility = 'kern'
+            else
+                @packet.facility = record['systemd']['u']['SYSLOG_FACILITY'] || @facility
+            end
+          elsif record.key?('_KERNEL_DEVICE')
+            @packet.facility = 'kern'
+          else
+            @packet.facility = record['facility'] || @facility
+          end
         else
+          @packet.hostname = hostname
           @packet.facility = @facilty
           @packet.severity = @severity
         end
@@ -81,17 +97,31 @@ module Fluent
           time = Time.now
         end
         @packet.time = time
-        @packet.tag = if @tag_key
-                        begin
-                          record[@tag_key][0..31].gsub(/[\[\]\s]/,'') # tag is trimmed to 32 chars for syslog_protocol gem compatibility
-                        rescue
-                          tag[0..31] # tag is trimmed to 32 chars for syslog_protocol gem compatibility
-                        end
-                      else
-                          tag[0..31] # tag is trimmed to 32 chars for syslog_protocol gem compatibility
-                      end
+        if @tag_key.any?
+          @tag_key.each { |tkey|
+            mytag = record
+            tkey.split('.').each { |p|
+              break if ! mytag.key?(p)
+              mytag = mytag[p]
+            }
+            next if ! mytag.is_a? String
+            next if mytag.empty?
+            @packet.tag = mytag[0..31].gsub(/[\[\]\s]/,'') # tag is trimmed to 32 chars for syslog_protocol gem compatibility
+            break
+          }
+        end
+        if @packet.tag.nil? || @packet.tag.empty?
+          @packet.tag = tag[0..31] # tag is trimmed to 32 chars for syslog_protocol gem compatibility
+        end
         packet = @packet.dup
-        packet.content = record[@payload_key]
+        if @use_record && (record.key?('kubernetes')) && @record[@payload_key]
+            packet.content = (((record['kubernetes']).key?('namespace_name')) ? 'namespace_name=' + record['kubernetes']['namespace_name'] + ', ' : '' ) + \
+                             (((record['kubernetes']).key?('container_name')) ? 'container_name=' + record['kubernetes']['container_name'] + ', ' : '' ) + \
+                             (((record['kubernetes']).key?('pod_name')) ? 'pod_name=' + record['kubernetes']['pod_name'] + ', ' : '' ) + \
+                             @payload_key + '=' + record[@payload_key]
+        else
+            packet.content = record[@payload_key]
+        end
         @socket.send(packet.assemble, 0, @remote_syslog, @port)
     }
     end
--- a/fluentd/out_syslog_buffered.rb
+++ b/fluentd/out_syslog_buffered.rb
@@ -14,7 +14,7 @@ module Fluent
     config_param :port, :integer, :default => 25
     config_param :hostname, :string, :default => ""
     config_param :remove_tag_prefix, :string, :default => nil
-    config_param :tag_key, :string, :default => nil
+    config_param :tag_key, :array, default: nil
     config_param :facility, :string, :default => 'user'
     config_param :severity, :string, :default => 'debug'
     config_param :use_record, :string, :default => nil
@@ -91,11 +91,27 @@ module Fluent
 
     def send_to_syslog(tag, time, record)
       tag = tag.sub(@remove_tag_prefix, '') if @remove_tag_prefix
-      @packet.hostname = hostname
       if @use_record
-        @packet.facility = record['facility'] || @facilty
-        @packet.severity = record['severity'] || @severity
+        @packet.hostname = record['hostname'] || hostname
+        @packet.severity = (record['level'].eql? 'warning')? 'warn' : record['level'] || @severity
+        if @use_record && record.key?('systemd') && (record['systemd']).key?('u') && (record['systemd']['u']).key?('SYSLOG_FACILITY')
+          fval = record['systemd']['u']['SYSLOG_FACILITY'].to_i
+          if (1..23).include?(fval)
+              @packet.facility = fval
+          elsif record['systemd']['u']['SYSLOG_FACILITY'].eql? '0'
+              @packet.facility = 0
+          elsif record['systemd']['u']['SYSLOG_FACILITY'].eql? 'kern'
+              @packet.facility = 'kern'
+          else
+              @packet.facility = record['systemd']['u']['SYSLOG_FACILITY'] || @facility
+          end
+        elsif record.key?('_KERNEL_DEVICE')
+          @packet.facility = 'kern'
+        else
+          @packet.facility = record['facility'] || @facility
+        end
       else
+        @packet.hostname = hostname
         @packet.facility = @facilty
         @packet.severity = @severity
       end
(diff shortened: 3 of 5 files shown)


nhosoi:
As described in the bug 1543761 - "openshift_logging_fluentd_remote_syslog_use_record=true is dysfunctional", severity and facility are not respected. I'm suspecting we do not store them in the record hash? (I.e., record['severity'] and record['facility']) I noticed we set record['level'] in viaq plugin. But replacing record['severity'] in the remote-syslog plugin did not work... @richm, do you have any idea how to get them in the remote-syslog plugin? Also, the original tag is lost at this phase since we had replaced each tag with output_tag. Can we keep the original tag somewhere to pass to remote-syslog plugin?

richm:
There is a crazy test failure going on here.  the `.operations` index is being created in the non es-ops instance!  https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_origin-aggregated-logging/955/test_pull_request_origin_aggregated_logging_ansible/1733/artifacts/scripts/entrypoint/artifacts/es.indices.before

The es instance contains `.operations*` but not es-ops?????

richm:
I don't see how it is related to this PR - it must be a test flake that snuck in somehow.

nhosoi:
Finally...  There were some issues but the hardest to catch was caused by this:
Bug 1536651 - logging-mux not working in 3.7.z when logging installed with openshift_logging_use_mux=true

jenkins CI test runs with USE_MUX=true, but no openshift_logging_mux_client_mode.  So, mux pod is generated, but not being used...
```
+ ansible-playbook -vv --become --become-user root --connection local 
--inventory sjb/inventory/ 
-e debug_level=2 
-e openshift_logging_elasticsearch_proxy_image_prefix=docker.io/openshift/ 
-e openshift_hosted_logging_hostname=kibana.127.0.0.1.nip.io 
x -e openshift_deployment_type=origin 
x -e openshift_logging_install_logging=True 
-e openshift_logging_image_prefix=openshift/origin- 
-e openshift_logging_master_public_url=https://localhost:8443 
-e openshift_master_logging_public_url=https://kibana.127.0.0.1.nip.io 
-e openshift_logging_use_mux=True 
-e openshift_logging_mux_allow_external=True 
-e openshift_logging_es_allow_external=True 
-e openshift_logging_es_ops_allow_external=True 
-e openshift_logging_image_version=latest 
/usr/share/ansible/openshift-ansible/playbooks/openshift-logging/config.yml --skip-tags=update_master_config
```
To learn whether the test is running with mux or without mux in remote-syslog.conf, I was checking the existence of the mux pod.  Instead, I did on the muxkeys dir in the fluentd pod.  Then, the test passed. :)
https://github.com/openshift/origin-aggregated-logging/pull/955/commits/532836f9b94dfe04faf06c9bc4dd612420b21159

To make the jenkins CI test more robust, we could add "-e openshift_logging_mux_client_mode=maximal" to the xml files in aos-cd-jobs, in which "-e openshift_logging_use_mux=True" is set?

Or once this PR is merged, mux_client_mode=maximal would be set by default...
https://github.com/openshift/openshift-ansible/pull/7192

nhosoi:
No not a problem.  It was a very good practice for me.  And I agree with you we should wait for the openshift/openshift-ansible#7192 merge.

Another good thing was I found this bug... :p  Since it's an independent bug from this pr, I'm going to submit a new pr for this patch and clean up this pr...
https://github.com/openshift/origin-aggregated-logging/pull/955/commits/e455e856c9e04852a80c086320ca8d87908b0024

nhosoi:
Rebased and squashed.

fluentd/configs.d/openshift/output-operations.conf
```diff
@@ -1,6 +1,7 @@
-<match output_ops_tag journal.** system.var.log** mux.ops audit.log** %OCP_FLUENTD_TAGS%>
+<match ops_fluentd_tag journal.system** system.var.log** mux.ops audit.log** %OCP_FLUENTD_TAGS%>
```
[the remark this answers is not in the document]
nhosoi: The tag name is being visible in the forwarded rsyslog output.
The QE filed bug [https://bugzilla.redhat.com/show_bug.cgi?id=1543761] complained that "The output_tag is also wrong".  I wanted to make it a little more user friendly.  Could there be a better/clearer keyword you could think of?

This is the new behaviour:
If use_record=true, operations logs would have systemd.u.SYSLOG_IDENTIFIER value.
Other messages from the containers have the tag "fluentd_tag".
```
Sample logs in /var/log/messages in the precise format:
3,17,Feb 15 14:52:25,ip-172-18-5-234.ec2.internal,rsyslogTestTag:, rsyslogTestMessage-20180215-145124
6,16,Feb 15 14:52:41,ip-172-18-5-234.ec2.internal,fluentd_tag:, namespace_name=logging, container_name=kibana, pod_name=logging-kibana-1-qr6pl, message=GET /testKibanaMessage-20180215-145225 404 4ms - 9.0B
```
If use_record=false, operations log has "ops_fluentd_tag" and container log has "fluentd_tag"

By default, use_record=false.  (I have to fix "use_record true, by default." in the git comment, which is wrong... :p)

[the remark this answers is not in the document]
richm: Changing the tags is going to make backporting anything very difficult and error prone.

Changing the tags also might cause problems for customers if they are already using `output_tag` and `output_ops_tag`.

```
$ logger -t newYear -p local0.notice happy_new_year

# cat /var/log/rsyslog.log | grep -i year
local3 info Feb  9 03:21:58 logging-fluentd-pzgrk output_tag: happy_new_year [[<158>Feb  9 03:21:58 logging-fluentd-pzgrk output_tag: happy_new_year]]

The output_tag is also wrong, it should be newYear in this case
```
The `-t` argument to logger is the `ident` field aka the `systemd.u.SYSLOG_IDENTIFIER` field, not the fluentd tag.  I don't think we should be sending the fluentd tag with the remote syslog protocol.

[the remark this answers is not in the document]
nhosoi: Ok.  I was hoping to put some meaningful word like retagged_by_fluentd or something like that.  But it seems it's rather harmful to change it.
- I'm reverting the tag name change and put them back to output_tag and output_ops_tag.
- remote-syslog plugin is implemented to send the fluentd tag to rsyslogd by default.  (Or you could specify a record key to alter.)  I could change the default behaviour to send ident/systemd.u.SYSLOG_IDENTIFIER as a tag to rsyslogd for the operations logs.  How about the container logs?  I could not find ident/systemd.u.SYSLOG_IDENTIFIER in the container logs.   Do you have any suggestion which field could be used for the container logs' tag?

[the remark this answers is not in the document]
richm: > remote-syslog plugin is implemented to send the fluentd tag to rsyslogd by default

In which field of the output does it add the fluentd tag?  Is this some sort of spec/rfc, or is it just by convention?

For example: https://tools.ietf.org/html/rfc3164#section-5.3
What they refer to as the `TAG` field is really the ident/systemd.u.SYSLOG_IDENTIFIER

https://tools.ietf.org/html/rfc5424#section-6.2.5 - here it is called the APP-NAME field, but it is really the same as the ident/systemd.u.SYSLOG_IDENTIFIER

If it really is specific to the remote-syslog plugin, and users of the remote-syslog plugin expect the fluentd tag to be in that field, then
- we have to do that by default so as not to break existing applications
- users should not care, by default, what that value is
- if the remote-syslog plugin already allows you to change that field value, then we should still honor that

Given that, I don't see why this is wrong:
```
local3 info Feb  9 03:21:58 logging-fluentd-pzgrk output_tag: happy_new_year [[<158>Feb  9 03:21:58 logging-fluentd-pzgrk output_tag: happy_new_year]]
```
if the remote-syslog plugin is doing exactly what it is supposed to do in that case, it isn't "wrong".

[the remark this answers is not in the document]
richm: re: container logs - when reading from journald, container logs typically use `docker.service` (which will probably change when cri-o is used instead of docker).  When reading from json-file, correct, there is no ident/systemd.u.SYSLOG_IDENTIFIER field.  Does the remote-syslog plugin already have the ability to specify what to put in the tag/ident field if there is no appropriate value in the record?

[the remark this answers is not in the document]
nhosoi: > In which field of the output does it add the fluentd tag?

This code assigns the fluentd tag to the tag field of the packet to be sent to rsyslogd.
https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/out_syslog_buffered.rb#L115

Regarding this question:
> Given that, I don't see why this is wrong:
> local3 info Feb  9 03:21:58 logging-fluentd-pzgrk output_tag: happy_new_year [[<158>Feb  9 03:21:58 logging-fluentd-pzgrk output_tag: happy_new_year]]

The test case in bz1543761 sets use_record=true.
```openshift_logging_fluentd_remote_syslog_use_record=true```
If it is true, the facility and the severity from this logger command are supposed to be respected.
```$ logger -t newYear -p local0.notice happy_new_year```
However, the log found in /var/log/rsyslog.log did not contain ```local0 notice``` but ```local3 info``` which pair comes from output-remote-syslog.conf [1].  Junqi claimed the value pair is different from the expected one.  So is the tag.  First, I thought the tag part is not a bug but a spec and it could be acceptable.  But rethinking of it, I had to agree losing ```-t newYear``` was not nice.  So, I added the code "if use_record is true, pass ident/systemd.u.SYSLOG_IDENTIFIER to rsyslogd."
```
# cat /var/log/rsyslog.log | grep -i year
local3 info Feb  9 03:21:58 logging-fluentd-pzgrk output_tag: happy_new_year [[<158>Feb  9 03:21:58 logging-fluentd-pzgrk output_tag: happy_new_year]]
```
The facility, severity bug is caused here in the remote-syslog plugin code, where it tries to get the facility, severity values as follows.
https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/out_syslog_buffered.rb#L95-L97

In reality, record['facility'] and record['severity'] do not exist and some tweaks were needed to pass the values to rsyslogd.
https://github.com/nhosoi/origin-aggregated-logging/blob/bz1515715/fluentd/out_syslog_buffered.rb#L96-L103

[1]
```
<store>
@type syslog_buffered
remote_syslog 172.18.9.41
port 514
hostname logging-fluentd-####
facility local3
severity info
use_record false
</store>
```
> Does the remote-syslog plugin already have the ability to specify what to put in the tag/ident field if there is no appropriate value in the record?

This is the current tag retrieving code.
https://github.com/nhosoi/origin-aggregated-logging/blob/bz1515715/fluentd/out_syslog_buffered.rb#L120-L136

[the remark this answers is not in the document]
Comment (richm) [reply fluentd/configs.d/openshift/output-operations.conf]: > This code assigns the fluentd tag to the tag field of the packet to be sent to rsyslogd.
https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/out_syslog_buffered.rb#L115

- by default it uses the fluentd tag - this is the existing behavior and should not be changed
- user can specify tag_key to use a different value from the record, so is customizable already

> However, the log found in /var/log/rsyslog.log did not contain local0 notice but local3 info which pair comes from output-remote-syslog.conf [1]. Junqi claimed the value pair is different from the expected one.

That sounds like a real bug that we should fix.

> So is the tag.

I don't see this is a bug.  The tag is set correctly to the fluentd tag.  This is not affected by `use_record=true` or `false`.  If the user doesn't want the fluentd tag to be in this field, the user can specify `tag_key`

> First, I thought the tag part is not a bug but a spec and it could be acceptable. But rethinking of it, I had to agree losing -t newYear was not nice.

But that's the way the remote-syslog plugin works - it isn't a bug, it is a feature.  If this is not desirable, then either use `tag_key`, or QE can file a separate RFE to make the `-t` `ident` field preserved (somewhere?) by default.

> So, I added the code "if use_record is true, pass ident/systemd.u.SYSLOG_IDENTIFIER to rsyslogd."

I don't think that should be done.  From what I can see, the user can set `tag_key` if they want the `ident` value instead of the fluentd tag.

If QE insists that `ident` should be used here, they should file a separate bz/RFE.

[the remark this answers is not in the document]
Comment (nhosoi) [reply fluentd/configs.d/openshift/output-operations.conf]: > But that's the way the remote-syslog plugin works - it isn't a bug, it is a feature. If this is not desirable, then either use tag_key, or QE can file a separate RFE to make the -t ident field preserved (somewhere?) by default.

Sounds good.  I have some question here... 

In filter-viaq-data-model.conf, ident and SYSLOG_IDENTIFIER are being removed as follows.  Could it be ok to leave and refer from remote-syslog plugin with ```tag_key ident``` or ```tag_key SYSLOG_IDENTIFIER```?  Unfortunately, only one tag_key is accepted, though.  Do we want to extend it to multiple?
```
  <formatter>
    tag "system.var.log**"
    remove_keys host,pid,ident
  </formatter>
  <formatter>
    tag "journal.system**" as well as "kubernetes.journal.container**"
    remove_keys ...,SYSLOG_IDENTIFIER,...
  </formatter>
```
Another question is in the current remote-syslog plugin gode, tag_key is referred as record[@tag_key] and there is no way to access ```record[syslogd][u][SYSLOG_IDENTIFIER]```.  Do you think we need some change to accept ```tag_key systemd.u.SYSLOG_IDENTIFIER``` and convert it to ```record[syslogd][u][SYSLOG_IDENTIFIER]```?  Or once we could make ident and SYSLOG_IDENTIFIER work as tag_key, we don't need this ```tag_key systemd.u.SYSLOG_IDENTIFIER``` support?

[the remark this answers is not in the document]
Comment (richm) [reply fluentd/configs.d/openshift/output-operations.conf]: > In filter-viaq-data-model.conf, ident and SYSLOG_IDENTIFIER are being removed as follows.

Right, because `ident` from /var/log/messages (which is deprecated) and `SYSLOG_IDENTIFIER` from the journal are converted to `systemd.u.SYSLOG_IDENTIFIER`.  Once that is done, we don't need the original `ident` and `SYSLOG_IDENTIFIER` fields any more.

Multiple tag keys would also be a good enhancement.

I guess the problem is that you cannot use a nested `tag_key`.  I think then the fix should be that you should be able to specify something like `tag_key systemd.u.SYSLOG_IDENTIFIER` which will use `record['systemd']['u']['SYSLOG_IDENTIFIER']`.  That would be a good enhancement to this plugin.   rewrite-tag-filter added something like this: https://github.com/fluent/fluent-plugin-rewrite-tag-filter#nested-attributes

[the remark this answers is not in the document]
Comment (nhosoi) [reply fluentd/configs.d/openshift/output-operations.conf]: Ok. Thanks!

**Comment by nhosoi** [comment]
Squashed and updated the comments.

File: test/docker_audit.sh [hunk 2]
```diff
@@ -36,11 +36,12 @@ esopspod=${esopspod:-$espod}
 # 
 logs_before=$( get_logs_count $espod '/project.*/' )
 ops_logs_before=$( get_logs_count $esopspod '/.operations.*/' )
+ops_logs_expected=$( ops_logs_before+5 )
```
richm: you have to use `((`
```
 +ops_logs_expected=$(( ops_logs_before+5 ))
```

nhosoi: Ah...  Thanks, Rich!  Pushed the fix...

fluentd/out_syslog.rb
```diff
@@ -67,11 +67,23 @@ def emit(tag, es, chain)
...
       chain.next
       es.each {|time,record|
-        @packet.hostname = hostname
         if @use_record
-          @packet.facility = record['facility'] || @facilty
-          @packet.severity = record['severity'] || @severity
+          @packet.hostname = record['hostname'] || hostname
+          @packet.severity = (record['level'].eql? 'warning')? 'warn' : record['level'] || @severity
+          if @use_record && record.key?('systemd') && (record['systemd']).key?('u') && (record['systemd']['u']).key?('SYSLOG_FACILITY')
+            fval = record['systemd']['u']['SYSLOG_FACILITY'].to_i
```
richm: `to_i` on a non-integer string value will return a `0` - so it will never hit the `else` below.
see https://stackoverflow.com/questions/24980295/strictly-convert-string-to-integer-or-nil

fluentd/out_syslog_buffered.rb
```diff
@@ -91,11 +91,23 @@ def write(chunk)
...
     def send_to_syslog(tag, time, record)
       tag = tag.sub(@remove_tag_prefix, '') if @remove_tag_prefix
-      @packet.hostname = hostname
       if @use_record
-        @packet.facility = record['facility'] || @facilty
-        @packet.severity = record['severity'] || @severity
+        @packet.hostname = record['hostname'] || hostname
+        @packet.severity = (record['level'].eql? 'warning')? 'warn' : record['level'] || @severity
+        if @use_record && record.key?('systemd') && (record['systemd']).key?('u') && (record['systemd']['u']).key?('SYSLOG_FACILITY')
+          fval = record['systemd']['u']['SYSLOG_FACILITY'].to_i
```
richm: see above

fluentd/out_syslog.rb
nhosoi: You are right, @richm ...  I've committed another patch.

nhosoi:
[(5) Include and support fluentd-syslog-plugin[logging]](https://trello.com/c/LCte3O0n/469-5-include-and-support-fluentd-syslog-pluginlogging)

nhosoi:
Test failures are depending on this PR: openshift/origin-aggregated-logging#993

nhosoi:
Rebased against master to pick up openshift/origin-aggregated-logging#993.

nhosoi:
Waiting for this PR merged: https://github.com/openshift/origin-aggregated-logging/pull/1008

Wondering this flake disappears once the pr/1008 is picked up???
```
Running test/zzzz-bulk_rejection.sh:131: executing 'grep -q BulkIndexQueueFull /var/log/fluentd.log' expecting success; re-trying every 0.2s until completion or 300.000s...
FAILURE after 299.700s: test/zzzz-bulk_rejection.sh:131: executing 'grep -q BulkIndexQueueFull /var/log/fluentd.log' expecting success; re-trying every 0.2s until completion or 300.000s: the command timed out
```

test/zzzz-bulk_rejection.sh
```diff
@@ -128,7 +128,7 @@ fpod=$( get_running_pod fluentd )
 
 # wait for BulkIndexQueueFull errors in fluentd log
 flog=/var/log/fluentd.log
-os::cmd::try_until_success "grep -q BulkIndexQueueFull /var/log/fluentd.log" $(( 300 * second ))
+os::cmd::try_until_success "grep -q BulkIndexQueueFull /var/log/fluentd.log" $(( 600 * second ))
```
richm: it should not take this long - is it possible that fluentd is configured to be a mux client, and is sending logs to mux instead of to es directly?

nhosoi: Ahhh, I see...  Thanks, @richm .

That is, if mux is configured, we have to enable DEBUG on mux instead of the client fluentd?  Probably, in that case, the mux's fluentd.log is not visible as /var/log/fluentd.log on the host since it's mounted to the client fluentd not mux?  Let me play with it a little more...

richm: @nhosoi For the bulk index test, MUX_CLIENT_MODE should not be set - fluentd should send logs directly to Elasticsearch - mux should not be used.
If this is what is happening, that means there is a problem with some other test - the other test that ran earlier did not unset MUX_CLIENT_MODE from fluentd.
MUX_CLIENT_MODE should be off by default - it is up to each test that tests mux to set MUX_CLIENT_MODE as needed, and reset it back to deleted after the test.

nhosoi: Yes, @richm , that was it...  remote-syslog.sh did not clean up the environment variable in the cleanup function.  I pushed the fixed version.

fluentd/out_syslog.rb
```diff
@@ -81,17 +97,31 @@ def emit(tag, es, chain)
...
-                        rescue
-                          tag[0..31] # tag is trimmed to 32 chars for syslog_protocol gem compatibility
-                        end
-                      else
-                          tag[0..31] # tag is trimmed to 32 chars for syslog_protocol gem compatibility
-                      end
+        if @tag_key.any?
+          @tag_key.each { |tkey|
+            mytag = record
+            tkey.split('.').each { |p|
```
richm: for example - if record contains `{'a' => 'b'}` and the user specifies the key like `a.b` then you will see an error like this:
```
NoMethodError: undefined method `key?' for "a":String
```
That is, the user specified a key that does not exist.
I guess in this case, we should use the default value - `tag[0..31]`.
This could be a misconfiguration, but we don't do any checking for any of the other key parameters that the user can specify. For example, below, if the user misspells the `@payload_key` then `packet.content` will always be `nil`.

But we don't want to emit a big error every time - what if the user knows that not every record will have that field?  Then we are spamming the fluentd error log with many, many messages.

I'd like to say that we have some sort of flag e.g. `@report_missing_key` that defaults to `true`, and we emit an error message when the record key is not found if this is set to `true`.  And, that value can be set via an env. var. e.g. `report_missing_key '#{REMOTE_SYSLOG_REPORT_MISSING_KEY}'`.

If I were an admin, I would like to know if I misconfigured something, but I would also like to have a knob with which to suppress log messages.

@portante - too many knobs?  or too much spam?

fluentd/out_syslog.rb
```diff
@@ -67,11 +67,27 @@ def emit(tag, es, chain)
...
       es.each {|time,record|
-        @packet.hostname = hostname
         if @use_record
-          @packet.facility = record['facility'] || @facilty
-          @packet.severity = record['severity'] || @severity
+          @packet.hostname = record['hostname'] || hostname
+          @packet.severity = (record['level'].eql? 'warning')? 'warn' : record['level'] || @severity
+          if @use_record && record.key?('systemd') && (record['systemd']).key?('u') && (record['systemd']['u']).key?('SYSLOG_FACILITY')
+            fval = record['systemd']['u']['SYSLOG_FACILITY'].to_i
+            if (1..23).include?(fval)
```
jcantrill: This is more an FYI that a case statement is unlike other languages and can switch on any conditions that evaluate to true.  I would prefer there use in future to multiple if/else blocks

[1] https://www.skorks.com/2009/08/how-a-ruby-case-statement-works-and-what-you-can-do-with-it/

fluentd/out_syslog.rb
jcantrill: This could be more elegantly expressed using #keep_if or #select

https://ruby-doc.org/core-2.2.0/Array.html#method-i-keep_if

fluentd/out_syslog.rb
```diff
@@ -81,17 +97,31 @@ def emit(tag, es, chain)
...
+            @packet.tag = mytag[0..31].gsub(/[\[\]\s]/,'') # tag is trimmed to 32 chars for syslog_protocol gem compatibility
+            break
+          }
+        end
+        if @packet.tag.nil? || @packet.tag.empty?
+          @packet.tag = tag[0..31] # tag is trimmed to 32 chars for syslog_protocol gem compatibility
+        end
         packet = @packet.dup
-        packet.content = record[@payload_key]
+        if @use_record && (record.key?('kubernetes')) && @record[@payload_key]
```
jcantrill: this can be expressed like:

```
packet.content = if @user_record ....
 ....
else
  record[@payload_key]
end
```

fluentd/out_syslog_buffered.rb
```diff
@@ -91,11 +91,27 @@ def write(chunk)
 
     def send_to_syslog(tag, time, record)
       tag = tag.sub(@remove_tag_prefix, '') if @remove_tag_prefix
-      @packet.hostname = hostname
       if @use_record
-        @packet.facility = record['facility'] || @facilty
-        @packet.severity = record['severity'] || @severity
+        @packet.hostname = record['hostname'] || hostname
```
jcantrill: There is a bunch of this code that can be refactored into a module that could be included into both files.  We should consider doing that if time allows.

fluentd/out_syslog.rb
richm: in the interest of moving this along - don't add the config parameter - just catch the exception and break - treat it just like `! mytag.is_a? String`

fluentd/out_syslog_buffered.rb
nhosoi: I agree.  Could you please teach me how to make it happen?   Just creating a method in one class and "require" it from another did not work.  I was hoping to do so using macro but so far no luck...

fluentd/out_syslog.rb
nhosoi: I added this debug print for the purpose.
https://github.com/nhosoi/origin-aggregated-logging/blob/bz1515715/fluentd/out_syslog_buffered.rb#L71

The following is a snippet of the fluentd log from my test run, when I ssue the message with the warning level.  This is too noisy and not so useful... :(
[2018-03-30T00:25:44.216+0000] 2018-03-30 00:25:29 +0000 [warn]: out:syslog_buffered: u from systemd.u.SYSLOG_IDENTIFIER in tag_key ["ident", "systemd.u.SYSLOG_IDENTIFIER", "err.local1"] is not a key of record.
[2018-03-30T00:25:44.216+0000] 2018-03-30 00:25:29 +0000 [warn]: out:syslog_buffered: err from err.local1 in tag_key ["ident", "systemd.u.SYSLOG_IDENTIFIER", "err.local1"] is not a key of record.

richm: ok, but please change the log level to debug instead of warn - we don't want this logged by default

richm: hmm - I see below that it is logged at `debug` level - ok

PR #958, opened 2018-02-15
Fix `.kibana` regex

josefkarasek:
I was trying to fix `' --exclude .searchguard* --exclude .kibana*'` and did not realize that the kibana index is named just `.kibana` and the regex `.kibana*` worked only by accident.

### DIFF
--- a/curator/src/curator_cmd.py
+++ b/curator/src/curator_cmd.py
@@ -95,7 +95,7 @@ class CuratorCmd():
                     + ' --older-than ' + str(count) \
                     + ' --time-unit ' + unit \
                     + ' --exclude ' + shellquote('^' + re.escape('.searchguard.') + '.*$') \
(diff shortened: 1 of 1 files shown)


curator/src/curator_cmd.py
```diff
@@ -95,7 +95,7 @@ def default_index(self):
                     + ' --older-than ' + str(count) \
                     + ' --time-unit ' + unit \
                     + ' --exclude ' + shellquote('^' + re.escape('.searchguard.') + '.*$') \
-                    + ' --exclude ' + shellquote('^' + re.escape('.kibana.') + '.*$')
+                    + ' --exclude ' + shellquote('^' + re.escape('.kibana') + '$')
```
jcantrill: actually... there can be any number of kibana indexes depending on the mode so it should really be '.kibana.*'

josefkarasek:
@richm @jcantrill what do you think now?

PR #960, opened 2018-02-16
Bug 1536651 - logging-mux not working in 3.7.z when logging installed…

nhosoi:
… with openshift_logging_use_mux=true

set MUX_CLIENT_MODE to maximal by default for the mux client.
I.e., if /etc/fluent/muxkeys dir exists (configuring mux system) and
      USE_MUX is not set or false (it is not a mux fluentd) and
      MUX_CLIENT_MODE is null/empty, maximal is set to MUX_CLIENT_MODE.

This pr requires https://github.com/openshift/openshift-ansible/pull/7192.

### DIFF
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -63,8 +63,12 @@ else
     export USE_JOURNAL=true
 fi
 
-if [ ! -d /etc/fluent/muxkeys ]; then
+if [ ! -d /etc/fluent/muxkeys -o "${USE_MUX:-}" = "true" ] ; then
+    # mux is not configured or starting fluentd as mux
     unset MUX_CLIENT_MODE
+else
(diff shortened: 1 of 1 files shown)


richm:
I think this should be handled in ansible by https://github.com/openshift/openshift-ansible/pull/7192 - that is, if openshift_logging_mux_client_mode and MUX_CLIENT_MODE are set correctly in ansible, then no changes to fluentd/run.sh are required.

nhosoi:
> I think this should be handled in ansible by openshift/openshift-ansible#7192 - that is, if openshift_logging_mux_client_mode and MUX_CLIENT_MODE are set correctly in ansible, then no changes to fluentd/run.sh are required.

Ok. Closing this pr.  Thanks, @richm.

PR #961, opened 2018-02-19
Upgrade index mappings and index pattern for ES 5.x

lukas-vlcek:
- update-viaq-data-model.sh now accepts ES version as an argument
- Extracted settings into new common index mapping files


lukas-vlcek:
@ewolinetz I do not follow, to me it looks the script is looking into valid location `${ES_HOME}/index_templates/*.json`. See
https://github.com/openshift/origin-aggregated-logging/blob/9f97b2d836569eaa6f85fcf87d3084f6fae0d14d/elasticsearch/run.sh#L135

This location should contain all mappings including those added in this PR called `common.settings.*template.json` that now holds the pattern to replace.

lukas-vlcek:
We probably could change the line to:

```
   for template_file in ${ES_HOME}/index_templates/common.settings.*.json 
```
to narrows down the number of files to process but I would call it optimization not an issue.

ewolinetz:
@lukas-vlcek ack you're correct... for some reason i thought it was pointing to those specific files... yes, that would be an optimization and not a blocker.
/hold cancel

ewolinetz:
Tests are failing because kibana doesn't start up
https://github.com/openshift/origin-aggregated-logging/pull/962 should resolve that.

ewolinetz:
/retest

Should we update this test or even remove it?
```
name                            component               version
logging-es-data-master-e8hr6fve discovery-kubernetes    5.5.2
logging-es-data-master-e8hr6fve openshift-elasticsearch 5.5.2.1-SNAPSHOT
logging-es-data-master-e8hr6fve prometheus-exporter     5.5.2.0
[INFO] Installed plugin: discovery-kubernetes
[INFO] Installed plugin: openshift-elasticsearch
[INFO] Installed plugin: prometheus-exporter[@]
[FATAL] Elasticsearch pod is missing expected plugin(s). Exp cloud-kubernetes, openshift-elasticsearch, prometheus-exporter, found: discovery-kubernetes
[FATAL] openshift-elasticsearch
[FATAL] prometheus-exporter
```

ewolinetz:
looking at the es logs, it seems that things are using the templates correctly but we are seeing errors due to kubelet flake and acl

PR #968, opened 2018-02-21
Add rawtcp plugin to fluentd

antoineHC:
Hello, 

In order to send logs to external logstash service from ```user/output-extra-*.conf```, I added https://github.com/uken/fluent-plugin-out_rawtcp fluentd plugin. 
This plugin implements a simple TCP forward, without heartbeats or any other out-of-band checks. 

Regards,

### DIFF
--- a/test/out_rawtcp.sh
+++ b/test/out_rawtcp.sh
@@ -1,0 +1,117 @@
+#!/bin/bash
+
+# This is a test suite for the fluentd raw_tcp feature
+
+source "$(dirname "${BASH_SOURCE[0]}" )/../hack/lib/init.sh"
+source "${OS_O_A_L_DIR}/hack/testing/util.sh"
+os::util::environment::use_sudo
+
+FLUENTD_WAIT_TIME=${FLUENTD_WAIT_TIME:-$(( 2 * minute ))}
+
+os::test::junit::declare_suite_start "test/raw-tcp"
+
+update_current_fluentd() {
+    # undeploy fluentd
+    os::log::debug "$( oc label node --all logging-infra-fluentd- )"
+    os::cmd::try_until_text "oc get daemonset logging-fluentd -o jsonpath='{ .status.numberReady }'" "0" $FLUENTD_WAIT_TIME
+
+    # update configmap logging-fluentd
+    # edit so we don't send to ES
+    oc get configmap/logging-fluentd -o yaml | sed '/## matches/ a\
+    <match **>\
+      @type copy\
+      @include configs.d/user/raw-tcp.conf\
+    </match>' | oc replace -f -
+      oc patch configmap/logging-fluentd --type=json --patch '[{ "op": "add", "path": "/data/raw-tcp.conf", "#": "generated config file raw-tcp.conf" }]' 2>&1
+      oc patch configmap/logging-fluentd --type=json --patch '[{ "op": "replace", "path": "/data/raw-tcp.conf", "value": "\
+  <store>\n\
+   @type rawtcp\n\
+   flush_interval 1\n\
+    <server>\n\
+      name logstash\n\
+      host logstash.openshift-logging.svc.cluster.local\n\
+      port 9400\n\
+    </server>\n\
+  </store>\n"}]'
+
+    # redeploy fluentd
+    os::cmd::expect_success flush_fluentd_pos_files
+    os::log::debug "$( oc label node --all logging-infra-fluentd=true )"
+    os::cmd::try_until_text "oc get pods -l component=fluentd" "^logging-fluentd-.* Running "
+    fpod=$( get_running_pod logstash )
+    if [ -n "${fpod:-}" ] ; then
+      os::cmd::try_until_text "oc logs $fpod 2>&1" ".*kubernetes.*" $FLUENTD_WAIT_TIME
+    fi
+    fpod=$( get_running_pod fluentd ) || :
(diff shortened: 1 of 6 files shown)


richm:
Where is this source code from?  What is the license?  We also require a test to demonstrate this functionality.  There is no fluentd config for this so a user would have to know how to hack and configure the fluent.conf configmap to use this.  Finally, it looks as though this is using msgpack to encode the payload - does logstash understand msgpack?

antoineHC:
Hello, 

@richm I added the source and licence in the commit https://github.com/openshift/origin-aggregated-logging/pull/968/commits/b670d5c4d421a408e559ccadeaecf7ecf85d939e

In order to make it works you can use the default config map from ansible playbook and put the following code in `configs.d/user/output-extra-1.conf`

```
<store>
    type rawtcp 
    flush_interval 1
    <server>
      name logstash
      host my-logstash.external.com
      port 9400
    </server>
</store>
```
The msgpack must be used in logstash using the built in [fluent codec](https://www.elastic.co/guide/en/logstash/current/plugins-codecs-fluent.html) 
Logstash example configuration: 
```
input {
    tcp {
        port => 9400
        codec => fluent
    }
}

output {
    elasticsearch {
        hosts => ["localhost:9200"]
        index => "logstash-%{+YYYY.MM.dd}"
        document_type => "logs"
    }
}
```
@jcantrill About the secure communication to the destination it is the same as syslog protocol 

Regards,

richm:
Why would we consume the source code rather than the rubygem?  https://rubygems.org/gems/fluent-plugin-out_rawtcp

antoineHC:
The gem install doesn't work because in the gemspec, the latest version of fluentd is referenced and it depends on Ruby 2.1 which is not a Ruby version available in the RHEL Fluentd image. 
https://github.com/uken/fluent-plugin-out_rawtcp/blob/master/fluent-plugin-out_rawtcp.gemspec#L18 
As we rely on the Fluentd install in the container, the plug-in works and the only way to install it is to copy directly the sources in the image.

richm:
@jcantrill @portante This gives us the ability to forward logs to logstash.  I vote for including this.  Approve?

antoineHC:
Ping @jcaianirh @portante any news on the subject ?

jcantrill:
Is there a test we can add for this plugin?  My concern is we merge it, someone uses it, we have no test that validates it functionally works as advertised and then we are on the hook to support it.

antoineHC:
@openshift-bot /retest

test/out_rawtcp.sh
```diff
@@ -0,0 +1,115 @@
...
+  fi
+
+  $mycmd raw-tcp test finished at $( date )
+
+  # Clean up only if it's still around
+  os::log::debug "$( oc delete service/logstash 2>&1 || : )"
+  os::log::debug "$( oc delete deploymentconfig/logstash 2>&1 || : )"
+  
+  os::cmd::expect_success flush_fluentd_pos_files
+  os::log::debug "$( oc label node --all logging-infra-fluentd=true 2>&1 || : )"
```
richm: need to wait for fluentd to be running here
```
os::cmd::try_until_text "oc get pods -l component=fluentd" "^logging-fluentd-.* Running "
```

test/out_rawtcp.sh
```diff
@@ -0,0 +1,115 @@
...
+os::log::info Starting raw-tcp test at $( date )
+
+# make sure fluentd is working normally
+os::cmd::try_until_text "oc get pods -l component=fluentd" "^logging-fluentd-.* Running "
+fpod=$( get_running_pod fluentd )
+os::cmd::expect_success wait_for_fluentd_to_catch_up
+
+create_forwarding_logstash
+update_current_fluentd
+cleanup
```
richm: remove this call to `cleanup` - the `trap EXIT` will call this upon exit from the script, no matter how that exit happens

test/out_rawtcp.sh
antoineHC: Added the command, thanks

antoineHC:
@richm Do I need to get the tests green ? They seem to run an incomplete test set when I launch them manually :(

richm:
@antoineHC looks like the CI system is currently undergoing maintenance - should be fixed soon

stevekuznetsov:
v

richm:
you keep settin'm up, I'll keep settin'm down . . .
/retest

antoineHC:
One last time 
/retest

richm:
Whatever the failure is here, it doesn't appear to be a flake - it is persistent, always in the same place in the utf8 test.  I have no idea why.

richm:
@antoineHC looks good - please rebase to the latest master and squash your commits into a single commit

richm:
lots of fluentd failures
/retest

antoineHC:
Oh yeah all green 🎉 🍾

richm:
@jcantrill @portante ok?  lgtm

richm:
@portante do you still require changes?

antoineHC:
@richm Is there any change in the namespace name ? Before having the lgtm label, the tests were green with `namespace: logging` in my logstash deployment. 

Since then, I get an error message 
> namespaces "logging" not found

richm:
yes - the default logging namespace is now `openshift-logging`

richm:
strange curator flakes . . .
/retest

PR #973, opened 2018-02-23
Upgrade common data model to 0.0.13 for release 3.7

### LINKED ISSUE #972: Back-port `.all` alias into releases 3.6 up to 3.9, and master
There is a PR https://github.com/openshift/origin-aggregated-logging/pull/961 that brings new index templates into `es5.x` branch that includes index alias `.all`. We want to back-port `.all` alias into index templates for `release-3.6` up to `release-3.9`. However, we can not cherry-pick the commit from the PR directly because index templates were upgraded to ES 5.x format and older releases are using ES 2.x.

We also want to bring this into `master` branch (which is as of writing still on ES 2.x).

Instead we need to create new PRs to upgrade Common Data Model in previous releases. Technically we need to bring in updated version of [`hack/update-viaq-data-model.sh`](https://github.com/openshift/origin-aggregated-logging/pull/961/files#diff-5972ceb7a972e33b22cd148305f8b172) script that can correctly pull updated index templates from the latest release of common data model and use this script to download and update relevant json and docs files (*). This way we will get the `.all` alias for ES 2.x index templates.

To upgrade Common Data Model to 0.0.13 we have created tickets for individual releases:

- [x] release 3.6 (#1001)
- [x] release 3.7 (#973)
- no need to back-port to release 3.8 (#1002)
- [x] release 3.9 (#1003)
- [x] master (#1004)

(*) _Note that the docs files for ES 2.x are impacted by https://github.com/ViaQ/elasticsearch-templates/issues/89. Technically this means that it will use ES 5.x field types instead of ES 2.x types._

lukas-vlcek:
See #972


lukas-vlcek:
I think better title (and commit message) would be upgrade common data model to 0.0.13 release. I am going to update it.

PR #974, opened 2018-02-23
Generate uuid on ingestion to reduce duplicates

jcantrill:
This PR generates record uuids at the source to resolve:

https://bugzilla.redhat.com/show_bug.cgi?id=1548104

where when ES is underload, it fails bulk indexing and retries can produce duplicate entries

### DIFF
--- a/fluentd/configs.d/user/fluent.conf
+++ b/fluentd/configs.d/user/fluent.conf
@@ -33,4 +33,3 @@
   # no post - applications.conf matches everything left
 ##
 </label>
-
--- a/test/fluentd-forward.sh
+++ b/test/fluentd-forward.sh
@@ -55,6 +55,10 @@ update_current_fluentd() {
       else
     # edit so we don't send to ES
     oc get configmap/logging-fluentd -o yaml | sed '/## matches/ a\
+      <filter **>\
+        @type record_transformer\
+        remove_keys _id, viaq_msg_id\
+      </filter>\
       <match **>\
         @type copy\
         @include configs.d/user/secure-forward1.conf\
--- a/fluentd/configs.d/openshift/filter-post-genid.conf
+++ b/fluentd/configs.d/openshift/filter-post-genid.conf
@@ -1,0 +1,6 @@
+# generate ids at source mitigate creating duplicate records
+# requires fluent-plugin-elasticsearch
+<filter **>
+  @type elasticsearch_genid
+  hash_id_key viaq_msg_id
+</filter>
--- a/fluentd/configs.d/openshift/output-es-config.conf
+++ b/fluentd/configs.d/openshift/output-es-config.conf
@@ -4,6 +4,7 @@
       port "#{ENV['ES_PORT']}"
       scheme https
       target_index_key viaq_index_name
+      id_key viaq_msg_id
       remove_keys viaq_index_name
       user fluentd
       password changeme
--- a/fluentd/configs.d/openshift/output-es-ops-config.conf
+++ b/fluentd/configs.d/openshift/output-es-ops-config.conf
@@ -4,6 +4,7 @@
       port "#{ENV['OPS_PORT']}"
       scheme https
       target_index_key viaq_index_name
+      id_key viaq_msg_id
       remove_keys viaq_index_name
       user fluentd
       password changeme


richm:
please test with https://github.com/openshift/origin-aggregated-logging/pull/933
or, let's merge 933, and change it to look for expected count == actual count instead of expected count <= actual count

fluentd/configs.d/user/fluent.conf
```diff
@@ -24,6 +24,13 @@
   @include configs.d/openshift/filter-syslog-record-transform.conf
   @include configs.d/openshift/filter-viaq-data-model.conf
   @include configs.d/openshift/filter-post-*.conf
+
+  # generate uuids at source mitigate creating duplicate records
+  # requires fluent-plugin-elasticsearch
+  <filter **>
+    @type elasticsearch_genid
+    hash_id_key _openshift_es_genid
+  </filter>
```
mffiedler: Follow up to IRC discussion - this should go above filter-post-*.conf.   My successful (no duplicates) config was:

    <label @INGRESS>
    ## filters
      @include configs.d/openshift/filter-pre-*.conf
      @include configs.d/openshift/filter-retag-journal.conf
      @include configs.d/openshift/filter-k8s-meta.conf
      @include configs.d/openshift/filter-kibana-transform.conf
      @include configs.d/openshift/filter-k8s-flatten-hash.conf
      @include configs.d/openshift/filter-k8s-record-transform.conf
      @include configs.d/openshift/filter-syslog-record-transform.conf
      @include configs.d/openshift/filter-viaq-data-model.conf
      <filter **>
        @type elasticsearch_genid
        hash_id_key _openshift_es_genid
      </filter>
      @include configs.d/openshift/filter-post-*.conf
    </label>

jcantrill: fixed

fluentd/configs.d/user/fluent.conf
```diff
@@ -23,6 +23,12 @@
   @include configs.d/openshift/filter-k8s-record-transform.conf
   @include configs.d/openshift/filter-syslog-record-transform.conf
   @include configs.d/openshift/filter-viaq-data-model.conf
+  # generate uuids at source mitigate creating duplicate records
+  # requires fluent-plugin-elasticsearch
+  <filter **>
+    @type elasticsearch_genid
+    hash_id_key _openshift_es_genid
```
ewolinetz: is there a reason we are using this particular key only to map and remove it later?

jcantrill: This is what is recommended by the filter.  The filter generates the key, copies the value to the '_id' field, and then comments that ES does not like records with '_' so it is removed prior to submission

richm: I don't think this file is used.
The "real" fluent.conf comes from ansible.
If we want this in the product by default, we'll need to create configs.d/openshift/filter-post-esid.conf or something like that with this config.  And - if we want to easily disable it, we'll need to make the inclusion of this file conditional based on an env var.

ewolinetz: ack, i read the documentation for the filter type. I didn't realize that `id_key` in the output plugin denotes what should be used as that key... I read it as it was setting another key to be that value.

richm:
```
Running test/fluentd-forward.sh:258: executing 'wait_for_fluentd_to_catch_up '' '' 2' expecting success...
FAILURE after 1245.828s: test/fluentd-forward.sh:258: executing 'wait_for_fluentd_to_catch_up '' '' 2' expecting success: the command returned the wrong error code
1
... repeated 853 times
```
/test json-file

richm:
I think the fluentd-forward test may be getting confused - it may be doing the esid in both fluentd - @jcantrill take a look and see if you need to disable the esid plugin in the client fluentd or the server fluentd

richm:
I see what the problem is - the test is failing here:
https://github.com/openshift/origin-aggregated-logging/blob/master/test/fluentd-forward.sh#L258

The test with `FORWARDCNT=2` is testing that you can have a client fluentd sending to _2_ server fluentd and each one of them will write the same record, but with a different id, to Elasticsearch.  However, with esid, the id is generated in the client fluentd, which is passed through to the 2 servers, one of them gets through, the other is rejected with a duplicate id.
```
FAILURE after 1243.062s: test/fluentd-forward.sh:258: executing 'wait_for_fluentd_to_catch_up '' '' 2' expecting success: the command returned the wrong error code
1
```
That is, the test is expecting _2_ matching records, but there is only 1.

I think the best fix is to disable the esid plugin on the client fluentd.  If that's not possible, then the next best is to create some sort filter to strip out the esid field from the record after generating it but before sending it to the `match`s.

test/fluentd-forward.sh
```diff
@@ -75,6 +75,10 @@ update_current_fluentd() {
     port 24284\n\
    </server>\n\
   </store>\n\
+  <filter **>\n\
+    @type record_modifier\n\
+    remove_keys _id\n\
```
richm: should this be _openshift_es_genid ?

jcantrill: No I dont believe so.  The gen plugin creates the id and stuffs it in a key that you define.  It then uses that to populate the _id field which is what ES uses.  We then tell it to remove the original key because apparently ES doesnt like that.  In our case are not getting a duplicate record so we need to remove the '_id' field

richm: Also, use a `record_transformer` filter instead of `record_modifier`

fluentd/configs.d/openshift/filter-post-genuuid.conf
```diff
@@ -0,0 +1,6 @@
+# generate uuids at source mitigate creating duplicate records
+# requires fluent-plugin-elasticsearch
+<filter **>
+  @type elasticsearch_genid
+  hash_id_key _openshift_es_genid
```
richm: this should be `openshift_es_genid` and it should be part of the official schema, part of the published api
This field will be useful for users to be able to correlate their own data with the data in Elasticsearch, and will be very useful in alerting applications (to alert on ES data) and for message queue applications.

test/fluentd-forward.sh
```diff
@@ -61,6 +61,10 @@ update_current_fluentd() {
       </match>' | oc replace -f -
         oc patch configmap/logging-fluentd --type=json --patch '[{ "op": "add", "path": "/data/secure-forward1.conf", "#": "generated config file secure-forward1.conf" }]' 2>&1
         oc patch configmap/logging-fluentd --type=json --patch '[{ "op": "replace", "path": "/data/secure-forward1.conf", "value": "\
+  <filter **>\n\
```
richm: Sorry, I missed this earlier - this won't work - you cannot have a `<filter>` inside of a `<match>` block - this oc patch is putting the patch inside a `<match>` block - you might need a separate oc patch in order to edit fluent.conf to add the filter

PR #977, opened 2018-02-26
Bug 1544815 - logging-fluentd SEGV_ACCERR crash on init when libjemalloc is used in scalability CI lab cluster

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1544815
Use

    export LD_BIND_NOW=1

to workaround the problem when using jemalloc.


richm:
/test json-file
curator flake?

PR #993, opened 2018-03-05
Docker audit test flake

josefkarasek:
The test that checks if es contains at least 5 docker events succeeds.
It's the test that calculates the delta before and after the run of a container which fails.
Check if there are any events before the run and print them to stdout, if found.

### DIFF
--- a/test/docker_audit.sh
+++ b/test/docker_audit.sh
@@ -12,11 +12,36 @@ function get_logs_count() {
     curl_es $es_pod "${index}"_count?q=docker.user:* | get_count_from_json
 }
 
+function get_logs_count_timerange() {
+    local es_pod=$1
+    local index="$2"
+    local timestamp="$3"
+    curl_es $es_pod "${index}"_count -d '{
+        "query": {
+            "bool": {
+                "must": {
+                    "exists": {
+                        "field": "docker.user"
+                    }
+                },
+                "filter": {
+                    "range" : {
+                        "@timestamp" : {
+                            "gt" : '\""$timestamp"\"'
+                        }
+                    }
+                }
+            }
+        }
+    }' | get_count_from_json
+}
+
 function logs_count_is_ge() {
     local es_pod=$1
     local index="$2"
     local expected=$3
-    local actual=$( get_logs_count $es_pod "$index" )
+    local timestamp="$4"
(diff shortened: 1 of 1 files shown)


test/docker_audit.sh
```diff
@@ -20,6 +20,12 @@ function logs_count_is_ge() {
     test $actual -ge $expected
 }
 
+function print_logs() {
+    local es_pod=$1
+    local index=$2
+    os::log::info "curl_es $es_pod /$index/q=docker.user:*"
```
richm: I suggest using artifact_out instead of `os::log::info`
```
curl_es $es_pod /$index/q=docker.user:* | jq . | artifact_out
```

nhosoi:
@josefkarasek 
> Check if there are any events before the run and print them to stdout, if found.

I also suspected it and modified the test like this in https://github.com/openshift/origin-aggregated-logging/pull/955, but it still fails due to the same issue - delta == 4. (;_;)
https://github.com/nhosoi/origin-aggregated-logging/blob/bz1515715/test/docker_audit.sh#L38-L44

Could there be any way to check if audit.log itself successfully logged 5 events or not?

test/docker_audit.sh
```diff
@@ -71,11 +71,11 @@ ops_logs_before=$( get_logs_count $esopspod '/.operations.*/' )
 os::log::info "ops diff before:  $ops_logs_before"
 os::log::info "proj diff before: $logs_before"
 
-# ping,create,attach,start,delete generates 5 docker audit messages
+# ping,create,attach,start generates 4 docker audit messages
```
richm: What happened to `delete`?  did it just disappear?

richm:
yep - new bug/flake in the viaq test for journald - fix forthcoming
@jcantrill please go ahead and merge this PR as it will unblock CI, at least for json-file, then I will have a journald fix for the viaq test asap

test/docker_audit.sh
josefkarasek: See the below comparison, both cases run this command `docker run -it --rm centos:7 echo "foo"`
docker-1.12.6-71.git3e8e77d.el7.x86_64:
====
create, attach, start, resize, wait, remove

docker-1.13.1-53.git774336d.el7.x86_64:
====
_ping, create, attach, start, resize

PR #1000, opened 2018-03-07
disable multi-tenancy test until we figure out this oc exec -i hang problem

richm:


richm:
I think this should unblock us on master until we figure out why oc exec -i is hanging

richm:
hit the access control hang
/retest

PR #1006, opened 2018-03-07
Modify the default project namespace

jcantrill:
* Modify the default project namespace to openshift-logging

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1535300

### DIFF
--- a/test/multi_tenancy.sh
+++ b/test/multi_tenancy.sh
@@ -16,6 +16,8 @@ os::util::environment::use_sudo
 
 os::test::junit::declare_suite_start "test/multi_tenancy"
 
+LOGGING_PROJECT=${LOGGING_NS:-openshift-logging}
+
 espod=$( get_es_pod es )
 esopspod=$( get_es_pod es-ops )
 esopspod=${esopspod:-$espod}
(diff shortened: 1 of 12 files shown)


jcantrill:
/test json-file
/test/logging

test/multi_tenancy.sh
```diff
@@ -154,11 +156,11 @@ done
 # use different usernames, otherwise you'll get this odd error:
 # # oc login --username=loguser --password=loguser
 # error: The server was unable to respond - verify you have provided the correct host and port and that the server is currently running.
-LOG_NORMAL_USER=${LOG_NORMAL_USER:-loguser}
-LOG_NORMAL_PW=${LOG_NORMAL_PW:-loguser}
+LOG_NORMAL_USER=${LOG_NORMAL_USER:-loguser-$RANDOM}
```
ewolinetz: where is `$RANDOM` coming from?

jcantrill: std bash function

richm:
I must admit I'm a bit confused as to how this is supposed to work.
There are many places in the tests where we are checking for an index like `/project.openshift-logging.*`.  But according to fluentd, _there should never be such an index_ - that is, because of https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/run.sh#L4 - _all records from namespace openshift-logging should go into .operations_.

So, how is it that the tests are working?

PR #1007, opened 2018-03-08
Shard operations indices

jcantrill:
This PR
* Adds support for sharding the operation indices
* Adds support for sharding the orphaned indices
* Add template for orphaned indices.  Requires a change in viaq to support template generation

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1553257

### DIFF
--- a/elasticsearch/index_templates/common.settings.operations.orphaned.json
+++ b/elasticsearch/index_templates/common.settings.operations.orphaned.json
@@ -1,0 +1,11 @@
+{
+  "order": 5,
+  "settings": {
+    "index.refresh_interval": "5s",
+    "index.number_of_replicas": %REPLICA_SHARDS%,
+    "index.number_of_shards": %PRIMARY_SHARDS%,
+    "index.translog.flush_threshold_size": "256mb",
+    "index.unassigned.node_left.delayed_timeout": "2m"
+  },
+  "template": ".orphaned*"
+}
--- a/elasticsearch/run.sh
+++ b/elasticsearch/run.sh
@@ -17,6 +17,9 @@ LOG_FILE=${LOG_FILE:-elasticsearch_connect_log.txt}
 RETRY_COUNT=${RETRY_COUNT:-300}		# how many times
 RETRY_INTERVAL=${RETRY_INTERVAL:-1}	# how often (in sec)
 
+PRIMARY_SHARDS=${PRIMARY_SHARDS:-3}
+REPLICA_SHARDS=${REPLICA_SHARDS:-0}
+
 retry=$RETRY_COUNT
 max_time=$(( RETRY_COUNT * RETRY_INTERVAL ))	# should be integer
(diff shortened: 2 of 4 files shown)


elasticsearch/index_templates/common.settings.operations.orphaned.json
```diff
@@ -0,0 +1,11 @@
+{
+  "order": 5,
+  "settings": {
+    "index.refresh_interval": "5s",
+    "index.number_of_replicas": $REPLICA_SHARDS,
```
richm: I think you should use something else as a delimiter here for the token - `$REPLICA_SHARDS` makes it look too much like a shell variable - suggest using `%REPLICA_SHARDS%` with the token delimiter at the front and back.

elasticsearch/run.sh
```diff
@@ -133,6 +136,8 @@ verify_or_add_index_templates() {
     shopt -s failglob
     for template_file in ${ES_HOME}/index_templates/*.json
     do
+        sed -i "s,\$REPLICA_SHARDS,$REPLICA_SHARDS," $template_file
```
richm: `"s,%REPLICA_SHARDS%,$REPLICA_SHARDS,"`

richm:
Is it ok to have `PRIMARY_SHARDS:-3` even if there is only 1 ES server?  It just means all 3 shards will live inside the 1 server?  If replicas are added, will ES automatically spread out the shards so that we can have the benefit of sharding for load balancing?

jcantrill:
@portante Can you respond to @richm comment: https://github.com/openshift/origin-aggregated-logging/pull/1007#issuecomment-371526057

I believe this is still functionally acceptable even in a single node cluster.  Sharding is simply dividing the data.  We leave it to ES functionality to spread it accordingly for load balancing which is its nature.  Obviously in a single node cluster when replicas are set to a value greater than zero we dont get any benifit.

portante:
> Is it ok to have PRIMARY_SHARDS:-3 even if there is only 1 ES server?

Yes.

> It just means all 3 shards will live inside the 1 server?

Yes.

> If replicas are added, will ES automatically spread out the shards so that we can have the benefit of sharding for load balancing?

Yes.

> I believe this is still functionally acceptable even in a single node cluster. Sharding is simply dividing the data. We leave it to ES functionality to spread it accordingly for load balancing which is its nature.

Correct.

Lucene already has multiple shards internally when detailing with writes, where it periodically merges them together (lucene indices are immutable, so to "write" to an index, it writes a new index with just the new data, and then later create another new index with the merge of the previous two indices).

Explicit shards are just a way of telling Elasticsearch the unit of parallelism you want for this index.

> Obviously in a single node cluster when replicas are set to a value greater than zero we don't get any benefit.

Right, no parallelism on a single node cluster the way we typically deploy one (one ES DC to one disk).

PR #1008, opened 2018-03-08
remove CONTAINER_TAG field

richm:
tell the viaq test to ignore the CONTAINER_TAG field


richm:
curator rollout problem
/test logging

PR #1016, opened 2018-03-13
Removing an unnecessary line from the rebase mistake in pr/#926.

nhosoi:
Bug 1534588 - Size of the emitted data exceeds buffer_chunk_limit


richm:
this looks like the oc exec -i hang flake
/test json-file

PR #1021, opened 2018-03-16
Take both forward and secure_forward into account for NUM_OUTPUTS

fcami:
We currently compute NUM_OUTPUTS using the number of secure_outputs, but core fluentd can use the forward plugin instead, so we can miss a number of outputs.
This PR changes the grep logic to take both secure_forward and forward entries.

### DIFF
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -214,12 +214,12 @@ if [ $TOTAL_LIMIT -le 0 ]; then
     exit 1
 fi
 
-# If secure-forward outputs are configured, add them to NUM_OUTPUTS.
-sec_forward_files=$( grep -l "@type *secure_forward" ${CFG_DIR}/*/* 2> /dev/null || : )
-for afile in ${sec_forward_files} ; do
+# If forward and secure-forward outputs are configured, add them to NUM_OUTPUTS.
+forward_files=$( grep -l "@type .*forward" ${CFG_DIR}/*/* 2> /dev/null || : )
+for afile in ${forward_files} ; do
     file=$( basename $afile )
     if [ "$file" != "${mux_client_filename:-}" ]; then
-        grep "@type *secure_forward" $afile | while read -r line; do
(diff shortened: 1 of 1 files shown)


fluentd/run.sh
```diff
@@ -214,12 +214,12 @@ if [ $TOTAL_LIMIT -le 0 ]; then
     exit 1
 fi
 
-# If secure-forward outputs are configured, add them to NUM_OUTPUTS.
-sec_forward_files=$( grep -l "@type *secure_forward" ${CFG_DIR}/*/* 2> /dev/null || : )
-for afile in ${sec_forward_files} ; do
+# If forward and secure-forward outputs are configured, add them to NUM_OUTPUTS.
+forward_files=$( grep -l "@type *forward" ${CFG_DIR}/*/* 2> /dev/null || : )
```
richm: The original was looking for the string "@type" followed by 0 or more spaces followed by the string "secure_forward".
I think the intention now is to look for the string "@type" followed by a space, followed by zero or more of any character, followed by "forward" - so
```
forward_files=$( grep -l "@type .*forward" ${CFG_DIR}/*/* 2> /dev/null || : )
```

fluentd/run.sh
```diff
@@ -214,12 +214,12 @@ if [ $TOTAL_LIMIT -le 0 ]; then
...
+for afile in ${forward_files} ; do
     file=$( basename $afile )
     if [ "$file" != "${mux_client_filename:-}" ]; then
-        grep "@type *secure_forward" $afile | while read -r line; do
+        grep "@type *forward" $afile | while read -r line; do
```
richm: same here

richm:
/test json-file
curator flake

richm:
this looks like the "oc exec -i hangs waiting for input" bug
/test json-file

jcantrill:
@richm Do you know if this made it into 1.5?

richm:
> @richm Do you know if this made it into 1.5?

It did not

PR #1022, opened 2018-03-17
ES5 disallows empty field names

richm:
If you attempt to submit an doc with an empty field name, you'll get
an error like this:
```
Object field starting or ending with a [.] makes object resolution ambiguous: []
```
which is very unhelpful.
See https://discuss.elastic.co/t/object-field-starting-or-ending-with-a-makes-object-resolution-ambiguous/123351


ewolinetz:
LGTM, thanks @richm

PR #1023, opened 2018-03-19
gather more information when docker_audit test fails

richm:


richm:
sorry - left in some debugging code - had docker_audit test timeout set to 1 second to force failure case - reset to 60 seconds for real testing

Please lgtm again - thanks

PR #1035, opened 2018-03-22
Updating the es output fluentd plugin request_timeout to be nil so th…

ewolinetz:
…at we dont set a timeout on the client


richm:
`nil` doesn't work:
```
2018-03-23 02:50:05 +0000 [warn]: temporarily failed to flush the buffer. next_retry=2018-03-23 02:50:03 +0000 error_class="Fluent::ElasticsearchOutput::ConnectionFailure" error="Can not reach Elasticsearch cluster ({:host=>\"logging-es-ops\", :port=>9200, :scheme=>\"https\", :user=>\"fluentd\", :password=>\"obfuscated\"})! read timeout reached" plugin_id="object:3f9c7f5d900c"
```

richm:
same with `0`

I think we're just going to have to pick a very large number . . . or change fluent-plugin-elasticsearch

ewolinetz:
@portante since we can't make it actually infinite (as it stands)... do you have a large enough timeout value we should use instead?

richm:
@ewolinetz I suggest $MAX_SIGNED_INT e.g. 2^31

fluentd/configs.d/openshift/output-es-config.conf
```diff
@@ -27,5 +27,5 @@
       buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT'] || '32' }"
       buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT'] || '8m' }"
       buffer_queue_full_action "#{ENV['BUFFER_QUEUE_FULL_ACTION'] || 'block'}"
-      request_timeout 600
+      request_timeout (2 ** 31)
```
richm: 2147483648

fluentd/configs.d/openshift/output-es-ops-config.conf
```diff
@@ -27,5 +27,5 @@
       buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT'] || '32' }"
       buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT'] || '8m' }"
       buffer_queue_full_action "#{ENV['BUFFER_QUEUE_FULL_ACTION'] || 'block'}"
-      request_timeout 600
+      request_timeout (2 ** 31)
```
richm: 2147483647

ewolinetz: you mean 2147483648?

ewolinetz:
@richm i don't think we need to pick to 3.8 since we didn't provide an image for that release...

ewolinetz:
/cherrypick release-3.8
just in case?

willemveerman:
Sorry to dig up such an old issue.

Can you explain why exactly you need to set the timeout to such a large figure? This will mean that fluentd's requests essentially never time out.

The bug report suggests that sometimes the timeout needs to be > 10 minutes, but it doesn't explain why.

We have seen some issues around the fluentd ES plugin's retry behaviour, and we're wondering whether we should also configure it to never time out.

PR #1062, opened 2018-04-04
support logging-xxx5 images

richm:


richm:
also requires a change to https://github.com/openshift/aos-cd-jobs/blob/master/sjb/config/test_cases/push_origin_aggregated_logging_release.yml#L24 in order to actually do anything

richm:
curator flake
/test json-file

PR #1064, opened 2018-04-05
Curator continually restarts

josefkarasek:
Perform a readiness check by contacting elasticsearch's root endpoint.
If elasticsearch is not available for more than 1 hour, skip the run
and resume next day.

An error message is written to stderr.
No exception / explosion / hard & fast failure.
Just an error message.

Resolves https://bugzilla.redhat.com/show_bug.cgi?id=1557483


curator/src/run_cron.py
```diff
@@ -18,23 +20,62 @@ def __init__(self):
...
+            + os.getenv('ES_HOST', 'logging-es') + ':' \
+            + os.getenv('ES_PORT', '9200')
 
     def setup_cron(self):
         for cmd in self.cmd_list:
             job = self.job_list.new(command=cmd, comment='Generated job based on settings')
             job.every().day()
 
+    def server_ready(self):
+        retry_max = 11  # (2**(11+1)-1) / 60 = 68 minutes of exponential back-off
```
richm: I think this should be unlimited - curator should keep trying to connect indefinitely

curator/src/run_cron.py
```diff
@@ -18,23 +20,62 @@ def __init__(self):
...
+                if code == 200:
+                    return True
+                # else:
+                    # This is probably not a transient error
+                    # The server is reacheable but in a failed state
+            except (urllib2.URLError, urllib2.HTTPError) as error:
+                self.logger.error('Can\'t connect to elasticsearch at [%s]. %s failed retries', self.es_url, retry_count)
+            if retry_count < retry_max:
+                time.sleep(until_next_retry)
+                until_next_retry *= 2
```
richm: There should be a max `until_next_retry` - 300 (5 minutes) is customary

curator/src/util.py
```diff
@@ -3,16 +3,21 @@
...
+    return logging.getLogger(name)
+
+def create_auth_opener(key, cert, ca):
+    handlers = []
+    handlers.append( https.HTTPSClientAuthHandler(
+        key = key,
+        cert = cert,
+        ca_certs = ca,
+        ssl_version = ssl.PROTOCOL_SSLv23,
+        ciphers = 'TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256' ) )
```
richm: is it necessary to provide `ssl_version` and `ciphers`?  If so - will have to figure out how to get the "best" default values from the library - I don't think we should hardcode these here - @tiran ?

tiran: I advise against hard-coding cipher strings. This code will break as soon as TLS 1.3 is available. ``PROTOCOL_SSLv23`` will negotiate the highest available TLS version, which may be TLS 1.3. But ``TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256`` is not a valid TLS 1.3 cipher suite.

In fact ``TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256`` is not a valid OpenSSL cipher suite at all.

curator/src/https.py
```diff
@@ -0,0 +1,59 @@
...
+        httplib.HTTPSConnection.__init__(self,host,**kwargs)
+
+    def connect(self):
+        sock = socket.create_connection( (self.host, self.port), self.timeout )
+
+        if self._tunnel_host:
+            self.sock = sock
+            self._tunnel()
+
+        self.sock = ssl.wrap_socket( sock,
```
tiran: Don't use ``ssl.wrap_socket``. It doesn't validate host names and is therefore vulnerable to MitM attacks. You want to use a SSLContext object from ``ssl.create_default_context`` and then use its wrap_socket() method with ``server_hostname``.

curator/src/util.py
richm: thanks @tiran - so `ssl_version = ssl.PROTOCOL_SSLv23` is ok - can we just omit `ciphers`?

tiran: @richm The code has security bugs. I left some comments.

richm:
still using httplib instead of requests
I suppose if this is still not suitable, you could just use the python elasticsearch library, which is already on the image, which is used by curator: https://github.com/elastic/curator/blob/3.5/curator/api/utils.py

https://github.com/elastic/elasticsearch-py/tree/2.3.0

josefkarasek:
@richm yeah those libraries look perfect, thanks
```
import curator, elasticsearch
>>> es = Elasticsearch(
...     [es_url],
...     # turn on SSL
...     use_ssl=True,
...     # make sure we verify SSL certificates (off by default)
...     verify_certs=True,
...     # provide a path to CA certs on disk
...     ca_certs=ca,
...     # PEM formatted SSL client certificate
...     client_cert=cert,
...     # PEM formatted SSL client key
...     client_key=key
... )
curator.get_version(es)
```

ewolinetz:
Is there really a reason that we should wait for ES to be ready?
Could we just check the resulting output of the curator run and if we weren't able to connect to ES maybe try again after a set period of time?

richm:
Please squash into a single commit
@tiran - looks ok now?

PR #1067, opened 2018-04-05
check logging namespace for actual resources

jcantrill:
Check for the logging namespace with resources in case there is just an empty namespace

### DIFF
--- a/hack/testing/entrypoint.sh
+++ b/hack/testing/entrypoint.sh
@@ -26,7 +26,7 @@ source "$(dirname "${BASH_SOURCE[0]}" )/../lib/init.sh"
 source "${OS_O_A_L_DIR}/hack/testing/util.sh"
 
 LOGGING_NS=openshift-logging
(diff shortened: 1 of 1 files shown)


hack/testing/entrypoint.sh
```diff
@@ -26,7 +26,7 @@ source "$(dirname "${BASH_SOURCE[0]}" )/../lib/init.sh"
 source "${OS_O_A_L_DIR}/hack/testing/util.sh"
 
 LOGGING_NS=openshift-logging
-if oc get project logging -o name > /dev/null ; then
+if oc get project logging -o name > /dev/null && [ $(oc get dc -n logging -o name | wc -l) -gt 0 ]  ; then
```
ewolinetz: isn't the `oc get` command missing an object type?

richm:
@jcantrill I say we force merge this to unblock CI - and force merge on es5.x branch as well

PR #1070, opened 2018-04-06
bug 1562004. Provide utility to sanitize chunks

jcantrill:
This PR:
* Provides a utility to manually sanitize fluentd buffers for when fluentd is stuck on buffers that are corrupt and it can no longer process.

https://bugzilla.redhat.com/show_bug.cgi?id=1562004

cc @portante

### DIFF
--- a/fluentd/Dockerfile.centos7
+++ b/fluentd/Dockerfile.centos7
@@ -58,6 +58,7 @@ RUN mkdir -p /etc/fluent/configs.d/{dynamic,user} && \
     chmod 777 /etc/fluent/configs.d/dynamic && \
     ln -s /etc/fluent/configs.d/user/fluent.conf /etc/fluent/fluent.conf
 
+COPY utils/** ${HOME}/utils
 ADD jemalloc/ ${HOME}/jemalloc/
 # note - make dist "fails" here
 # false -o doc/jemalloc.html doc/html.xsl doc/jemalloc.xml
--- a/fluentd/utils/sanitize_msg_chunks
+++ b/fluentd/utils/sanitize_msg_chunks
@@ -1,0 +1,185 @@
+#!/bin/env ruby
+
+# Sanitize the buffer message chunks by unpacking.Verify the file buffers to confirm they
+# are deserializable and include the necessary
+# keys/values to push to Elasticsearch
+#
+require 'logger'
+require 'msgpack'
+require 'optparse'
+require 'ostruct'
+require 'fileutils'
+
+DEFAULT_BUFFER_DIR='/var/lib/fluentd'
+
+options = OpenStruct.new(
+  dir: '/var/lib/fluentd',
+  log_level: Logger::INFO,
+  dlq_dir: '/var/lib/fluentd/dlq',
+  index_field: 'viaq_index_name',
+  require: nil
+)
+
+$parser = OptionParser.new do |opts|
+  opts.banner = %Q(
+    Usage: #{File.basename(__FILE__)} [options]
+
+    Sanitize the buffer chunks to remove messages that
+    would otherwise block fluentd because the chunk is
+    corrupt and can not be properly unpacked.
+
+    Note: This operation is destructive and will replace
+    the existing buffer chunks.
+  )
+
+
+  opts.on("-d", "--directory DIRECTORY", "The directory where to find buffers to sanitize", "Default: #{options.dir}") do |dir|
+    options.dir = dir
+  end
+
+  opts.on("-i", "--index-field INDEX_FIELD", "The name of the key in the record that contains the index name",
+          "Default: #{options.index_field}") do | i |
+    options.index_field = i
+  end
+
+
+  opts.on("-q", "--dead-letter-queue DLQ", "The directory where to write unprocessable messages.",
+         "'DROP' - drop messages from buffer chunk", "Default: #{options.dlq_dir}") do |dir|
+    options.dlq_dir = dir
+  end
+
+  opts.on("-r", "--require EXTRA_HANDLER", "The full path to additional handler to further evaluate",
+          "each record. The file should contain a single function with the following name and args:",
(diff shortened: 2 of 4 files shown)


fluentd/utils/sanitize_msg_chunks
```diff
@@ -0,0 +1,178 @@
...
+  end
+  if options.index_field.strip.empty?
+      $logger.error("The index_field can not be empty.")
+      return false
+  end
+  if options.dlq_dir != 'DROP'
+     unless options.dlq_dir.start_with?('/')
+        $logger.error("The dead letter queue must be either 'DROP' or an absolute path: #{options.dlq_dir}")
+        return false
+     end
```
richm: check for `Dir.exists?(options.dlq_dir)` here too?

fluentd/utils/sanitize_msg_chunks
```diff
@@ -0,0 +1,178 @@
...
+  clean_file = file + '.clean'
+  p = MessagePack::Packer.new(File.open(clean_file, 'w'))
+  u = MessagePack::Unpacker.new(File.new(file, 'r'))
+  u.each do |obj|
+    begin
+      time = obj[0]
+      record = obj[1]
+      target_index = record[opts.index_field]
+      target_index = target_index.downcase
+      check_record($logger, time, record) if defined?(:check_record)
```
richm: I don't think this works - `if defined?(:check_record)` will always return true

ewolinetz:
I may be missing something regarding this utility... is this causing Fluentd to dump from its current heap or are we reading/cleaning up its DLQ?

richm:
This is a quick&dirty utility that admins can use to clean up corrupted file buffers. It will remove corrupted records from the file buffers and dump them in a dlq, and rewrite the buffers to have only "clean" records.
Usage:
- shutdown fluentd
- run this utility - move .clean files to their original names
- startup fluentd

fluentd should now be able to flush its buffers to elasticsearch without getting stuck on the corrupted records

Yes, the next step is to change the fluent-plugin-elasticsearch to do this on the fly

fluentd/utils/sanitize_msg_chunks
jcantrill: corrected.  The right way is `check_record($logger, time, record) if defined?(check_record)`

fluentd/utils/sanitize_msg_chunks
jcantrill: We create it if it doesnt exist

ewolinetz:
>Usage:
>
>    shutdown fluentd
>    run this utility - move .clean files to their original names
>    startup fluentd

ack ack
@jcantrill can you also add to the repo's README how to use this please as part of this PR?

richm:
problem with fluentd - run.sh says fluentd not found?
/retest

richm:
failed to deploy es?
```
openshift-logging                   30m         50m          21        logging-es-data-master-1kbcn9px-1.15265171d131f4fc                ReplicationController                                            Warning   FailedCreate                  replication-controller                         Error creating: No API token found for service account "aggregated-logging-elasticsearch", retry after the token is automatically created and added to the service account
```
/retest

richm:
```
        "stderr": "The DeploymentConfig \"logging-es-data-master-rd09ovtn\" is invalid: \n* spec.template.spec.affinity.podAntiAffinity.preferredDuringSchedulingIgnoredDuringExecution[0].podAffinityTerm.topologyKey: Required value: can not be empty\n* spec.template.spec.affinity.podAntiAffinity.preferredDuringSchedulingIgnoredDuringExecution[0].podAffinityTerm.topologyKey: Invalid value: \"\": name part must be non-empty\n* spec.template.spec.affinity.podAntiAffinity.preferredDuringSchedulingIgnoredDuringExecution[0].podAffinityTerm.topologyKey: Invalid value: \"\": name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')\n", 
```
@jcantrill @ewolinetz does this look familiar?

jcantrill:
@richm @ewolinetz That would be because of this https://github.com/openshift/openshift-ansible/pull/7864 I think.  Logging is no longer gating for ansible because of 'flakes'.  They are run but not readily visible without digging.  We need to revisit this with @smarterclayton and @sdodson because the whole point of CI blocking or atleast making it more obvious.

richm:
more flakes?
/retest

richm:
still getting very strange flakes 
/retest

richm:
the failure here is what https://github.com/openshift/origin-aggregated-logging/pull/1090 solves
/test logging
see if we can get it to pass anyway

fluentd/Dockerfile.centos7
```diff
@@ -58,6 +58,7 @@ RUN mkdir -p /etc/fluent/configs.d/{dynamic,user} && \
     chmod 777 /etc/fluent/configs.d/dynamic && \
     ln -s /etc/fluent/configs.d/user/fluent.conf /etc/fluent/fluent.conf
 
+COPY utils/** ${HOME}/bin
```
richm: I think this is wiping out fluentd - it seems to overwrite or remove ${HOME}/bin/fluentd
```
run.sh: line 348: exec: fluentd: not found
```

jcantrill:
@richm maybe we no longer need this pr if we have proper handling of dlq in the elastic plugin

richm:
> @richm maybe we no longer need this pr if we have proper handling of dlq in the elastic plugin

how far back are we planning to backport that plugin?  We have customers with this problem in 3.6.

jcantrill:
Closing as changes to the Elasticsearch fluentd plugin will handle those issues now.

PR #1073, opened 2018-04-10
Sort events by timestamp

fbac:
This minimizes the amount of data to sort and helps debugging.


fbac:
@jcantrill can you check and merge this PR?

richm:
I don't understand why there is no option for `oc get events` to sort by timestamp (and provide high resolution timestamps, for that matter . . .)

PR #1074, opened 2018-04-10
Adding tests to verify rolled journald files are released

ewolinetz:
While we wait for a fix for journald file closing on rollover to be released

### DIFF
--- a/test/access_control.sh
+++ b/test/access_control.sh
@@ -25,7 +25,7 @@ function cleanup() {
         done
     fi
     if [ -n "${espod:-}" ] ; then
-        curl_es $espod /project.access-control-* -XDELETE > /dev/null
+        curl_es_pod $espod /project.access-control-* -XDELETE > /dev/null
     fi
     for proj in access-control-1 access-control-2 access-control-3 ; do
         oc delete project $proj 2>&1 | artifact_out
@@ -66,7 +66,7 @@ function add_message_to_index() {
     local project_uuid=$( oc get project $1 -o jsonpath='{ .metadata.uid }' )
     local index="project.$1.$project_uuid.$(date -u +'%Y.%m.%d')"
     local espod=$3
-    curl_es "$espod" "/$index/access-control-test/" -XPOST -d '{"message":"'${2:-"access-control message"}'"}' | python -mjson.tool 2>&1 | artifact_out
+    curl_es_pod "$espod" "/$index/access-control-test/" -XPOST -d '{"message":"'${2:-"access-control message"}'"}' | python -mjson.tool 2>&1 | artifact_out
 }
 
 # test the following
@@ -108,11 +108,11 @@ function test_user_has_proper_access() {
             continue
         fi
         os::log::info See if user $user $negverb read /project.$proj.*
-        nrecs=$( curl_es_with_token $espod "/project.$proj.*/_count" $test_name $test_token | \
+        nrecs=$( curl_es_pod_with_token $espod "/project.$proj.*/_count" $test_name $test_token | \
                      get_count_from_json )
         if ! os::cmd::expect_success "test $nrecs = $expected" ; then
             os::log::error $user $verb access project.$proj.* indices from es
-            curl_es_with_token $espod "/project.$proj.*/_count" $test_name $test_token | python -mjson.tool
+            curl_es_pod_with_token $espod "/project.$proj.*/_count" $test_name $test_token | python -mjson.tool
             exit 1
         fi
         nrecs=$( curl_es_from_kibana "$kpod" logging-es "/project.$proj.*/_count" $test_name $test_token | \
@@ -123,11 +123,11 @@ function test_user_has_proper_access() {
             exit 1
         fi
         # no user name - allow it - look up username corresponding to token
-        nrecs=$( curl_es_with_token $espod "/project.$proj.*/_count" "" $test_token | \
+        nrecs=$( curl_es_pod_with_token $espod "/project.$proj.*/_count" "" $test_token | \
                      get_count_from_json )
         if ! os::cmd::expect_success "test $nrecs = $expected" ; then
             os::log::error $user $verb access project.$proj.* indices from es
-            curl_es_with_token $espod "/project.$proj.*/_count" "" $test_token | python -mjson.tool
+            curl_es_pod_with_token $espod "/project.$proj.*/_count" "" $test_token | python -mjson.tool
             exit 1
         fi
         nrecs=$( curl_es_from_kibana "$kpod" logging-es "/project.$proj.*/_count" "" $test_token | \
@@ -139,11 +139,11 @@ function test_user_has_proper_access() {
         fi
         # if wrong user name is given, allow it - server will look up and use correct user name, so
         # not possible to specify a privileged username to use for access with an unprivileged token
-        nrecs=$( curl_es_with_token $espod "/project.$proj.*/_count" $LOG_ADMIN_USER $test_token | \
+        nrecs=$( curl_es_pod_with_token $espod "/project.$proj.*/_count" $LOG_ADMIN_USER $test_token | \
                      get_count_from_json )
         if ! os::cmd::expect_success "test $nrecs = $expected" ; then
             os::log::error $user $verb access project.$proj.* indices from es
-            curl_es_with_token $espod "/project.$proj.*/_count" $LOG_ADMIN_USER $test_token | python -mjson.tool
+            curl_es_pod_with_token $espod "/project.$proj.*/_count" $LOG_ADMIN_USER $test_token | python -mjson.tool
             exit 1
         fi
         nrecs=$( curl_es_from_kibana "$kpod" logging-es "/project.$proj.*/_count" $LOG_ADMIN_USER $test_token | \
@@ -156,23 +156,23 @@ function test_user_has_proper_access() {
         if [ "$expected" = 1 ] ; then
             # make sure no access with incorrect auth
             # username with no token
-            os::cmd::expect_success_and_text "curl_es_with_token $espod '/project.$proj.*/_count' '$test_name' '' -w '%{response_code}\n'" '^401$'
+            os::cmd::expect_success_and_text "curl_es_pod_with_token $espod '/project.$proj.*/_count' '$test_name' '' -w '%{response_code}\n'" '^401$'
             os::cmd::expect_success_and_text "curl_es_from_kibana $kpod $eshost '/project.$proj.*/_count' '$test_name' '' -w '%{response_code}\n'" '^401$'
             # username and bogus token
-            os::cmd::expect_success_and_text "curl_es_with_token $espod '/project.$proj.*/_count' '$test_name' BOGUS -w '%{response_code}\n'" '^401$'
+            os::cmd::expect_success_and_text "curl_es_pod_with_token $espod '/project.$proj.*/_count' '$test_name' BOGUS -w '%{response_code}\n'" '^401$'
             os::cmd::expect_success_and_text "curl_es_from_kibana $kpod $eshost '/project.$proj.*/_count' '$test_name' BOGUS -w '%{response_code}\n'" '^401$'
             # no username, no token
-            os::cmd::expect_success_and_text "curl_es_with_token $espod '/project.$proj.*/_count' '' '' -w '%{response_code}\n'" '^401$'
+            os::cmd::expect_success_and_text "curl_es_pod_with_token $espod '/project.$proj.*/_count' '' '' -w '%{response_code}\n'" '^401$'
             os::cmd::expect_success_and_text "curl_es_from_kibana $kpod $eshost '/project.$proj.*/_count' '' '' -w '%{response_code}\n' -o /dev/null" '^403$'
         fi
     done
 
     os::log::info See if user $user is denied /.operations.*
-    nrecs=$( curl_es_with_token $esopspod "/.operations.*/_count" $test_name $test_token | \
+    nrecs=$( curl_es_pod_with_token $esopspod "/.operations.*/_count" $test_name $test_token | \
                  get_count_from_json )
     if ! os::cmd::expect_success "test $nrecs = 0" ; then
         os::log::error $LOG_NORMAL_USER has improper access to .operations.* indices from es
-        curl_es_with_token $esopspod "/.operations.*/_count" $test_name $test_token | python -mjson.tool
+        curl_es_pod_with_token $esopspod "/.operations.*/_count" $test_name $test_token | python -mjson.tool
         exit 1
     fi
     esopshost=logging-es-ops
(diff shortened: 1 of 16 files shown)


richm:
What about downstream?  We can't use gem-patch.

ewolinetz:
@richm The change was accepted and tagged for fluent-systemd-plugin. The release is `0.0.10`

richm:
@portante so the patch accepted upstream will cause log record duplication?

portante:
Yes, that is what we found debugging the problem from the rsyslog side.  One can't close and re-open the journal with out saving the cursor, and saving the cursor so tightly can be problematic.  The fix is to not close and reopen the journal, but just properly setup the inotify FD in the systemd APIs by calling `sd_journal_get_fd()` immediately after calling `sd_journal_open()`.  You should be able to see that in the rsyslog patch.

ewolinetz:
@portante would it be a better fix if we closed the old journal object after https://github.com/reevoo/fluent-plugin-systemd/blob/0.0.x/lib/fluent/plugin/in_systemd.rb#L39 instead?

portante:
I don't think so, @ewolinetz, because it is the fact that the journal is closed at all, and then re-opened without tracking a cursor to remember where you are in the journal.

The fact is, we don't need to close the journal all.  We just need to call sd_journal_get_fd() after the first time we open and then can ignore invalidates entirely.  That is what we do with rsyslog.

ewolinetz:
@portante so you're saying for the gem, instead of :
```ruby
@journal.close() if @journal
@journal = Systemd::Journal.new(path: @path)
```
we need to update this logic so that we can compare the currently used fd to the one we would receive from calling `file_descriptor` to mirror https://github.com/rsyslog/rsyslog/commit/34b2d733699954e2ac770e8607c1694aacb51862#diff-b1ea6478b8060f07cd30ecde78bfdc49L698 ?

I'm still trying to figure out the correct location to update that logic... because the `fluent-systemd-plugin` would need to change how its calling the `systemd/journal` gem

test/zzzz-duplicate-entries.sh
```diff
@@ -0,0 +1,80 @@
...
+    os::test::junit::reconcile_output
+    exit $return_code
+}
+trap "cleanup" EXIT
+
+# turn off fluentd
+oc label node --all logging-infra-fluentd- 2>&1 | artifact_out || :
+os::cmd::try_until_text "oc get daemonset logging-fluentd -o jsonpath='{ .status.numberReady }'" "0" $((second * 120))
+
+sudo sed -i.bak "s/^.MaxRetentionSec=.*$/ MaxRetentionSec=30/" /etc/systemd/journald.conf
```
richm: `s/^.*MaxRetentionSec=.*$/ MaxRetentionSec=30/`
In case it is commented out, or it is at the beginning of the line, or line begins with `##` or etc.

test/zzzz-duplicate-entries.sh
```diff
@@ -0,0 +1,80 @@
...
+}
+sleep 10
+# we're actually waiting for rsyslog instead . . .
+wait_for_fluentd_to_catch_up get_logmessage get_logmessage2
+
+# dump random messages into journal for the next 35 seconds to ensure that we've rolled over
+MESSAGES=""
+TIMES=0
+while [ $TIMES -lt 35 ]; do
+  MESSAGE="$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)"
```
richm: `$( openssl rand -hex 16 )`

test/zzzz-duplicate-entries.sh
```diff
@@ -0,0 +1,80 @@
...
+sudo systemcl restart systemd-journal
+
+get_logmessage() {
+    logmessage="$1"
+}
+get_logmessage2() {
+    logmessage2="$1"
+}
+sleep 10
+# we're actually waiting for rsyslog instead . . .
```
richm: remove this

test/zzzz-duplicate-entries.sh
```diff
@@ -0,0 +1,80 @@
...
+wait_for_fluentd_to_catch_up get_logmessage get_logmessage2
```
richm: You don't need to use the `get_logmessage get_logmessage2` parameters and functions if you don't need the messages created by `wait_for_fluentd_to_catch_up`

test/zzzz-duplicate-entries.sh
ewolinetz: is there an advantage using one over the other?

test/zzzz-duplicate-entries.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
+
+# This is a test suite for testing basic log processing
+# functionality and existance of duplicate records when journal rolls over
+
+source "$(dirname "${BASH_SOURCE[0]}" )/../hack/lib/init.sh"
+source "${OS_O_A_L_DIR}/hack/testing/util.sh"
+os::util::environment::use_sudo
```
richm: I think this test should only run if `log-driver=journald` - the `add_test_message` function will add messages to kibana, and works by having fluentd get the kibana logs.  If using `log-driver=json-file` the logs will go to the json files, not the journal.

test/zzzz-duplicate-entries.sh
```diff
@@ -0,0 +1,80 @@
...
+sudo mv /etc/systemd/journald.conf.bak /etc/systemd/journald.conf
+sudo systemctl restart systemd-journal
+
+sleep 10
+
+wait_for_fluentd_to_catch_up get_logmessage get_logmessage2
+
+es_pod=$( get_es_pod es )
+for message in $MESSAGES; do
+  qs='{"query":{"match_phrase":{"message":"'"${message}"'"}}}'
```
richm: https://github.com/openshift/origin-aggregated-logging/blob/master/hack/testing/util.sh#L204
```
  fullmsg="GET /${message} 404 "
  qs='{"query":{"match_phrase":{"message":"'"${fullmsg}"'"}}}'
```
otherwise, you may get spurious hits

test/zzzz-duplicate-entries.sh
ewolinetz: is there a way i can restrict which this runs against? or is there a clear and easy way for me to push messages straight to journal?

ewolinetz: There isn't really a detriment if it runs against both though I suppose... it just doesn't test what we're trying to test otherwise...

test/zzzz-duplicate-entries.sh
richm: actually, this will always error out - because fluentd is shutdown - I think you need to restart fluentd after restarting the journal, then call `wait_for_fluentd_to_catch_up`

test/zzzz-duplicate-entries.sh
richm: no - they will both get the job done

richm:
```
hack/testing/entrypoint.sh: line 156: /data/src/github.com/openshift/origin-aggregated-logging/hack/testing/test-zzzz-duplicate-entries.sh: Permission denied
```
you need to chmod +x both hack/testing/test-zzzz-duplicate-entries.sh and test/zzzz-duplicate-entries.sh

richm:
I think we're stuck until we get the backports for the openshift-logging namespace fix in - I'm working on "forward porting" @jcantrill 's patch - unless he gets there first

richm:
```
Failed to restart systemd-journal.service: Unit not found.
```
systemd-journal`d`.service

test/zzzz-duplicate-entries.sh
```diff
@@ -37,18 +40,11 @@ trap "cleanup" EXIT
 oc label node --all logging-infra-fluentd- 2>&1 | artifact_out || :
 os::cmd::try_until_text "oc get daemonset logging-fluentd -o jsonpath='{ .status.numberReady }'" "0" $((second * 120))
 
-sudo sed -i.bak "s/^.MaxRetentionSec=.*$/ MaxRetentionSec=30/" /etc/systemd/journald.conf
-sudo systemcl restart systemd-journal
+sudo sed -i.bak "s/^.*MaxRetentionSec=.*$/ MaxRetentionSec=30/" /etc/systemd/journald.conf
+sudo systemctl restart systemd-journal
```
richm: systemd-journald

richm:
I think this test should only run against the es-ops instance.  That instance will always get records via journald, even when log-driver=json-file, so it can be run always.  That means it will work very similar to the new https://github.com/richm/origin-aggregated-logging/blob/9a8849099710ea42ad8636cdd13f28ab0fdfc600/test/zzzz-bulk_rejection.sh test which only runs against the es-ops instance, adds test records like this: https://github.com/richm/origin-aggregated-logging/blob/9a8849099710ea42ad8636cdd13f28ab0fdfc600/test/zzzz-bulk_rejection.sh#L166 - that is, creating a file with 1 message per line, then feeding that file to the `logger` command.
Then, check for the messages like this: https://github.com/richm/origin-aggregated-logging/blob/9a8849099710ea42ad8636cdd13f28ab0fdfc600/test/zzzz-bulk_rejection.sh#L171

You could check for messages by either doing an individual curl_es search for each message, or search like ^^^ for all messages to a single json file, then use something like jq to check the json file for each message.

ewolinetz:
failed waiting for fluentd to catch up...?
/retest

richm:
I suggest renaming the file to test-zzz-duplicate-entries.sh and zzz-duplicate-entries.sh, so that this test runs before the rsyslog and bulk index tests.

When should this test be run?  Is it going to alter the test environment so much that it will cause failures in other tests which are subsequently run?  If it is "moderately" stressful, I suggest that it runs before the z* tests - perhaps yyy-duplicate-entries.sh?

richm:
The test should only run against es-ops - since this test is really about the journal functionality - using es-ops will guarantee to test journal functionality no matter what the log driver is, since operations logs will always be read from the journal and will always be written to the .operations index in the es-ops instance.

richm:
When using es-ops/.operations, use the `logger` command to write the messages to look for.  You can use the `logger -f` option to create a file containing a lot of messages to write.  See https://github.com/openshift/origin-aggregated-logging/pull/1089/files#diff-1e94ccb79adaf3b43bc5ad62f61efd0bL161 for an example of how the bulk rejection test loads a bunch of ops records and looks for them.

test/zzz-duplicate-entries.sh
```diff
@@ -49,10 +49,12 @@ wait_for_fluentd_to_catch_up
 # dump random messages into journal for the next 35 seconds to ensure that we've rolled over
 MESSAGES=""
 TIMES=0
+opslogline=$( mktemp )
 while [ $TIMES -lt 35 ]; do
   MESSAGE="$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)"
   MESSAGES="$MESSAGES $MESSAGE"
-  add_test_message $MESSAGE
+  echo -e $MESSAGE > opslogline
+  logger -i -p local6.info -t $MESSAGE -f $opslogline
```
richm: so yes, a little different use case than the bulk rejection test.  In your case, you want systemd.u.SYSLOG_IDENTIFIER to be the same as the message field:
```
 while [ $TIMES -lt 35 ]; do
   MESSAGE="$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)"
   logger -i -p local6.info -t $MESSAGE $MESSAGE
   let TIMES=TIMES+1
 done
```
Then you don't need opslogline at all

ewolinetz: awesome, I wasn't sure if that was required or not... I'll update to remove the `-f` and just use `$MESSAGE`

richm:
```
[INFO] Logging test suite test-zzz-duplicate-entries started at Thu Apr 19 22:36:18 UTC 2018
/data/src/github.com/openshift/origin-aggregated-logging/hack/testing/test-zzz-duplicate-entries.sh: line 5: /data/src/github.com/openshift/origin-aggregated-logging/test/zzzz-duplicate-entries.sh: No such file or directory
```

test/zzz-duplicate-entries.sh
```diff
@@ -68,9 +68,8 @@ wait_for_fluentd_to_catch_up
 
 es_pod=$( get_es_pod es )
 for message in $MESSAGES; do
-  full_msg="GET /${message} 404 "
-  qs='{"query":{"match_phrase":{"message":"'"${full_msg}"'"}}}'
-  if os::cmd::try_until_text "curl_es ${es_pod} /project.logging.*/_count -X POST -d '$qs' | get_count_from_json_from_search" 1 $(( 10 * minute )); then
+  qs='{"query":{"term":{"systemd.u.SYSLOG_IDENTIFIER":"'"${message}"'"}}}'
+  if os::cmd::try_until_text "curl_es ${es_ops_pod} /.operations.*/_count -X POST -d '$qs' | get_count_from_json_from_search" 1 $(( 10 * minute )); then
```
richm: Use `get_count_from_json` because you are doing a `_count` operation as opposed to a `_search` operation for which you would use `get_count_from_json_from_search`

richm:
On line 40 you shutdown fluentd: https://github.com/ewolinetz/origin-aggregated-logging/blob/bd89bbd47bd9205d781ee926726a972a714e3edf/test/yyy-duplicate-entries.sh#L40

But you never restart it, which is why the `wait_for_fluentd_to_catch_up` fails at line 47

ewolinetz:
> But you never restart it, which is why the wait_for_fluentd_to_catch_up fails at line 47

For some reason I saw that I was just restarting journald and thought it was the same thing... thanks @richm

richm:
this is a new one
```
[2018-04-20T21:19:23.193+0000] SUCCESS after 0.226s: test/zzz-correct-index-names.sh:40: executing 'oc get project openshift-this-is-a-test' expecting success; re-trying every 0.2s until completion or 180.000s
[2018-04-20T21:19:23.463+0000] Error from server (Forbidden): pods "test-pod" is forbidden: error looking up service account openshift-this-is-a-test/default: serviceaccount "default" not found
```
maybe timing?  Even though `oc get project openshift-this-is-a-test` returned success, perhaps the namespace is still not ready to accept pods?

richm:
nice!

richm:
```
[2018-04-21T02:03:47.743+0000] [ERROR] PID 23971: hack/testing/util.sh:109: `curl --silent --insecure "${args[@]}" --key "${secret_dir}/admin-key" --cert "${secret_dir}/admin-cert" "https://${svc_name}.${LOGGING_NS}.svc:9200${endpoint}"` exited with status 6.
```
add `-v` to curl to see what the error is - usually means cert or key is not valid

richm:
not sure why https://github.com/ewolinetz/origin-aggregated-logging/blob/8a2ca68dbe4eb961d6316530a170c0be4b3bdb17/hack/testing/util.sh#L388 would return a non-numeric value . . .

richm:
origin install flake?
/retest

richm:
```
ValueError: No JSON object could be decoded
[ERROR] PID 16151: test/access_control.sh:115: `nrecs=$( curl_es_with_token $essvc "/project.$proj.*/_count" $test_name $test_token |                      get_count_from_json
https://github.com/notifications )` exited with status 1.
```
means `curl_es_with_token` did not return JSON, which means a curl error.  I suppose you could put a ` tee -a $ARTIFACT_DIR/access_control_curl_errors.log |` before `get_count_from_json` to see what the curl output looks like.

test/access_control.sh
```diff
@@ -27,8 +24,8 @@ function cleanup() {
             oc delete user $user 2>&1 | artifact_out
         done
     fi
-    if [ -n "${essvc:-}" ] ; then
-        curl_es $essvc /project.access-control-* -XDELETE > /dev/null
+    if [ -n "${espod:-}" ] ; then
+        curl_es $espod /project.access-control-* -XDELETE > /dev/null
```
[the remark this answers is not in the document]
ewolinetz: I foolishly only did a s/ on `curl_es_with_token`

ewolinetz:
Looks like for the journald and json-file tests we didn't get any duplicate records unless we need to stress it further.

thoughts @richm  @portante ?

portante:
In order to get duplicate records, fluentd needs to close and reopen the journal where it rereads journald records already sent.  To do that, the logging rate for the journal needs to be high, causing "large" journal files, but still small enough in size to cause journal log file rotations which will trigger the problem.

So what is the logging rate of the test, and how big are the journal files configured to be?

ewolinetz:
@portante not very high -- 1 message/sec and we are rolling over the journal files based on time not size.

I could update the test to hammer journal as fast as it can for 35 seconds and maintain the 30sec life of journal files for this test. Do you think that would satisfy the performance required to recreate this or is it required to have very large journal files to recreate this issue?

portante:
@ewolinetz, this kind of testing would probably be best served by a regression test of sorts.

richm:
I think this bug can be tested in a CI test - if one was going to test this outside of CI, what would one do?  Is that automatable?

richm:
```
/data/src/github.com/openshift/origin-aggregated-logging/hack/testing/util.sh: line 168: svc_name: unbound variable
```

richm:
origin install flake?
/retest

ewolinetz:
install flake
/retest

ewolinetz:
Noticed that we generated 3.3k messages and we weren't seeing any of the later ones so I moved the wait_for_fluentd up to after we generate all our messages

ewolinetz:
So far I don't see duplicate records, but I did see the last 213 records (of 2868) weren't showing up as found. Trying to verify that this isn't a result of the plugin update before being ready to merge this.

ewolinetz:
jenkins died for logging...
/test logging

ewolinetz:
nvm, guess it was just my console view that had an issue...

ewolinetz:
it looks like even with downgrading the plugin to a prior version it can't grab all the messages generated

ewolinetz:
journald job completed duplicate-entry test (json-file had flake)
`[INFO] Logging test suite test-zzz-duplicate-entries succeeded at Wed May  9 19:44:04 UTC 2018`

ewolinetz:
for both tests: `[ERROR] PID 19702: test/zzzz-bulk_rejection.sh:18: `es_ver=$( get_es_major_ver )` exited with status 6.` duplicate-entries passed though

ewolinetz:
looks like flakes for both tests
/retest

richm:
remote-syslog flake
/retest

ewolinetz:
prereqs failed to install...
/retest

richm:
flakes
/retest

richm:
remote-syslog flake
/retest

ewolinetz:
@jcantrill @richm @portante would it be useful to merge this test in? It seems we have trouble recreating the issue that its seeking to catch.

richm:
sure - rebase and squash - this will at least give us a base to work from as we work on adding an automated test for this issue

ewolinetz:
@richm squashed

ewolinetz:
@richm I'll create manual cherrypicks for these changes for `3.9` and `es5.x`

PR #1077, opened 2018-04-11
bug 1565909. Modify the default operation to be create to avoid delet…

jcantrill:
…ions

(cherry picked from commit e5ddcdf665bad46dc4a6310a98e9b76aa00fd404)


jcantrill:
3.9 was done in #1076

richm:
need a ci test for this - this bit us because when we switched to using the genid we did not have a ci test to confirm no duplicates and no deletions

jcantrill:
How are we to confirm 'no deletions'?  Essentially what is happening is we get deletions by 'update'?

portante:
> Not sure if you can query for deleted documents ...

Yes, you can.  The "_cat/indices" API [1] will give you the count of deleted documents.

[1] https://www.elastic.co/guide/en/elasticsearch/reference/2.4/cat-indices.html

PR #1081, opened 2018-04-13
[release-3.9] Fix access-control, mux-client-mode, docker hostmount, other issues

richm:
hack for docker hostmount problem
    better way to save and restore bulk config

    fix mux-client-mode by tagging raw messages with .raw suffix
    The code relies on the presence of the @timestamp field to indicate that
    the record has already been processed into the common data model. However,
    some applications actually use the @timestamp field, so we need some other
    way to denote to mux that the message is a raw message. Change fluentd
    running in MUX_CLIENT_MODE=minimal to add a `.raw` tag suffix to
    unprocessed log records before sending to mux so that mux will know the
    record is not processed.

    ignore errors for bulk index monitor
    improve mux-client-mode debugging
    efficiently look for source record to report missing records
    add errorfile to rsyslog test to capture elasticsearch errors
    add bulk indexing stats monitor to CI
    fix bugs
    sleep to allow cluster admin user to be processed; more debugging

### DIFF
--- a/docs/mux-logging-service.md
+++ b/docs/mux-logging-service.md
@@ -112,7 +112,7 @@ You will need the mux CA cert and shared_key.  You can obtain them from the
 logging cluster like this:
 
     # oc login --username=system:admin
-    # oc project logging
+    # oc project openshift-logging # if that project doesn't exist, use logging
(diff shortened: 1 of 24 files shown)


richm:
/retest
let's see if it picks up the openshift-ansible logging project revert

richm:
strange access_control flake - never seen it before
/test json-file

PR #1082, opened 2018-04-13
Modify the check for logging namespace

sureshgaikwad:
Signed-off-by: root <root@sgaikwad.pnq.csb>
The script fails to capture the logging data if the namespace openshift-logging does not exist or if the user is not a member of this project.
[root@master ~]# ./logging-dump.sh
error: You are not a member of project "openshift-logging".

[root@master ~]# ./logging-dump.sh
error: A project named "openshift-logging" does not exist on "https://master.example.com:443"

Changed the code to use logging namespace as a default namespace to capture the information.

### DIFF
--- a/hack/logging-dump.sh
+++ b/hack/logging-dump.sh
@@ -46,11 +46,11 @@ then
     components=( "kibana" "fluentd" "curator" "elasticsearch" "project_info" )
 fi
 
-if [ -z "${NAMESPACE:-''}" ] && oc get project logging > /dev/null ; then
+if [ -z "${NAMESPACE:-}" ] && oc get project logging > /dev/null ; then
   NAMESPACE=logging
 fi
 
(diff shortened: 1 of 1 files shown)


nekop:
I think the issue is previous lines, it seems the script is trying to check `logging` project but the test condition is wrong.

```
if [ -z "${NAMESPACE:-''}" ] && oc get project logging > /dev/null ; then
```

As a result of this line, the NAMESPACE is set to `''` (not an empty string, it's 2 quote chars). This quote chars are not needed. Fix would be:

```
if [ -z "${NAMESPACE:-}" ] && oc get project logging > /dev/null ; then
```

sureshgaikwad:
yes, the two quotes are not required. Removing the two quotes and leaving ${NAMESPACE:-openshift-logging} as is should be sufficient.

PR #1088, opened 2018-04-18
[es5.x] fix several CI issues; merge with latest master; bulk reject test fixes

richm:
make bulk rejection test work with ES5; use es-ops instead of es

Due to
https://www.elastic.co/guide/en/elasticsearch/reference/5.1/breaking_50_settings_changes.html#_threadpool_settings
we cannot dynamically update the bulk threadpool settings, so have to hack
the es configmap and rollout. Only run tests against es-ops to avoid having
to create another namespace/pod for the es instance (since logging could be
openshift-logging).

ES5 also changed the _cat api for the bulk thread_pool, so add some utility
functions to get the ES version and the bulk thread_pool url.

support logging-xxx5 images

debug remote-syslog test
(cherry picked from commit 78325ee4216126b13f6aeba215f5174131429fdd)

Sort events by timestamp
(cherry picked from commit e01bae2e0f864aecb68bc29fe02fcf27c49f0df1)

Take both forward and secure_forward into account for NUM_OUTPUTS
Reported-by: Ludovic Juaneda
(cherry picked from commit 1ee717be49cf31b5d9755e4addd6bdb10921f46f)

fluentd: determine if openshift uses cri-o from nodeconfig
Parse `etc/origin/node/node-config.yaml` config file for
`kubeletArguments -> container-runtime-endpoint` if it contains `crio`.
Change generation of `in_tail` plugin config to don't try parsing as json
but as a regex instead if crio is identified.
The regex used is:
/^(?<time>.+) (?<stream>stdout|stderr)( (?<logtag>.))? (?<message>.*)$/
and is based on the parsing logic in cri-o for 1.8 and 1.9:
https://github.com/kubernetes-incubator/cri-o/blob/release-1.8/oci/oci.go#L363
https://github.com/kubernetes-incubator/cri-o/blob/release-1.9/oci/oci.go#L363
NOTE: the logtag was introduced in 1.9 so it is put as optional here, and
currently doesn't appear to be working well with Elasticsearch logs so this
PR doesn't try to append and reconstruct multiline logs.
(cherry picked from commit d79019b5edcd42be98aab7b9777b253da9f65515)

Bug 1550842 - Embed json logs when using cri-o
(cherry picked from commit 49d82d2cfa417a84ae10ec343018b8e88e4a8ef7)

Bug 1548720 - fluent-plugin-secure-forward couldn't read the Environment
${HOSTNAME}
Replacing the self_hostname value ${HOSTNAME} with the placeholder
${hostname}.
(cherry picked from commit 99175ee3a99f5faf15b9b8e3b89b2cbf4d6d0bc1)

fluentd: cri-o config env variables
(cherry picked from commit f4ed20c8d56053585d2d12531473ba748f55b677)

many, many fixes for openshift-logging and related code cleanup
(cherry picked from commit 7db3a9426ae1e7a8d7ddbb7468161dd0115f2a4b)

bug 1548104. Generate record uuids upon ingestion to minimize duplicates
(cherry picked from commit 5e1e372658b57f916af9543efb3f096cc2936bfc)

check logging namespace for actual resources
(cherry picked from commit d68cc2126ddba1e938d58495dcb183993a84ee5a)

add openshift/origin-aggregated-logging/pull/1067

use LOGGING_NS
hack for docker hostmount problem
(cherry picked from commit 47bdd7fef7df63474adeb832db14a9c526f109e4)

better way to save and restore bulk config
(cherry picked from commit f4afda0e7c52ffb72e27a04c8f58270fa5a4558c)

bug 1535300. modify the default namespace for logging
(cherry picked from commit 06925e7159824df0c0c7fe63e6aeabb9aec2a166)

add bulk index rejection test
This tests the ability of fluentd to handle bulk index rejections. The test
starts up several curl processes which hammer Elasticsearch with bulk index
requests which causes fluentd to receive bulk rejections. The methodology
is to add some messages for both projects and operations indices which
should be rejected and queued and eventually resent successfully. The test
will ensure that Elasticsearch is rejecting bulk requests and fluentd is
receiving bulk rejections. The test has to run last because it will cause
other CI tests to have test flakes due to bulk index rejections, which
somehow become persistent after this test runs.
(cherry picked from commit 9c89068ac9dadfd3837781b196e992164bc8de80)

ignore errors for bulk index monitor
(cherry picked from commit d9b66001290a8b9a80791baaa6af655d02e191e9)

improve mux-client-mode debugging
(cherry picked from commit 17db82836e24252766b84bc8d71f7445552f1930)

efficiently look for source record to report missing records
(cherry picked from commit cb491a314ff2ae589a46e5a8c62ea8975b593548)

add errorfile to rsyslog test to capture elasticsearch errors
(cherry picked from commit d1cdaa83296bc56b4f08d41706460827a1210aa2)

new location for node kubeconfig; dump rsyslog log at end of test
(cherry picked from commit 41b6325a53d9d41b1de0beaa7296e2c1d3aa05f7)

rsyslog testing - test k8s, es tls, viaq formatting
This tests the ability of rsyslog to do kubernetes metadata annotation,
elasticsearch with client cert auth, and some basic viaq data model
formatting
This currently uses pre-release rsyslog built in fedora copr, but that
should move into more official repos soon.
(cherry picked from commit 2b6366b3b6b852b13e94f1a2707355289e05ba97)

add bulk indexing stats monitor to CI
(cherry picked from commit 4b24e0b1948bdcaacf3aae2d8b28acc54ee43059)

fix bugs
(cherry picked from commit ed147175e0e8fb86d7a36e5c0ba7c875a31e8475)

sleep to allow cluster admin user to be processed; more debugging
(cherry picked from commit 667ac5b5176b1da5880a8e10f74aa7391616fad4)

### DIFF
--- a/fluentd/configs.d/filter-pre-mux-client.conf
+++ b/fluentd/configs.d/filter-pre-mux-client.conf
@@ -1,7 +1,7 @@
 <match **>
   @type secure_forward
   log_level "#{ENV['FORWARD_INPUT_LOG_LEVEL'] || ENV['LOG_LEVEL'] || 'warn'}"
-  self_hostname ${HOSTNAME}
+  self_hostname ${hostname}
   shared_key "#{File.open('/etc/fluent/muxkeys/shared_key') do |f| f.readline end.rstrip}"
   secure yes
   ca_cert_path /etc/fluent/muxkeys/ca
--- a/fluentd/configs.d/user/secure-forward.conf
+++ b/fluentd/configs.d/user/secure-forward.conf
@@ -1,8 +1,8 @@
 # <store>
 #   @type secure_forward
 
-#   self_hostname ${HOSTNAME}
-#   shared_key ${SECRET_STRING}
+#   self_hostname ${hostname} # ${hostname} is a placeholder.
+#   shared_key <shared_key_between_forwarder_and_forwardee>
 
 #   secure yes
 #   enable_strict_verification yes
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -214,12 +214,12 @@ if [ $TOTAL_LIMIT -le 0 ]; then
     exit 1
 fi
 
-# If secure-forward outputs are configured, add them to NUM_OUTPUTS.
-sec_forward_files=$( grep -l "@type *secure_forward" ${CFG_DIR}/*/* 2> /dev/null || : )
-for afile in ${sec_forward_files} ; do
+# If forward and secure-forward outputs are configured, add them to NUM_OUTPUTS.
+forward_files=$( grep -l "@type .*forward" ${CFG_DIR}/*/* 2> /dev/null || : )
+for afile in ${forward_files} ; do
     file=$( basename $afile )
     if [ "$file" != "${mux_client_filename:-}" ]; then
-        grep "@type *secure_forward" $afile | while read -r line; do
+        grep "@type .*forward" $afile | while read -r line; do
             if [ $( expr "$line" : "^ *#" ) -eq 0 ]; then
                 NUM_OUTPUTS=$( expr $NUM_OUTPUTS + 1 )
             fi
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -25,6 +25,19 @@ else
   set -e
 fi
 
+#NOTE: USE_CRIO variable used in generate_throttle_configs.rb as well
+export USE_CRIO=false
+node_config=/etc/origin/node/node-config.yaml
+if [[ -f "$node_config" ]]; then
+    cre=$(ruby -e "require 'yaml'; y=YAML.load_file('${node_config}')
+        puts y['kubeletArguments']['container-runtime-endpoint']")
+    if [[ "$cre" =~ crio ]]; then
+        USE_CRIO=true
+    fi
+else
+    echo "WARNING: Unable to check for cri-o"
+fi
+
 docker_uses_journal() {
     # need to be able to handle cases like
     # OPTIONS='--log-driver=json-file ....' # or use --log-driver=journald
@@ -51,7 +64,9 @@ if [ -z "${USE_MUX:-}" -o "${USE_MUX:-}" = "false" ] ; then
             export JOURNAL_SOURCE=/run/log/journal
         fi
     fi
-    if docker_uses_journal ; then
+    if [[ "$USE_CRIO" == true ]]; then
+        export USE_JOURNAL=false
+    elif docker_uses_journal ; then
         export USE_JOURNAL=true
     else
         export USE_JOURNAL=false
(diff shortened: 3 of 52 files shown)


fluentd/configs.d/filter-pre-mux-client.conf
```diff
@@ -1,7 +1,7 @@
 <match **>
   @type secure_forward
   log_level "#{ENV['FORWARD_INPUT_LOG_LEVEL'] || ENV['LOG_LEVEL'] || 'warn'}"
-  self_hostname ${HOSTNAME}
+  self_hostname ${hostname}
```
ewolinetz: shouldn't this remain `HOSTNAME` so that we can get the name of the pod that's injected into our ENV by k8s? or does `hostname` also resolve to that on the pod?

fluentd/configs.d/user/secure-forward.conf
```diff
@@ -1,8 +1,8 @@
 # <store>
 #   @type secure_forward
 
-#   self_hostname ${HOSTNAME}
-#   shared_key ${SECRET_STRING}
+#   self_hostname ${hostname} # ${hostname} is a placeholder.
```
ewolinetz: same question about `hostname` vs `HOSTNAME` here

fluentd/configs.d/filter-pre-mux-client.conf
nhosoi: The lowercase ${hostname} is a placeholder provided by fluentd.  Please see this bug.  Thanks!
https://bugzilla.redhat.com/show_bug.cgi?id=1548720

ewolinetz: ack. that's strange though, when did that stop working? When we originally introduced secure forward `HOSTNAME` resolved correctly...

fluentd/run.sh
```diff
@@ -25,6 +25,19 @@ else
...
 fi
 
+#NOTE: USE_CRIO variable used in generate_throttle_configs.rb as well
+export USE_CRIO=false
+node_config=/etc/origin/node/node-config.yaml
+if [[ -f "$node_config" ]]; then
+    cre=$(ruby -e "require 'yaml'; y=YAML.load_file('${node_config}')
+        puts y['kubeletArguments']['container-runtime-endpoint']")
+    if [[ "$cre" =~ crio ]]; then
+        USE_CRIO=true
```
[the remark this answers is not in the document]
richm: yes - it is exported on line 29 - once it is exported, you don't have to re-export it

fluentd/configs.d/filter-pre-mux-client.conf
richm: Not really sure what `self_hostname` is used for, access control maybe? but we aren't using that

fluentd/run.sh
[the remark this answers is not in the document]
nhosoi: ok.  acked.

fluentd/configs.d/filter-pre-mux-client.conf
nhosoi: It looks it's being used at least for checking the forwarder vs the forwardee?  (Please note that @anpingli added just the line 176 to reveal the bug.)
https://bugzilla.redhat.com/show_bug.cgi?id=1548720#c0

richm:
can I get an /lgtm?

PR #1089, opened 2018-04-19
make bulk rejection test work with ES5; use es-ops instead of es

richm:
Due to https://www.elastic.co/guide/en/elasticsearch/reference/5.1/breaking_50_settings_changes.html#_threadpool_settings
we cannot dynamically update the bulk threadpool settings, so have
to hack the es configmap and rollout.
Only run tests against es-ops to avoid having to create another
namespace/pod for the es instance (since logging could be
openshift-logging).

ES5 also changed the _cat api for the bulk thread_pool, so add some
utility functions to get the ES version and the bulk thread_pool url.

The setting in ES2 is `threadpool` and in ES5 is `thread_pool`

(cherry picked from commit 9a8849099710ea42ad8636cdd13f28ab0fdfc600)


richm:
This should also help with recent test flakes in the bulk rejection test

richm:
had to fix a flake where es was not running by the time the bulk index monitor started - now the monitor will wait until es is running
please re-lgtm

PR #1108, opened 2018-04-24
bug 1568361. Move ES logs to persistent volume with max backup

jcantrill:
Fix: https://bugzilla.redhat.com/show_bug.cgi?id=1568361

This PR:

* Moves logs to persistent dir
* Adds custom appender to allow daily rotation with max backup
* defaults rotation to per day, max of 7
* Modifies logging.yml at run if needed
* Modifies elasticsearcy.yml if needed to get the proper logs dir

The modification to run.sh is not required for ES5.x because it relies on log4j2 that has a max rollover setting


elasticsearch/run.sh
```diff
@@ -24,6 +27,31 @@ retry=$RETRY_COUNT
 max_time=$(( RETRY_COUNT * RETRY_INTERVAL ))	# should be integer
 timeouted=false
 
+# modify configs as needed
+mkdir -p ${HOME}/config
+cp ${ES_CONF}/* ${HOME}/config
+pushd ${HOME}/config
+  if [ -f ${HOME}/config/logging.yml ] && ! grep -q "io\.openshift\.log4jextras" ${HOME}/config/logging.yml ; then
```
richm: use `logging.yml` instead of `${HOME}/config/logging.yml` since you already did the pushd, and to be consistent with the python script

elasticsearch/run.sh
```diff
@@ -24,6 +27,31 @@ retry=$RETRY_COUNT
...
+    python -c '
+import yaml
+with open("logging.yml") as file:
+  c = yaml.load(file)
+  for appender in ["file", "deprecation_log_file", "index_search_slow_log_file", "index_indexing_slow_log_file"]:
+    if c["appender"][appender]:
```
richm: `if "appender" in c and appender in c["appender"]`

elasticsearch/run.sh
```diff
@@ -24,6 +27,31 @@ retry=$RETRY_COUNT
...
+      f = c["appender"][appender]
+      f["type"] = "io.openshift.log4jextras.appender.DailyRollingFileAppender"
+      f["maxBackupIndex"] = "${ES_LOGS_MAX_BACKUP}"
+      f["datePattern"] = "${ES_LOGS_DATE_PATTERN}"
```
richm:
```
      f["maxBackupIndex"] = sys.argv[1]
      f["datePattern"] = sys.argv[2]
```

elasticsearch/run.sh
```diff
@@ -24,6 +27,31 @@ retry=$RETRY_COUNT
 max_time=$(( RETRY_COUNT * RETRY_INTERVAL ))	# should be integer
 timeouted=false
 
+# modify configs as needed
+mkdir -p ${HOME}/config
+cp ${ES_CONF}/* ${HOME}/config
+pushd ${HOME}/config
+  if [ -f ${HOME}/config/logging.yml ] && ! grep -q "io\.openshift\.log4jextras" ${HOME}/config/logging.yml ; then
+    python -c '
+import yaml
```
richm: `import yaml,sys`

elasticsearch/run.sh
```diff
@@ -24,6 +27,31 @@ retry=$RETRY_COUNT
...
+  for appender in ["file", "deprecation_log_file", "index_search_slow_log_file", "index_indexing_slow_log_file"]:
+    if c["appender"][appender]:
+      f = c["appender"][appender]
+      f["type"] = "io.openshift.log4jextras.appender.DailyRollingFileAppender"
+      f["maxBackupIndex"] = "${ES_LOGS_MAX_BACKUP}"
+      f["datePattern"] = "${ES_LOGS_DATE_PATTERN}"
+      f["append"] = True
+  with open("logging.new","w") as target:
+    yaml.dump(c, target)
+' ||:
```
richm:
```
' "${ES_LOGS_MAX_BACKUP}" "${ES_LOGS_DATE_PATTERN}"
```
And why do you need to disregard all errors?

elasticsearch/run.sh
jcantrill: with the construct of `f["maxBackupIndex"] = "${ES_LOGS_MAX_BACKUP}"` elastic will evaluate environment vars which we use as a pattern before this change

elasticsearch/run.sh
jcantrill: My thinking here is I don't want the pod to die if this operation fails

elasticsearch/run.sh
richm: ah, ok

elasticsearch/run.sh
richm: ok

elasticsearch/run.sh
```diff
@@ -24,6 +27,31 @@ retry=$RETRY_COUNT
 max_time=$(( RETRY_COUNT * RETRY_INTERVAL ))	# should be integer
 timeouted=false
 
+# modify configs as needed
+mkdir -p ${HOME}/config
+cp ${ES_CONF}/* ${HOME}/config
+pushd ${HOME}/config
+  if [ -f logging.yml ] && ! grep -q "io\.openshift\.log4jextras" ${HOME}/config/logging.yml ; then
```
richm: `${HOME}/config/logging.yml` -> `logging.yml`

ewolinetz:
Wouldn't the ES changes for the log file rotation be easier to implement in the ES configmap in the openshift-ansible role?

jcantrill:
@ewolinetz I think at some point later we should update ansible and remove the code in run.sh

ewolinetz:
> @ewolinetz yes but we dont know if the image or ansible will release first. I'm trying to avoid upgrade issues we have experienced in the past where we get errors or other when the config and image dont match

@jcantrill I would argue that this fix doesn't belong in this image though. It is a fix that is non-obvious and worries me that it will have issues if the ansible fix and the image fix are both present. This is something we could better do with just pushing it into ansible and creating a knowledge base doc to describe how to patch this on existing systems.

jcantrill:
@ewolinetz The tests all passed and I see no error in https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_origin-aggregated-logging/1108/test_pull_request_origin_aggregated_logging_json_file/351/artifacts/scripts/entrypoint/logs/containers/k8s_elasticsearch_logging-es-data-master-zie2mjg4-1-ss94k_openshift-logging_983f6f56-48a4-11e8-bbf5-0e5fb8a993a8_0.log that states it is unable to find the appender class

PR #1112, opened 2018-04-25
no way to guarantee fluentd will generate bulk index rejections

richm:
There seems to be no way to get fluentd to consistently get bulk
index rejections.  So, just report warn/info if there were no
bulk rejections during the test run.

### DIFF
--- a/test/zzzz-bulk_rejection.sh
+++ b/test/zzzz-bulk_rejection.sh
@@ -92,6 +92,12 @@ f_ds=$( mktemp )
 oc get cm/logging-fluentd -o yaml > $f_cm
 oc get ds/logging-fluentd -o yaml > $f_ds
 
+# stop fluentd to make sure the logs are clear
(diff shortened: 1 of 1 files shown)


test/zzzz-bulk_rejection.sh
```diff
@@ -183,6 +187,11 @@ fi
...
 wait $curlpid
 endtime=$( date +%s )
+end_bulk_rejections=$( curl_es $esopspod ${bulk_reject_url} )
+
+if ! os::cmd::expect_success "test ${start_bulk_rejections} -lt ${end_bulk_rejections}" ; then
+    os::log::warning No bulk rejections reported between $( date --date=@$starttime ) and $( date --date=@$endtime )
+fi
 
 # check the logs to see if there are bulk index rejection errors between starttime and endtime
 found=
```
nhosoi: A very minor thing, but it seems found is no longer used?

PR #1114, opened 2018-04-25
disable curator test

jcantrill:
Disabled curator test for time being since its blocking all other PRs

PR #1115, opened 2018-04-25
Disable curator test

jcantrill:
3.9 cherry pickof #1114

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ build-images:
 .PHONY: build-images
 
 test:
-	EXCLUDE_SUITE=upgrade hack/testing/entrypoint.sh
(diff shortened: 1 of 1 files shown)


richm:
do we need to disable curator on 3.9?  the problematic changes are in openshift-ansible 3.9?

ewolinetz:
This only exists on master for now. It is related to https://github.com/openshift/openshift-ansible/pull/7985 which hasn't merged into 3.9 yet.

Also https://github.com/openshift/openshift-ansible/pull/8139 should fix this

ewolinetz:
@richm we will want a cherrypick of 8139, but yes. So long as 7985's cherrypick lands after then we won't see this issue at all.

PR #1134, opened 2018-05-02
Bug 1515715 - [RFE]No record generator information in rsyslog server

nhosoi:
Fixing a test failure -- no tag_key

### DIFF
--- a/fluentd/out_syslog.rb
+++ b/fluentd/out_syslog.rb
@@ -14,7 +14,7 @@ module Fluent
     config_param :port, :integer, :default => 25
     config_param :hostname, :string, :default => ""
     config_param :remove_tag_prefix, :string, :default => nil
-    config_param :tag_key, :array, default: nil
+    config_param :tag_key, :array, default: []
     config_param :facility, :string, :default => 'user'
     config_param :severity, :string, :default => 'debug'
     config_param :use_record, :string, :default => nil
@@ -99,7 +99,7 @@ module Fluent
         end
         @packet.time = time
         @packet.tag = @random_string
-        if @tag_key.any? && record.any?
+        if !@tag_key.empty? && !record.empty?
           # tag_key is an array type
           # E.g., tag_key ident,systemd.u.SYSLOG_IDENTIFIER,ident
           #       tkey = ident
--- a/test/remote-syslog.sh
+++ b/test/remote-syslog.sh
@@ -418,3 +418,81 @@ sudo egrep "/${mymessage}" /var/log/messages 2>&1 | artifact_out || :
 if [ $rc -eq 1 ] ; then
     exit 1
 fi
+
+title="Test 7, no tag_key"
+os::log::info $title
+
+myhost=$( hostname )
+
+if [ "$fluentdtype" = "fluentd" ] ; then
+    # make sure fluentd is running after previous test
+    os::cmd::try_until_text "oc get pods -l component=fluentd" "^logging-fluentd-.* Running "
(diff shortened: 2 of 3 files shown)


fluentd/out_syslog.rb
```diff
@@ -99,7 +99,7 @@ def emit(tag, es, chain)
         end
         @packet.time = time
         @packet.tag = @random_string
-        if @tag_key.any? && record.any?
+        if @tag_key && @tag_key.any? && record.any?
```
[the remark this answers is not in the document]
nhosoi: Thank you, @richm.  I've updated the patch....

fluentd/out_syslog.rb
```diff
@@ -99,7 +99,7 @@ def emit(tag, es, chain)
         end
         @packet.time = time
         @packet.tag = @random_string
-        if @tag_key.any? && record.any?
+        if !@tag_key.nil? && @tag_key.any? && record.any?
```
richm: `if !@tag_key.nil? && @tag_key.respond_to?(:any?) && record.respond_to?(:any?)`
You have to use `respond_to?` to see if an object provides a method.  You cannot use `any?` if the object is not a string:
```
irb(main):001:0> ''.any?
NoMethodError: undefined method `any?' for "":String
	from (irb):1
	from /usr/bin/irb:11:in `<main>'
```

richm: Ah - I see from the test that you expect the code to raise an exception.  I think we should not raise exceptions in this code.
`@tag_key` is defined as an array.  To ensure that it is always an array, do the following:
```
 config_param :tag_key, :array, default: []
```
Why are you using `any?` here?  Are you trying to see if the array is not empty and the record is not empty?  Then just use `empty?`:
```
         if !@tag_key.empty? && !record.empty?

Comment (nhosoi) [reply fluentd/out_syslog.rb]: Thank you, @richm !!  I've pushed the new patch as you suggested.

File: fluentd/out_syslog.rb [hunk 3]
```diff
@@ -99,7 +99,7 @@ def emit(tag, es, chain)
         end
         @packet.time = time
         @packet.tag = @random_string
-        if @tag_key.any? && record.any?
+        if !@tag_key.empty? && record.any?
```
richm: What are you trying to determine about `record` here?  Are you trying to test to see if record has any fields in it?  If so, you should use `!record.empty?` instead.

nhosoi: Thanks, @richm !  Pushed a new patch...

test/remote-syslog.sh
```diff
@@ -415,3 +415,70 @@ fi
...
+    artifact_log failed - not found $mymessage
+fi
+os::cmd::try_until_text "sudo egrep \"${mymessage}$\" /var/log/messages" ".*${mymessage}.*" $MUX_WAIT_TIME
+artifact_log Log test message by logger: $mymessage
+sudo egrep "${mymessage}$" /var/log/messages 2>&1 | artifact_out || :
+
+oc logs $mypod 2>&1 | artifact_out || :
+no_tag_key_log=$( mktemp )
+oc logs $mypod > $no_tag_key_log 2>&1 || :
+grep NoMethodError $no_tag_key_log > /dev/null || :
```
richm: calling anything then doing `|| :` will always set `$?` to 0:
```
>foo() { return 1 ; }
>foo
>echo $?
1
>foo || :
>echo $?
0
```
What should this test do?  Is it sufficient to look for the message in elasticsearch and in /var/log/mesages?

nhosoi:
Had to solve a conflict...

PR #1136, opened 2018-05-04
ping doesnt return False, it throws exception

josefkarasek:
This was a surprise, but what can we do..

http://elasticsearch-py.readthedocs.io/en/2.4.0/api.html?highlight=ping#elasticsearch.Elasticsearch.ping

### DIFF
--- a/curator/src/run_cron.py
+++ b/curator/src/run_cron.py
@@ -67,8 +67,11 @@ class CuratorCronJob():
         until_next_retry = 1
         es = self.get_es_client()
         while True:
-            if es.ping():
-                break
+            try:
+                if es.ping():
+                    break
+            except elasticsearch.ElasticsearchException as err:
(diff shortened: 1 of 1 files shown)


josefkarasek:
@jcantrill can u pls merge asap so qe can have image tomorrow?

josefkarasek:
@richm we don't have any wrappers around curator in es5.x.
Also curator can not fall into a crashloop since it runs as a cron job. If one run fails, it will be scheduled for another run based on the cron job schedule (every 24 hours).

We can use [pod restrart policy](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy) to control how many retries will be tried before giving up. Generally it is not advised to retry failed jobs to infinity as the overhead for creating a pod is higher than our infinite while loop we have in 2.x development

PR #1153, opened 2018-05-10
bump dist-git dockerfiles

jcantrill:
This PR:
* bumps dist-git to 5.6.9
* Fixes the Dockerfile to rely on the look-aside cache

cc @adammhaile My concern is the FROM in downstream was changes to a non scl image which means even if the build succeeds I do not have confidence it will run.

### DIFF
--- a/curator/Dockerfile
+++ b/curator/Dockerfile
@@ -1,4 +1,13 @@
-FROM rhel7.3:7.3-released
+#oit## This file is managed by the OpenShift Image Tool: https://github.com/openshift/enterprise-images
+#oit## by the OpenShift Continuous Delivery team (#aos-cd-team on IRC).
+#oit## 
+#oit## Any yum repos listed in this file will effectively be ignored during CD builds.
+#oit## Yum repos must be enabled in the oit configuration files.
+#oit## The content of this file is managed from an external source.
+#oit## Changes made directly in distgit will be lost during the next
+#oit## reconciliation process.
+#oit## 
+FROM rhel7:7-released
 
 MAINTAINER OpenShift Development <dev@lists.openshift.redhat.com>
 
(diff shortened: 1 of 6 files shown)


curator/Dockerfile
```diff
@@ -17,22 +26,17 @@ ENV HOME=/opt/app-root/src \
...
     fi
 
-RUN yum-config-manager --enable rhel-7-server-ose-3.8-rpms \
-                       --enable rhel-7-server-extras-rpms && \
-    INSTALL_PKGS="elasticsearch-curator-5.2.0-1.x86_64.rpm \
-                  python2-ruamel-yaml" && \
+ADD elasticsearch-curator-5.2.0-1.x86_64.rpm /tmp/ocp/
+
+RUN yum install -y --setopt=tsflags=nodocs /tmp/ocp/elasticsearch-curator-5.2.0-1.x86_64.rpm && \
+    INSTALL_PKGS="python2-ruamel-yaml" && \
```
richm: I think you still need the
```
RUN yum-config-manager --enable rhel-7-server-ose-3.10-rpms \
--enable rhel-7-server-extras-rpms
```
in order to install python2-ruamel-yaml

elasticsearch/Dockerfile
```diff
@@ -23,31 +32,27 @@ ENV ES_CLOUD_K8S_VER=5.6.9 \
...
     fi
 
-RUN yum-config-manager --enable rhel-7-server-ose-3.8-rpms \
-                   --enable rhel-7-server-extras-rpms  && \
-    packages="elasticsearch-${ES_VER}.rpm \
-              java-${JAVA_VER}-openjdk-headless \
+ADD *.rpm /tmp/ocp/
+ADD *.zip /tmp/lib/
+
+RUN yum install -y --setopt=tsflags=nodocs /tmp/ocp/*.rpm && \
```
richm: same here - where will PyYAML and zip and java come from

curator/Dockerfile
jcantrill: Per info from @adammhaile, OSBS ignores those repo enablements and provides its own during the build process.  They dont provide value unless building outside of the build system.  The LOCAL_REPO allowed me to build locally by providing my own repo file.   I was able to make a scratch build of curator using this dockerfile without that statement.

richm:
/retry
major errors in many tests - not sure why

PR #1162, opened 2018-05-14
Set log level on module loggers instead of on root logger

josefkarasek:
The idea was to inherit log level from root logger but that doesn't seem to work.


richm:
If this code is applicable to 3.7, please cherry pick for 3.7

PR #1173, opened 2018-05-17
sync downstream with origin

jcantrill:
This PR brings in the changes from downstream that are needed to make build succeed.


jcantrill:
merging manually as nothing about this change will affect the test

PR #1177, opened 2018-05-18
Relies on single SG index 

jcantrill:
This PR pulls in the multi-tenant plugin that relies on a single SG index


jcantrill:
Seems to work best with https://github.com/openshift/openshift-ansible/pull/8589 which introduces headless discovery

jcantrill:
[(3) Implement Single Searchguard Index [epic-ois-es5]](https://trello.com/c/d3Fadd6D/557-3-implement-single-searchguard-index-epic-ois-es5)

docs/checking-efk-health.md
```diff
@@ -56,6 +56,32 @@ especially if the SearchGuard stack trace is repeated multiple times:
 [ERROR][com.floragunn.searchguard.auth.BackendRegistry] Not yet initialized
 ```
 
+The following error is caused from using a headless services to provide Elasticsearch node discovery.  This log message is present when an Elasticsearch node boots and attempts to discover the
+other members of the cluster before the Kublet is able to publish all the pod endpoints.  This is an acceptible message unless the cluster is not able to elect a master or form a quorum.
```
ewolinetz: Should we be specific and mention that the message "reason: zen-disco-elected-as-master" means that a quorum was able to be formed?

jcantrill: open to suggestions here.  If you have a specific message to add we probably should

ewolinetz: I think something along the lines of 
"This is an acceptable message unless the cluster is not able to elect a master of form a quorum. The election of a master is indicated by a mesage like `[INFO ][o.e.c.s.ClusterService   ] [logging-es-data-master-x6fia9t2] new_master {logging-es-data-master-x6fia9t2}{fEQ6fBt0QuGIC_JyPlLw_A}{GNaZBu9VS_OQcpkOBskfSw}{10.128.0.35}{10.128.0.35:9300}, reason: zen-disco-elected-as-master`"

jcantrill:
@ewolinetz left the discover plugin as it looks like the dependent ansible config isnt there yet

PR #1181, opened 2018-05-21
Fix up for downstream

jcantrill:
This PR fixes some issues in the issues built by @adammhaile where:

* hostname is missing from the base image
* The probe is checking for SG index

### DIFF
--- a/elasticsearch/probe/readiness.sh
+++ b/elasticsearch/probe/readiness.sh
@@ -19,10 +19,8 @@
 # TODO: try re-use code from ./run.sh
 ES_REST_BASEURL=https://localhost:9200
 EXPECTED_RESPONSE_CODE=200
-EXPECTED_SG_DOCUMENT_COUNT=5
 secret_dir=/etc/elasticsearch/secret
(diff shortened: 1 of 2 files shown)


elasticsearch/probe/readiness.sh
```diff
@@ -43,21 +41,3 @@ function check_if_ready() {
 }
 
 check_if_ready "/" "Elasticsearch node is not ready to accept HTTP requests yet"
-check_if_ready "/.searchguard.$DC_NAME" "Searchguard index '.searchguard.$DC_NAME' is missing in ES cluster"
```
ewolinetz: we're removing this because we're going to move to single SG index pattern?

jcantrill: We are removing it because it was never propagated to the ES5 branch.  We decided to significantly simply the readiness probe some time ago because all the checks we devised interfered with pod going into a readiness state.  I saw this behavior when testing the images produced downstream by @adammhaile but I don't know why we don't see issues in upstream CI.  It could be a timing issue

PR #1182, opened 2018-05-22
only make changes to curator when it is stopped

richm:


richm:
this should help with some of the curator test flakes we've been seeing

PR #1183, opened 2018-05-22
add fluentd unit tests

jcantrill:
This PR provides a 'proof' of concept of how we might utilize travis to 'unit test' parts of the logging stack.  The basic assumption here:
* Early testing of dependencies (e.g. start scripts)
* No cluster dependency (or use of mocks)

It provides a bases by which we can refactor if needed when unit tests are in place

If we desire to move forward we may wish to hold merging this until master is open for 3.11

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,11 @@
-sudo: required
+matrix:
+  include:
+    - env: NAME='fluentd - generate_throttle_configs test'
+      language: ruby
+      rvm:
+      - 2.0
+      before_script:
+        cd fluentd/lib/generate_throttle_configs
+      script:
+        rake test
 
-language: ruby
-
-services:
-  - docker
-
-script:
-  - cd hack && PREFIX="openshift/origin-" ./build-images.sh
--- a/fluentd/Dockerfile
+++ b/fluentd/Dockerfile
@@ -49,7 +49,8 @@ ADD configs.d/ /etc/fluent/configs.d/
 ADD filter_k8s_meta_for_mux_client.rb /etc/fluent/plugin/
 ADD out_syslog_buffered.rb out_syslog.rb out_rawtcp.rb /etc/fluent/plugin/
 ADD parser_viaq_docker_audit.rb viaq_docker_audit.rb /etc/fluent/plugin/
-ADD run.sh generate_throttle_configs.rb generate_syslog_config.rb ${HOME}/
+ADD run.sh generate_syslog_config.rb ${HOME}/
+ADD lib/*/lib/*.rb ${HOME}/
 
 RUN mkdir -p /etc/fluent/configs.d/{dynamic,user} && \
     chmod 777 /etc/fluent/configs.d/dynamic && \
--- a/fluentd/Dockerfile.centos7
+++ b/fluentd/Dockerfile.centos7
@@ -52,7 +52,8 @@ ADD configs.d/ /etc/fluent/configs.d/
 ADD filter_k8s_meta_for_mux_client.rb /etc/fluent/plugin/
 ADD out_syslog_buffered.rb out_syslog.rb out_rawtcp.rb /etc/fluent/plugin/
 ADD parser_viaq_docker_audit.rb viaq_docker_audit.rb /etc/fluent/plugin/
-ADD run.sh generate_throttle_configs.rb generate_syslog_config.rb ${HOME}/
+ADD run.sh generate_syslog_config.rb ${HOME}/
+ADD lib/*/lib/*.rb ${HOME}/
 
 RUN mkdir -p /etc/fluent/configs.d/{dynamic,user} && \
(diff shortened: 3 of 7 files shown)


.travis.yml
```diff
@@ -1,9 +1,17 @@
...
 services:
   - docker
 
-script:
-  - cd hack && PREFIX="openshift/origin-" ./build-images.sh
+matrix:
+  include:
+    - language: ruby
+      script:
+      - cd hack && PREFIX="openshift/origin-" ./build-images.sh
```
richm: I don't think we need build images here

jcantrill: its a sanity check... but we do that in our CI anyway and could remove.

jcantrill:
cc @openshift/sig-integration-services

.travis.yml
richm: yeah - I think having image build here runs counter to being able to run the unit tests quickly

ewolinetz:
Is there a particular reason why we're no longer using `log` as an instance var and instead passing it around, is this a travis limitation?

ewolinetz:
failing CI resolved by https://github.com/openshift/openshift-ansible/pull/8482

jcantrill:
@ewolinetz The only reason is that when introducing the unit tests `@log` failed.  I presume it was some sort of scoping between running as a script vs unit testing.  Ideally, I would like to expand upon these changes and maybe introduce a class or to which would allow us to not have to pass the logger around

nhosoi:
The following failure was logged in `https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_origin-aggregated-logging/1183/test_pull_request_origin_aggregated_logging_journald/540/artifacts/scripts/entrypoint/artifacts/remote-syslog-logging-fluentd-sv5n5.log`.  I don't understand why "log.debug" is not allowed here...
```
generate_throttle_configs.rb:163:in `block in create_default_docker': undefined method `debug' for #<Hash:0x00000000da6268> (NoMethodError)
	from generate_throttle_configs.rb:162:in `open'
	from generate_throttle_configs.rb:162:in `create_default_docker'
	from generate_throttle_configs.rb:274:in `generate_throttle_configs'
	from generate_throttle_configs.rb:284:in `<main>'
```

jcantrill:
@ewolinetz @richm @nhosoi Is this something we wish to pursue?  Do we feel it provides value?  I find some of our scripts and supporting code monolithic wiht no quick tests to be executed in a 'unit' test fashion.  My preference would be to do similar refactoring where possible to expose changes sooner then spinning up an entire cluster

nhosoi:
@jcantrill , I think this is a great addition.  I wonder this unit test could be run at the customer sites when they manually change their fluentd config files to give some guidance/assurance to them?

jcantrill:
@nhosoi they could its just a matter of executing the right command.  if we dont identify it in the docs we should

fluentd/Dockerfile
```diff
@@ -49,7 +49,8 @@ ADD configs.d/ /etc/fluent/configs.d/
 ADD filter_k8s_meta_for_mux_client.rb /etc/fluent/plugin/
 ADD out_syslog_buffered.rb out_syslog.rb out_rawtcp.rb /etc/fluent/plugin/
 ADD parser_viaq_docker_audit.rb viaq_docker_audit.rb /etc/fluent/plugin/
-ADD run.sh generate_throttle_configs.rb generate_syslog_config.rb ${HOME}/
+ADD run.sh generate_syslog_config.rb ${HOME}/
+ADD lib/*/lib/*.rb ${HOME}/
```
ewolinetz: What is this adding? I don't see any `lib/*/lib/` directories.

fluentd/Dockerfile.centos7
```diff
@@ -52,7 +52,8 @@ ADD configs.d/ /etc/fluent/configs.d/
 ADD filter_k8s_meta_for_mux_client.rb /etc/fluent/plugin/
 ADD out_syslog_buffered.rb out_syslog.rb out_rawtcp.rb /etc/fluent/plugin/
 ADD parser_viaq_docker_audit.rb viaq_docker_audit.rb /etc/fluent/plugin/
-ADD run.sh generate_throttle_configs.rb generate_syslog_config.rb ${HOME}/
+ADD run.sh generate_syslog_config.rb ${HOME}/
+ADD lib/*/lib/*.rb ${HOME}/
```
ewolinetz: same question as above

fluentd/Dockerfile
jcantrill: Its this file move from root: https://github.com/openshift/origin-aggregated-logging/pull/1183/files#diff-0ae11326c532c900e8e799ba61a2bec6

ewolinetz: Thanks, I didn't look close enough at the destination paths I guess

PR #1187, opened 2018-05-25
better describe error situations

josefkarasek:


josefkarasek:
As much as I don't see [this](https://bugzilla.redhat.com/show_bug.cgi?id=1581575) being an actual issue, I propose the following changes:
* steal the `'elasticseach'` logger and force it to log only `ERROR` messages (that's the only way how to configure the log level for this particular library)
* don't make the level configurable as we only use it for the `ping` operation.

```
ERROR:__main__:Connection to elasticsearch at [localhost:9200] failed. Number of failed retries: 1
ERROR:__main__:Connection to elasticsearch at [localhost:9200] failed. Number of failed retries: 2
ERROR:__main__:Connection to elasticsearch at [localhost:9200] failed. Number of failed retries: 3
ERROR:__main__:Connection to elasticsearch at [localhost:9200] failed. Number of failed retries: 4
```

josefkarasek:
@richm can you merge please? Thank you

PR #1189, opened 2018-05-29
mapper.allow_dots_in_name setting not needed (supported) in ES 5.x

### LINKED ISSUE #1188: Remove mapper.allow_dots_in_name settings for ES 5.x
I think we should remove (_cleanup_) the setting from Docker file(s):
```
ES_JAVA_OPTS="-Dmapper.allow_dots_in_name=true"
```
It is supported OOB in ES 5.x and later.

Some references:
- <https://www.elastic.co/guide/en/elasticsearch/reference/2.4/dots-in-names.html>
- <https://discuss.elastic.co/t/is-there-a-way-to-allow-dots-in-the-fieldname/90134>

Thoughts?

lukas-vlcek:
Closes #1188

### DIFF
--- a/elasticsearch/Dockerfile
+++ b/elasticsearch/Dockerfile
@@ -17,7 +17,6 @@ USER 0
 
 ENV ES_CLOUD_K8S_VER=5.6.9.2 \
     ES_CONF=/etc/elasticsearch/ \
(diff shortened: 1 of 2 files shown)


lukas-vlcek:
Do we need to

/retest

again?

PR #1190, opened 2018-05-29
Test to ensure documents are indexed properly

niconosenzo:
https://bugzilla.redhat.com/show_bug.cgi?id=1575662#c5

### DIFF
--- a/test/indexing_after_ns_removal.sh
+++ b/test/indexing_after_ns_removal.sh
@@ -1,0 +1,76 @@
+#!/bin/bash
+
+# This is a test suite for checking the index mapping after the project has
+# been deleted and another namespaces has been created with the same name. 
+
+source "$(dirname "${BASH_SOURCE[0]}" )/../hack/lib/init.sh"
+source "${OS_O_A_L_DIR}/hack/testing/util.sh"
+os::test::junit::declare_suite_start "test/indexing_after_ns_removal"
+
+
+create_pod(){
+# keep indentation clean
+    oc create -f - <<EOF &>/dev/null
+apiVersion: v1
+kind: Pod
+metadata:
+  name: bar
+  namespace: $NS
+spec:
+  containers:
+  - image: gcr.io/google_containers/busybox:latest
+    name: bar
+    command: [ "/bin/sh", "-c", "echo foo-bar $1" ]
+  restartPolicy: Never
+EOF
+}
+
+
+cleanup(){
+    local return_code="$?"
+
+    # delete all the $NS indices
+    curl_es $espod /project.$NS.* -XDELETE || :
+
+    # this will call declare_test_end, suite_end, etc.
+    os::test::junit::reconcile_output
+
+    exit $return_code
+}
+trap "cleanup" EXIT
+
+
+NS="pelle-foo"
+espod=$( get_es_pod es )
+
+# Create a namespace named pelle-foo containing a pod, which logs a line and
+# then remove the namespace, repeating this procedure 20 times
+for i in $( seq 1 20 ); 
+do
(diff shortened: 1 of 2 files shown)


richm:
@niconosenzo Does the latest failure mean the bug has not been fixed in the actual code?  That is, this isn't a test failure, it is the test correctly reporting a problem in the logging code?

test/indexing_after_ns_removal.sh
```diff
@@ -64,13 +75,38 @@ done
...
+# the indices.
+
+for index in $(curl_es $es_pod /_cat/indices?h=index | grep pelle-foo)
+do
+    namespace_uid=$(curl_es $es_pod /"${index}"/_search?q=kubernetes.namespace_id:* | jq . | grep -i namespace_id | awk -F'"' '{ print $4 }')
+    grep -m 1 "${namespace_uid}" temp_ns_uid &>/dev/null
+    [ "$?" != "0" ] && os::log::error ${namespace_uid} does not match any UID collected from the actual namespace list during namespace creation
+
+    pod_uid=$(curl_es $es_pod /"${index}"/_search?q=kubernetes.namespace_id:* | jq . | grep -i pod_id | awk -F'"' '{ print $4 }')
+    grep -m 1 "${pod_uid}" temp_ns_uid &>/dev/null
```
nhosoi: s/temp_ns_uid/temp_pod_uid/ ???

niconosenzo: 'Yes' to both of your comments.

richm:
@jcantrill This test failure means the fluent-plugin-kubernetes is not handling the namespace create/delete/create case correctly - i.e. there is still a bug in that plugin.

test/indexing_after_ns_removal.sh
```diff
@@ -64,13 +75,38 @@ done
...
+    [ "$?" != "0" ] && os::log::error ${namespace_uid} does not match any UID collected from the actual namespace list during namespace creation
+
+    pod_uid=$(curl_es $es_pod /"${index}"/_search?q=kubernetes.namespace_id:* | jq . | grep -i pod_id | awk -F'"' '{ print $4 }')
+    grep -m 1 "${pod_uid}" temp_pod_uid &>/dev/null
+    [ "$?" != "0" ] && os::log::error ${pod_id} does not match any UID collected from the actual pod list during pod creation
+done
+
+
+# .orphaned index  is not created by default, so if it exists, then it contains records
+[[ $(curl_es $es_pod /_cat/indices?h=index,docs.count | grep .orphaned) ]] && os::log::error  There are records in the .orphaned index.
```
nhosoi: Found one issue.  This last line returns 1 if .orphaned is not found.  That is, if successful, this test ends with (failure == 1); if .orphaned is found, (success == 0).
```
$ [[ $(cat /etc/hosts | grep .orphaned) ]] && echo "found"
$ echo $?
1
$ [[ $(cat /etc/hosts | grep localhost) ]] && echo "found"
found
$ echo $?
0
```
Can we do "if .orphaned is found, print error; exit 1; else exit 0"?

niconosenzo:
- Removed es_pod references (it was returning  non-zero exit code)
- Orphan indices check fixed.
- Fixed main while loop, now the it waits till pod is "Completed" to remove it.
- Adding "|| true" to uid matching loop to avoid failing early

test/indexing_after_ns_removal.sh
niconosenzo: @nhosoi Done, did some tests and throwing the expected result now.

jcantrill:
Closing as stale and most of these tests have since moved into the operator(s) code base

PR #1192, opened 2018-05-30
yacht - new tool for pre-emptive pre-creation of es indices

josefkarasek:
What do you think, guys?

It's initial version, let's make it better!


jcantrill:
I would prefer to not introduce yet another component considering we have talked about pushing this functionality possibly into the operator.  Could this not be a 'mode' of the curator image where we could define a separate kube cronjob to run in 'precreate' instead of 'delete' mode?

josefkarasek:
@jcantrill @portante @richm Let's figure out if it meets our functional requirements first.
* Is it robust enough?
* Is it configurable enough?
* Does it use long-living es connections well?
* Does it solve our performance issues?

Let's work on answering the above first. Burn energy on figuring out deployment strategy once we know we actually want it.

richm:
I think adding a new image for this functionality is going to add a tremendous amount of work both upstream and downstream . . .

josefkarasek:
@richm there are no tests for yacht, yet

yacht/lib/yacht.py
```diff
@@ -0,0 +1,121 @@
...
+        self.retry_period = int(os.getenv('RETRY_PERIOD_SECONDS', '5'))
+        self.index_source = IndexSource()
+        self.es = CuratorEsClientFactory.curator_es_client(False, 'ERROR')
+        self.es_util = EsUtil(self.es)
+
+    def create(self):
+        self.es_util.server_running()
+        self.server_busy()
+        indices_tomorrow = self.index_source.indices_tomorrow()
+        for i in range(len(indices_tomorrow)):
```
portante: Consider using `enumerate`, might be a bit more pythonic.  Consider calculating length of indices once, as well.

yacht/lib/yacht.py
```diff
@@ -0,0 +1,121 @@
+import elasticsearch
+import os
+import datetime
+import time
+from esclientfactory import EsClientFactory
+from curatoresclientfactory import CuratorEsClientFactory
+from esutil import EsUtil
+from util import create_logger
+
+class IndexSource():
```
portante: Consider using `class IndexSource(object):` for all your python classes.

yacht/lib/yacht.py
```diff
@@ -0,0 +1,121 @@
...
+        try:
+            current_queue = self.es.cluster.health()['task_max_waiting_in_queue_millis']
+        except elasticsearch.ElasticsearchException as err:
+            self.logger.error(err)
+        return current_queue
+
+class Yacht():
+
+    def __init__(self):
+        self.logger = create_logger(__name__,  os.getenv('YACHT_LOG_LEVEL', 'DEBUG'))
```
portante: Consider moving `os.getenv()` to be parameters to the constructor for Yacht.

yacht/lib/curatoresclientfactory.py
```diff
@@ -0,0 +1,15 @@
+import os
+from esclientfactory import EsClientFactory
+
+class CuratorEsClientFactory():
```
portante: Consider moving this to the `Yacht()` class constructor and passing the ES object to source and create objects.

PR #1193, opened 2018-06-01
bug 1584665. Remove prometheus plugin

jcantrill:
Removing this change in 3.10 as it is causing memory issues and is not targeted for 3.10 anyway.


jcantrill:
@dhananjaysathe I don't have any more information at this time other then we ran some short soak tests using our OOTB setting (512M I believe) and saw various OOM kills.  To be fair, this also co-incited with an introduction of a buffered retry queue which also contributed to the memory growth.  Removing the prometheus plugin, the container memory limit, and soak testing for a longer period demonstrated steady state in our tests around 512M which leads us to think we should default to something like 756M.  It could be that with the addition of the prometheus plugin we never found the new steady state.  Also note we are running on ruby 2.0 where there has been some measurable memory savings using ruby 2.4.  We are limited by the version available in our production releases so the newer version of ruby may also resolve some of these issues.

jcantrill:
[(5) Prometheus plugin for fluentd[logging][epic-diagnostics]](https://trello.com/c/uR6C3LHC/578-5-prometheus-plugin-for-fluentdloggingepic-diagnostics)

PR #1194, opened 2018-06-01
bug 1582232. Disable shard allocation for searchguard

jcantrill:
This PR resolves https://bugzilla.redhat.com/show_bug.cgi?id=1582232 by:

* disabling shard allocation for the searchguard index
* updates the replica count to 0
* modifies the shard setting to peg the index to the node

### DIFF
--- a/elasticsearch/utils/es_seed_acl
+++ b/elasticsearch/utils/es_seed_acl
@@ -2,18 +2,11 @@
 
 source "logging"
 
-MAX_WAIT_SEED=${MAX_WAIT_SEED:-$((60*60*24*7))} #one week
-
-info "Seeding the searchguard ACL index.  Will wait up to $MAX_WAIT_SEED seconds."
-
 config=${ES_CONF}/elasticsearch.yml
 index=$(python -c "import yaml; print yaml.load(open('${config}'))['searchguard']['config_index_name']" | xargs -i sh -c "echo {}")
 
-now=0
-numretries=0
-loginterval=${SG_ADMIN_LOG_INTERVAL:-100}
-while ! ${ES_HOME}/plugins/openshift-elasticsearch/sgadmin.sh \
-    -cd ${HOME}/sgconfig \
+function sgadmin {
+  ${ES_HOME}/plugins/openshift-elasticsearch/sgadmin.sh \
     -i $index \
     -ks /etc/elasticsearch/secret/searchguard.key \
     -kst JKS \
@@ -22,7 +15,18 @@ while ! ${ES_HOME}/plugins/openshift-elasticsearch/sgadmin.sh \
     -tst JKS \
     -tspass tspass \
     -nhnv \
-    -icl ${DEBUG:+-dg} && [ ${now} -lt ${MAX_WAIT_SEED} ]
+    -icl \
+    ${DEBUG:+-dg} ${@:-}
+}
+
+MAX_WAIT_SEED=${MAX_WAIT_SEED:-$((60*60*24*7))} #one week
+
+info "Seeding the searchguard ACL index.  Will wait up to $MAX_WAIT_SEED seconds."
+
+now=0
+numretries=0
+loginterval=${SG_ADMIN_LOG_INTERVAL:-100}
+while ! sgadmin -cd ${HOME}/sgconfig && [ ${now} -lt ${MAX_WAIT_SEED} ]
 do
     if ! expr $numretries % $loginterval ; then
         warn "Error seeding the searchguard ACL index... retrying in 10 seconds - $numretries retries so far"
@@ -39,3 +43,12 @@ if [ ${now} -ge ${MAX_WAIT_SEED} ]; then
     exit 1
 fi
 info "Seeded the searchguard ACL index"
+
+info "Disabling auto replication"
+sgadmin -dra
+
+info "Updating replica count to 0"
+sgadmin -us 0
+
+info "Setting filters to allocate shard to this node only"
+es_util --query="$index/_settings" -XPUT -d "{\"index.routing.allocation.include._name\": \"$index\"}"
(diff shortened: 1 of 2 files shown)


elasticsearch/utils/set_sg_replicas
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash -e
+
+source "logging"
+
+info "Modifying the Searchguard index replica settings."
+
+config=${ES_CONF}/elasticsearch.yml
+index=$(python -c "import yaml; print yaml.load(open('${config}'))['searchguard']['config_index_name']" | xargs -i sh -c "echo {}")
+
+info "Disabling auto replication"
```
ewolinetz: I think these changes should be made in `utils/es_seed_acl` instead

jcantrill: I prefer not.  I find it useful to be able to manually execute seeding.  The current impl allows us to run them separately.

ewolinetz: Right, shouldn't we always restrict our seeding to not use dynamic replica scaling? we can also default the number of replicas to 0 and allow a variable to configure that. We currently have docs that guide users to run `es_seed_acl` if necessary, they would be in an unfavorable state then.

jcantrill: I believe we are confusing ACL seeding vs cluster initialization.  We should always be able to seed the ACLs. I would expect only once to needing to really initialize the cluster and characteristics of the Searchguard index

portante: Searchguard indices, once created, have persistent state; that is to say that creating the .searchguard indices are part of cluster initialization.

However, as more ES DCs are created, each new DC will need its cluster initialization step run as well.

elasticsearch/utils/set_sg_replicas
```diff
@@ -0,0 +1,39 @@
...
+    -i $index \
+    -ks /etc/elasticsearch/secret/searchguard.key \
+    -kst JKS \
+    -kspass kspass \
+    -ts /etc/elasticsearch/secret/searchguard.truststore \
+    -tst JKS \
+    -tspass tspass \
+    -nhnv \
+    -icl \
+    -us 0 ${DEBUG:+-dg}
```
portante: Why can't we issue both sgadmin.sh options, `-us 0 -dra` in one sgadmin.sh script invocation?

elasticsearch/utils/set_sg_replicas
```diff
@@ -0,0 +1,39 @@
...
+
+info "Setting filters to allocate shard to this node only"
+es_util --query="$index/_settings" -XPUT -d "{\"index.routing.allocation.include._name\": \"$index\"}"
```
portante: Before we remove the replicas, shouldn't we first route the primary shard to the same node and then fix up the replicas?

ewolinetz:
Opened https://github.com/floragunncom/search-guard/issues/500 so we can eventually condense this into a single call hopefully

elasticsearch/utils/set_sg_replicas
ewolinetz: In the code, the will only update one of the operations and then exit. Its required to be two calls at this point. I've created an issue requesting the ability to do this.

elasticsearch/utils/es_seed_acl
```diff
@@ -39,3 +43,12 @@ if [ ${now} -ge ${MAX_WAIT_SEED} ]; then
     exit 1
 fi
 info "Seeded the searchguard ACL index"
+
+info "Disabling auto replication"
+sgadmin -dra
```
portante: For each of the following operations, what happens if they fail?  The user will see errors in the logs, but should they be concerned or not?

Seems like we need to direct them to find out why the `sgadmin` script failed to apply those changes, or why the `es_util` command failed, and then re-run this script?

elasticsearch/utils/es_seed_acl
```diff
@@ -39,3 +43,12 @@ if [ ${now} -ge ${MAX_WAIT_SEED} ]; then
...
+
+info "Updating replica count to 0"
+sgadmin -us 0
+
+info "Setting filters to allocate shard to this node only"
+es_util --query="$index/_settings" -XPUT -d "{\"index.routing.allocation.include._name\": \"$index\"}"
```
portante: Should `include` be `require`?

jcantrill: I source this information from the ES docs which did not offer an alternative.  Given we anticipate using this for the SG index that is based on the DC name, will this not fundamentally be the same?

elasticsearch/utils/es_seed_acl
jcantrill: If we can seed the ACL's I don't see why any of these steps would fail.  Given neither of these operations is drastically important IMO, I don't know how much thought we wish to put into error control nor if we have have a reasonable understanding of what they are or how to react.

elasticsearch/utils/es_seed_acl
portante: @jcantrill, do you mean this page?  https://www.elastic.co/guide/en/elasticsearch/reference/2.4/shard-allocation-filtering.html

elasticsearch/utils/es_seed_acl
portante: Checking for an error and emitting a message regarding the ramifications seems pretty easy to do.  The `sgadmin` tool is already going to emit message messages which are confusing.

richm:
remote-syslog flake
/test logging

richm:
remote-syslog flake
/test logging

richm:
remote-syslog flake
/test logging

I don't think this is a flake - it is very persistent - but I don't understand why this particular change causes this particular issue . . . and only in the journald case, not the json-file case

richm:
/retest
use latest remote-syslog debugging code

PR #1195, opened 2018-06-01
fix logs util to retrieve logs from persistent dir.  Ref #1108

jcantrill:
This fixes the log utility to get logs from /elasticsearch/persistent ref #1108


richm:
/test logging
remote-syslog flake

PR #1201, opened 2018-06-05
add script to provision machine and run tests

richm:


nhosoi:
/lgtm

Thanks, @richm !!

richm:
@jcantrill I think this should be merged anyway, if folks think it is useful, even if the tests fail.

hack/get-machine-run-tests.sh
```diff
@@ -0,0 +1,580 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+usage() {
+    local bn=$( basename $0 )
+    echo Usage: $0
+    echo Assumes you have $HOME/origin-aggregated-logging and $HOME/openshift-ansible
```
jcantrill: Can we not modify this to be a var other then $HOME.  I, for example, use $HOME/git

hack/get-machine-run-tests.sh
```diff
@@ -0,0 +1,580 @@
+#!/bin/bash
+
```
jcantrill: Can we add a comment here describing the purpose.

PR #1202, opened 2018-06-05
convert remote-syslog test to use standard test func; get more info when entries are not in es

richm:
convert remote-syslog to use wait_for_fluentd_to_catch_up
when wait_for_fluentd_to_catch_up fails, perform a search on
the index in question to get all records since the
function started
change entrypoint.sh to put each test in a separate LOG_DIR
and ARTIFACT_DIR


richm:
the main point of this commit is so that when wait_for_fluentd_to_catch_up fails and says the count of the matching records is 0, we can try to determine what actually is in elasticsearch

PR #1203, opened 2018-06-06
fix test failures

richm:


nhosoi:
> first version had syntax error, added when I added the usage() function

Yeah, I had to put double quotes around the echo'ed line to execute the script. :)  But your fix looks much smarter.

/lgtm

PR #1212, opened 2018-06-14
Allow separate config options for how to handle buffers from app and …

wshearn:
…ops logs

### DIFF
--- a/fluentd/configs.d/openshift/output-es-ops-config.conf
+++ b/fluentd/configs.d/openshift/output-es-ops-config.conf
@@ -27,9 +27,9 @@
       disable_retry_limit true
       buffer_type file
       buffer_path '/var/lib/fluentd/buffer-output-es-ops-config'
-      buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT'] || '32' }"
-      buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT'] || '8m' }"
-      buffer_queue_full_action "#{ENV['BUFFER_QUEUE_FULL_ACTION'] || 'block'}"
+      buffer_queue_limit "#{ENV['OPS_BUFFER_QUEUE_LIMIT'] || #{ENV['BUFFER_QUEUE_LIMIT'] || '32' }"
+      buffer_chunk_limit "#{ENV['OPS_BUFFER_SIZE_LIMIT'] || #{ENV['BUFFER_SIZE_LIMIT'] || '8m' }"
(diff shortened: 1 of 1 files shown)


jcantrill:
@nhosoi please review as these changes would impact the memory calculations in the run script

nhosoi:
> please review as these changes would impact the memory calculations in the run script

@jcantrill , it makes me a bit worry since wouldn't this change make buffer_queue_limit in output-es-ops-config.conf and output-es-config.conf asymmetric?  OPS_BUFFER_QUEUE_LIMIT is new and fluentd/run.sh does not modify the value, so the value is directly set to buffer_queue_limit in output-es-ops-config.conf.  But BUFFER_QUEUE_LIMIT is the calculated value and the user defined BUFFER_QUEUE_LIMIT won't be passed.

That's being said, there could be 2 issues...
1) OPS_BUFFER_QUEUE_LIMIT could be set by a user, while BUFFER_QUEUE_LIMIT would not.
2) Currently, BUFFER_QUEUE_LIMIT is calculated based on the available disk size (if it is smaller than the given FILE_BUFFER_LIMIT).  If given OPS_BUFFER_QUEUE_LIMIT is larger than that, it may cause disk full.

We could change fluentd/run.sh easily for (1), which may introduce the potential disk full issue, though.  Then, we also need to get rid of setting BUFFER_QUEUE_LIMIT in openshift-ansible.

fluentd/configs.d/openshift/output-es-ops-config.conf
```diff
@@ -27,9 +27,9 @@
       disable_retry_limit true
       buffer_type file
       buffer_path '/var/lib/fluentd/buffer-output-es-ops-config'
-      buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT'] || '32' }"
-      buffer_chunk_limit "#{ENV['BUFFER_SIZE_LIMIT'] || '8m' }"
-      buffer_queue_full_action "#{ENV['BUFFER_QUEUE_FULL_ACTION'] || 'block'}"
+      buffer_queue_limit "#{ENV['OPS_BUFFER_QUEUE_LIMIT'] || #{ENV['BUFFER_QUEUE_LIMIT'] || '32' }"
```
richm: `"#{ENV['OPS_BUFFER_QUEUE_LIMIT'] || ENV['BUFFER_QUEUE_LIMIT'] || '32'}"` `"#{ENV['OPS_BUFFER_SIZE_LIMIT'] || ENV['BUFFER_SIZE_LIMIT'] || '8m' }"`
`"#{ENV['OPS_BUFFER_QUEUE_FULL_ACTION'] || ENV['BUFFER_QUEUE_FULL_ACTION'] || 'block'}"`

richm:
remote-syslog flake
/test logging

PR #1210, opened 2018-06-12
[es5.x] Adding test to check for duplicate records with fluentd and journald …

ewolinetz:
…file rollover

Manual cherry-pick of https://github.com/openshift/origin-aggregated-logging/pull/1074

PR #1065, opened 2018-04-05
Bug 1538560 - [RFE]rename the project name mux-undefined

nhosoi:
mux-undefined ==> .orphaned
mux.sh: if there are any tests which create project "mux-undefined"
        just remove that project checking and creation.

PR #1214, opened 2018-06-15
5x into master

jcantrill:
This PR merges ES5.x work into master

cc @portante @t0ffel @nhosoi @lukas-vlcek


portante:
Do we really need to keep all 155 commits listed?  Can't we first push all 155 commits to a branch off of master, and then squash all the commits down to one?

jcantrill:
@portante there is nothing really to review... its a straight merge of 5->master. I dont understand why there are conflicts here because i dont have them locally.  squashing however would remove history

portante:
Can't we preserve the history by preserving the branch, and then squash?

If you don't squash, then the commits get sprinkled through all the other commits that get done at the same time.  If you rebase on top of master, and squash, then you have one commit that points to the point time the code base changed to pull in ES5 support.

Does that seem beneficial?

ewolinetz:
I was actually wondering something similar. Can we rebase `master` back onto the `es5.x` branch to isolate all the 5.x specific changes prior to merging back into `master`? (We have a couple commits above that are where we merged `master` back into `es5.x`)

Going along with Peter's point above, we could then in theory squash all the 5.x specific changes as an all encompassing switch from 2.x to 5.x on master... I'm not sure if we plan to retain the `es5.x` branch after this merge though. So we'd likely lose all that history and I'm not in favor of that happening.

elasticsearch/Dockerfile.centos7
```diff
@@ -15,16 +15,16 @@ ENV ES_CLOUD_K8S_VER=2.4.4_01 \
     INSTANCE_RAM=512G \
     JAVA_VER=1.8.0 \
     NODE_QUORUM=1 \
-    OSE_ES_VER=2.4.4.13 \
-    PROMETHEUS_EXPORTER_VER=2.4.4.0 \
+    OSE_ES_VER=2.4.4.22 \
```
ewolinetz: is this what we have in the es5.x branch?

jcantrill:
@ewolinetz Maybe since es5.x base was master we should rebase on master then merge which will be a FF

elasticsearch/Dockerfile.centos7
lukas-vlcek: I do second @ewolinetz this is looking strange.

docs/metrics.md
```diff
@@ -0,0 +1,121 @@
...
+  <oauth token>     ---------------   <username/passwd>   -----------------
+    Request   ----> | oauth-proxy | ----- Request ------> | Elasticsearch |
+    Response  <---- |             | <---- Respose ------- |               |
+                    ---------------                       -----------------
+```
+### Scrape Rules
+A service is deployed for the metrics endpoint which is [annotated](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging_elasticsearch/tasks/main.yaml#L395) to allow auto-discovery of the
+endpoint by Prometheus.  The scrape rule must be defined such that it uses the Service Account specified during the deployment of the logging stack.
+
+## Fluentd
```
lukas-vlcek: May be I missed something but haven't we decided to drop Prometheus plugin for Fluentd due to some issues?

ewolinetz:
@jcantrill yeah, or maybe we just do a merge like how you had before... 
Also, I tried doing a merge as well and it seems like the ES5.x branch is missing a cherry pick [1]

[1] https://github.com/openshift/origin-aggregated-logging/pull/1210

elasticsearch/lib/elasticsearch-prometheus-exporter/README.md
```diff
@@ -1,168 +0,0 @@
-# Prometheus Exporter Plugin for ElasticSearch
```
ewolinetz: @lukas-vlcek did this never get cherry-picked to es5.x? or do we actually want to remove this here?

ewolinetz:
Just a question regarding prometheus docs, otherwise LGTM

elasticsearch/lib/elasticsearch-prometheus-exporter/README.md
jcantrill: I think the README is possibly valid but the plugin is available without vendoring

hack/build-images.sh
```diff
@@ -16,9 +16,10 @@ if [ "${RELEASE_STREAM:-}" = 'prod' ] ; then
 fi
 dockerfile="Dockerfile${docker_suffix}"
 
+name_suf="5"
```
richm: so do we still want the version 5 images published to e.g. logging-elasticsearch5 instead of logging-elasticsearch?

hack/push-release.sh
```diff
@@ -33,16 +33,16 @@ if [[ -z "${source_tag}" ]]; then
...
-  ${PREFIX}logging-kibana
+  ${PREFIX}logging-elasticsearch5
+  ${PREFIX}logging-kibana5
   ${PREFIX}logging-auth-proxy
   ${PREFIX}logging-eventrouter
 )
 
 curbranch=$( git rev-parse --abbrev-ref HEAD )
-if [[ "${curbranch:-master}" == es5.x ]] ; then
+if [[ "${curbranch:-master}" != master ]] ; then
```
richm: same question as above - we are going to have different named images for the es 5 images?

ewolinetz:
I thought that was what we had agreed on, that we'd use
logging-${component}5

richm:
the big test/mux.sh diff is due to missing this cherry pick: https://github.com/openshift/origin-aggregated-logging/pull/1065

richm:
> I thought that was what we had agreed on, that we'd use logging-${component}5

does that mean we'll need logging-elasticsearch6, logging-elasticsearch7, etc.?

What about downstream?

jcantrill:
@richm I believe the answer is yes, we have to publish to {component}5 and we are required to continue that pattern because of the workflow of our release process.  There has to be a 1:1 relationship between the downstream repo and the published image.  Additionally, I presume we will need to continue to support the 2.x images for bug fixes, etc and there is no way to that without changing the tagging structure.  Unless we conceive of an alternate tagging scheme that is separate for OCP tagging I dont see an alternative.

test/mux.sh
```diff
@@ -367,9 +334,9 @@ cleanup() {
         os::log::debug "$( oc replace --force -f $saveds )"
     fi
     # delete indices created by this test
-    curl_es $es_svc /project.testproj.* -XDELETE
-    curl_es $es_svc /project.default.* -XDELETE
-    curl_es $es_svc /project..orphaned.* -XDELETE
+    curl_es $es_pod /project.testproj.* -XDELETE
```
richm: I think these should be es_svc instead of es_pod

ewolinetz: They need to be, or they should be `curl_es_pod`

elasticsearch/lib/elasticsearch-prometheus-exporter/README.md
jcantrill: added back in

PR #1217, opened 2018-06-20
cannot have empty field name in es5

richm:
change test so record sent to es5 does not have an empty field name


ewolinetz:
we may need to manually merge this

richm:
@jcantrill please merge this to unblock ci

PR #1218, opened 2018-06-20
Fixing date option for hack/testing/util

ewolinetz:


ewolinetz:
this will also need a manual merge until https://github.com/openshift/openshift-ansible/pull/8783 is merged in

PR #1222, opened 2018-06-22
fix downstream to depend on rpm

jcantrill:
This PR productizes downstream Kibana to rely on rpms instead of look aside cache:
* resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1590947
* addresses https://trello.com/c/406ULKGV

### DIFF
--- a/kibana/Dockerfile
+++ b/kibana/Dockerfile
@@ -31,9 +31,7 @@ RUN if [ -n "${LOCAL_REPO}" ] ; then \
      curl -s -o /etc/yum.repos.d/local.repo ${LOCAL_REPO} ; \
     fi
 
-ADD *.rpm /tmp/lib/
-WORKDIR /tmp/lib
-RUN INSTALLED_PKGS="kibana-${KIBANA_VER}-x86_64.rpm" && \
+RUN INSTALLED_PKGS="kibana-${KIBANA_VER}*" && \
     yum install -y --setopt=tsflags=nodocs  ${INSTALLED_PKGS} && \
     yum clean all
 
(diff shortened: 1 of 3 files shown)


jcantrill:
cc @adammhaile

jcantrill:
[(3) Productize kibana [epic-ois-es5]](https://trello.com/c/406ULKGV/684-3-productize-kibana-epic-ois-es5)

kibana/Dockerfile
```diff
@@ -43,8 +41,8 @@ ADD kibana.yml ${KIBANA_CONF_DIR}/
 ADD lib/* ${HOME}/
 ADD run.sh utils install.sh prep-install.${RELEASE_STREAM} ${HOME}/
 ADD logo-OCP-console-hdr-thin.svg ${KIBANA_HOME}/installedPlugins/origin-kibana/public/images/logo-origin-thin.svg
-RUN sh ${HOME}/install.sh
-RUN rm -r /tmp/lib
+RUN (which node || ln -s $(which $NODE_BIN) "/usr/bin/node") && \
```
richm: I've had problems using `which` in the past since it seems to be in an optional package - use `type -p` instead which is built-in to the shell.

kibana/install.sh
```diff
@@ -6,10 +6,10 @@ set -o pipefail
...
 source "${HOME}/prep-install.${RELEASE_STREAM}"
 
-ORIGIN_KIBANA_PLUGIN=$(ls -t1 ${HOME}/origin-kibana-v$KIBANA_VER-*.zip | head -1)
+ORIGIN_KIBANA_PLUGIN=$(ls -t1 ${HOME}/origin-kibana-v${KIBANA_VER}*.zip | head -1)
 ${KIBANA_HOME}/bin/kibana-plugin install file://${ORIGIN_KIBANA_PLUGIN}
 
 chmod -R og+w "${HOME}"
 chmod -R og+rw "${KIBANA_HOME}"
-chmod -R og+rw /var/lib/kibana
+mkdir -m og+w -p /var/lib/kibana ||
```
richm: Why would this fail?  that is, why do you need the `||`?

jcantrill: if yo udont have permissions to do so?  but i guess if you are building the image perms are out the window... or if the directory already exists?

kibana/Dockerfile
jcantrill: +1

kibana/install.sh
jcantrill: removed

PR #1228, opened 2018-06-29
bug 1596099. Use IndexRequest instead of UpdateRequest

jcantrill:
This PR fixes https://bugzilla.redhat.com/show_bug.cgi?id=1596099 by pulling in a new version of the openshift-elasticsearch-plugin to use IndexRequest instead of UpdateRequest


jcantrill:
/hold

Until release of 3.10

jcantrill:
@ewolinetz No because it was determined to not be a 3.10 blocker bug so it has to go in a z stream.  Also, Im not quite sure why the tests were skipped?

smarterclayton:
Jobs are fixed in https://github.com/openshift/release/pull/1027, might take a few minutes before /refresh will work for you.

PR #1232, opened 2018-07-06
sync downstream es with upstream for openshift-elasticsearch-plugin-5.…

jcantrill:
…6.9.1


ewolinetz:
/lgtm
we probably could update our .gitignore to do wildcard expansion for all our versions though couldn't we?
Maybe take that on as tech debt, its non blocking for me

PR #1234, opened 2018-07-06
change rsyslog test to use viaq/rsyslog image

richm:
Use the viaq/rsyslog container image instead of the copr
repo packages for the rsyslog tests.

Fix the k8s_filename.rulebase to workaround
https://github.com/rsyslog/rsyslog/pull/2824

rsyslog 8.30 uses case insensitive variable names, so we have
to do a little hacking around the way we handle MESSAGE/message
in order to make sure the final record sent uses "message", and
in order to support older versions of rsyslog if necessary.

kube-* projects are operations projects

### DIFF
--- a/hack/testing/rsyslog/viaq_formatting.conf
+++ b/hack/testing/rsyslog/viaq_formatting.conf
@@ -159,16 +159,31 @@ if strlen($!_MACHINE_ID) > 0 then {
         }
     }
     unset $!PRIORITY;
-    if strlen($!message) == 0 then {
-        if strlen($!MESSAGE) > 0 then {
-            set $!message = $!MESSAGE;
+    # rsyslog 8.30.0 and later does case insensitive variable name comparison
+    # which means $!MESSAGE is the same as $!message - HOWEVER - the case
+    # of the variable name is preserved when outputting, so we need to "normalize"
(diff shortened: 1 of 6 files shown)


hack/testing/rsyslog/viaq_formatting.conf
```diff
@@ -159,16 +159,31 @@ if strlen($!_MACHINE_ID) > 0 then {
...
+            set $.save_message = $!message;
+            unset $!message;
+            set $!message = $.save_message;
+            unset $.save_message;
         } else {
-            if strlen($!log) > 0 then {
-                set $!message = $!log;
+            if strlen($!message) == 0 then {
+                # case sensitive
+                set $!message = $!MESSAGE;
```
nhosoi: Could you confirm that we don't have to worry about this case?  $!MESSAGE is just trashed.
`(strlen($!message) > 0) and (strlen($!MESSAGE) > 0) and ($!message != $!MESSAGE)`

richm: yes - in this case, `$!message` was somehow set, which means we need to preserve its value.  This case shouldn't happen in this section of code, which deals with the raw data read from journald.

PR #1240, opened 2018-07-11
Adding prometheus plugin, sources, and exposing endpoints

jcantrill:
This PR re-enables the prometheus endpoint for fluentd.  @mffiedler initial test results indicate 756m default seems to be enought to include the plugin.  Will need to up the default OOTB https://bugzilla.redhat.com/show_bug.cgi?id=1598990

The PR allows disabling the plugin via env var

### DIFF
--- a/fluentd/README.md
+++ b/fluentd/README.md
@@ -6,6 +6,7 @@ Following are the environment variables that can be modified to adjust the confi
 
 | Environment Variable | Description |Example|
 |----------------------|-------------|---|
+| `ENABLE_PROMETHEUS_ENDPOINT`| Enable or diable the prometheus endpoint (Default: true)| `ENABLE_PROMETHEUS_ENDPOINT=false`|
 | `MERGE_JSON_LOG`     | **DEPRECATED** Parse JSON log messages and merge them into the JSON payload to be indexed to Elasticsearch. **Default:** True | `MERGE_JSON_LOG=true`|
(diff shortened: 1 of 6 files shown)


jcantrill:
[(5) Prometheus plugin for fluentd[logging][epic-diagnostics]](https://trello.com/c/uR6C3LHC/578-5-prometheus-plugin-for-fluentdloggingepic-diagnostics)

fluentd/run.sh
```diff
@@ -2,6 +2,7 @@
 
 export MERGE_JSON_LOG=${MERGE_JSON_LOG:-true}
 CFG_DIR=/etc/fluent/configs.d
+ENABLE_PROMETHEUS_ENDPOINT=${ENABLE_PROMETHEUS_ENDPOINT:-"true"}
```
ewolinetz: should we default this to false? Since it potentially will increase the memory footprint of Fluentd, this way a team can properly capacity plan (e.g. ops)

fluentd/README.md
```diff
@@ -6,6 +6,7 @@ Following are the environment variables that can be modified to adjust the confi
 
 | Environment Variable | Description |Example|
 |----------------------|-------------|---|
+| `ENABLE_PROMETHEUS_ENDPOINT`| Enable or diable the prometheus endpoint (Default: true)| `ENABLE_PROMETHEUS_ENDPOINT=false`|
```
ewolinetz: if we change the default, also making sure this line gets changed

fluentd/run.sh
jcantrill: My opinion is we should enable it and modify https://bugzilla.redhat.com/show_bug.cgi?id=1590920 to increase the memory as it seems this will resolve the memory issue.

PR #1243, opened 2018-07-13
allow es to build against an es base image

richm:
Clean up and simplify downstream Dockerfile
allow es to build against an es base image


elasticsearch/Dockerfile
```diff
@@ -61,22 +36,7 @@ COPY utils/** /usr/local/bin/
 RUN ln -s /usr/local/bin/logging ${HOME}/logging
 
 RUN ${HOME}/install.sh
-RUN rm -r /tmp/lib
 
 WORKDIR ${HOME}
 USER 1000
 CMD ["sh", "/opt/app-root/src/run.sh"]
-
```
jcantrill: I would add all this generated bits back in

elasticsearch/Dockerfile
```diff
@@ -1,12 +1,3 @@
-#oit## This file is managed by the OpenShift Image Tool: https://github.com/openshift/enterprise-images
```
jcantrill: I would add all this generated bits back in

elasticsearch/install.sh
```diff
@@ -8,19 +8,27 @@ source ${HOME}/prep-install.${RELEASE_STREAM}
 echo "ES plugins: ${es_plugins[@]}"
 for es_plugin in ${es_plugins[@]}
 do
-  ${ES_HOME}/bin/elasticsearch-plugin install -b $es_plugin
+  if [ -x ${ES_HOME}/bin/elasticsearch-plugin ] ; then
```
jcantrill: Is this so we can use the install script between 2.x and 5.x?

elasticsearch/install.sh
```diff
@@ -8,19 +8,27 @@ source ${HOME}/prep-install.${RELEASE_STREAM}
...
 #fix location from config
 if [[ "${ES_HOME}" != "/usr/share/elasticsearch" ]]; then
   ln -s ${ES_HOME}/index_templates /usr/share/elasticsearch/index_templates
   ln -s ${ES_HOME}/index_patterns /usr/share/elasticsearch/index_patterns
   ln -s ${ES_HOME}/kibana_ui_objects /usr/share/elasticsearch/kibana_ui_objects
 fi
 
-mkdir /elasticsearch
-mkdir -p $ES_CONF
+if [ ! -d /elasticsearch ] ; then
```
jcantrill: I don't know if its worth it here? but we should consider moving this to a place that is modifiable when we have to run on atomic/coreos.  That would be something like `/var/lib/elasticsearch` right?

elasticsearch/prep-install.prod
```diff
@@ -1,11 +0,0 @@
-# prep the plugins we need
```
jcantrill: If we no longer are using this script don't we need to make some adjustment either in the origin version or have this version for production?

elasticsearch/Dockerfile
richm: @jcantrill why?

elasticsearch/Dockerfile
richm: @jcantrill why?

elasticsearch/install.sh
richm: no - the rpm install names it `elasticsearch-plugin` but the es base image uses `plugin`

elasticsearch/install.sh
richm: Does it really matter if this is a volume mount anyway?

elasticsearch/prep-install.prod
richm: My plan is to have a prep-install.prod that lives only in the downstream repo

elasticsearch/Dockerfile
jcantrill: I defer to @adammhaile  here.  I believe the downstream move copies the files from this repo into distgit directly.  Maybe it doesnt matter since we dont actually merge and I can't speak to if its regenerated.

elasticsearch/install.sh
jcantrill: Why does the base image call it plugin?  If you download the artifact for 5.x its called `elasticsearch-plugin`? Is it being renamed by us?

elasticsearch/install.sh
jcantrill: It only matters if it matters to how it interacts with atomic/coreos.  If mounted volumes are allowed to be in this location the its a non-issue.

elasticsearch/Dockerfile
adammhaile: Definitely leave it removed. The source Dockerfiles in your github repos should absolutely not have any of the OIT stuff in them. That could potentially break things.

elasticsearch/install.sh
richm: @jcantrill yes it is being renamed by us - what difference does it make?

elasticsearch/install.sh
richm: I think mounted volumes are allowed in almost every location

elasticsearch/install.sh
jcantrill: I would suspect that we would prefer to modify the original source as little as possible but f it works then I guess its immaterial

richm:
@jcantrill if you have no further objections, can I get a lgtm?

PR #1245, opened 2018-07-13
[WIP] Creation of aggregated-logging-operator

ewolinetz:
The work for the aggregated-logging-operator for OCP 4.0

refs:
* https://docs.google.com/document/d/1T0yWe0w39ikyZtkAeUNwk-b7Qp5pg-_VcWU_P3sAZhE/
* https://github.com/operator-framework/operator-sdk


t0ffel:
maybe merge to a branch? 6k+ files is not easy to view and these are mostly dependencies that don't require any review.

ewolinetz:
@t0ffel ack

jcantrill:
@t0ffel @ewolinetz We may wish to do this like some of our other projects and establish its own repo and vendor in the submodule.  I think that would be more appropriate and then maybe build the image with our build script when needed.

ewolinetz:
@jcantrill the only concern I have with that is that we are then in a situation where we have configurations separated out from the image changes like we have now. they're in separate repos and, to me, that adds another layer that we need to keep sync'd.

I'm alright with moving this work into a private branch of my own for the time being, but I think we should have a discussion around where we would want these object to live. Maybe we can get some buy in from other teams to see what they are planning to do.

ewolinetz:
@richm this operator would be the one that we tell customers to kick off if they want to install an aggregated logging solution. My initial thought was that this operator would install the elasticsearch-operator, however per Clayton's comments in the arch document, that is not how that should happen. So I believe, this operator would be responsible for telling the OLM how to configure the elasticsearch-operator.

cluster-logging-operator/pkg/apis/logging/v1alpha1/types.go
```diff
@@ -0,0 +1,110 @@
...
+
+type ClusterLogging struct {
+  metav1.TypeMeta   `json:",inline"`
+  metav1.ObjectMeta `json:"metadata"`
+  Spec              ClusterLoggingSpec   `json:"spec"`
+  Status            ClusterLoggingStatus `json:"status,omitempty"`
+}
+
+type ClusterLoggingSpec struct {
+  AllInOne          bool `json:allinone`
```
ewolinetz: I'm open to changing this name. The idea I'm trying to convey is that we would send `app` and `infra` logs to the same place. I want to distance us from the term `ops` as that may be a misnomer

cluster-logging-operator/pkg/apis/logging/v1alpha1/types.go
```diff
@@ -0,0 +1,110 @@
...
+  Resources         v1.ResourceRequirements `json:"resources"`
+  NodeSelector      map[string]string `json:"nodeSelector,omitempty"`
+  Replicas          int32 `json:"replicas"`
+}
+
+// This is the struct that will contain information pertinent to Log storage (Elasticsearch)
+type LogStoreSpec struct {
+  Disabled          bool `json:"disabled"`
+  Resources         v1.ResourceRequirements `json:"resources"`
+  NodeSelector      map[string]string `json:"nodeSelector,omitempty"`
```
ewolinetz: Should we have a storage option here at all? We expose that on the elasticsearch operator side currently https://github.com/ViaQ/elasticsearch-operator/blob/master/pkg/apis/elasticsearch/v1alpha1/types.go#L33

cluster-logging-operator/pkg/apis/logging/v1alpha1/types.go
ewolinetz: @jcantrill I'm not sure about making this an annotation instead. I don't forsee us allowing customers to specify their specific end points unless they were doing an external log ship. That seems to be overly complicated for the 90% use case.

t0ffel: imho a view from a point of view of a pipeline would be better.
i.e. we have 2 log pipelines:
app -> logging-es
infra -> logging-es

so in yaml terms smth like:
```
appLogs:
  outputType: elasticsearch
  elasticsearchEndpoint: logging-es
infraLogs:
  outputType: elasticsearch
  elasticsearchEndpoints: logging-es
```

That way if we want logs forwarded to a different place - we simply point to a different endpoint.

t0ffel:
I can't see the files on review, it seems too large :-(

`Visualization` seem to be a bit out of place I can easily add it to elasticsearch-operator, since Kibana only interact w/ ES. wdyt?

 From the point of view of what users were able to configure I think the following is missing:
- eventrouter
- external route for ES
- custom certs
- specific versions of images to use to deploy various components

ewolinetz:
@t0ffel I haven't started adding logic for the handler with those types yet...

My thought with having the `Visualization` part separate is to isolate by functionality. 
If in the future we choose to stop using `Kibana` or offer an alternative to customers, it would make sense to me that not be tied to the `LogStore`.

I'm not sure we should allow customers to specify different versions since we are going to lock in image versions with the configs we are providing. I can see that becoming very complex if we allow the image to be specified. In my mind, we have a defined list of image versions that will be configured with an operator image version. This way when the CVO updates our operator, we can know how to update the cluster too if necessary.

Certs I'm trying to figure out if we should actually manage here as the generation of our signer certs is an admin command. But we could just expect them to exist?

I think the event router and external ES will be a stretch goal for 4.0 as I believe that would fall outside of the 80% deployment.

ewolinetz:
> I agree that in future it might be decoupled from ES, but I think it'll be easier to completely skip this section in cluster-logging operator for now since ES almost fully defines Kibana(image version, endpoint to use). Later if there is a need to decouple - we can revisit and add the section for Visualization. I'm fine either way, I just think it's redundant.

I'm not following. Our ES operator doesn't currently configure/define Kibana does it? We're going to have some redundancy I think unless we don't want this operator to create the CR for Elasticsearch at all... 

> I think this will be necessary for testing though. i.e. something goes wrong - you need to change just the version of a specific image you wouldn't want to rebuild the entire operator.

I think I need to adjust my thinking of what exposing something means. I agree with what you're saying here, that we need to be able to specify specific image versions for testing. However, I think we would just omit that from the normal deploy template for the CR so we didn't have customers updating that by default.

ewolinetz:
Closing, tracking this work here intead: https://github.com/openshift/cluster-logging-operator

PR #1248, opened 2018-07-16
Fixing unbound var in our test

ewolinetz:
Addresses error seen in output of logging job here:
https://github.com/openshift/openshift-ansible/pull/9210


richm:
and please fix line 398 too - es_ops_pod -> es_ops_svc

richm:
note that this code is in the error handling section, so something already went wrong, and we may have another problem to fix somewhere else

ewolinetz:
@richm added

PR #1252, opened 2018-07-19
Updating ES to create jks files upon start up if not present in secret dir

ewolinetz:
Should precede https://github.com/openshift/openshift-ansible/pull/9252

### DIFF
--- a/elasticsearch/run.sh
+++ b/elasticsearch/run.sh
@@ -87,6 +87,81 @@ sg_role_prometheus:
     - "${PROMETHEUS_USER:-system:serviceaccount:prometheus:prometheus}"
 CONF
 
+build_jks_from_pem() {
+
+  jks_name=$1
+  key_name=$2
+  cert_name=$3
+  ca_name=$4
+
+  openssl                                   \
+    pkcs12                                  \
+    -export                                 \
+    -in $secret_dir/$cert_name              \
+    -inkey $secret_dir/$key_name            \
+    -out $secret_dir/$jks_name.p12
+
+  keytool                                   \
+    -importkeystore                         \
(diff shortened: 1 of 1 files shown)


richm:
This will create the following .jks files in /etc/elasticsearch/secret - admin.jks elasticsearch.jks logging-es.jks - I see admin.jks referred to in the elasticsearch.yml: 
```
    openshift.searchguard:
      keystore.path: /etc/elasticsearch/secret/admin.jks
```
but I don't see elasticsearch.jks or logging-es.jks referred to anywhere in the current config or in https://github.com/openshift/openshift-ansible/pull/9252

ewolinetz:
@richm ack, i was creating them to match what ansible created even though we then use different names in the secret. I'll soft link to or mv them with the intended names.

elasticsearch/run.sh
```diff
@@ -87,6 +87,81 @@ sg_role_prometheus:
...
+  if [[ -e $secret_dir/logging-es.key && -e $secret_dir/logging-es.crt ]]; then
+    build_jks_from_pem "logging-es" "logging-es.key" "logging-es.crt" "admin-ca"
+    mv $secret_dir/logging-es.jks $secret_dir/key
+  fi
+
+  if [[ ! -e $secret_dir/truststore ]]; then
+    keytool                                        \
+      -import                                      \
+      -file $secret_dir/admin-ca                   \
+      -keystore $secret_dir/truststore.jks         \
```
richm: This file is called just `truststore` in the elasticsearch config:
```
    http:
      truststore_filepath: /etc/elasticsearch/secret/truststore
```

ewolinetz: I even had the check above correct... thanks for catching that

PR #1253, opened 2018-07-19
add back prep-install.prod

richm:
User can specify alternate maven repo at build time


richm:
@jcantrill please manually merge - downstream only

PR #1255, opened 2018-07-19
master now requires ansible-2.6

richm:
test script configures ansible-2.6 repo

### DIFF
--- a/hack/get-machine-run-tests.sh
+++ b/hack/get-machine-run-tests.sh
@@ -268,6 +268,30 @@ EOF
 scp $runfile openshiftdevel:/tmp
 ssh -n openshiftdevel "bash $runfile"
 
+#      title: "enable ansible 2.6 repo"
+cat > $runfile <<EOF
+set -euxo pipefail
+pushd $OS_O_A_L_DIR > /dev/null
+curbranch=\$( git rev-parse --abbrev-ref HEAD )
+popd > /dev/null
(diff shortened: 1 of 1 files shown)


hack/get-machine-run-tests.sh
```diff
@@ -268,6 +268,30 @@ EOF
 scp $runfile openshiftdevel:/tmp
 ssh -n openshiftdevel "bash $runfile"
 
+#      title: "enable ansible 2.6 repo"
+cat > $runfile <<EOF
+set -euxo pipefail
+pushd $OS_O_A_L_DIR > /dev/null
+curbranch=\$( git rev-parse --abbrev-ref HEAD )
+popd > /dev/null
+if [[ "\${curbranch}" == master ]] ; then
```
nhosoi: It'd be nice if you could allow ${curbranch} contains some keyword, e.g., "4.0" to execute the following if-clause in addition to just "master".  It'd help individual's debugging branch (which has to have the keyword in it, though :) to share the logic without modifying the script.

richm: yeah - somehow parse the branch into a number, and do `[ $branchnumber -gt 311 ]` - not sure how to do that - we have the same problem in aos-cd-jobs . . .

nhosoi: Ok.  It's not a big deal.  I'm setting /lgtm.

PR #1260, opened 2018-07-24
Revert "Updating ES to create jks files upon start up if not present in secret dir"

ewolinetz:
Reverts openshift/origin-aggregated-logging#1252

There are a few things that aren't as simple as I make it seem here. Backing it out until I can resolve it all locally first.


richm:
viaq-data-model flake
/test logging

PR #1263, opened 2018-07-25
Updating ES to generate jks if keys do not already exist

ewolinetz:
A more complete version of this work.
Part of this effort may also be to leave the certs in pkcs12 format instead of creating JKS files.

### DIFF
--- a/elasticsearch/run.sh
+++ b/elasticsearch/run.sh
@@ -28,6 +28,7 @@ mkdir -p /elasticsearch/$CLUSTER_NAME
 # the deployment mounts the secrets at this location - not necessarily the same
 # as $ES_CONF
 secret_dir=/etc/elasticsearch/secret
+provided_secret_dir=/etc/openshift/elasticsearch/secret
 
 BYTES_PER_MEG=$((1024*1024))
 BYTES_PER_GIG=$((1024*${BYTES_PER_MEG}))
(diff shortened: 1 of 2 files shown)


elasticsearch/run.sh
```diff
@@ -87,6 +88,100 @@ sg_role_prometheus:
...
+
+  info "Building required jks files and truststore"
+
+  # check for lack of admin.jks
+  if [[ ! -e $secret_dir/admin.jks ]]; then
+    build_jks_from_pem "admin" "admin-key" "admin-cert" "admin-ca"
+  fi
+
+  # check for elasticsearch.key and elasticsearch.crt
+  if [[ -e $secret_dir/elasticsearch.key && -e $secret_dir/elasticsearch.crt ]]; then
```
richm: you check for the existence of `admin.jks` before writing it - do you need to check for `elasticsearch.jks` and `logging-es.jks` before writing them?

ewolinetz: so my thought here was if we have the .key and .crt then that would mean we shouldn't have the jks as generated by ansible. There shouldn't be any harm in adding the jks check though... I'll do that.

ewolinetz:
I want to wait on this change just a little bit -- I may end updating the pem crt and key file names to match what the cert signing service would provide

t0ffel:
PEM format seem to be supported by SearchGuard https://docs.search-guard.com/v5/configuring-tls

ewolinetz:
will update this to follow the pattern in https://github.com/openshift/origin-aggregated-logging/pull/1266

ewolinetz:
@jcantrill Kibana is failing to roll out now?

richm:
> @jcantrill Kibana is failing to roll out now?

Looks like it - and the CI logs/artifacts are all messed up, so we can't even figure out why?

PR #1266, opened 2018-07-30
fixes #1251. Move init to scripts

### LINKED ISSUE #1251: Cluster initialization and modifications proposal
# Overview
Initialization of the ES cluster is currently executed by the run script by backgrounding a process which is contrary to recommendations for running containers.  Kubernetes does not offer (AFAIK) a way to run post initialization outside of using something like a Job.  In anticipation of the introduction of operators, this proposal identifies how we may modify and structure the initialization and migration tasks so they could be executed by something like a Job or the operator.  It would move us more in line with the original deployer which allows us to add version the initialization and migration tasks within our elasticsearch image.

# Proposal
* Create directories specifically for initialization tasks and migration tasks
* Add an `init` (or `cluster-init`) and `migrate` entrypoints
## Directory Structure
Each directory includes the scripts needed to perform the init or migrate tasks:
* prefixed with a 4 digit number to identify the order in which they are executed
* idempotent
```
/etc/elasticsearch
  |- init/
     |- 0001-seed-acl
     |- 0010-seed-index-templates
  |-migrate/
    |- 0100-clean-kibana-index-patterns
    |- 0150-remove-old-searchguard-indices
```
## Entry points
```
${HOME}/init

  #!/bin/bash
  source logging
  for file in $(/etc/elasticsearch/init/*); do
     if [ -f $file ] ; then
       $("./${file}")
     fi  
  done

${HOME}/migrate
```
## run.sh
Initially, modify the `run.sh` script to background the `init` and `migrate` scripts

jcantrill:
This PR resolves #1251:
* creates init script that can be called from something like an operator
* moves init scripts to single scripts that can be called in order

### DIFF
--- a/elasticsearch/init.sh
+++ b/elasticsearch/init.sh
@@ -1,0 +1,27 @@
+#!/bin/bash -e
+#
+# Copyright 2018 Red Hat, Inc. and/or its affiliates
+# and other contributors as indicated by the @author tags.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+source "logging"
+
+pushd "${ES_HOME}/init"
+  files=$(ls | sort)
+  for init_file in ${files}; do
+    if [ -f "${init_file}" ] ; then
+      ./"${init_file}"
+    fi
+  done
+popd
--- a/elasticsearch/init/0010-seed-acl
+++ b/elasticsearch/init/0010-seed-acl
@@ -1,0 +1,41 @@
+#!/bin/bash -e
+#
+# Copyright 2018 Red Hat, Inc. and/or its affiliates
+# and other contributors as indicated by the @author tags.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+source "logging"
+
+force=${FORCE_SEED_ACL:-''}
+config=${ES_CONF}/elasticsearch.yml
+index=$(python -c "import yaml; print yaml.load(open('${config}'))['searchguard']['config_index_name']" | xargs -i sh -c "echo {}")
+
+if [ -n "${force}" ] ; then
+  docs=(roles rolesmapping actiongroups internalusers config)
+  for d in $docs; do
+    resp=$(es_util --query=${index}/$d/0 -w %{response_code} | cut -d '}' -f2)
+    #re-seed acls if any of them are missing
+    if [ "${resp}" != "200"]; then
+      info "Missing ACL document '${d}'. Reseeding all ACL documents"
+      force="true"
+      break
+    fi
+  done
+else
+  info "Forcing the seeding of ACL documents"
+fi
+
+if [ -z "${force}" ] ; then
+  es_seed_acl
+fi
--- a/elasticsearch/init/0100-seed-index-templates
+++ b/elasticsearch/init/0100-seed-index-templates
@@ -1,0 +1,43 @@
+#!/bin/bash -e
+#
+# Copyright 2018 Red Hat, Inc. and/or its affiliates
+# and other contributors as indicated by the @author tags.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+source "logging"
+
+info Adding index templates
+shopt -s failglob
+for template_file in ${ES_HOME}/index_templates/*.json
+do
+    sed -i "s,\$REPLICA_SHARDS,$REPLICA_SHARDS," $template_file
+    sed -i "s,\$PRIMARY_SHARDS,$PRIMARY_SHARDS," $template_file
+    template=`basename $template_file`
+    # Check if index template already exists
+    response_code=$(es_util --query=_template/$template \
+        ${DEBUG:+-v} -s \
+        --request HEAD --head --output /dev/null \
+        -w '%{response_code}')
+    if [ "${response_code}" == "200" ]; then
+        info "Index template '$template' found in the cluster, overriding it"
+    else
+        info "Create index template '$template'"
+    fi
+    es_util --query=_template/$template \
+        ${DEBUG:+-v} -s -X PUT \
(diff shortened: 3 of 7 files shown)


elasticsearch/init.sh
```diff
@@ -0,0 +1,61 @@
...
+        info Elasticsearch is ready and listening
+        return 0
+    fi
+    cat $LOG_FILE
+    rm -f $LOG_FILE
+    exit 1
+}
+
+pushd "${ES_HOME}/init"
+  files=$(ls ${ES_HOME}/init | sort)
```
richm: `files` is unused?

elasticsearch/init.sh
```diff
@@ -0,0 +1,61 @@
...
+  for init_file in $(ls ${ES_HOME}/init | sort); do
```
richm: you're already in the directory, you could just do `for init_file in $(ls | sort); do`

elasticsearch/init.sh
```diff
@@ -0,0 +1,61 @@
...
+    fi
+    cat $LOG_FILE
+    rm -f $LOG_FILE
+    exit 1
+}
+
+pushd "${ES_HOME}/init"
+  files=$(ls ${ES_HOME}/init | sort)
+  for init_file in $(ls ${ES_HOME}/init | sort); do 
+      set +e
```
richm: what is the purpose of disabling error handling here?  Do you care if one of the init scripts returns an error?

elasticsearch/init.sh
```diff
@@ -0,0 +1,61 @@
...
+      if [ -f $init_file ] ; then
+        $(./init-file)
```
richm: what about `./$init_file`?
Also, maybe you want to execute all init scripts, and return an error if any of them returns an error?  If so, then use something like this
```
    rc=0
    ....
   ./$init_file || rc=1
   ....
   exit $rc
```
then you can leave `set -e` on all the time - the `||` will "catch" the error

elasticsearch/init.sh
```diff
@@ -0,0 +1,61 @@
...
+
+source "logging"
+
+LOG_FILE=${LOG_FILE:-elasticsearch_connect_log.txt}
+
+wait_for_port_open() {
+    rm -f $LOG_FILE
+    # test for ES to be up first and that our SG index has been created
+    info "Checking if Elasticsearch is ready on $ES_REST_BASEURL"
+    while ! response_code=$(es_util --query=/ \
```
t0ffel: `es_util` hits `$ES_BASE`, not `$ES_REST_BASEURL`

elasticsearch/init.sh
jcantrill: fixed.  was added without completing my thought to check file size and then wait for port to open if the length is greater then zero

elasticsearch/init.sh
jcantrill: done

elasticsearch/init/0010-seed-acl
```diff
@@ -0,0 +1,41 @@
...
+      force="true"
+      break
+    fi
+  done
+else
+  info "Forcing the seeding of ACL documents"
+fi
+
+if [ -z "${force}" ] ; then
+  es_seed_acl
```
t0ffel: sgadmin needs to be passed ES address as a parameter

elasticsearch/init.sh
jcantrill: my thought here is for the process to continue regardless of the outcome so no one script kills ES from initializing.  This could allow it to be manually fixed but maybe its better for it to simply fail outright.  That would be consistent to the behavior we have now.  I'm open to suggestions.

elasticsearch/init/0100-seed-index-templates
```diff
@@ -0,0 +1,43 @@
...
+        ${DEBUG:+-v} -s \
+        --request HEAD --head --output /dev/null \
+        -w '%{response_code}')
+    if [ "${response_code}" == "200" ]; then
+        info "Index template '$template' found in the cluster, overriding it"
+    else
+        info "Create index template '$template'"
+    fi
+    es_util --query=_template/$template \
+        curl ${DEBUG:+-v} -s -X PUT \
```
t0ffel: is `$ES_REST_BASEURL` missing here?

elasticsearch/init.sh
jcantrill: removed ref.  not significant though maybe it is if the script is running in a container to access some random ES node instead of local?

elasticsearch/init/0010-seed-acl
jcantrill: same here?  if run outside as part of an init container then it would need to know how to find es?

elasticsearch/init/0100-seed-index-templates
jcantrill: yes i believe it is.  This assumes you are running against local host which will not be the case of run as an init container

elasticsearch/init/0010-seed-acl
t0ffel: right, I'm looking from the point of view of running it in the separate container.

t0ffel:
I guess change to run init script as a separate container can be done later on.

elasticsearch/init/0100-seed-index-templates
jcantrill: We will need to pass `$ES_BASE` as an env var to the init container which should be "https://logging-es:9200" for template initialization and 
$ES_HOST AND $ES_PORT for the seeding to run. $ES_HOST is `logging-es-cluster` service and port defaults to `9300`.  Maybe we can source these from env vars

jcantrill: We should pass the values of:
```
ES_HOST=${LOGGING_ES_CLUSTER_SERVICE_HOST}
ES_PORT=${LOGGING_ES_CLUSTER_SERVICE_PORT}
```
I refrain from adding those here until we move init out of each ES node since then it would force it to go through the service instead of `localhost`.  The vars passed by the kublet are more appropriately named.

jcantrill: Further looking we will also need to base $ES_BASE so the es_util script will succeed which can also be changed for completeness to:
```
ES_BASE="${LOGGING_ES_SERVICE_HOST}:${LOGGING_ES_SERVICE_PORT}"
```

elasticsearch/init.sh
```diff
@@ -17,47 +17,11 @@
...
-  if [ ${#files[@]} -gt 0 ] ; then
-    wait_for_port_open
-    for init_file in ${files}; do
-      if [ -f $init_file ] ; then
-        ./${init-file}
-      fi
-    done
-  fi
+  for init_file in ${files}; do
+    if [ -f ${init_file} ] ; then
```
richm: `-f "${init_file}"`
have to use double quotes in case dir is empty

elasticsearch/utils/es_seed_acl
```diff
@@ -2,6 +2,8 @@
 
 source "logging"
 
+ES_HOST=${ES_HOST:-"localhost"}
+ES_PORT=${ES_PORT:-9300}
```
richm: sg uses transport port?  If so, that may be a problem as ES_PORT is usually the http port

jcantrill: updating to be 'ES_CLUSTER_**' but should be https://github.com/openshift/origin-aggregated-logging/pull/1266#discussion_r206537804 in future

elasticsearch/init.sh
```diff
@@ -19,9 +19,9 @@ source "logging"
 
 pushd "${ES_HOME}/init"
   files=$(ls | sort)
-  for init_file in ${files}; do
-    if [ -f ${init_file} ] ; then
-      ./${init_file}
+  for init_file in "${files}"; do
```
richm: use `${files}` here, not `"${files}"`

richm:
```
FAILURE after 0.530s: test/cluster/functionality.sh:114: executing 'curl_es_pod 'logging-es-data-master-16fr32mp-1-qvf5d' '/_template/com.redhat.viaq-openshift-orphaned.template.json' --request HEAD --head --output /dev/null --write-out '%{response_code}'' expecting success and text '200': the output content test failed
404
```
That is, the index template was not found - do we need to make the timeout for this test a lot longer, or otherwise wait until the index templates have been loaded?

jcantrill:
@richm fixed issue with improper change to the template seeding.  it was trying to use both es_util and curl

richm:
mux flake?
/test json-file

ewolinetz:
/lgtm 
this cleans up our run script nicely and gives us an easy way to remove scripts for when the operator takes over functionality

PR #1267, opened 2018-07-31
Changed analyzer on project, operations and orphaned index templates

niconosenzo:
Using the "whitespace" analyzer [0] on the message field instead of the standard one used by ES (when no analyzer is set [1]) allows to perform search for upper-case characters.

[0] https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-whitespace-analyzer.html
[1] https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-standard-analyzer.html

BZ [1610204](https://bugzilla.redhat.com/show_bug.cgi?id=1610204)

### DIFF
--- a/elasticsearch/index_templates/com.redhat.viaq-openshift-operations.template.json
+++ b/elasticsearch/index_templates/com.redhat.viaq-openshift-operations.template.json
@@ -39,6 +39,7 @@
           "message_field": {
             "mapping": {
               "index": "analyzed",
+              "analyzer": "whitespace",
               "norms": {
                 "enabled": false
               },
@@ -439,6 +440,7 @@
         "message": {
           "doc_values": false,
           "index": "analyzed",
+          "analyzer": "whitespace",
           "norms": {
             "enabled": false
           },
--- a/elasticsearch/index_templates/com.redhat.viaq-openshift-orphaned.template.json
+++ b/elasticsearch/index_templates/com.redhat.viaq-openshift-orphaned.template.json
@@ -11,6 +11,7 @@
           "message_field": {
             "mapping": {
               "index": "analyzed",
+              "analyzer": "whitespace",
               "omit_norms": true,
               "type": "string"
             },
@@ -436,6 +437,7 @@
         "message": {
           "doc_values": false,
           "index": "analyzed",
+          "analyzer": "whitespace",
           "norms": {
             "enabled": false
           },
--- a/elasticsearch/index_templates/com.redhat.viaq-openshift-project.template.json
+++ b/elasticsearch/index_templates/com.redhat.viaq-openshift-project.template.json
@@ -39,6 +39,7 @@
           "message_field": {
             "mapping": {
               "index": "analyzed",
+              "analyzer": "whitespace",
               "norms": {
                 "enabled": false
               },
@@ -439,6 +440,7 @@
         "message": {
           "doc_values": false,
           "index": "analyzed",
+          "analyzer": "whitespace",
           "norms": {
             "enabled": false
           },


lukas-vlcek:
@niconosenzo @jcantrill Correct, this PR needs to go against <https://github.com/ViaQ/elasticsearch-templates>.

But I am not sure I understand the reasoning behind this change in general. Linked BZ does not include enough details about why this change is really needed. IMO keeping lower case filter in token stream is very valuable.

Why is case sensitive search so important in this case? If the `TEST` term is so important then I assume it describes some kind of document category and in such case it should be extracted into separate document field and indexed without lowercasing. Then you could search it using `important_field:TEST` or equivalent ES query language expression.

niconosenzo:
@lukas-vlcek 
The request arises for the use case where an alarming solution needs to search for patterns in the message field.   

Said that, I certainly understand the advantages of having all stored in lower-case

Although there is another issue, the standard analyzer does not work well or I'm not able to perform a search for special characters (like #, : , []). Which I actually do with the whitespace analyzer

lukas-vlcek:
> Although there is another issue, the standard analyzer does not work well or I'm not able to perform a search for special characters (like #, : , []). Which I actually do with the whitespace analyzer

@niconosenzo that is expected. Standard analyzer uses [standard tokenizer](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-standard-tokenizer.html) which removes many special characters. While the whitespace analyzer uses [whitespace tokenizer](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-whitespace-tokenizer.html) which uses only whitespaces to produce tokens.

As for the alerting on case sensitive patterns in messages, if that is really needed then I would suggest the customer to modify the default mapping or add a custom mapping to introduce a new field (see [fields](https://www.elastic.co/guide/en/elasticsearch/reference/master/multi-fields.html) or [copy-to](https://www.elastic.co/guide/en/elasticsearch/reference/master/copy-to.html)) and setup case sensitive mapping for it.

May be you could elaborate more about this use case in the BZ, right now it is not obvious if the funcionality will be general enough for others to benefit from. Keep in mind that case sensitive mapping for field like message (i.e. a lot of free texts) will have impact on index size, thus it should be considered carefully.

niconosenzo:
@lukas-vlcek 
Thanks for the answers, let me request Loic  to join into this conversation.

yeplaa:
Hello,
Thank you for your feedback. Indeed, currently we can not do research on some special characters as well as for case sensitive characters  (either with an alarming tool or to search with Kibana).
With this configuration it is therefore difficult to implement search patterns / regexp on logs.

With the whitespace analyzer on the message field, what would be the impact on the indexes size? 
Do you already have feedback on this use case? 
Do you think of another solution that meets our needs without modifying the analyzer?

Thanks
Loïc

richm:
Is it the `message` field that you need to perform case sensitive searches in?  If so, @lukas-vlcek if they followed directions like this: https://github.com/richm/docs/releases/tag/20180726172001 without the index settings but with a different mapping for the `message` field at a higher priority, would that merge into and replace the existing mapping for whatever indices were matched in the template?

lukas-vlcek:
@richm yes, they can _override_ mapping of just the single field by adding custom index template at the higher level (i.e. higher `order` value) that contains just this single field mapping, see [multiple-templates](https://www.elastic.co/guide/en/elasticsearch/reference/5.5/indices-templates.html#multiple-templates) doc.

Though, notice that the `message` field is actually defined "indirectly" via [dynamic template](https://www.elastic.co/guide/en/elasticsearch/reference/5.5/dynamic-templates.html), so better test how this will work first.

niconosenzo:
@lukas-vlcek @richm 
Would be there any significant impact (in performance) if customer changes the analyzer as proposed on this PR instead of defining a custom mapping ? In such a case, I think the template will be overwritten during next upgrade, is it correct ?

richm:
> Would be there any significant impact (in performance) if customer changes the analyzer as proposed on this PR instead of defining a custom mapping ?

I don't know, but it seems as if the customer has no choice - they must have the correct matching, and will be willing to trade off some performance degradation.

> In such a case, I think the template will be overwritten during next upgrade, is it correct ?

No, because you will be adding a completely new template instead of attempting to edit the existing one.

niconosenzo:
Closing PR. As this is an isolated case that can be worked out by customizing the .project* template instead of changing the mapping for all new releases.

PR #1271, opened 2018-08-01
Read in /etc/podinfo/mem_limit if it exists and use that over INSTANCE_RAM env var

ewolinetz:
ES image change for https://github.com/openshift/openshift-ansible/pull/9394

To resolve RFE https://bugzilla.redhat.com/show_bug.cgi?id=1602819

### DIFF
--- a/elasticsearch/run.sh
+++ b/elasticsearch/run.sh
@@ -20,6 +20,9 @@ RETRY_INTERVAL=${RETRY_INTERVAL:-1}	# how often (in sec)
 PRIMARY_SHARDS=${PRIMARY_SHARDS:-1}
 REPLICA_SHARDS=${REPLICA_SHARDS:-0}
 
+MEM_LIMIT_FILE=/etc/podinfo/mem_limit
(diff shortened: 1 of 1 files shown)


elasticsearch/run.sh
```diff
@@ -35,11 +38,18 @@ BYTES_PER_GIG=$((1024*${BYTES_PER_MEG}))
 MAX_ES_MEMORY_BYTES=$((64*${BYTES_PER_GIG}))
 MIN_ES_MEMORY_BYTES=$((256*${BYTES_PER_MEG}))
 
+if [[ -e $MEM_LIMIT_FILE ]]; then
```
jcantrill: Do we need to check the existence of the file?

ewolinetz: yes, if the file doesn't exist because the DC doesn't have the volume mount defined then we don't want to fail

richm:
looks like a hard error - could not deploy elasticsearch

ewolinetz:
yup, reran to get the container logs `/opt/app-root/src/run.sh: line 42: local: can only be used in a function`

PR #1272, opened 2018-08-02
rsyslog - fix retry logic - use local vars in templates

richm:
When retrying, the _id and _index fields should already be
set - use them - otherwise, create new values for them.
Also fixes tabs/spaces.

### DIFF
--- a/hack/testing/rsyslog/elasticsearch.conf.j2
+++ b/hack/testing/rsyslog/elasticsearch.conf.j2
@@ -2,7 +2,7 @@ template(name="viaq_template" type="list") {
     property(name="$!all-json-plain")
 }
 
-template(name="index_pattern" type="list") {
+template(name="prefix_index_template" type="list") {
     property(name="$.viaq_index_prefix")
     property(name="$!@timestamp" dateFormat="rfc3339" position.from="1" position.to="4")
     constant(value=".")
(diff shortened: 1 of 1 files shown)


hack/testing/rsyslog/elasticsearch.conf.j2
```diff
@@ -11,33 +11,44 @@ template(name="index_pattern" type="list") {
...
+        }
+        # else fall through to retry operation
+    }
+    if strlen($!viaq_msg_id) > 0 then {
+	    set $.viaq_msg_id = $!viaq_msg_id;
+    } else if strlen($.omes!_id) > 0 then {
+        # retry
+        set $.viaq_msg_id = $.omes!_id;
+    } else {
+	    # NOTE: depends on rsyslog being compiled with --enable-uuid
```
nhosoi: Where --enable-uuid is set?  I cannot find it in the spec files...
An indentation issue in line 43? ;)

richm: > Where --enable-uuid is set? I cannot find it in the spec files...

It is set by default:
```
>./configure --help|grep uuid
  --enable-uuid           Enable support for uuid generation [default=yes]
```

fixed the indentation issue

richm: see also https://kojipkgs.fedoraproject.org//packages/rsyslog/8.36.0/1.fc28/data/logs/x86_64/build.log
```
rsyslog will be compiled with the following settings:
...
    uuid support enabled:                     yes
```

nhosoi: I see.  If it's set by default, it's even better.
/lgtm

PR #1275, opened 2018-08-08
use standalone json parser instead of k8s meta filter

richm:
We already had the functionality in filter_k8s_meta_for_mux_client.rb
so that was renamed to filter_parse_json_field.rb to give us a
general purpose parser filter for json valued fields.  That filter
was also enhanced to look for either the `log` or the `MESSAGE`
field so that we could get rid of some of the `USE_JOURNAL`
related logic and simplify the mux pipeline cases.


fluentd/filter_parse_json_field.rb
```diff
@@ -0,0 +1,75 @@
...
+      json_fields.each do |merge_json_log_key|
+        if record.has_key?(merge_json_log_key)
+          value = record[merge_json_log_key].strip
+          if value[0].eql?('{') && value[-1].eql?('}')
+            begin
+              record = JSON.parse(value).merge(record)
+              unless @preserve_json_log
+                record.delete(merge_json_log_key)
+              end
+            rescue JSON::ParserError
```
jcantrill: Maybe we should add a trace or debug message that parse failed.  Minimal message, not the entire stack

jcantrill:
/lgtm 
Does @ebalsumgo need to document anything here or should this be transparent to our users?

richm:
>  Does @ebalsumgo need to document anything here or should this be transparent to our users?

It should be transparent to our users - MERGE_JSON_LOG is still used the same way - default value is still `true` - setting `MERGE_JSON_LOG=false` will disable json parsing as before.

PR #1283, opened 2018-08-13
Added s3 plugin to Es5.x to enable s3 backups

shgriffi:
Added s3 plugin in order to enable backups and restores of indexes to S3.

### DIFF
--- a/elasticsearch/Dockerfile.centos7.s3repository
+++ b/elasticsearch/Dockerfile.centos7.s3repository
@@ -1,0 +1,66 @@
+FROM centos:centos7
+
+MAINTAINER OpenShift Development <dev@lists.openshift.redhat.com>
+
+EXPOSE 9200
+EXPOSE 9300
+USER 0
+
+ENV ES_CLOUD_K8S_VER=5.6.9.2 \
+    ES_CONF=/etc/elasticsearch/ \
+    ES_HOME=/usr/share/elasticsearch \
+    ES_VER=5.6.9 \
+    HOME=/opt/app-root/src \
+    INSTANCE_RAM=512G \
+    JAVA_VER=1.8.0 \
+    JAVA_HOME=/usr/lib/jvm/jre \
+    NODE_QUORUM=1 \
+    OSE_ES_VER=5.6.9.0 \
+    PROMETHEUS_EXPORTER_VER=5.6.9.0 \
+    PLUGIN_LOGLEVEL=INFO \
+    RECOVER_AFTER_NODES=1 \
+    RECOVER_EXPECTED_NODES=1 \
+    RECOVER_AFTER_TIME=5m \
+    RELEASE_STREAM=origin \
+    ADDITIONAL_PLUGINS=repository-s3
+
+ARG ES_CLOUD_K8S_VER=5.6.9.2
+ARG OSE_ES_VER=5.6.9.0
+ARG SG_VER=5.6.9-19
+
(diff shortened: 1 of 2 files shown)


elasticsearch/prep-install.origin
```diff
@@ -8,4 +8,7 @@ fi
 if [ -z "${PROMETHEUS_EXPORTER_URL:-}" ] ; then
     PROMETHEUS_EXPORTER_URL=https://distfiles.compuscene.net/elasticsearch/elasticsearch-prometheus-exporter-${PROMETHEUS_EXPORTER_VER}.zip
 fi
-es_plugins=($ES_CLOUD_K8S_URL $OSE_ES_URL $PROMETHEUS_EXPORTER_URL)
+if [ -z "${ADDITIONAL_PLUGINS:-}" ] ; then
+    ADDITIONAL_PLUGINS=
+fi
+es_plugins=($ES_CLOUD_K8S_URL $OSE_ES_URL $PROMETHEUS_EXPORTER_URL $ADDITIONAL_PLUGINS)
```
richm: You don't need the `if` clause, you can just write it as one line:
`es_plugins=($ES_CLOUD_K8S_URL $OSE_ES_URL $PROMETHEUS_EXPORTER_URL ${ADDITIONAL_PLUGINS:-})`

richm:
Do you need another Dockerfile?  Can you pass in the additional plugins to use like this, with your first commit?
```
docker build -e ADDITIONAL_PLUGINS=repository-s3 ....
```
or using `build_arg`:
```
docker build --build-arg=ADDITIONAL_PLUGINS=repository-s3 ....
```
?

elasticsearch/prep-install.origin
```diff
@@ -8,4 +8,7 @@ fi
 if [ -z "${PROMETHEUS_EXPORTER_URL:-}" ] ; then
     PROMETHEUS_EXPORTER_URL=https://distfiles.compuscene.net/elasticsearch/elasticsearch-prometheus-exporter-${PROMETHEUS_EXPORTER_VER}.zip
 fi
-es_plugins=($ES_CLOUD_K8S_URL $OSE_ES_URL $PROMETHEUS_EXPORTER_URL)
+if [ -z "${ADDITIONAL_PLUGINS:-}" ] ; then
```
jcantrill: In general, I like this addition to allow users to customize their own images.

elasticsearch/Dockerfile.centos7.s3repository
```diff
@@ -0,0 +1,66 @@
+FROM centos:centos7
```
jcantrill: @richm @ewolinetz In general, I don't like the idea of adding a whole other Dockerfile.  Shouldn't we be able to build upon one that exists by adding env vars and args to one which already exists?

richm: > Shouldn't we be able to build upon one that exists by adding env vars and args to one which already exists?

Yes, we should

ewolinetz: I agree, I feel that if we have another Dockerfile that implies we fully support and will provide images for that Dockerfile.

If we provide the means to support custom images where a customer can specify and build with additional plugins I would prefer to go that route for now.

shgriffi:
Thanks for the feedback everyone.  Makes perfect sense to not introduce another docker file.  I'll rework this to take that into account.

PR #1284, opened 2018-08-14
Bug 1489533 - logging-fluentd needs to periodically reconnect to logging-mux or elasticsearch to help balance sessions

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1489533

https://github.com/uken/fluent-plugin-elasticsearch/pull/459
implements support for reloading connections when the
Elasticsearch is behind a proxy/load balancer, as in our case,
and allows specifying the reload interval in terms of the
number of operations.

This PR adds support for the following env. vars which can be
set in the fluentd daemonset/mux deployment.  The ability to
set these is provided primarily for experimentation, not something
which will ordinarily require tuning in production.
`ES_RELOAD_CONNECTIONS` - boolean - default `true`
`ES_RELOAD_AFTER` - integer - default `100`
`ES_SNIFFER_CLASS_NAME` - string - default `Fluent::Plugin::ElasticsearchSimpleSniffer`
There are also `OPS_` named env. vars which will override the
corresponding `ES_` named env. var.

That is, by default, fluentd will reload connections to
Elasticsearch every 100 operations (NOTE: not every 100 records!)
These include internal `ping` operations, so will not exactly
correspond to each bulk index request.

### DIFF
--- a/fluentd/configs.d/openshift/output-es-config.conf
+++ b/fluentd/configs.d/openshift/output-es-config.conf
@@ -18,9 +18,11 @@
       type_name com.redhat.viaq.common
       retry_tag "retry_es"
 
-      # there is currently a bug in the es plugin + excon - cannot
-      # recreate/reload connections
-      reload_connections false
+      reload_connections "#{ENV['ES_RELOAD_CONNECTIONS'] || 'true'}"
+      # https://github.com/uken/fluent-plugin-elasticsearch#reload-after
+      reload_after "#{ENV['ES_RELOAD_AFTER'] || '100'}"
+      # https://github.com/uken/fluent-plugin-elasticsearch#sniffer-class-name
+      sniffer_class_name "#{ENV['ES_SNIFFER_CLASS_NAME'] || 'Fluent::ElasticsearchSimpleSniffer'}"
       reload_on_failure false
       flush_interval "#{ENV['ES_FLUSH_INTERVAL'] || '1s'}"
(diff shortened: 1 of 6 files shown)


fluentd/configs.d/openshift/output-es-config.conf
```diff
@@ -18,9 +18,9 @@
       type_name com.redhat.viaq.common
       retry_tag "retry_es"
 
-      # there is currently a bug in the es plugin + excon - cannot
-      # recreate/reload connections
-      reload_connections false
+      reload_connections "#{ENV['ES_RELOAD_CONNECTIONS'] || 'true'}"
+      reload_after "#{ENV['ES_RELOAD_AFTER'] || '100'}"
```
jcantrill: We may want to consider including a 'unit' to explicitly identify what it is for clarity: '_OPERATIONS'

ewolinetz: I think we could get away with just adding a comment above this or expanding our docs or both

ewolinetz:
Overall LGTM

fluentd/configs.d/openshift/output-es-config.conf
richm: I put in a link to the upstream docs - ok?

jcantrill: wfm

richm:
can I get a /lgtm?

PR #1290, opened 2018-08-15
use rpm to find sniffer downstream

richm:
use rpm to find sniffer downstream


richm:
@jcantrill please manually merge to unblock downstream builds

PR #1303, opened 2018-08-21
Use correct logo for OCP distribution

lukas-vlcek:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1615293

### DIFF
--- a/kibana/Dockerfile
+++ b/kibana/Dockerfile
@@ -24,7 +24,7 @@ RUN if [ -n "${LOCAL_REPO}" ] ; then \
     fi
 
 RUN INSTALLED_PKGS="kibana-${KIBANA_VER}*" && \
-    yum install -y --setopt=tsflags=nodocs  ${INSTALLED_PKGS} && \
+    yum install -y --setopt=tsflags=nodocs  ${INSTALLED_PKGS} zip && \
     yum clean all
 
 ADD nodescl-node /usr/bin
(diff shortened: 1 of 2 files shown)


jcantrill:
/lgtm
image

lukas-vlcek:
@jcantrill 

Still there is a visual and `css:style` issue here. The logo file is placed as a `background-image` using the [following style](https://github.com/openshift/origin-kibana/blob/master/lib/styles/main.less#L149-L150):

```
  height: 11px;
  background-image: url("@{img-path}/logo-origin-thin.svg");
  background-size: 111px 11px
```
Now, the [original file](https://github.com/openshift/origin-kibana/blob/master/lib/images/logo-origin-thin.svg) has internal ratio of **15.7** (`width="160" height="10.159"`).
The [OCP file](https://github.com/openshift/origin-aggregated-logging/blob/master/kibana/logo-OCP-console-hdr-thin.svg) has internal ratio of **22** (`width="320" height="14"`).

This means the OCP logo is wider but still it is forced to fit the came 111px/11px dimension.

IMO we should make sure both the logos have same ratio. We can probably add more white space on both the left and right hand side of the original logo and change the ccs style as well. BTW the css ratio is about **10** so both the logos are bit deformed, though the OCP one is deformed a lot more now.

jcantrill:
@lukas-vlcek https://github.com/openshift/origin-kibana/blob/master/lib/styles/main.less

lukas-vlcek:
@jcantrill Who to talk to wrt svg files? I can change them myself but I assume there are some original templates buried deep on hard drive of some designer.

jcantrill:
@spadgett  Do you know who has these images?

spadgett:
@rhamilto can you help?

PR #1304, opened 2018-08-21
add Red Hat specific versions, repos

richm:
add Red Hat specific versions, repos


richm:
@jcantrill @adammhaile @tbielawa

PR #1310, opened 2018-08-23
disable non-PIC build; build only components used by fluentd

richm:
disable non-PIC build; build only components used by fluentd
The builds of the non-PIC libraries are throwing off build
security detectors, so disable them.
jemalloc 5.1.0 makes it easy to build only the components we
actually use - the shared library and the jemalloc-config script


richm:
flake?
/test json-file

PR #1311, opened 2018-08-23
fix local devel to be faster by modifying image layer

jcantrill:
Move the compile of jemelloc up in the Dockerfile to take advantage of using cached layers in local devel since it unlikely to change as frequently as the other files.

### DIFF
--- a/fluentd/Dockerfile.centos7
+++ b/fluentd/Dockerfile.centos7
@@ -30,6 +30,11 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
       redhat-rpm-config && \
     yum clean all
 
+ADD jemalloc/ ${HOME}/jemalloc/
+RUN cd ${HOME}/jemalloc && EXTRA_CFLAGS="$( rpm --eval '%{optflags}' )" ./autogen.sh && \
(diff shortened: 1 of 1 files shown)


fluentd/Dockerfile.centos7
```diff
@@ -30,6 +30,16 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
...
+ADD jemalloc/ ${HOME}/jemalloc/
+# note - make dist "fails" here
+# false -o doc/jemalloc.html doc/html.xsl doc/jemalloc.xml
+# so cannot use &&
+# cannot use make install - will try to install docs - not only does that fail
+# for some reason, but we also don't want them here - use install_lib install_bin
+RUN cd ${HOME}/jemalloc && EXTRA_CFLAGS="$( rpm --eval '%{optflags}' )" ./autogen.sh && \
+    make dist ; make && make install_lib install_bin && cp COPYING ${HOME}/COPYING.jemalloc && \
+    cd .. && rm -rf jemalloc
+
```
richm: looks like you removed my changes - it should look like this:
```
    make install_lib_shared install_bin && cp COPYING ${HOME}/COPYING.jemalloc && \
```
Also, we can get rid of the comment - there is no error anymore

PR #1314, opened 2018-08-23
Use debops framework to deploy viaq configuration.

nhosoi:
Introduced nocomment key.

mmkubernetes: moved options from module to action.

Adding hack/testing/rsyslog/README.md


hack/testing/rsyslog/roles/openshift_logging_rsyslog/tasks/main.yaml
```diff
@@ -0,0 +1,122 @@
+---
+- name: see if openshift-logging namespace exists
```
richm: These tasks through line 79 should only be run on the masters[0] node, and are specific to openshift.  The remaining tasks are not specific to openshift and should be run on all nodes.

hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml
```diff
@@ -0,0 +1,175 @@
...
+    createhome: False
+    system: True
+  when: rsyslog__unprivileged|bool and rsyslog__user != 'root'
+
+# Creating a work directory
+- name: Create a work directory
+  file:
+    state: directory
+    path: "{{ rsyslog__work_directory }}"
+    mode: 0700
```
richm: should add `owner` and `group` here, then I'm not sure if you need the spool dir task below

hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml
```diff
@@ -0,0 +1,175 @@
...
+# Moving the contents of /etc/rsyslog.d to the backup dir
+- name: Moving the contents of /etc/rsyslog.d to the backup dir
+  shell: mv /etc/rsyslog.d/* "{{ tempdir }}"
+
+- name: create rsyslog viaq subdir
+  file: path="{{ rsyslog__viaq_path }}" state=directory mode=0700
+
+- name: Fix directory permissions if needed
+  file:
+    path: '/var/spool/rsyslog'
```
richm: I think this is a debian-ism.  We use `rsyslog__work_directory` (which is usually `/var/lib/rsyslog`) on rhel/fedora/centos.

hack/testing/rsyslog/roles/rsyslog/defaults/main.yaml
```diff
@@ -0,0 +1,1509 @@
...
+#
+# The unprivileged system group account used by the ``rsyslogd`` daemon, when
+# unprivileged operation is enabled.
+rsyslog__group: '{{ "syslog" if rsyslog__unprivileged|bool else "root" }}'
+
+# .. envvar:: rsyslog__home
+#
+# The home directory of the :envvar:`rsyslog__user` user, dependent on the OS
+# defaults. Takes effect only when the unprivileged mode is enabled.
+rsyslog__home: '{{ "/var/log" }}'
```
richm: not sure what this is for rhel/fedora/centos - might be same as work_directory

hack/testing/rsyslog/roles/rsyslog/defaults/main.yaml
```diff
@@ -0,0 +1,1509 @@
...
+  - '/var/log/user.log'
+  - '/var/log/daemon.log'
+  - '/var/log/messages'
+  - '/var/log/mail.log'
+  - '/var/log/mail.info'
+  - '/var/log/mail.warn'
+  - '/var/log/mail.err'
+  - '/var/log/cron.log'
+  - '/var/log/lpr.log'
+  - '/var/log/debug'
```
richm: this looks very debian specific

hack/testing/rsyslog/roles/rsyslog/defaults/main.yaml
```diff
@@ -0,0 +1,1509 @@
...
+              property(name="$!@timestamp" dateFormat="rfc3339" position.from="6" position.to="7")
+              constant(value=".")
+              property(name="$!@timestamp" dateFormat="rfc3339" position.from="9" position.to="10")
+          }
+          
+          template(name="index_template" type="string" string="%$.viaq_index_name%")
+          template(name="id_template" type="string" string="%$.viaq_msg_id%")
+          
+          ruleset(name="error_es") {
+              action(type="omfile" template="viaq_template_nl" file="/var/lib/rsyslog/es-bulk-errors.log")
```
richm: {{ rsyslog__work_directory }}/es-bulk-errors.log

richm:
for CI, you'll either need to get the rpm packages onto the machine somehow, or add support for using the rsyslog-container image

richm:
> when I test this patch, I had to create a local repo and put rsyslog-*8.24.0-33.el7.x86_64.rpm and its depending packages...

I'm not sure how we can do that in CI

hack/testing/rsyslog/roles/rsyslog/defaults/main.yaml
nhosoi: @richm, rsyslog__default_logfiles is used in logrotate, which I haven't looked in yet.  (currently, commented out...)  Do we want to support it in this ansible work?

richm: I don't think so - the rsyslog rpm configures logrotate by default

richm: @nhosoi see https://src.fedoraproject.org/rpms/rsyslog/tree/f28 - the logrotate config file is rsyslog.log

richm: @nhosoi also note that rhel 7 and rhel 8 rsyslog are different than this

nhosoi: @richm, yes, it looks rhel7 expects /etc/logrotate.d/syslog and rhel8 (and new fedoras) do /etc/logrotate.d/rsyslog.
```
$ rpm -ql rsyslog | egrep logrotate
/etc/logrotate.d/rsyslog
```
The contents look identical...
```
/var/log/cron
/var/log/maillog
/var/log/messages
/var/log/secure
/var/log/spooler
{
    missingok
    sharedscripts
    postrotate
	/bin/kill -HUP `cat /var/run/syslogd.pid 2> /dev/null` 2> /dev/null || true
    endscript
}
```
This ansible role is hoping to add:
```
          notifempty
          delaycompress
          compress
```
We could add them to the syslog or rsyslog whichever exists on the host?

richm: > We could add them to the syslog or rsyslog whichever exists on the host?

I don't think so.  For now, the rsyslog rpm should be the source of truth for rsyslog.conf and logrotate config for rsyslog.  If anything, by default, deploying the rsyslog playbook should configure the system _exactly_ as it is currently deployed by the rpm.

nhosoi: Ok.  I'm convinced that logrotate should not be touched in this ansible deployment...  I'm removing the logrotate related code from the tasks.

hack/testing/rsyslog/roles/rsyslog/defaults/main.yaml
```diff
@@ -832,8 +832,8 @@ rsyslog__conf_viaq_mmk8s:
           input(type="imfile" file="{{ rsyslog__viaq_log_path }}/*.log" tag="kubernetes" addmetadata="on" reopenOnTruncate="on")
           
           if ((strlen($!CONTAINER_NAME) > 0) and (strlen($!CONTAINER_ID_FULL) > 0)) or
-              ($!metadata!filename startswith "{{ rsyslog__viaq_log_path }}/") then {
-              if $!metadata!filename startswith "{{ rsyslog__viaq_log_path }}/" then {
+              ((strlen($!metadata) > 0) and (strlen($!metadata!filename) > 0) and ($!metadata!filename startswith "/var/log/containers/")) then {
```
richm: Does it crash if you omit the `(strlen($!metadata) > 0)`, and only test for `(strlen($!metadata!filename) > 0)`?  Or does it crash if `$!metadata` is unset?
The reason I ask is because I think rsyslog has to actually convert `$!metadata` to string, so the expensive json-to-string functions.  Same with `(strlen($!kubernetes) > 0)` below.

nhosoi: > Does it crash if you omit the (strlen($!metadata) > 0), and only test for (strlen($!metadata!filename) > 0)?
Yes, it crashes. :(  I tested with ```(strlen($!metadata!filename) > 0)``` only, first, but it did not solve the crash problem.

Regarding $!kubernetes, I just did copy&paste when I ran into the 2nd crash.  So, I haven't tested without ```(strlen($!kubernetes) > 0)```... :p

hack/testing/rsyslog/roles/rsyslog/defaults/main.yaml
```diff
@@ -2,16 +2,35 @@
...
 # .. envvar:: rsyslog__enabled
 #
 # Enable or disable ``rsyslog`` management.
 rsyslog__enabled: True
 
-rsyslog__work_directory: /var/lib/rsyslog
+# .. envvar:: rsyslog__home
+#
+# Top directory to place logs
+rsyslog__home: '{{ "/var/log" }}'
```
sradco: No brackets required .
It can be

rsyslog_home: '/var/log'

Also for all the other params like 
rsyslog_dir, rsyslog_work_directory and so on...

hack/testing/rsyslog/roles/openshift_logging_rsyslog/tasks/main.yaml
sradco: Do you mean until line 70?

hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml
```diff
@@ -0,0 +1,159 @@
...
+    - '{{ rsyslog__default_viaq_rules }}'
+  when: (rsyslog__enabled|bool and rsyslog__viaq|bool and (item.filename|d() or item.name|d()) and
+         (item.state is undefined or item.state != 'absent') and
+         (item.options|d() or item.sections|d()))
+  notify: [ 'Restart rsyslogd' ]
+
+# rsyslog__viaq_rules:
+- name: Generate rsyslog viaq configuration configuration files
+  template:
+    src: 'etc/rsyslog.d/rules.conf.j2'
```
sradco: Need to update /etc/rsyslog.d/ with rsyslog__dir parameter

hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml
```diff
@@ -0,0 +1,159 @@
...
+    - '{{ rsyslog__dependent_rules }}'
+  when: (rsyslog__enabled|bool and rsyslog__example|bool and (item.filename|d() or item.name|d()) and
+         (item.state is undefined or item.state != 'absent') and
+         (item.options|d() or item.sections|d()))
+  notify: [ 'Restart rsyslogd' ]
+
+# rsyslog__default_viaq_rules:
+- name: Generate rsyslog configuration rules
+  template:
+    src: 'etc/rsyslog.d/rules.conf.j2'
```
sradco: Need to update /etc/rsyslog.d/ with rsyslog__dir parameter

hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml
```diff
@@ -0,0 +1,159 @@
...
+    dest: '/etc/rsyslog.conf'
+    owner: 'root'
+    group: 'root'
+    mode: '0400'
+  notify: [ 'Restart rsyslogd' ]
+  when: rsyslog__enabled|bool
+
+- name: Generate rsyslog configuration rules
+  template:
+    src: 'etc/rsyslog.d/rules.conf.j2'
```
sradco: Need to update /etc/rsyslog.d/ with rsyslog__dir parameter

hack/testing/rsyslog/vars.yaml
```diff
@@ -0,0 +1,10 @@
+rsyslog__enabled: true
+# install viaq packages & config files
+rsyslog__viaq: true
+rsyslog__capabilities: [ 'viaq' ]
+rsyslog__group: root
+rsyslog__user: root
+# to share rsyslog__dir with roles/openshift_logging_rsyslog
+rsyslog__dir: /etc/rsyslog.d
+rsyslog__viaq_path: /etc/rsyslog.d/viaq
```
sradco: I believe it should be 
rsyslog__viaq_path: '{{ rsyslog__dir }}/viaq'

test/zzz-rsyslog.sh
```diff
@@ -30,7 +30,7 @@ cleanup() {
     sudo journalctl -u $rsyslog_service --since="-1hour" > $ARTIFACT_DIR/rsyslog-rsyslog.log 2>&1
     if [ -n "${rsyslog_save}" -a -d "${rsyslog_save}" ] ; then
         sudo rm -rf /etc/rsyslog.d/*
```
sradco: Should this and the raw bellow use the path from the params as well?

hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml
```diff
@@ -0,0 +1,159 @@
+---
+- name: Install required packages
```
sradco: This installs the packages and not affected by the "use_rsyslog_image" parameter. Should the when condition be:
when: rsyslog__enabled|bool and not use_rsyslog_image|bool  ?

sradco: Are the packages in the container image need to be installed or they are already installed and this should not be run when use_rsyslog_image is set to true?

Are the steps below relevant when we use the container image?

hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml
```diff
@@ -0,0 +1,159 @@
...
+  yum:
+    name: '{{ item }}'
+    state: 'present'
```
sradco: Should we also try to update the packages or only have them present?

hack/testing/rsyslog/roles/openshift_logging_rsyslog/tasks/setup_certs.yaml
```diff
@@ -0,0 +1,106 @@
...
+    name: rsyslog.service
+    enabled: no
+    state: stopped
+  when: use_rsyslog_image | default(False)
+
+- name: install rsyslog container systemd service
+  template: src=rsyslog-container.service.j2 dest=/usr/lib/systemd/system/rsyslog-container.service mode=0400
+  when: use_rsyslog_image | default(False)
+
+- name: enable and run rsyslog-container.service
```
sradco: I think you are missing disabling rsyslog-container.service when: not use_rsyslog_image | default(False)

hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml
richm: update too

hack/testing/rsyslog/roles/openshift_logging_rsyslog/tasks/main.yaml
richm: No, I mean these tasks - the `oc` command is only guaranteed to be present on the master nodes with the typical openshift installation

hack/testing/rsyslog/roles/rsyslog/defaults/main.yaml
nhosoi: Yep, you are right.  It's mistakenly brought in by the porting effort of debops...  Fixing them.

hack/testing/rsyslog/vars.yaml
nhosoi: That was my first intention, but ansible did not like to have {{ ... }} in vars.yaml. :(

hack/testing/rsyslog/roles/openshift_logging_rsyslog/tasks/main.yaml
sradco: There is no line 79...
Only 70 lines in the file.

hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml
sradco: So this should be updated to latest

hack/testing/rsyslog/vars.yaml
sradco: Not sure why. It should work. Maybe you were missing '{{ }}' ?

hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml
nhosoi: So, this task is good, right?
Had I better change the "name" to "Install/Update required packages"?

hack/testing/rsyslog/roles/openshift_logging_rsyslog/tasks/main.yaml
richm: hmm - ok - then - Any task which uses `oc` must be executed on node `masters[0]`

hack/testing/rsyslog/roles/rsyslog/defaults/main.yaml
```diff
@@ -2,16 +2,35 @@
...
+
+# .. envvar:: rsyslog__viaq_path
+#
+# Viaq configuration directory
+rsyslog__viaq_path: '{{rsyslog__dir}}/viaq'
+
+# .. envvar:: rsyslog__viaq_path
+#
+# Viaq log directory
+rsyslog__viaq_log_path: '{{ "{{rsyslog__home}}/containers" }}'
```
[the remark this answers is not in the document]
nhosoi: Ok!  I'm replacing:
rsyslog__home with rsyslog__systemlog_dir, and
rsyslog__dir with rsyslog__config_dir.

hack/testing/rsyslog/vars.yaml
richm: try
```
 rsyslog__viaq_path: "{{ rsyslog__dir }}/viaq"
```
note the double quotes - https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#hey-wait-a-yaml-gotcha

hack/testing/rsyslog/roles/openshift_logging_rsyslog/tasks/setup_certs.yaml
nhosoi: Thanks!  Adding it...

hack/testing/rsyslog/roles/openshift_logging_rsyslog/tasks/main.yaml
nhosoi: I run the openshift related tasks only on master[0] here in the playbook (at least, I meant... :)
https://github.com/nhosoi/origin-aggregated-logging/blob/rsyslog_deploy/hack/testing/rsyslog/playbook.yaml#L11-L16

sradco: I'm not familier with this syntax in the playbook.
When I usually want to run once on a specific machine I either have a new separate play for the specific hosts 
or delegate the tasks to the machine and add run_ once: true flag to the tasks.

Does the syntax you are using work for you?

nhosoi: Well, it was not working as I expected... :(
The current source tree passes the rsyslog ci-test.  So, I did not notice this warning was issued from ansible.
``` [WARNING]: Ignoring invalid attribute: hosts```
Obviously, ```hosts: master[0]``` was ignored.  I'm going to follow your suggestion.  
Thanks, @sradco !

File: hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml (hunk 2 above)
Comment (nhosoi) [reply hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml]: I've added ```owner``` and ```group```.

It seems /var/spool/rsyslog used to be used for a WorkDirectory.
```$WorkDirectory /var/spool/rsyslog  # default location for work (spool) files```
We use ```/var/lib/rsyslog``` by default, which is owned by the rsyslog package:
```
$ rpm -qf /var/lib/rsyslog
rsyslog-8.24.0-33.el7.x86_64
```
Note: we created the work directory (rsyslog__work_dir) only when the owner of rsyslog was not root. 
``` 
when: rsyslog__unprivileged|bool and rsyslog__user != 'root'
```
I think we should not have the condition here since if I deploy rsyslog as a root and set the work dir to somewhere else like /var/spool/rsyslog, which won't be created...

File: test/zzz-rsyslog.sh (hunk 13 above)
Comment (nhosoi) [reply test/zzz-rsyslog.sh]: Ok.  I'm updating the test script zzz-rsyslog.sh to generate vars file to sync the parameter.

File: hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml (hunk 15 above)
Comment (nhosoi) [reply hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml]: I see.  Done.  Thanks!

File: hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml (hunk 2 above)
Comment (richm) [reply hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml]: > It seems /var/spool/rsyslog used to be used for a WorkDirectory.

That's a debian-ism - default on rhel/fedora/centos is `/var/lib/rsyslog`

> We use /var/lib/rsyslog by default, which is owned by the rsyslog package:

So by default, we should use the same mode, owner, and group that the rsyslog package uses.  That way you will get exactly the same thing if you run the ansible playbook after installing the rpm rsyslog package.

File: hack/testing/rsyslog/roles/openshift_logging_rsyslog/tasks/main.yaml (hunk 1 above)
Comment (richm) [reply hack/testing/rsyslog/roles/openshift_logging_rsyslog/tasks/main.yaml]: Where did you use hosts? I would recommend keeping with the openshift way of doing things for the openshift specific playbooks.

File: hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml (hunk 2 above)
Comment (nhosoi) [reply hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml]: Yep, we don't change the mode, owner and group of the directory.

File: hack/testing/rsyslog/roles/openshift_logging_rsyslog/tasks/main.yaml (hunk 1 above)
Comment (nhosoi) [reply hack/testing/rsyslog/roles/openshift_logging_rsyslog/tasks/main.yaml]: I think these 2 lines would do the trick.  Could you please review it?
https://github.com/nhosoi/origin-aggregated-logging/blob/rsyslog_deploy/hack/testing/rsyslog/playbook.yaml#L15-L16

Comment (richm) [reply hack/testing/rsyslog/roles/openshift_logging_rsyslog/tasks/main.yaml]: something like this: https://gist.github.com/richm/f76f723ebd1ab1ec8d34f5b225abed20

That is, before doing the main rsyslog playbook, do the openshift specific bits to get the certs, tokens, etc.

In the future, we'll have `rsyslog_` or `logging_` ansible variables for certs, tokens, etc. e.g. `logging_mmk8s_token`, `logging_mmk8s_cacert`, etc. that we will specify in an inventory or vars.yaml before passing to the rsyslog playbook.

Comment (nhosoi) [reply hack/testing/rsyslog/roles/openshift_logging_rsyslog/tasks/main.yaml]: Thanks, @richm .  Verified your playbook.yaml and pushed it.

Comment (nhosoi) [reply hack/testing/rsyslog/roles/openshift_logging_rsyslog/tasks/main.yaml]: And added these variables...
+logging_mmk8s_token: "{{rsyslog__viaq_path}}/mmk8s.token"
+logging_mmk8s_ca_cert: "{{rsyslog__viaq_path}}/mmk8s.ca.crt"
+logging_elasticsearch_ca_cert: "{{rsyslog__viaq_path}}/es-ca.crt"
+logging_elasticsearch_cert: "{{rsyslog__viaq_path}}/es-cert.pem"
+logging_elasticsearch_key: "{{rsyslog__viaq_path}}/es-key.pem"

File: hack/testing/rsyslog/README.md [hunk 18]
```diff
@@ -0,0 +1,64 @@
...
+Sample vars.yaml to deploy viaq scripts 
+```
+rsyslog__enabled: true
+# install viaq packages & config files
+rsyslog__viaq: true
+rsyslog__group: root
+rsyslog__user: root
+rsyslog__capabilities: viaq
+```
+The following configuration files are deployed in /etc by the set of variables.
```
sradco: I'm probably missing something, but where are the configuration files that should be deployed actually get deployed?

hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml
```diff
@@ -0,0 +1,159 @@
+---
+- name: Install/Update required packages
+  yum:
+    name: '{{ item }}'
+    state: 'latest'
+  with_flattened:
+    - '{{ rsyslog__base_packages }}'
+    - '{{ rsyslog__viaq_prereq_packages if rsyslog__viaq|bool else [] }}'
+    - '{{ rsyslog__viaq_packages if rsyslog__viaq|bool else [] }}'
```
sradco: Does this package place the configuration files for viaq?
If yes, where is the repo for it ?

hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml
```diff
@@ -0,0 +1,159 @@
...
+    group: '{{ item.group | d("root") }}'
+    mode:  '{{ item.mode  | d("0400") }}'
+  with_flattened:
+    - '{{ rsyslog__default_viaq_rules }}'
+  when: (rsyslog__enabled|bool and rsyslog__viaq|bool and (item.filename|d() or item.name|d()) and
+         (item.state is undefined or item.state != 'absent') and
+         (item.options|d() or item.sections|d()))
+  notify: [ 'Restart rsyslogd' ]
+
+# rsyslog__viaq_rules:
```
sradco: What is the difference between this task and the one above it?

hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml
```diff
@@ -0,0 +1,159 @@
...
+- name: Generate rsyslog viaq configuration configuration files
```
sradco: is having "configuration configuration" intentional?

hack/testing/rsyslog/README.md
nhosoi: I'm in the middle of rewriting README.md.  
Please give me some time for reviewing it...  Thanks.

hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml
nhosoi: The configuration templates are these 2 files.  They are shared among all deployment.
```
https://github.com/nhosoi/origin-aggregated-logging/blob/rsyslog_deploy/hack/testing/rsyslog/roles/rsyslog/templates/etc/rsyslog.conf.j2
https://github.com/nhosoi/origin-aggregated-logging/blob/rsyslog_deploy/hack/testing/rsyslog/roles/rsyslog/templates/etc/rsyslog.d/rules.conf.j2
```
If the question is about the rpm packages, the list is in the defaults/main.yaml:
```
https://github.com/nhosoi/origin-aggregated-logging/blob/rsyslog_deploy/hack/testing/rsyslog/roles/rsyslog/defaults/main.yaml#L56-L69
```
If "the repo" is about the yum repo, it is not maintained in this roles.  (Maybe, I might have misunderstood your questions...?)

hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml
nhosoi: Ahhh, a good catch. Thanks!

hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml
```diff
@@ -0,0 +1,159 @@
...
+    - '{{ rsyslog__group_rules }}'
+    - '{{ rsyslog__host_rules }}'
+    - '{{ rsyslog__dependent_rules }}'
+  when: (rsyslog__enabled|bool and rsyslog__example|bool and (item.filename|d() or item.name|d()) and
+         (item.state is undefined or item.state != 'absent') and
+         (item.options|d() or item.sections|d()))
+  notify: [ 'Restart rsyslogd' ]
+
+# rsyslog__default_viaq_rules:
+- name: Generate rsyslog configuration rules
```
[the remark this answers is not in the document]
nhosoi: > Why do we include them in separate includes and not include the entire directory?

Hmmm, I created this task at the early stage of my porting.  Yes, ```https://github.com/nhosoi/origin-aggregated-logging/blob/rsyslog_deploy/hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml#L111-L124``` may not be needed independently.  Let me retest it.

> Where can I see the rsyslog__default_viaq_rules templates?

Basically, all the configuration files under rsyslog.d share this template.
 https://github.com/nhosoi/origin-aggregated-logging/blob/rsyslog_deploy/hack/testing/rsyslog/roles/rsyslog/templates/etc/rsyslog.d/rules.conf.j2

richm:
@stevekuznetsov any idea why we are not getting the logs and artifacts?  https://openshift-gce-devel.appspot.com/build/origin-/pr-logs/pull/openshift_origin-aggregated-logging/1314/test_pull_request_origin_aggregated_logging_json_file/999/

hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml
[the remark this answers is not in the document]
nhosoi: Unfortunately, I could not merge the two tasks into one...
This roles currently support 3 configuration types (more precisely 4, but the 4th does not work.  please bare with me...): rsyslog__example: true, rsyslog__viaq: true, and both are false.  I've updated the task names with more detailed info (hopefully).

hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml
richm: @sradco https://github.com/openshift/origin-aggregated-logging/blob/master/hack/testing/rsyslog/mmk8s.conf.j2#L12

This will parse a single line JSON string into individual fields in the record

https://github.com/openshift/origin-aggregated-logging/blob/master/hack/testing/rsyslog/viaq_formatting.conf#L15

This will parse a JSON valued string field in the record called `$!MESSAGE` into individual fields in the record

Here is the parse_json rulebase: https://github.com/openshift/origin-aggregated-logging/blob/master/hack/testing/rsyslog/parse_json.rulebase

In general, rsyslog tries to avoid using regexp to parse, because regexp are slow and expensive.  Instead, they have their own high performance log parsing and normalizing library liblognorm http://www.liblognorm.com/ and their own high performance json parsing and formatting library https://github.com/rsyslog/libfastjson/

The rest of the file https://github.com/openshift/origin-aggregated-logging/blob/master/hack/testing/rsyslog/viaq_formatting.conf shows how to convert journald and kubernetes file logs into the viaq common data model format.

For elasticsearch - https://github.com/openshift/origin-aggregated-logging/blob/master/hack/testing/rsyslog/elasticsearch.conf.j2#L55
The template field controls how the record is formatted into the JSON blob to send in the bulk index request
```
 template="viaq_template"
```
The template is defined here: https://github.com/openshift/origin-aggregated-logging/blob/master/hack/testing/rsyslog/elasticsearch.conf.j2#L1
The `$!all-json-plain` is a special property which means "format all of the message fields in $! as a json string in the usual JSON format".

stevekuznetsov:
@richm @nhosoi can you please submit an [issue](https://github.com/openshift/release/issues/new) detailing what happened with details? DPTP will look into it

nhosoi:
CI-tests fail unless it has an access to the repo containing the latest rsyslog rpm packages including rsyslog-mmkubernetes...
```
TASK [rsyslog : Install/Update required packages] ******************************
task path: /data/src/github.com/openshift/origin-aggregated-logging/hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml:2
...
failed: [localhost] (item=[u'rsyslog', u'nmap-ncat', u'systemd-python', u'policycoreutils', u'checkpolicy', u'policycoreutils-python', u'rsyslog-elasticsearch', u'rsyslog-mmjsonparse', u'rsyslog-mmkubernetes', u'rsyslog-mmnormalize'])
```

richm:
if `use_rsyslog_image=True` then the task `rsyslog : Install/Update required packages` should be skipped.

nhosoi:
Ok.  That's the bug in playbook/roles...  Let me fix it.  Thanks!

hack/testing/rsyslog/playbook.yaml
```diff
@@ -97,153 +45,13 @@
...
-    when: use_rsyslog_image | default(False)
-
-  - name: reload and restart rsyslog.service
-    systemd:
-      name: rsyslog.service
-      daemon_reload: yes
-      enabled: yes
-      state: restarted
-    when: not use_rsyslog_image | default(False)
+  tasks:
```
richm: you don't need `tasks` here since this is already part of the task list of the play `- name: install and configure rsyslog on the nodes` - you are getting the following error: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_origin-aggregated-logging/1314/test_pull_request_origin_aggregated_logging_json_file/1021/artifacts/scripts/test-zzz-rsyslog/artifacts/zzz-rsyslog-ansible.log
```
 [WARNING]: While constructing a mapping from /data/src/github.com/openshift
/origin-aggregated-logging/hack/testing/rsyslog/playbook.yaml, line 32, column
3, found a duplicate dict key (tasks). Using last defined value only.
```
not sure where the `roles` should go, before or after the list of tasks

hack/testing/rsyslog/roles/rsyslog/tasks/main.yaml
```diff
@@ -9,7 +9,7 @@
     - '{{ rsyslog__viaq_packages if rsyslog__viaq|bool else [] }}'
     - '{{ rsyslog__tls_packages if rsyslog__pki|bool else [] }}'
     - '{{ rsyslog__packages }}'
-  when: rsyslog__enabled|bool
+  when: rsyslog__enabled|bool and use_rsyslog_image|bool
```
richm: sorry, this should be `not use_rsyslog_image|default(False)|bool` - that is, if you are using the image, you do not want to install the rsyslog packages such as rsyslog, rsyslog-mmnormalize, etc.

richm:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_origin-aggregated-logging/1314/test_pull_request_origin_aggregated_logging_json_file/1025/artifacts/scripts/test-zzz-rsyslog/artifacts/zzz-rsyslog-ansible.log
```
ERROR! Syntax Error while loading YAML.
  did not find expected key

The error appears to have been in '/data/src/github.com/openshift/origin-aggregated-logging/hack/testing/rsyslog/playbook.yaml': line 52, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:


  - name: Run tasks/setup_certs.yaml instead of 'main'
  ^ here
```

hack/testing/rsyslog/README.md
```diff
@@ -0,0 +1,224 @@
...
+TASK [rsyslog : Generate exaple configuration files in rsyslog.d] 
+```
+WARNING: Pre-existing rsyslog.conf and configuration files in /etc/rsyslog.d are moved to the backup directory /tmp/rsyslog.d-XXXXXX.  If the pre-existing files need to be merged with the newly deployed files, you need to do it manually.
+
+-defaults/main.yaml defines variables to switch the deployment paths, variables to specify the locations to deploy and the configurations to be deployed.
+
+Describing how the configuration files are defined to be deployed using the viaq case.
+
+Viaq configuration files are defined in {{rsyslog__viaq_rules}} in defaults/main.yaml.  The set is made from the generic modules{rsyslog__conf_global_options, rsyslog__conf_local_modules, rsyslog__conf_network_modules, rsyslog__conf_network_modules} and viaq specific configurations.  
+
```
sradco: You have here "rsyslog__conf_network_modules" twice

nhosoi: Thank you, @sradco !  Indeed...  It was a copy&paste error. :p  Fixed it.

richm:
on hold until 4.0 is open - just want to make sure this doesn't somehow destabilize the 3.11 tests (even though CI tests are passing)

richm:
however, I think the non-openshift specific parts should be submitted as a PR against https://github.com/linux-system-roles/logging

sradco:
I believe the OpenShift specific parts can remain as long as they are set by variables, which I think they are no? @richm

nhosoi:
@richm,
> however, I think the non-openshift specific parts should be submitted as a PR against https://github.com/linux-system-roles/logging

Trying to port hack/testing/rsyslog/roles/rsyslog to logging/roles/rsyslog.  I have one question.  This [1] is a snippet from vars.yaml and they are referred from the config files.  In the ci-test, roles in openshift_logging_rsyslog get and place them in the expected location.  Without openshift_logging_rsyslog, what we should do to deploy them?  For instance, shall we a task to check some location and if they are there, copy them to  {{rsyslog__viaq_config_dir}}?
```
[1]
logging_mmk8s_token: "{{rsyslog__viaq_config_dir}}/mmk8s.token"
logging_mmk8s_ca_cert: "{{rsyslog__viaq_config_dir}}/mmk8s.ca.crt"
logging_elasticsearch_ca_cert: "{{rsyslog__viaq_config_dir}}/es-ca.crt"
logging_elasticsearch_cert: "{{rsyslog__viaq_config_dir}}/es-cert.pem"
logging_elasticsearch_key: "{{rsyslog__viaq_config_dir}}/es-key.pem"
```

richm:
> I believe the OpenShift specific parts can remain as long as they are set by variables, which I think they are no?

Yes. I meant any code which uses the `oc` command to extract data from openshift should not be in the linux system role.

> Without openshift_logging_rsyslog, what we should do to deploy them? For instance, shall we a task to check some location and if they are there, copy them to {{rsyslog__viaq_config_dir}}?

Note that for an external viaq client such as RHV or OpenStack, they will not need the logging_mmk8s_* parameters.

> For instance, shall we a task to check some location

It depends.  If we have a flag that says "I want to use elasticsearch output with client cert auth" then we can check for ca cert, cert, and key.  But we will want to support cases where the user wants to output to elasticsearch with no auth at all.  We could do some simple checks like the user has provided a client cert with no key, or vice versa.

nhosoi:
@richm , @sradco ,
I've submitted a PR: https://github.com/linux-system-roles/logging/pull/1

PR #1325, opened 2018-08-28
Fluentd support to log into file and rotate it

VladMasarik:
New feature ticket LOG-30 done by:
- Replacing fluentd source files by customized ones which do the following.
- Using environment variables "LOGGING_FILE_SIZE, LOGGING_FILE_PATH, LOGGING_FILE_AGE" as options for fluentd. If no "LOGGING_FILE_PATH" environment variable is set, the logs will be outputted as usually.
- Created global variable "openShiftLogger" which is used to output logs into the specified file and rotate them.
- There are 2 "Log" class methods("puts", "write") that are used for writing the logs into the STDOUT. By redirecting them to the "openShiftLogger" (when set to exist), all the logs are written into file and rotated when needed.


fluentd/Dockerfile.centos7
```diff
@@ -7,7 +7,10 @@ ENV DATA_VERSION=1.6.0 \
     GEM_HOME=/opt/app-root/src \
     HOME=/opt/app-root/src \
     PATH=/opt/app-root/src/bin:/opt/app-root/bin:$PATH \
-    RUBY_VERSION=2.0
+    RUBY_VERSION=2.0 \
+    LOGGING_FILE_PATH=/opt/app-root/src/fluentd.log \
```
jcantrill: This should default to the same volume used by fluent to store its buffers.

fluentd/logger/log.rb
```diff
@@ -0,0 +1,384 @@
+#
```
jcantrill: Lets move this into the fluentd/patch/logger directory to better identify what it is

fluentd/logger/supervisor.rb
```diff
@@ -0,0 +1,582 @@
+#
```
jcantrill: same here. please move

fluentd/logger/supervisor.rb
```diff
@@ -0,0 +1,582 @@
...
+      if group.to_i.to_s == group
+        Etc.getgrgid(group.to_i)
+      else
+        Etc.getgrnam(group)
+      end
+    end
+
+    class LoggerInitializer
+      def initialize(path, level, chuser, chgroup, opts)
+        envPath = ENV["LOGGING_FILE_PATH"]
```
jcantrill: Should provide defaults here for safety.  maybe `ENV["LOGGING_FILE_PATH"]||/var/log/fluentd/`

fluentd/logger/supervisor.rb
```diff
@@ -0,0 +1,582 @@
...
+        $openShiftLogger = nil
+        if !envPath.nil?
+          path = envPath
+          age = Integer(ENV["LOGGING_FILE_AGE"])
```
jcantrill: provide default

fluentd/logger/supervisor.rb
```diff
@@ -0,0 +1,582 @@
...
+      else
+        Etc.getgrnam(group)
+      end
+    end
+
+    class LoggerInitializer
+      def initialize(path, level, chuser, chgroup, opts)
+        envPath = ENV["LOGGING_FILE_PATH"]
+        $openShiftLogger = nil
+        if !envPath.nil?
```
jcantrill: This if block wont be needed if we default.  for future ref, this is also valid `unless envPath.nil?`

fluentd/logger/log.rb
```diff
@@ -0,0 +1,384 @@
...
+      dump_stacktrace(backtrace, LEVEL_FATAL)
+    end
+
+    def puts(msg)
+      if $openShiftLogger.nil?  
+        @out.puts(msg)
+        @out.flush
+        msg
+      else
+        $openShiftLogger.info("#{msg}")
```
richm: `$openShiftLogger.info(msg)`
no need to convert `msg` to a string - it should already be a string

richm: https://ruby-doc.org/stdlib-2.0.0/libdoc/logger/rdoc/Logger.html
I think you should use the `<<` method - it looks like the message is already formatted - not sure.  If the message is not formatted, then you'll need to configure the formatter to format the messages exactly like fluentd does.  Same with `write` below.

fluentd/Dockerfile.centos7
VladMasarik: Is that `/var/log/fluentd/` or `/var/log/` or `/var/lib/fluentd/`?

fluentd/logger/supervisor.rb
VladMasarik: Won't the user want fluentd to write logs to `STDOUT`?

fluentd/Dockerfile.centos7
jcantrill: So maybe not [1] since this is in `/var/lib`.  We should defer then to using conventions: `/var/log/fluentd`

[1] https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/run.sh#L207

fluentd/Dockerfile.centos7
```diff
@@ -7,7 +7,10 @@ ENV DATA_VERSION=1.6.0 \
     GEM_HOME=/opt/app-root/src \
     HOME=/opt/app-root/src \
     PATH=/opt/app-root/src/bin:/opt/app-root/bin:$PATH \
-    RUBY_VERSION=2.0
+    RUBY_VERSION=2.0 \
+    LOGGING_FILE_PATH=/var/log/fluentd \
```
jcantrill: if 'console'   write to stdout

jcantrill:
This will have to wait till branching in order to merge.  Also, please squash once you are satisfied with the reviews you have received

fluentd/patch/logger/supervisor.rb
```diff
@@ -56,9 +56,9 @@ def initialize(path, level, chuser, chgroup, opts)
         $openShiftLogger = nil
         if !envPath.nil?
           path = envPath
-          age = Integer(ENV["LOGGING_FILE_AGE"])
-          size = Integer(ENV["LOGGING_FILE_SIZE"])
-          $openShiftLogger = Logger.new(path, age, size)
+          age = Integer(ENV["LOGGING_FILE_AGE"]) || 10
```
richm: You can use `irb` to test out small pieces of ruby code:
```
irb
irb(main):005:0> Integer(nil)
TypeError: can't convert nil into Integer
        from (irb):5:in `Integer'
        from (irb):5
        from /usr/bin/irb:12:in `<main>'
irb(main):006:0> Integer(ENV['FOO'])
TypeError: can't convert nil into Integer
        from (irb):6:in `Integer'
        from (irb):6
        from /usr/bin/irb:12:in `<main>'
irb(main):007:0> Integer(ENV['FOO'] || 10)
=> 10
```
so you need to use
```
          age = Integer(ENV["LOGGING_FILE_AGE"] || 10)
```
to avoid a possible exception

jcantrill: or `(ENV["LOGGING_FILE_AGE"] || 10).to_i`

VladMasarik:
> I think we are closing this PR in favor of #1336 ?

Yes I believe it would be the best. Shall we also delete it?

PR #1330, opened 2018-08-31
[3.11] More readable format for logs from curator config generator

josefkarasek:


jcantrill:
/lgtm

@josefkarasek no need to cherrypick pack to 3.11 since they are fastforwarding those branches

PR #1338, opened 2018-09-10
ignore red cluster when seeding

jcantrill:
This PR adds a switch to ignore red clusters when seeding which we have seen on some of the online and dedicated clusters where the seed script hangs on YELLOW.  This looks to unblock the health probe

### DIFF
--- a/elasticsearch/utils/es_seed_acl
+++ b/elasticsearch/utils/es_seed_acl
@@ -23,6 +23,7 @@ function sgadmin {
     -tst JKS \
     -tspass tspass \
     -nhnv \
+    -arc \
     -icl \
     ${DEBUG:+-dg} ${@:-}
 }


elasticsearch/utils/es_seed_acl
```diff
@@ -23,6 +23,7 @@ function sgadmin {
     -tst JKS \
     -tspass tspass \
     -nhnv \
+    -arc \
```
richm: Should this be conditional, or is it ok to ignore red cluster at all times?

jcantrill: I was wondering the same thing but I can't honestly say.  Red simply means you dont have placed replicas for the cluster as a whole.  ES should be smart enough to figure out how to reconcile changes if there are multiple replicas for this particular index.

ewolinetz: The SG admin tool is also smart enough that it won't try to create this index if it already exists.

PR #1341, opened 2018-09-12
increase default reload_after to 200

richm:
increase default reload_after to 200


ewolinetz:
@richm is there and performance recommendation that supports `200` over the plugin default of `10000`?

richm:
@ewolinetz yes, from the AOS scale team - see https://bugzilla.redhat.com/show_bug.cgi?id=1489533#c16

PR #1349, opened 2018-09-18
add tools from jcantrill/openshift-logging-tools

richm:
add tools from jcantrill/openshift-logging-tools
https://github.com/jcantrill/openshift-logging-tools

### DIFF
--- a/hack/tools/allocate-all-shards
+++ b/hack/tools/allocate-all-shards
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+# This script allocates all unassigned primary shards to a given node using
+# the openshift binary.  The binary must be in the path and the user executing the script
+# must have access to logging project.  The inputs are:
+#   pod   An Elasticsearch pod name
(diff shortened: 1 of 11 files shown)


jcantrill:
@richm do we want to submodule the repo instead?  might make it easier to pull it in as they update

richm:
> @richm do we want to submodule the repo instead? might make it easier to pull it in as they update

who is "they"?

jcantrill:
@richm  ...'we'

richm:
I don't understand - wouldn't it be easier for us to contribute PRs directly against the single shared openshift/origin-aggregated-logging repo?

jcantrill:
Closing this PR.  My preference is to either submodule them in or create an openshift repo.  The issue with either of these implies there is some official support which I'm not sure we want to take on .  Additionally, I like the idea of not pulling them in here because my experience has been to simply clone them onto a target host which is significantly easier with a dedicated repo vs all of logging

PR #1350, opened 2018-09-19
Including fluent-plugin-concat to support cri-o multiline logs.

nhosoi:
Backporting the partial code in fluent-plugin-concat v2.3.0 to v1.0.0.
Note: removing @seperator = '\n' from join.  It adds '\n' at the end of
      the partial message.  The original code has the issue.
  - @key => lines.join(@separator)
  + @key => lines.join

### DIFF
--- a/fluentd/lib/generate_throttle_configs/lib/generate_throttle_configs.rb
+++ b/fluentd/lib/generate_throttle_configs/lib/generate_throttle_configs.rb
@@ -171,7 +171,6 @@ def create_default_container_input(input_conf_file, excluded, log, options={})
 <source>
   @type tail
   @id container-input
-  @label @INGRESS
   path "#{options[:cont_logs_path] || cont_logs_path}"
   pos_file "#{options[:cont_pos_file] || cont_pos_file}"
   tag kubernetes.*
@@ -179,7 +178,20 @@ def create_default_container_input(input_conf_file, excluded, log, options={})
   keep_time_key true
   read_from_head "#{options[:read_from_head] || read_from_head}"
   exclude_path #{excluded}
+  @label @CONCAT
 </source>
+<label @CONCAT>
+  <filter kubernetes.**>
+    @type concat
+    key log
+    partial_key logtag
+    partial_value P
+  </filter>
+  <match kubernetes.**>
+    @type relabel
+    @label @INGRESS
+  </match>
+</label>
     CONF
   }
     log.debug "Created default container input config file"
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,3 +32,14 @@ matrix:
         cd fluentd/lib/parser_json_or_crio
       script:
         rake test
+    - name: fluentd - filter_concat test
+      env:
+        - FLUENTD_VERSION=0.12.0
+      language: ruby
+      rvm:
+      - 2.0
+      gemfile: fluentd/lib/filter_concat/Gemfile
+      before_script:
+        cd fluentd/lib/filter_concat
+      script:
+        rake test
--- a/fluentd/Dockerfile
+++ b/fluentd/Dockerfile
@@ -55,6 +55,7 @@ ADD parser_viaq_docker_audit.rb viaq_docker_audit.rb /etc/fluent/plugin/
 ADD run.sh generate_syslog_config.rb ${HOME}/
 ADD lib/generate_throttle_configs/lib/*.rb ${HOME}/
 ADD lib/filter_parse_json_field/lib/*.rb /etc/fluent/plugin/
+ADD lib/filter_concat/lib/filter_concat.rb /etc/fluent/plugin/
 ADD patch/in_prometheus.rb ${GEM_HOME}/gems/fluent-plugin-prometheus-0.4.0/lib/fluent/plugin/
 ADD lib/parser_json_or_crio/lib/*.rb /etc/fluent/plugin/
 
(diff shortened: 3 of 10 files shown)


fluentd/lib/generate_throttle_configs/lib/generate_throttle_configs.rb
```diff
@@ -171,15 +171,27 @@ def create_default_container_input(input_conf_file, excluded, log, options={})
...
   read_from_head "#{options[:read_from_head] || read_from_head}"
   exclude_path #{excluded}
+  @label @CONCAT
 </source>
+<label @CONCAT>
+  <filter kubernetes.**>
+    @type concat
+    key log
+    partial_key logtag
+    partial_value P
```
richm: when reading from a docker json-file log, and there is no field `logtag`, what happens?  I'm assuming it just passes the record through.

nhosoi: Good question.  I ran the ci-test with "-e openshift_use_crio=false ", i.e., with docker.  The tests themselves passed.  But "oc logs FLUENTDPOD" returns nothing, which is unusual.  I was suspecting too many errors/warnings might have been dumped.

This is the line where the logtag value is checked.  It looks to me it causes an exception...
https://github.com/nhosoi/origin-aggregated-logging/blob/crio-0.12/fluentd/filter_concat.rb#L134

I'm going to test this modified code.
```
--- a/fluentd/filter_concat.rb
+++ b/fluentd/filter_concat.rb
@@ -131,7 +131,7 @@ module Fluent
         end
       when :partial
         @buffer[stream_identity] << [tag, time, record]
-        unless @partial_value == record[@partial_key]
+        unless record.key?(@partial_key) || @partial_value == record[@partial_key]
           new_time, new_record = flush_buffer(stream_identity)
           time = new_time if @use_first_timestamp
           new_record.delete(@partial_key)
```

richm: no, the old code should be fine:
```
        unless @partial_value == record[@partial_key]
```
`record['logtag']` returns `nil` - in ruby, if you access a Hash by a non-existent key, `nil` is returned by default: http://ruby-doc.org/core-2.0.0/Hash.html#method-c-new
It is only a problem if `record` is not a Hash (and that will cause many more problems . . .)

richm: since `"P" == nil` will always return `false`, you will always go through the `unless` block

richm: > "oc logs FLUENTDPOD" returns nothing,

do you have a log file for the fluentd pod under `/var/log/containers/`?  Or, if using journald, do you have a journalctl -o export log matching `CONTAINER_NAME=.*fluentd`?

nhosoi: Good to hear about the original code.  I'll revert it.
Sorry, I've already terminated the instance...  
I'm running with "LOG_DRIVER=json-file".

nhosoi: Yep, I do.  And every time fluentd pod is restarted, an empty log file logging-fluentd-*.log is created like this...
```-rw-r-----. 1 root root 0 Sep 19 21:58 /var/log/containers/logging-fluentd-lpq4s_openshift-logging_fluentd-elasticsearch-136d496ae499b34e415a6e998fa9b95dda1e5171494cb007a02d60686511f96c.log```
Next, I set DEBUG=true.
```oc set env ds/logging-fluentd DEBUG=true```
but still the log is empty...
```
# oc rsh logging-fluentd-k27xq
sh-4.2# ls -l /var/log/fluentd.log 
-rw-r--r--. 1 root root 0 Sep 19 22:04 /var/log/fluentd.log
```

**Comment by richm** [comment]
try /var/log/fluentd/fluentd.log

File: fluentd/filter_concat.rb [hunk 2]
```diff
@@ -0,0 +1,254 @@
+module Fluent
```
jcantrill: We should move this into the `lib` dir so that we can add tests if needed based on any changes we need.  I assume this is otherwise a straight copy

nhosoi:
The unit test has come back to work for me...  *puzzled*
```
$ GEM_HOME=.vendor bundle exec rake test
(in /home/nhosoi/origin-aggregated-logging/fluentd/lib/filter_concat)
Loaded suite /home/nhosoi/.gem/ruby/gems/rake-11.3.0/lib/rake/rake_test_loader
Started
....
Finished in 3.086177988 seconds.
---------------------------------------------------------------------------------------------------------------------
4 tests, 7 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
```
-------------------------------------------------------------------

nhosoi:
@richm , @jcantrill , could you please review this pr once again?   Thanks!

richm:
@nhosoi did you submit a manual cherrypick pr for this one for release-3.11?

nhosoi:
@richm,
>  did you submit a manual cherrypick pr for this one for release-3.11?
I've just opened this pr for 3.11.
https://github.com/openshift/origin-aggregated-logging/pull/1386

PR #1351, opened 2018-09-20
make tests pass when using crio

richm:
make tests pass when using crio
couple of minor changes in viaq-data-model and rsyslog ansible
@nhosoi

### DIFF
--- a/test/viaq-data-model.sh
+++ b/test/viaq-data-model.sh
@@ -98,7 +98,7 @@ cat > $cfg <<EOF
     undefined4 undefined4
     undefined5 undefined5
   </record>
-  remove_keys CONTAINER_TAG
+  remove_keys CONTAINER_TAG,logtag
 </filter>
 EOF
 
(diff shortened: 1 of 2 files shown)


richm:
> Does this require the multiline changes to go in?

No

> Or does that rely on this PR?

No, they are independent.

I guess we've been lucky with our crio testing, that our CI tests do not generate multi-line crio logs . . .

test/viaq-data-model.sh
```diff
@@ -98,7 +98,7 @@ cat > $cfg <<EOF
     undefined4 undefined4
     undefined5 undefined5
   </record>
-  remove_keys CONTAINER_TAG
+  remove_keys CONTAINER_TAG,logtag
```
nhosoi: I've pushed this patch to https://github.com/openshift/origin-aggregated-logging/pull/1350, which adds missing keep_partial_key.  Since the default value is false, logtag is not supposed to be kept in the record.
https://github.com/openshift/origin-aggregated-logging/pull/1350/commits/d7d0ba7384f44134a1572d3e385a013cb96986ff

richm:
I think github is throttling us - could not clone https://github.com/jemalloc/jemalloc

richm:
please re-lgtm

nhosoi:
@richm , do we want to port the patch for hack/testing/rsyslog/roles/rsyslog/defaults/main.yaml to linux-system-roles/logging/roles/rsyslog?

PR #1361, opened 2018-09-25
Capture OpenShift audit logging into Fluentd and send it to ES.

renatoppuccini:
This conf file have the configuration capture the OpenShift Audit logs and send them to ElasticSearch.
It creates an index also.

### DIFF
--- a/fluentd/configs.d/input-ocp-audit.conf
+++ b/fluentd/configs.d/input-ocp-audit.conf
@@ -15,7 +15,7 @@
     enable_ruby true
     auto_typecast true
     <record>
-      audit ${record}
+      kubernetes {"audit":"${record}"}
       message "this is an audit event"
       level info
       time ${Time.parse(record['timestamp']).to_datetime.rfc3339(6)}
--- a/fluentd/configs.d/input-ocp-audit.conf
+++ b/fluentd/configs.d/input-ocp-audit.conf
@@ -27,12 +27,12 @@
 
     type_name com.redhat.ocp.auditlog
 
-    # there is currently a bug in the es plugin + excon - cannot
-    # recreate/reload connections
-    reload_connections false
-    reload_on_failure false
-    flush_interval 5s
-    max_retry_wait 300
+    reload_connections "#{ENV['OPS_RELOAD_CONNECTIONS'] || ENV['ES_RELOAD_CONNECTIONS'] || 'true'}"
+    # https://github.com/uken/fluent-plugin-elasticsearch#sniffer-class-name
+    reload_after "#{ENV['OPS_RELOAD_AFTER'] || ENV['ES_RELOAD_AFTER'] || '200'}"
+    # https://github.com/uken/fluent-plugin-elasticsearch#reload-after
+    flush_interval "#{ENV['OPS_FLUSH_INTERVAL'] || ENV['ES_FLUSH_INTERVAL'] || '1s'}"
+    max_retry_wait "#{ENV['OPS_RETRY_WAIT'] || ENV['ES_RETRY_WAIT'] || '300'}"
     disable_retry_limit true
     buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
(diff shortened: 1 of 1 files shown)


fluentd/configs.d/input-ocp-audit.conf
```diff
@@ -0,0 +1,44 @@
...
+<match audit-ocp.log>
+  # @type stdout
+   @type copy
+  <store>
+    @type elasticsearch_dynamic
+    log_level debug
+    host "#{ENV['OPS_HOST']}"
+    port "#{ENV['OPS_PORT']}"
+    scheme https
+    index_name operations.ocp.auditlog.
```
jcantrill: @richm maybe we should take this as an opportunity to separate some of our log traffic in order not to bloat the indices.  I can see us putting this in some 'infra' index and aliasing it to 'operations' or  maybe it becomes a particular document type in 'operations' and we need to take advantage of rollover because of the volume this will add.

fluentd/configs.d/input-ocp-audit.conf
```diff
@@ -0,0 +1,44 @@
+<source>
```
jcantrill: We;ll have to figure out how to make this optional as  i expect this will add considerable load that we will want to 'opt-in'

fluentd/configs.d/input-ocp-audit.conf
```diff
@@ -0,0 +1,44 @@
...
+    type_name com.redhat.ocp.auditlog
+
+    # there is currently a bug in the es plugin + excon - cannot
+    # recreate/reload connections
+    reload_connections false
+    reload_on_failure false
+    flush_interval 5s
+    max_retry_wait 300
+    disable_retry_limit true
+    buffer_queue_limit "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
```
jcantrill: @richm this will have an impact on memory will it not?

fluentd/configs.d/input-ocp-audit.conf
richm: > @richm maybe we should take this as an opportunity to separate some of our log traffic in order not to bloat the indices. I can see us putting this in some 'infra' index and aliasing it to 'operations' 

not sure what you mean

> maybe it becomes a particular document type in 'operations' 

Afaik types will be going away in Elasticsearch soon - deprecated in 5.x and unsupported in 6.x - so we'll need to use some mechanism other than types

Also - what about multi-tenancy and access control?  changes to the openshift-elasticsearch-plugin?  What about curator?

In general, I think this PR is premature and needs an actual design doc.

richm:
another problem with the proposed solution -  # From 3.9, the advanced audits logs comes in json format by default.  `format json`

This will make the fields into top level fields in the record e.g.
```
{"kind": "Event",
"apiVersion": "audit.k8s.io/v1beta1",
"metadata":{"creationTimestamp":"2018-09-25T19:13:03Z"},
"level":"Metadata",
"timestamp":"2018-09-25T19:13:03Z"
...
}
```
These field names are ambiguous, and conflict with the existing schema.  There needs to be some sort of mapping so that it is easily known that these field names are associated with OpenShift auditing events.  We already have some of these fields designed for Kubernetes events, so perhaps a good place to start would be to add additional audit fields to https://github.com/ViaQ/elasticsearch-templates/blob/master/namespaces/kubernetes.yml#L92 or perhaps create an `audit` subfield of `kubernetes` for audit specific data.

fluentd/configs.d/input-ocp-audit.conf
```diff
@@ -0,0 +1,29 @@
+<source>
+  @type tail
+  path /var/log/audit/audit-ocp.log
```
JAORMX: This path is no longer available in OCP4.X.
However, there are audit logs available in the following locations:
* /var/log/kube-apiserver/audit.log
* /var/log/openshift-apiserver/audit.log

richm:
I think this PR is obsoleted by https://github.com/openshift/origin-aggregated-logging/pull/1762

PR #1363, opened 2018-09-25
LOG-192. Impl bearer token for  openshift-elasticsearch-plugin

jcantrill:
This PR imples https://jira.coreos.com/browse/LOG-192

which adds a native bearer token authentication and should allow us to remove the proxy for metrics

Blocked by https://github.com/fabric8io/openshift-elasticsearch-plugin/pull/158

### DIFF
--- a/test/cluster/functionality.sh
+++ b/test/cluster/functionality.sh
@@ -50,6 +50,16 @@ test_token="$( oc whoami -t )"
 os::cmd::expect_success "oc login --username=system:admin"
 os::cmd::expect_success "oc project ${LOGGING_NS}"
 
+result=$(oc get sa prometheus-scraper -n ${LOGGING_NS} ||:)
+if [ "$result" == "" ] ; then
+  oc create sa prometheus-scraper -n ${LOGGING_NS}
+fi
+result=$(oc get rolebinding prometheus-scraper -n ${LOGGING_NS} prometheus-scraper ||:)
+if [ "$result" == "" ] ; then
+  oc create rolebinding prometheus-scraper --serviceaccount=${LOGGING_NS}:prometheus-scraper --role=prometheus-metrics-viewer
+fi
+SATOKEN=$(oc -n ${LOGGING_NS} serviceaccounts get-token prometheus-scraper)
+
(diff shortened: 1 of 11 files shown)


ewolinetz:
prelim lgtm

jcantrill:
We should be able to backport this PR to 3.11.  I expect it to function properly even without the desired cleanup to ansible to remove the oauth-proxy from the es pod

test/cluster/functionality.sh
```diff
@@ -133,12 +133,12 @@ for elasticsearch_pod in $( oc get pods --selector component="${OAL_ELASTICSEARC
 	fi
 
 	os::log::info "Checking that Elasticsearch pod ${elasticsearch_pod} exports Prometheus metrics"
-    oc create sa prometheus-scraper -n ${LOGGING_NS} ||
-    oc create rolebinding prometheus-scraper --serviceaccount=${LOGGING_NS}:prometheus-scraper --role=prometheus-metrics-viewer ||
+    os::cmd_exect_success "oc create sa prometheus-scraper -n ${LOGGING_NS}"
```
richm: os::cmd::expect_success

jcantrill:
Tests should pass once the dependent plugin is visible in Maven: https://fabric8-ci.fusesource.com/view/release%20builds/job/openshift-elasticsearch-plugin-release/76/

Reamer:
How is the plan for monitoring Elasticsearch with Prometheus? HTTP-Get over oauth-proxy or direct to rest-api endpoint?
With this patch, the direct connection to rest-api endpoint works.

jcantrill:
This merge of this PR removes the need to use the oauth-proxy and allows direct connection to the ES endpoint over HTTPS.  Prometheus is configured with a scrape rule that uses a ServiceAccount that has a role binding which satisfies the SubjectAccessReview [1].  The SAR is subject to change.

[1] https://github.com/openshift/origin-aggregated-logging/pull/1363/files#diff-84bb2a16fa453407379d6048d001a362R19

Reamer:
Thanks for clarification

PR #1364, opened 2018-09-26
Bug 1632648 - Record is missing kubernetes field when Docker log driver is journald

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1632648
In /etc/sysconfig/docker, the log-driver may be specified
as --log-driver=journald or --log-driver journald delimited
by one or more whitespace

### DIFF
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -43,7 +43,7 @@ docker_uses_journal() {
         if grep -q '^[^#].*"log-driver":.*journald' /etc/docker/daemon.json 2> /dev/null ; then
             return 0
         fi
-    elif grep -q "^OPTIONS='[^']*--log-driver=journald" /etc/sysconfig/docker 2> /dev/null ; then
+    elif grep -q "^OPTIONS='[^']*--log-driver[   =][     ]*journald" /etc/sysconfig/docker 2> /dev/null ; then
         return 0
     fi
     return 1


raatti:
Hi,

Please cherrypick this to 3.6 aswell, we are having issue with 3.6 fluentd and this seems to be solution for problem.

richm:
Does it work when you do this:
`oc set env ds/logging-fluentd USE_JOURNAL=true`

raatti:
Hi,

Nope.

raatti:
3.6 code:

  if docker_uses_journal ; then
        export USE_JOURNAL=true
    else
        export USE_JOURNAL=false
    fi

if check fails, then it ignores that variable and just sets it to false anyway.

richm:
:-(

richm:
then - the other workaround is to manually edit `/etc/sysconfig/docker` to remove the extraneous whitespace - does that work?

raatti:
Hi,

As workaround, I changed the format to: --log-driver=journald so the grep works. But that function should be fixed in the long run for all (supported) versions of fluentd. Something like OPTIONS.*journal* match would be better.

richm:
What would be best is for fluentd to auto-detect based on the incoming data, and not have to mess with hacks like mounting /etc/sysconfig/docker and /etc/docker/daemon.json in the container (or, shudder, mounting the docker unix socket in the container . . .)

PR #1366, opened 2018-10-01
Added info message about redirection of fluentd logs

VladMasarik:
Added a message that will be printed out when `oc logs <fluentd_pod>` is used. Informing user that logs are written to log files and what to do if they want to see them.

I am unsure whether to put the message at the end of the `run.sh` script or somewhere at the start.
Also `logs` file is not in the PATH so you need to type `./utils/logs` or the absolute path variant to use it. Is it a problem?

### DIFF
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -16,6 +16,12 @@ for file in ${ocp_fluentd_files} ; do
     sed -i -e "s/%OCP_FLUENTD_TAGS%/${OCP_FLUENTD_TAGS}/" $file
 done
 
+echo "============================="
+echo "Fluentd logs have been redirected to: $LOGGING_FILE_PATH"
+echo "If you want to print out the logs, use command:"
+echo "oc exec <pod_name> $HOME/utils/logs"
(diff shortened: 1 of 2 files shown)


richm:
> I am unsure whether to put the message at the end of the run.sh script or somewhere at the start.

I think what you did is good.

> Also logs file is not in the PATH so you need to type ./utils/logs or the absolute path variant to use it. Is it a problem?

@jcantrill In the elasticsearch image, we install the utils into /usr/local/bin so that they are available in PATH automatically - should we do the same for fluentd?

ewolinetz:
> In the elasticsearch image, we install the utils into /usr/local/bin so that they are available in PATH automatically - should we do the same for fluentd?

I was about to suggest that we do this :)

jcantrill:
@richm @VladMasarik yes, install /usr/local/bin....

richm:
@VladMasarik so change Dockerfile and Dockerfile.centos7 to 
```
COPY utils/** /usr/local/bin/
```

PR #1367, opened 2018-10-01
look for /var/log/fluentd/fluentd.log for pod logs

richm:
add utility functions get_fluentd_pod_log and get_mux_pod_log to
dump the pod logs from wherever they are
also added sudo rm -rf /var/log/fluentd/fluentd.log so that
tests which look for exact strings in the fluentd pod logs
will work correctly and not find older matches
fix incorrect usage of artifact_out


nhosoi:
lgtm

PR #1371, opened 2018-10-02
move fluent utils to /usr/local/bin

jcantrill:
This PR moves the fluent utils to /usr/local/bin so they can be executed from `oc exec` without full path


ewolinetz:
Should we be copying things to `/usr/local/bin` or adding `${HOME}/utils` to `PATH`?

richm:
I think we should be consistent with what elasticsearch does https://github.com/openshift/origin-aggregated-logging/blob/master/elasticsearch/Dockerfile#L38
```
COPY utils/** /usr/local/bin/
```

ewolinetz:
lgtm

richm:
manual 3.9 cherrypick https://github.com/openshift/origin-aggregated-logging/pull/1411

richm:
manual 3.10 cherrypick https://github.com/openshift/origin-aggregated-logging/pull/1412

PR #1377, opened 2018-10-03
create container.yaml for odcs multi-arch builds

jeffdyoung:
create container.yaml for odcs multi-arch builds
@dorzel @adammhaile Please review.

### DIFF
--- a/elasticsearch/container.yaml
+++ b/elasticsearch/container.yaml
@@ -1,0 +1,6 @@
+---
+compose:
+  packages:
+    - elasticsearch
+    - elasticsearch-cloud-kubernetes
+    - openshift-elasticsearch-plugin
--- a/curator/container.yaml
+++ b/curator/container.yaml
@@ -1,0 +1,6 @@
+---
+compose:
+  packages:
+    - elastic-curator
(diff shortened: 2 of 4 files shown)


elasticsearch/container.yaml
```diff
@@ -0,0 +1,6 @@
+---
+compose:
+  packages:
+    - elasticsearch
```
richm: actually, none of these are needed . . . we don't build elasticsearch from elasticsearch rpm packages anymore, which are essentially wrappers around the maven artifacts, we build it directly from the maven artifacts

jeffdyoung: Hey @richm 
The ART team needs this content so they can consume the corresponding Dockerfiles, and pass them to OSBS to build OCP 3.10. We are having to go back to do this because we need to build x86 and ppc64le containers with ODCS enabled. 
More on the osbs/odcs things:
https://osbs.readthedocs.io/en/latest/users.html#odcs-compose

Once we get this done, we will be able to put the right things in release-3.11 and master branches. 
You don't need these things upstream because OKD doesn't use OSBS to build their containers.

PR #1378, opened 2018-10-03
Updating to create secret_dir if it doesnt exist for es cert generation

ewolinetz:
To resolve currently seeing the following with cluster-logging-operator generated secret:
```
[2018-10-03 20:29:01,311][INFO ][container.run            ] Copying certs from /etc/openshift/elasticsearch/secret to /etc/elasticsearch/secret
cp: target '/etc/elasticsearch/secret/' is not a directory
```

### DIFF
--- a/elasticsearch/utils/logging
+++ b/elasticsearch/utils/logging
@@ -185,6 +185,11 @@ copy_keys_to_secretdir() {
 
   if [ -d $provided_secret_dir ] ; then
     info "Copying certs from ${provided_secret_dir} to ${secret_dir}"
+
+    if [ ! -d $secret_dir ] ; then
+      mkdir $secret_dir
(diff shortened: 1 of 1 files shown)


richm:
which directory is being mounted into the elasticsearch pod now?  It used to be this:
https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging_elasticsearch/templates/es.j2#L137

ewolinetz:
It is expected to be mounted at `/etc/openshift/elasticsearch/secret`[1]  -- the elasticsearch-operator needs to be updated as well.

[1] https://github.com/openshift/origin-aggregated-logging/pull/1378/files#diff-711c501edf306d5359cc2802df95ec40R19

PR #1379, opened 2018-10-04
Support fluentd 1.2.6 using rh-ruby25 scl base image

richm:
Use the rh-ruby25 scl base image for fluentd.  This allows us
to use the latest fluentd 1.2.6 and the latest components.

This means we do not have to carry the patches for prometheus
and concat.  We can also use the log rotation built-in to
fluentd 1.x.  There is no longer a separate secure_forward
plugin - the built-in forward plugin has TLS and secret_key
support.

Some of our internal plugins like parser_json_or_crio had to
be ported to use the fluentd 1.x internal api.

Another big change is that there is a new format for the
fluent-plugin-system pos file - It is in JSON format and
must have a .json suffix.  The fluentd run.sh attempts to
convert the old file to the new format file.

The config file format is a little stricter - it is
apparently now an error to have a @type copy plugin
with no <store> sections.  You also cannot have a
rewrite_tag plugin with an empty pattern.


richm:
centos repo failure
/test logging

richm:
This is ready for merge.  Please review.

test/fluentd-forward.sh
```diff
@@ -33,63 +32,75 @@ update_current_fluentd() {
     # update configmap secure-forward#.conf
     if [ $cnt -eq 1 ]; then
       # edit so we don't send to ES
-      oc get configmap/logging-fluentd -o yaml | sed '/## matches/ a\
+      oc get configmap/logging-fluentd -o yaml | sed -e '/## matches/ a\
```
nhosoi: Do we need '-e'???

richm: not strictly, but will save some confusion if another sed expression is added later

test/docker_audit.sh
```diff
@@ -72,19 +72,16 @@ esopssvc=$( get_es_svc es-ops )
...
 ops_logs_before=$( get_logs_count $esopssvc '/.operations.*/' )
 
 os::log::info "ops diff before:  $ops_logs_before"
 os::log::info "proj diff before: $logs_before"
 
-os::cmd::expect_success flush_fluentd_pos_files
-sudo rm -f /var/log/fluentd/fluentd.log
-os::log::debug "$( oc label node --all logging-infra-fluentd=true 2>&1 || : )"
-os::cmd::try_until_text "oc get pods -l component=fluentd" "^logging-fluentd-.* Running "
+start_fluentd cleanfirst $((second * 180)) 2>&1 | artifact_out
```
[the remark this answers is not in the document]
richm: yes

test/fluentd-forward.sh
richm: sorry - yes - it is needed because there are more than 1 expression for this sed now

nhosoi: ok.

richm:
@jcantrill @ewolinetz yea or nay?  If you guys approve, I will change the base image used by the prow/images job so that it will pass.

PR #1386, opened 2018-10-08
[release-3.11] Including fluent-plugin-concat to support cri-o multiline logs.

nhosoi:
Backporting the partial code in fluent-plugin-concat v2.3.0 to v1.0.0.
Note: removing @seperator = '\n' from join.  It adds '\n' at the end of
      the partial message.  The original code has the issue.
  - @key => lines.join(@separator)
  + @key => lines.join

Added missing keep_partial_key code.
Moved filter_concat.rb to lib/filter_concat and added a unit test.
Fixed broken unit test in test/generate_throttle_configs_test.rb.
Enabling filter_concat unit test.
Capturing exception in shutdown.

(cherry picked from commit 30efccb7a7bebbdb09193248934186f54c4294f2)

### DIFF
--- a/fluentd/lib/generate_throttle_configs/lib/generate_throttle_configs.rb
+++ b/fluentd/lib/generate_throttle_configs/lib/generate_throttle_configs.rb
@@ -174,7 +174,20 @@ def create_default_docker(input_conf_file, excluded, log, options={})
   keep_time_key true
   read_from_head "#{options[:read_from_head] || 'true'}"
   exclude_path #{excluded}
+  @label @CONCAT
 </source>
+<label @CONCAT>
+  <filter kubernetes.**>
+    @type concat
+    key log
+    partial_key logtag
+    partial_value P
+  </filter>
+  <match kubernetes.**>
+    @type relabel
+    @label @INGRESS
+  </match>
+</label>
     CONF
   }
     log.debug "Created default docker input config file"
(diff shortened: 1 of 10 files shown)


richm:
updated aos cd jobs
/retest

richm:
If we need this in 3.11, you'll need to change it to use the `use_crio` flag - the automatic detection of crio or json-file was not backported to 3.11

nhosoi:
> If we need this in 3.11, you'll need to change it to use the use_crio flag - the automatic detection of crio or json-file was not backported to 3.11

Yes, that's what I manually backported pr/1350 since /cherrypick release-3.11 failed because of it.

fluentd/lib/generate_throttle_configs/lib/generate_throttle_configs.rb
```diff
@@ -168,13 +168,26 @@ def create_default_docker(input_conf_file, excluded, log, options={})
   @label @INGRESS
   path "#{cont_logs_path}"
   pos_file "#{cont_pos_file}"
-  time_format #{use_crio == 'true' ? '%Y-%m-%dT%H:%M:%S.%N%:z' : '%Y-%m-%dT%H:%M:%S.%N%Z'}
+  time_format '%Y-%m-%dT%H:%M:%S.%N%Z'
```
richm: I don't understand this.  Don't we need this in 3.11 for crio support?

nhosoi: Sorry, I missed the time_format change...  Pushed the fix.

richm:
These problems with kibana not starting are persistent . . .
/retest

PR #1388, opened 2018-10-09
remove building of logging-auth-proxy

jcantrill:
This PR removes the conditional to build the logging-auth-proxy since it is not used in this branch.


ewolinetz:
did something happen with the fluentd image?

jcantrill:
Tests passed. Manually merging.

PR #1389, opened 2018-10-09
remove building of logging-auth-proxy

jcantrill:
Backport of https://github.com/openshift/origin-aggregated-logging/pull/1388.  This pr fixes the tagging and building of the images along with removing the logging-auth-proxy


richm:
also remove the git submodule if we don't need it anymore
also remove it from the hack/push-release.sh script

jcantrill:
@richm the submodule was deleted a while ago.  updated to remove from the push script

PR #1396, opened 2018-10-10
add get_all_logging_pod_logs function

richm:
Add a function get_all_logging_pod_logs so tests can dump all
of the logging pod logs. This function takes into account
the various logs commands and log files that different pods
have.

### DIFF
--- a/test/cluster/rollout.sh
+++ b/test/cluster/rollout.sh
@@ -23,19 +23,18 @@ cleanup() {
   local return_code="$?"
   set +e
   for r in dc ds cronjob ; do
-    oc describe -n ${LOGGING_NS} $r > $ARTIFACT_DIR/$r.describe 2>&1
+    for it in $( oc get -n ${LOGGING_NS} $r -o name ) ; do
+      oc describe -n ${LOGGING_NS} $it > $ARTIFACT_DIR/$it.describe 2>&1
+    done
   done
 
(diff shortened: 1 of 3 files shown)


test/cluster/rollout.sh
```diff
@@ -23,19 +23,18 @@ cleanup() {
...
-  done
-  for p in $(oc get pods -n ${LOGGING_NS} -l component!=es -o jsonpath={.items[*].metadata.name}) ; do
-    oc logs -n ${LOGGING_NS} $p || :  > $ARTIFACT_DIR/$p.stdout.logs 2>&1
-  done
+  get_all_logging_pod_logs
   oc get -n ${LOGGING_NS} --all 2>&1 | artifact_out
-  
+  sudo docker images|grep logging 2>&1 | artifact_out
+  sudo docker images|grep oauth 2>&1 | artifact_out
+  sudo docker images|grep eventrouter 2>&1 | artifact_out
```
nhosoi: Just curious.  Does "docker images" work for cri-o?

richm: no, it does not - so there will be a failure reported in artifact_out in the crio case - if this is a problem we can revisit this

nhosoi: ok. thanks!
/lgtm

PR #1418, opened 2018-10-16
bug 1638602. kibana-5.6.12 to resolve cross scripting error

jcantrill:
This bz resolves https://bugzilla.redhat.com/show_bug.cgi?id=1638602 

by moving the logging stack to 5.6.12

cc @openshift/sig-logging


richm:
@stevekuznetsov the problem is that the jenkins jobs have finished successfully, but the status still says Pending

richm:
`/test all` is the magic word

PR #1424, opened 2018-10-17
Bug 1613722 - Eventrouter creates duplicated events every 30 min with…

nhosoi:
To resolve https://bugzilla.redhat.com/show_bug.cgi?id=1613722
… verb UPDATE

Adding a plugin filter_elasticsearch_genid_ext.rb, which extends filter_
elasticsearch_genid.rb from fluent-plugin-elasticsearch.  The extended plugin
takes config parameter alt_key as shown in filter-post-genid.conf.
  <filter **>
    @type elasticsearch_genid_ext
    hash_id_key viaq_msg_id
    alt_key kubernetes.event.metadata.uid
  </filter>
The plugin checks if record[kubernetes][event][metadata][uid] exists or not.
If it does, viaq_msg_id is set to the value.  Otherwise, the hash value is used.

Note: the value of kubernetes.event.metadata.uid is shared among the event logs
originated from the same one event.  By using the value for the elasticsearch
primary key, duplicated logs are not to be indexed in the elasticsearch.

For testing, code to check there is no duplicated value of kubernetes.event.
metadata.uid is added to test/eventrouter.sh.

### DIFF
--- a/fluentd/configs.d/openshift/filter-retag-journal.conf
+++ b/fluentd/configs.d/openshift/filter-retag-journal.conf
@@ -59,18 +59,20 @@
   # k8s_kibana.a67f366_logging-kibana-1-d90e3_logging_26c51a61-2835-11e6-ad29-fa163e4944d5_f0db49a2
   # we filter these logs through the kibana_transform.conf filter
   rewriterule1 CONTAINER_NAME ^k8s_kibana\. kubernetes.journal.container.kibana
+  # mark eventrouter events
+  rewriterule2 CONTAINER_NAME ^k8s_[^_]+_logging-eventrouter-[^_]+_ kubernetes.journal.container._default_.kubernetes.event
   # mark logs from default namespace for processing as k8s logs but stored as system logs
-  rewriterule2 CONTAINER_NAME ^k8s_[^_]+_[^_]+_default_ kubernetes.journal.container._default_
+  rewriterule3 CONTAINER_NAME ^k8s_[^_]+_[^_]+_default_ kubernetes.journal.container._default_
   # mark logs from kube-* namespaces for processing as k8s logs but stored as system logs
-  rewriterule3 CONTAINER_NAME ^k8s_[^_]+_[^_]+_kube-(.+)_ kubernetes.journal.container._kube-$1_
+  rewriterule4 CONTAINER_NAME ^k8s_[^_]+_[^_]+_kube-(.+)_ kubernetes.journal.container._kube-$1_
   # mark logs from openshift-* namespaces for processing as k8s logs but stored as system logs
-  rewriterule4 CONTAINER_NAME ^k8s_[^_]+_[^_]+_openshift-(.+)_ kubernetes.journal.container._openshift-$1_
+  rewriterule5 CONTAINER_NAME ^k8s_[^_]+_[^_]+_openshift-(.+)_ kubernetes.journal.container._openshift-$1_
   # mark logs from openshift namespace for processing as k8s logs but stored as system logs
-  rewriterule5 CONTAINER_NAME ^k8s_[^_]+_[^_]+_openshift_ kubernetes.journal.container._openshift_
+  rewriterule6 CONTAINER_NAME ^k8s_[^_]+_[^_]+_openshift_ kubernetes.journal.container._openshift_
   # mark fluentd container logs
-  rewriterule6 CONTAINER_NAME ^k8s_.*fluentd kubernetes.journal.container.fluentd
+  rewriterule7 CONTAINER_NAME ^k8s_.*fluentd kubernetes.journal.container.fluentd
   # this is a kubernetes container
-  rewriterule7 CONTAINER_NAME ^k8s_ kubernetes.journal.container
+  rewriterule8 CONTAINER_NAME ^k8s_ kubernetes.journal.container
   # not kubernetes - assume a system log or system container log
-  rewriterule8 _TRANSPORT .+ journal.system
+  rewriterule9 _TRANSPORT .+ journal.system
 </match>
--- a/fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
+++ b/fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
@@ -1,0 +1,78 @@
+require 'securerandom'
+require 'base64'
+require 'fluent/filter'
+
+begin
+  GenidMatchClass = Fluent::Match
+rescue
+  # Fluent::Match not provided with 0.14
+  class GenidMatchClass
+    def initialize(pattern_str, unused)
+      patterns = pattern_str.split(/\s+/).map {|str|
+        Fluent::MatchPattern.create(str)
+      }
+      if patterns.length == 1
+        @pattern = patterns[0]
+      else
+        @pattern = Fluent::OrMatchPattern.new(patterns)
+      end
+    end
+    def match(tag)
+      @pattern.match(tag)
+    end
+    def to_s
+      "#{@pattern}"
+    end
+  end
+end
+
+module Fluent
+  class ElasticsearchGenidExtFilter < Filter
+    Fluent::Plugin.register_filter('elasticsearch_genid_ext', self)
+
+    desc 'key to store generated unique id or the value of alt_key if specified'
+    config_param :hash_id_key, :string, :default => '_hash'
+    desc 'key for the hash "record" (optional)'
+    config_param :alt_key, :string, default: ""
+    desc 'process alt_key in records with this tag pattern'
+    config_param :alt_tags, :string, default: ""
+
+    def configure(conf)
+      super
+      @alt_keys = @alt_key.split('.')
+      @alt_tag_matcher = GenidMatchClass.new(@alt_tags, nil)
+    end
+
+    def filter(tag, time, record)
+      record[@hash_id_key] = ""
+      if @alt_tag_matcher.match(tag)
+        myid = nil
+        unless @alt_key.to_s.strip.empty? || record.empty?
+          myid = record
+          @alt_keys.each do |p|
+            unless myid.key?(p)
+              unless p.eql? @alt_key
+                log.on_debug do
+                  log.debug "filter:elasticsearch_genid_ext: #{p} in alt_key #{@alt_key} is not a key of record."
+                end
+                myid = nil
+              end
+              break
+            end
+            myid = myid[p]
+          end
+        end
+        record[@hash_id_key] = if myid.is_a? String
+                                 myid
+                               else
+                                 Base64.strict_encode64(SecureRandom.uuid)
+                               end
+      end
+      if record[@hash_id_key].to_s.strip.empty?
+          record[@hash_id_key] = Base64.strict_encode64(SecureRandom.uuid)
+      end
+      record
+    end
+
+  end
+end
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,3 +43,14 @@ matrix:
         cd fluentd/lib/filter_concat
       script:
         rake test
+    - name: fluentd - filter_elasticsearch_genid_ext test
+      env:
+        - FLUENTD_VERSION=0.12.0
+      language: ruby
+      rvm:
+      - 2.0
+      gemfile: fluentd/lib/filter_elasticsearch_genid_ext/Gemfile
+      before_script:
+        cd fluentd/lib/filter_elasticsearch_genid_ext
+      script:
+        rake test
--- a/fluentd/Dockerfile
+++ b/fluentd/Dockerfile
@@ -56,6 +56,7 @@ ADD run.sh generate_syslog_config.rb ${HOME}/
 ADD lib/generate_throttle_configs/lib/*.rb ${HOME}/
 ADD lib/filter_parse_json_field/lib/*.rb /etc/fluent/plugin/
 ADD lib/filter_concat/lib/filter_concat.rb /etc/fluent/plugin/
+ADD lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb /etc/fluent/plugin/
 ADD patch/in_prometheus.rb ${GEM_HOME}/gems/fluent-plugin-prometheus-0.4.0/lib/fluent/plugin/
 ADD lib/parser_json_or_crio/lib/*.rb /etc/fluent/plugin/
 
--- a/fluentd/Dockerfile.centos7
+++ b/fluentd/Dockerfile.centos7
@@ -63,6 +63,7 @@ ADD run.sh generate_syslog_config.rb ${HOME}/
 ADD lib/generate_throttle_configs/lib/*.rb ${HOME}/
 ADD lib/filter_parse_json_field/lib/*.rb /etc/fluent/plugin/
 ADD lib/filter_concat/lib/filter_concat.rb /etc/fluent/plugin/
+ADD lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb /etc/fluent/plugin/
 ADD patch/logger/log.rb ${GEM_HOME}/gems/fluentd-0.12.43/lib/fluent/
 ADD patch/logger/supervisor.rb ${GEM_HOME}/gems/fluentd-0.12.43/lib/fluent/
 COPY utils/** /usr/local/bin/
--- a/fluentd/configs.d/openshift/filter-post-genid.conf
+++ b/fluentd/configs.d/openshift/filter-post-genid.conf
@@ -1,6 +1,8 @@
 # generate ids at source mitigate creating duplicate records
 # requires fluent-plugin-elasticsearch
 <filter **>
-  @type elasticsearch_genid
+  @type elasticsearch_genid_ext
   hash_id_key viaq_msg_id
+  alt_key kubernetes.event.metadata.uid
+  alt_tags "#{ENV['GENID_ALT_TAG'] || 'kubernetes.var.log.containers.kube-eventrouter-*.** kubernetes.journal.container.kubernetes-event'}"
 </filter>
--- a/fluentd/lib/filter_elasticsearch_genid_ext/Gemfile
+++ b/fluentd/lib/filter_elasticsearch_genid_ext/Gemfile
@@ -1,0 +1,5 @@
+source 'https://rubygems.org'
+
+gem 'codeclimate-test-reporter', :group => :test, :require => nil
+
+gemspec
--- a/fluentd/lib/filter_elasticsearch_genid_ext/Rakefile
+++ b/fluentd/lib/filter_elasticsearch_genid_ext/Rakefile
@@ -1,0 +1,11 @@
+#require "bundler/gem_tasks"
+require "rake/testtask"
+
+Rake::TestTask.new do |t|
+    t.test_files = FileList['test/**/*_test.rb']
+    t.warning = false
+    #t.verbose = true
+end
+desc "Run tests"
+
+task default: :test
--- a/fluentd/lib/filter_elasticsearch_genid_ext/filter_elasticsearch_genid_ext.gemspec
+++ b/fluentd/lib/filter_elasticsearch_genid_ext/filter_elasticsearch_genid_ext.gemspec
@@ -1,0 +1,24 @@
+# coding: utf-8
+lib = File.expand_path('../lib', __FILE__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+
+# can override for testing
+FLUENTD_VERSION = ENV['FLUENTD_VERSION'] || "0.12.0"
+
+Gem::Specification.new do |gem|
+  gem.name          = "filter_elasticsearch_genid_ext"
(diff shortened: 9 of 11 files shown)


fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
```diff
@@ -0,0 +1,45 @@
+require 'securerandom'
+require 'base64'
+require 'fluent/filter'
+
+module Fluent
+  class ElasticsearchGenidExtFilter < Filter
```
jcantrill: Consider actually extending the other filter and calling its filter method if needed.

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
```diff
@@ -0,0 +1,45 @@
...
+    def configure(conf)
+      super
+      @alt_key = conf['alt_key']
+    end
+
+    def filter(tag, time, record)
+      myid = nil
+      unless @alt_key.to_s.strip.empty? || record.empty?
+        myid = record
+        @alt_key.split('.').each { |p|
```
jcantrill: prefer using do/end over braces when coding multi-line

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
```diff
@@ -0,0 +1,45 @@
...
+          if ! myid.key?(p)
```
jcantrill: consider using `unless` over `if ! `

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
```diff
@@ -0,0 +1,45 @@
...
+            if ! p.eql? @alt_key
+              log.debug "filter:elasticsearch_genid_ext: #{p} in alt_key #{@alt_key} is not a key of record."
```
jcantrill: use 
```
 log.on_debug do
   log.debug(msg) 
end
```

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
```diff
@@ -0,0 +1,45 @@
...
+    def initialize
+      super
+    end
+
+    def configure(conf)
+      super
+      @alt_key = conf['alt_key']
+    end
+
+    def filter(tag, time, record)
```
jcantrill: Can we add a check here to only use the `@alt_key` for a specific tag.  I can imagine this could become expensive for all records when we only need to check events

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
```diff
@@ -0,0 +1,45 @@
...
+            if ! p.eql? @alt_key
+              log.debug "filter:elasticsearch_genid_ext: #{p} in alt_key #{@alt_key} is not a key of record."
+              myid = nil
+            end
+            break
+          end
+          myid = myid[p]
+        }
+      end
+      record[@hash_id_key] = if myid.is_a? String
```
jcantrill: Call `super` on the base class once you extend what already exists

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
nhosoi: Thanks!  Done.

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
nhosoi: Done.

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
nhosoi: Done.

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
nhosoi: If LOG_DRIVER is json-file, the tag contains "eventrouter" in the tag, I'd think.  But journald does not have the knowledge (for now)?  Do we need to add it?

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
```diff
@@ -0,0 +1,47 @@
...
+    config_param :hash_id_key, :string, :default => '_hash'
+    config_param :alt_key, :string, default: ""
+
+    def initialize
+      super
+    end
+
+    def configure(conf)
+      super
+      @alt_key = conf['alt_key']
```
richm: `config_param` already does this - it creates an instance variable called `@alt_key` and sets it to the configured value.  You can omit the `configure` method too

nhosoi: Thanks!  Done.

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
```diff
@@ -0,0 +1,38 @@
...
+    Fluent::Plugin.register_filter('elasticsearch_genid_ext', self)
+
+    config_param :hash_id_key, :string, :default => '_hash'
+    config_param :alt_key, :string, default: ""
+
+    def filter(tag, time, record)
+      myid = nil
+      unless @alt_key.to_s.strip.empty? || record.empty?
+        myid = record
+        @alt_key.split('.').each do |p|
```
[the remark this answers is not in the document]
nhosoi: Is configure called for each record?

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
richm: Yes.  You'll need to add some logic to filter-retag-journal.conf to create a special tag for eventrouter e.g. something like 
```
# mark eventrouter events
rewriterule6 CONTAINER_NAME ^k8s_[^_]+_kube-eventrouter-[^_]+_ kubernetes.journal.container._kube-eventrouter_
```
That is, if the pod-name part of CONTAINER_NAME is kube-eventrouter-*, tag as kubernetes.journal.container._kube-eventrouter_

Then, you'll need to implement programmatic tag matching in this plugin.  You can probably grab the code from the viaq plugin: https://github.com/ViaQ/fluent-plugin-viaq_data_model/blob/master/lib/fluent/plugin/filter_viaq_data_model.rb#L106
and
https://github.com/ViaQ/fluent-plugin-viaq_data_model/blob/master/lib/fluent/plugin/filter_viaq_data_model.rb#L166
and
https://github.com/ViaQ/fluent-plugin-viaq_data_model/blob/master/lib/fluent/plugin/filter_viaq_data_model.rb#L371

I suggest adding a new config param:
```
      desc 'only do the alt_tag processing on records which match these tag patterns'
      config_param :alt_tag, :string
```
Then you can specify them in the config like this:
```
   @type genid-ext
   alt_tag "#{ENV['GENID_ALT_TAG'] || 'kubernetes.var.log.containers.kube-eventrouter-*.** kubernetes.journal.container._kube-eventrouter_'}"
```

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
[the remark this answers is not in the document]
richm: no

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
jcantrill: Should we consider modifying the tag to not be 'eventrouter' and calling it kuberenetes_events or something like it.  Technically, we are processing events that so happen to be produced by the event router.  Why not tag it for what it is

nhosoi: @richm , regarding this suggestion...
```
# mark eventrouter events
rewriterule6 CONTAINER_NAME ^k8s_[^_]+_kube-eventrouter-[^_]+_ kubernetes.journal.container._kube-eventrouter_
```
1) could this rewriterule6 be `^k8s_[^_]+[^_]+_kube-eventrouter-[^_]+_ kubernetes.journal.container._kube-eventrouter_`?
2) if so, this existing rewriterule already does the rewrite?  (i.e., the answer to my original question about journald was "no"?)
```
# mark logs from kube-* namespaces for processing as k8s logs but stored as system logs
rewriterule3 CONTAINER_NAME ^k8s_[^_]+_[^_]+_kube-(.+)_ kubernetes.journal.container._kube-$1_
```

richm: > could this rewriterule6 be ^k8s_[^_]+[^_]+_kube-eventrouter-[^_]+_ kubernetes.journal.container._kube-eventrouter_?

No.  That will match a _namespace name_ that starts with "kube-eventrouter".  We want to match on the _pod name_.

> Should we consider modifying the tag to not be 'eventrouter' and calling it kuberenetes_events or something like it. Technically, we are processing events that so happen to be produced by the event router. Why not tag it for what it is

Sure.  That will work for log-driver=journald.  But for json-file or crio, we have no control over the tag, unless we want to do a retag.

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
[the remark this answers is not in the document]
nhosoi: hmmm, then checking if the key-value in the record has to be done in the filter, anyway...  Are you suggesting to do `split` only in `configure` and save the result in an array?

fluentd/configs.d/openshift/filter-retag-journal.conf
```diff
@@ -67,10 +67,12 @@
   rewriterule4 CONTAINER_NAME ^k8s_[^_]+_[^_]+_openshift-(.+)_ kubernetes.journal.container._openshift-$1_
   # mark logs from openshift namespace for processing as k8s logs but stored as system logs
   rewriterule5 CONTAINER_NAME ^k8s_[^_]+_[^_]+_openshift_ kubernetes.journal.container._openshift_
+  # mark eventrouter events
+  rewriterule6 CONTAINER_NAME ^k8s_[^_]+_kube-eventrouter-[^_]+_ kubernetes.journal.container._kube-eventrouter_
```
richm: sorry - looks like the pod name is this:
```
oc get -n default pods
NAME                          READY     STATUS    RESTARTS   AGE
logging-eventrouter-1-nvkvw   1/1       Running   0          1h
```
so the pattern should be `^k8s_[^_]+_logging-eventrouter-[^_]+_

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
```diff
@@ -2,35 +2,85 @@
 require 'base64'
 require 'fluent/filter'
 
+begin
+  ViaqMatchClass = Fluent::Match
```
richm: probably don't want to use Viaq . . .

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
```diff
@@ -2,35 +2,85 @@
...
 module Fluent
   class ElasticsearchGenidExtFilter < Filter
     Fluent::Plugin.register_filter('elasticsearch_genid_ext', self)
 
+    desc 'key to store generated unique id or the value of alt_key if specified'
     config_param :hash_id_key, :string, :default => '_hash'
+    desc 'key for the hash "record" (optional)'
     config_param :alt_key, :string, default: ""
+    desc 'process alt_key in records with this tag pattern'
+    config_param :alt_tags, :array, default: []
```
richm: `alt_tags` should be a `string` - that's the way the MatchClass works - you pass it a string containing space delimited tag patterns.

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
```diff
@@ -2,35 +2,85 @@
...
+
+    def configure(conf)
+      super
+      @alt_keys = @alt_key.split('.')
+      if ENV['GENID_ALT_TAG']
```
richm: By using `alt_tags "#{ENV['GENID_ALT_TAG'] || 'tag patterns string'}"` in the config, you can get rid of this

fluentd/configs.d/openshift/filter-retag-journal.conf
richm: Also - make the tag `kubernetes.journal.container.kubernetes-event` as per @jcantrill 's suggestion

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
nhosoi: Ahh, ok.  Thanks!

fluentd/configs.d/openshift/filter-retag-journal.conf
nhosoi: Thanks for the reminder.  Done!

This suggestion by @jcantrill is not taken care yet...
> Consider actually extending the other filter and calling its filter method if needed.

Does this mean instead of adding filter_elasticsearch_genid_ext, we are having filter_elasticsearch_genid in our code and make it something like this?
```
module Fluent
  class ElasticsearchGenidFilter < Filter
    ....
  end
  class ElasticsearchGenidExtFilter < ElasticsearchGenidFilter
    ....
  end
end
```
And could you tell me where I can take advantage of calling `super`?
> Call super on the base class once you extend what already exists

nhosoi:
Currently, pretty badly broken... :(
```
  File "<string>", line 1, in <module>
  File "/usr/lib64/python2.7/json/__init__.py", line 338, in loads
    return _default_decoder.decode(s)
  File "/usr/lib64/python2.7/json/decoder.py", line 366, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib64/python2.7/json/decoder.py", line 384, in raw_decode
    raise ValueError("No JSON object could be decoded")
ValueError: No JSON object could be decoded
[WARNING]  events from eventrouter in index /.operations.*/ were not processed by ViaQ fluentd plugin
Running test/eventrouter.sh:49: executing 'curl_es logging-es-ops /.operations.*/_count?q=kubernetes.event.verb:* | get_count_from_json' expecting success and not text '^0$'...
FAILURE after 0.240s: test/eventrouter.sh:49: executing 'curl_es logging-es-ops /.operations.*/_count?q=kubernetes.event.verb:* | get_count_from_json' expecting success and not text '^0$': the output content test failed
0
```

richm:
> Currently, pretty badly broken... :(

I have a fix for that too - https://github.com/openshift/origin-aggregated-logging/pull/1425

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
jcantrill: Is that expensive?  I don't know if that means there are significant change to the pipeline?

richm:
> This commit b46e8f9 is meant for changing the tag to kubernetes.journal.container.kubernetes-event, which seems to break the logging test. If I revert it, it passes. puzzled

I think it has something to do with the fact that `%OCP_FLUENTD_TAGS%` matches too much.  That is - `kubernetes.journal.container._kube-eventrouter_` matches `%OCP_FLUENTD_TAGS%` but `kubernetes.journal.container.kubernetes-event` does not match.

I'm not sure where in the pipeline it is causing problems.  It may be in the viaq plugin.  It should match this for the formatting part:
```
  <formatter>
    tag "kubernetes.journal.container**"
    type k8s_journal
    remove_keys "..."
  </formatter>
```
to format the record like a k8s record from the journal.  Then I think this is a problem, to get the index name:
```
  <elasticsearch_index_name>
    enabled "#{ENV['ENABLE_ES_INDEX_NAME'] || 'true'}"
    tag "journal.system** system.var.log** audit.log** %OCP_FLUENTD_TAGS%"
    name_type operations_full
  </elasticsearch_index_name>
  <elasticsearch_index_name>
    enabled "#{ENV['ENABLE_ES_INDEX_NAME'] || 'true'}"
    tag "**"
    name_type project_full
  </elasticsearch_index_name>
```
I think we want the record stored in the operations log (which is where it is now since it comes from the `default` namespace which is listed in `%OCP_FLUENTD_TAGS%`.  The problem is that `kubernetes.journal.container.kubernetes-event` does not match and falls through to the last `project_full` index name.  In the json-file case, since the eventrouter runs in the `default` namespace, the tag will match `%OCP_FLUENTD_TAGS%`.

I think the best course of action is to use a tag like `kubernetes.journal.container._default_.kubernetes-event` for the journald case.  That way, we can still use the `**_default_**` match in `%OCP_FLUENTD_TAGS%` to route operations logs.

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
richm: > Is that expensive?

Do you mean doing the retag for the json-file/crio case?  Yes, in terms of both performance and complexity.  Which is why I'm reluctant to do that.

> its from journal or its from a container is not significant

It is - in both the viaq filter formatting, and the routing to figure out if it is a project/apps log or an operations/infra log.

jcantrill:
@richm @nhosoi if the tag change is causing issue then we should forgo modifying it now.  At some point we wish to add events to individual namespaces.  Maybe that is a better place to resolve the tag change since this pr is supposed to be solving the dup events

nhosoi:
Thanks a lot, @richm.  Replaced the tag: 
```
# mark eventrouter events
-  rewriterule2 CONTAINER_NAME ^k8s_[^_]+_logging-eventrouter-[^_]+_ kubernetes.journal.container.kubernetes-event
+  rewriterule2 CONTAINER_NAME ^k8s_[^_]+_logging-eventrouter-[^_]+_ kubernetes.journal.container._default_.kubernetes-event
```

fluentd/configs.d/openshift/filter-retag-journal.conf
```diff
@@ -60,7 +60,7 @@
   # we filter these logs through the kibana_transform.conf filter
   rewriterule1 CONTAINER_NAME ^k8s_kibana\. kubernetes.journal.container.kibana
   # mark eventrouter events
-  rewriterule2 CONTAINER_NAME ^k8s_[^_]+_logging-eventrouter-[^_]+_ kubernetes.journal.container._default_.kubernetes-event
+  rewriterule2 CONTAINER_NAME ^k8s_[^_]+_logging-eventrouter-[^_]+_ kubernetes.journal.container._default_.kubernetes.event
```
richm: Is there a problem with using `kubernetes-event`?

nhosoi: No.  Just I noticed @jcantrill mentioned the suffix. :)

fluentd/lib/filter_elasticsearch_genid_ext/lib/filter_elasticsearch_genid_ext.rb
richm: The problem with `kubernetes.event` is that the viaq filter plugin won't know that this is a record from the journal, and so will not add all of the journal metadata to the record . . . that is a regression from the current behavior.

nhosoi:
@jcantrill , @richm, If cherrypick does not happen, had I better manually cherrypick for 3.10 and 3.11?  Or we'd better wait?

nhosoi:
Instead of cherrypick release-3.11, manually cherry-picked:

https://github.com/openshift/origin-aggregated-logging/pull/1436

nhosoi:
Manually cherry-picked for release-3.10:
https://github.com/openshift/origin-aggregated-logging/pull/1437

PR #1436, opened 2018-10-23
[3.11] Bug 1613722 - Eventrouter creates duplicated events every 30 min with…

nhosoi:
… verb UPDATE

Adding a plugin filter_elasticsearch_genid_ext.rb, which extends filter_
elasticsearch_genid.rb from fluent-plugin-elasticsearch.  The extended plugin
takes config parameter alt_key as shown in filter-post-genid.conf.
  <filter **>
    @type elasticsearch_genid_ext
    hash_id_key viaq_msg_id
    alt_key kubernetes.event.metadata.uid
    alt_tags "#{ENV['GENID_ALT_TAG'] || 'kubernetes.var.log.containers.kube-eventrouter-*.** kubernetes.journal.container._default_.kubernetes.event'}"
  </filter>
The plugin does -
  1. If the tag matches the alt_tags list or not,
  2. If it does, it checks the alt_key exists in the record hash as the key.
     As the key is kubernetes.event.metadata.uid, it checks
     record[kubernetes][event][metadata][uid] exists or not.
  3. If it exists, viaq_msg_id is set to the value.
  4. Otherwise, the generated hash value is set. This is the default behaviour.

Note: the value of kubernetes.event.metadata.uid is shared among the event logs
originated from the same one event.  By using the value for the elasticsearch
primary key, duplicated logs are not to be indexed in the elasticsearch.

For testing, code to check there is no duplicated value of kubernetes.event.
metadata.uid is added to test/eventrouter.sh.

Additionally, the tag of the event logs via journald used to be
  kubernetes.journal.container._kube-eventrouter_
which is changed to
  kubernetes.journal.container._default_.kubernetes.event

(cherry picked from commit a995992b85ce4fd53e1ea7354ebd95a7dd992174)

https://bugzilla.redhat.com/show_bug.cgi?id=1613722

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,3 +32,14 @@ matrix:
         cd fluentd/lib/filter_concat
       script:
         rake test
+    - name: fluentd - filter_elasticsearch_genid_ext test
+      env:
+        - FLUENTD_VERSION=0.12.0
+      language: ruby
+      rvm:
+      - 2.0
+      gemfile: fluentd/lib/filter_elasticsearch_genid_ext/Gemfile
+      before_script:
+        cd fluentd/lib/filter_elasticsearch_genid_ext
+      script:
+        rake test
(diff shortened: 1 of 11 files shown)


nhosoi:
Cherry-picked: https://github.com/openshift/origin-aggregated-logging/pull/1424

PR #1450, opened 2018-11-14
(release-3.11) Bug 1643340 - Fluentd pod should log its logs to /var/log/fluentd/flu…

nhosoi:
…entd.log by default

Output logs in /var/log/fluentd/fluentd.log via stdout...

### DIFF
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -19,8 +19,13 @@ done
 echo "============================="
 echo "Fluentd logs have been redirected to: $LOGGING_FILE_PATH"
 echo "If you want to print out the logs, use command:"
-echo "oc exec <pod_name> $HOME/utils/logs"
+echo "oc exec <pod_name> -- logs"
 echo "============================="
+
+if [ ! -d `dirname $LOGGING_FILE_PATH` ]; then
+  mkdir -p `dirname $LOGGING_FILE_PATH`
+fi
+touch $LOGGING_FILE_PATH; exec >> $LOGGING_FILE_PATH 2>&1
 
 fluentdargs="--no-supervisor"
 if [[ $VERBOSE ]]; then


nhosoi:
Not sure this is acceptable or not, but at least "oc logs" outputs fluentd logs...
```
# oc logs logging-fluentd-POD
=============================
Fluentd logs have been redirected to: /var/log/fluentd/fluentd.log
=============================
umounts of dead containers will fail. Ignoring...
umount: /var/lib/docker/containers/0711f835e745201b8388ece086e2445f2edc0701c537f1551504bd2633f9a516/shm: not mounted
.....
# Logfile created on 2018-11-13 23:34:59 +0000 by logger.rb/41954
2018-11-13 23:34:59 +0000 [info]: reading config file path="/etc/fluent/fluent.conf"
2018-11-13 23:34:59 +0000 [warn]: 'block' action stops input process until the buffer full is resolved. Check your pipeline this action is fit or not
.....
2018-11-13 23:37:05 +0000 [warn]: 'block' action stops input process until the buffer full is resolved. Check your pipeline this action is fit or not
```

fluentd/run.sh
```diff
@@ -18,8 +18,6 @@ done
 
 echo "============================="
 echo "Fluentd logs have been redirected to: $LOGGING_FILE_PATH"
-echo "If you want to print out the logs, use command:"
-echo "oc exec <pod_name> $HOME/utils/logs"
```
richm: I don't think we want to get rid of this - the message should read
`If you want to print out the logs, use command:`
`oc exec <pod_name> -- logs`

fluentd/run.sh
```diff
@@ -366,6 +364,10 @@ fi
 mkdir -p /var/log/fluentd/
 
 issue_deprecation_warnings
+# Output logs via stdout
+if [ -f $LOGGING_FILE_PATH ]; then
+  tail -n 1000 $LOGGING_FILE_PATH
+fi
```
richm: There probably isn't a log at this point - fluentd will create this log file

richm:
> Not sure this is acceptable or not, but at least "oc logs" outputs fluentd logs...

No, that is not what we want.  The only thing `oc logs $fluentd_pod` should show is the output of the `run.sh` script itself.  All of the output of fluentd should go to `/var/log/fluentd/fluentd.log`

fluentd/run.sh
nhosoi: Thanks, @richm.
Yeah, I also thought it cannot be this simple... :p  I verified the above `tail` just outputs the content of the log file at the startup...  Reverted it.

I wish "oc logs" took an option to tail the file or another option to pass a command line to run inside the pod...  but it seems there's on such trick (probably, it'd add a security issue, though...)

fluentd/run.sh
```diff
@@ -19,7 +19,7 @@ done
 echo "============================="
 echo "Fluentd logs have been redirected to: $LOGGING_FILE_PATH"
 echo "If you want to print out the logs, use command:"
-echo "oc exec <pod_name> $HOME/utils/logs"
+echo "oc exec <pod_name> /usr/local/bin/logs"
```
richm: Does `oc exec <pod_name> -- logs` work?  If not, we need to find out why the `logs` command is not in the default PATH.

fluentd/run.sh
richm: not sure what you are trying to do here

nhosoi:
@richm 
> Does oc exec <pod_name> -- logs work?

Yes, it does.  Actually, it works with and without '--'.  We can change the echo line to:
`echo "oc exec <pod_name> logs"`

Yet another silly idea...  With this `tail -f`, oc logs logging-fluentd-POD works, too.  Not sure how reliable it'd be, though.  Once the tail process dies, it stops there. (i.e., no more new logs...)
https://github.com/nhosoi/origin-aggregated-logging/blob/release-3.11-1643340.0/fluentd/run.sh#L369-L371

richm:
I don't think we want `oc logs $fluentd_pod` to work.  That defeats the purpose of having fluentd log to a file only, which is to prevent feedback loops caused by fluentd logging to `oc logs $fluentd_pod`.

fluentd/run.sh
```diff
@@ -19,7 +19,7 @@ done
 echo "============================="
 echo "Fluentd logs have been redirected to: $LOGGING_FILE_PATH"
 echo "If you want to print out the logs, use command:"
-echo "oc exec <pod_name> $HOME/utils/logs"
+echo "oc exec <pod_name> logs"
```
richm: ok - you only need to use the `oc exec $fluentd_pod -- $command` form if you need to pass command line arguments beginning with `-` to `$command` - in this case, `logs` doesn't have any arguments

fluentd/run.sh
nhosoi: Ok.  Done.
So, for bz1643340, there's no more to fix!?

nhosoi:
Sorry, I misunderstood the goal of the fixes (one of 2 issues).  One is fixing the utility path.  Could you update me about the another one?  (I thought making `oc logs` work agin, but I was wrong...)

fluentd/run.sh
richm: There were a couple of other issues reported in that bz

> rsh into a fluentd pod, there is no fluentd.log file in /var/log/fluentd directory

Can you confirm that this is working?

> No output shows after run "oc exec $fluentd-pod -- logs".

Can you confirm that this is working?

One more thing you might do if you want to learn some bash tricks - make the output of the run.sh script go to the /var/log/fluentd/fluentd.log file.  That is - the only thing going to stdout will be the message about the logging file path.  Then, close stdout/stderr, and redirect stdout/stderr to /var/log/fluentd/fluentd.log for the rest of the process.

```bash
exec >> $LOGGING_FILE_PATH 2>&1
```
that will make all stdout/stderr from all commands go to `$LOGGING_FILE_PATH`

nhosoi: >There were a couple of other issues reported in that bz
` rsh into a fluentd pod, there is no fluentd.log file in /var/log/fluentd directory`
> Can you confirm that this is working?

Yes, there are fluentd.log file and its rotated files in /var/log/fluentd directory in v3.11 openshift:
```
# oc exec logging-fluentd-dflrc -- ls -l /var/log/fluentd
total 9904
-rw-r--r--. 1 root root  887959 Nov 14 18:53 fluentd.log
-rw-r--r--. 1 root root 1024387 Nov 14 18:52 fluentd.log.0
-rw-r--r--. 1 root root 1025148 Nov 14 18:52 fluentd.log.1
-rw-r--r--. 1 root root 1024609 Nov 14 18:51 fluentd.log.2
```
`No output shows after run "oc exec $fluentd-pod -- logs".`
> Can you confirm that this is working?

Yes, it is working as expected.

> One more thing you might do if you want to learn some bash tricks - make the output of the run.sh script go to the /var/log/fluentd/fluentd.log file. That is - the only thing going to stdout will be the message about the logging file path. Then, close stdout/stderr, and redirect stdout/stderr to /var/log/fluentd/fluentd.log for the rest of the process.
`exec >> $LOGGING_FILE_PATH 2>&1`
> that will make all stdout/stderr from all commands go to $LOGGING_FILE_PATH

All right.  Let me test it...

richm:
needs https://github.com/openshift/origin-aggregated-logging/pull/1452 to fix failing tests

richm:
/retest
try again with logstash fix

nhosoi:
Thanks, @richm .  Now "oc exec $fluentd_pod -- logs" includes output from run.sh.

```
=============================
Fluentd logs have been redirected to: /var/log/fluentd/fluentd.log
If you want to print out the logs, use command:
oc exec <pod_name> -- logs
=============================
umounts of dead containers will fail. Ignoring...
umount: /var/lib/docker/containers/27ca70f0adc0733f04154ddd290d396213f1b71d586a265e9a858dcf07c4fa54/shm: not mounted
.....
2018-11-14 21:32:14 +0000 [info]: reading config file path="/etc/fluent/fluent.conf"

```

fluentd/run.sh
```diff
@@ -16,10 +16,12 @@ for file in ${ocp_fluentd_files} ; do
     sed -i -e "s/%OCP_FLUENTD_TAGS%/${OCP_FLUENTD_TAGS}/" $file
 done
 
+mkdir -p `dirname $LOGGING_FILE_PATH`; touch $LOGGING_FILE_PATH; exec >> $LOGGING_FILE_PATH 2>&1
```
richm: This has to go after line 25, after the last `echo "============================="`
We want the notice about the the `$LOGGING_FILE_PATH` and the `logs` command to be available via `oc logs $fluentd_pod`.  Everything after that, we do not want to be available via `oc logs $fluentd_pod`, only via `logs` or `$LOGGING_FILE_PATH`

nhosoi: Ahh, I see.  Sorry.   Fixed.

nhosoi:
https://bugzilla.redhat.com/show_bug.cgi?id=1643340

nhosoi:
https://bugzilla.redhat.com/show_bug.cgi?id=1649997
Bug 1649997 - [3.11] Fluentd pod should log its logs to /var/log/fluentd/fluentd.log by default

richm:
> Next time, should I file a bug or fix it right away?

Submit a PR right away

PR #1458, opened 2018-11-15
Fixes for Mac OS users.

pierreozoux:
As a mac user, you get a bash version 3.* and [line 158](https://stackoverflow.com/questions/39726784/syntax-error-near-unexpected-token-when-using-redirection-operator) is failing.
Then you need to install it with `brew`.
And bash installed with `brew` leaves in `/usr/local/bin/bash`

### DIFF
--- a/hack/logging-dump.sh
+++ b/hack/logging-dump.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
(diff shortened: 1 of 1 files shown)


richm:
works for me on centos7 - @jcantrill  any objections?

jcantrill:
none
/lgtm

PR #1462, opened 2018-11-20
support running tests against local libvirt machine

richm:
Now that api.ci is regularly publishing the latest rpm packages
and images, we no longer require an aws devenv.  This commit
allows devs to stand up and run tests against a local "devenv"
style libvirt machine.

Requirements: a CentOS7 VM with at least 12GB RAM, 40GB disk,
and 4 CPUs.  Can probably be made to work on a smaller vm.
The VM should be set up with the following:

- the machine has an fqdn, not just a local 192.168.122.x IP address e.g.
  you have added an /etc/hosts entry for the machine - this is useful
  when you don't want to or cannot use an external resolver like xip.io
- can ssh as root into the machine with no password, which usually
  means the machine /root/.ssh/authorized_keys has your ssh pub key
- your host has NFS exported your $HOME (or source root) directory, and
  has mapped root to your local uid/gid (default 1000) e.g.

```
  /etc/exports.d/home.exports:
  /home/myusername	192.168.0.0/255.255.0.0(rw,all_squash,anonuid=1000,anongid=1000)
```

- your machine has NFS mounted this export as `/share` inside the VM:

```
  /etc/fstab:
  192.168.122.1:/home/myusername    /share          nfs     nfsvers=3       0 0
```

### DIFF
--- a/hack/get-machine-run-tests.sh
+++ b/hack/get-machine-run-tests.sh
@@ -51,6 +51,26 @@ EXTRA_ENV - (default none) - extra environment variables to pass to the test
 EXTRA_ANSIBLE - (default none) - extra options to add to the ansible-playbook
   command line when running the logging playbook e.g.
   EXTRA_ANSIBLE="-e openshift_logging_use_ops=False -e other=True"
+USE_LOGGING - (default true) - if true, install logging after installing
+  OpenShift using openshift-ansible logging playbook
+TEST_LOGGING - (default true) - if true, run logging CI after installing
+  logging
+PROVIDER - (default aws) - `aws` uses oct and your aws credentials to
+  provision a machine, install OpenShift, and run tests in it
+  `libvirt` assumes you have already provisioned a local libvirt
+  machine running CentOS7 with the following:
+  - the machine has an fqdn, not just a local 192.168.122.x IP address e.g.
+    you have added an /etc/hosts entry for the machine - this is useful
+    when you don't want to or cannot use an external resolver like xip.io
+  - can ssh as root into the machine with no password, which usually
+    means the machine /root/.ssh/authorized_keys has your ssh pub key
+  - your host has NFS exported your $HOME (or source root) directory, and
+    has mapped root to your local uid/gid (default 1000) e.g.
+  /etc/exports.d/home.exports:
+  /home/myusername	192.168.0.0/255.255.0.0(rw,all_squash,anonuid=1000,anongid=1000)
+  - your machine has NFS mounted this export as `/share` inside the VM:
+  /etc/fstab:
+  192.168.122.1:/home/myusername    /share          nfs     nfsvers=3       0 0
 
 see the file $0 for other variables which can be set
 You can also pass these in as environment variables on the command line:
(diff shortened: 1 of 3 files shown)


richm:
I'd like to know what other local vm use cases we should target e.g. I will be responsible for creating the VM and getting the source code onto it. I just want to log in to the vm and run the tests.
I like using NFS because it allows me to use my IDE on my host, and immediately test/run the code in the VM - no sync required.

richm:
another thing - this PR removes the cumbersome and error prone steps to build rpms from openshift-ansible source and install openshift-ansible - instead, we just run the openshift ansible playbooks directly from the source directory

hack/get-machine-run-tests.sh
```diff
@@ -96,22 +116,29 @@ if echo "${EXTRA_ANSIBLE_OPENSHIFT:-}" | grep -q -i "use_crio=true" ; then
 else
     USE_CRIO=${USE_CRIO:-false}
 fi
-export REMOTE_GOPATH=${REMOTE_GOPATH:-/data}
+GOPATH=${GOPATH:-$HOME/go}
```
ewolinetz: we use `go` instead of `golang` ?

hack/get-machine-run-tests.sh
```diff
@@ -123,108 +150,146 @@ ROOT_VOLUME_SIZE=${ROOT_VOLUME_SIZE:-75}
...
-source .venv/bin/activate
-if [ "${NO_SKIP:-0}" = 1 ] ; then
-    if pip show origin-ci-tool > /dev/null ; then
-#        pip install --upgrade git+file://$HOME/origin-ci-tool --process-dependency-links
-        pip install --upgrade git+https://github.com/openshift/origin-ci-tool.git --process-dependency-links
-    else
-#        pip install git+file://$HOME/origin-ci-tool --process-dependency-links
-        pip install git+https://github.com/openshift/origin-ci-tool.git --process-dependency-links
+if [ $PROVIDER = aws ] ; then
+    export REMOTE_GOPATH=${REMOTE_GOPATH:-/data}
```
ewolinetz: is there a reason we `export` the same value twice? (also line 122)

hack/get-machine-run-tests.sh
```diff
@@ -123,108 +150,146 @@ ROOT_VOLUME_SIZE=${ROOT_VOLUME_SIZE:-75}
...
+    # to login to the VM as root with no password e.g. ssh root@the.local.vm "ls"
+    # should not prompt for a password
+    TESTHOST=${TESTHOST:-$1}
+    SSH_HOST=root@$TESTHOST
+    OS_O_A_L_DIR=${OS_O_A_L_DIR:-/share/origin-aggregated-logging}
+    OS_O_A_DIR=${OS_O_A_DIR:-/share/openshift-ansible}
+    OS_A_C_J_DIR=${OS_A_C_J_DIR:-/share/aos-cd-jobs}
+    ip=$( getent hosts $TESTHOST | awk '{print $1}' )
+    fqdn=$( getremotefqdn $ip )
+    export REMOTE_GOPATH=${REMOTE_GOPATH:-/share/go}
```
ewolinetz: this is a different value than line 125?

hack/get-machine-run-tests.sh
richm: I thought the convention was to use `$HOME/go`?  https://golang.org/doc/code.html#GOPATH
" The GOPATH environment variable specifies the location of your workspace. It defaults to a directory named go inside your home directory, so $HOME/go on Unix, $home/go on Plan 9, and %USERPROFILE%\go (usually C:\Users\YourName\go) on Windows. "

hack/get-machine-run-tests.sh
richm: fixed

hack/get-machine-run-tests.sh
richm: fixed

hack/get-machine-run-tests.sh
ewolinetz: oh, I may just not be following the convention then.

hack/get-machine-run-tests.sh
```diff
@@ -123,110 +150,146 @@ ROOT_VOLUME_SIZE=${ROOT_VOLUME_SIZE:-75}
...
+    # /etc/exports.d/home.exports:
+    # /home/myusername	192.168.0.0/255.255.0.0(rw,all_squash,anonuid=1000,anongid=1000)
+    # and inside the VM - /etc/fstab:
+    # 192.168.122.1:/home/myusername    /share      nfs nfsvers=3   0 0
+    # assumes you have set up the VM so that you can use your user SSH identity
+    # to login to the VM as root with no password e.g. ssh root@the.local.vm "ls"
+    # should not prompt for a password
+    TESTHOST=${TESTHOST:-$1}
+    SSH_HOST=root@$TESTHOST
+    OS_O_A_L_DIR=${OS_O_A_L_DIR:-/share/origin-aggregated-logging}
```
nhosoi: We cannot rely on this default setting since if OS_O_A_L_DIR is not set in ~/.config/get-machine-run-tests.sh, the following default value is already assigned at line 146...?
`OS_O_A_L_DIR=${OS_O_A_L_DIR:-$REMOTE_GOPATH/src/github.com/openshift/origin-aggregated-logging}`

I.e., I had to have these in ~/.config/get-machine-run-tests.sh.
```
export OS_O_A_L_DIR=/share/origin-aggregated-logging
export OS_O_A_DIR=/share/openshift-ansible
export OS_A_C_J_DIR=/share/aos-cd-jobs
```

richm:
looks like a bug in openshift-ansible, that the need to add support for 4.0 . . . but they are not likely to do that, since 4.0 has moved to the new openshift-installer

PR #1477, opened 2018-12-04
add replica shards to the config map too

josefkarasek:


elasticsearch/utils/logging
```diff
@@ -18,8 +18,9 @@ esac
 secret_dir=/etc/elasticsearch/secret
 provided_secret_dir=/etc/openshift/elasticsearch/secret
 
-if [ -f /usr/share/java/elasticsearch/config/primary_shards ]; then
-    source /usr/share/java/elasticsearch/config/primary_shards
+# if present, the following file contains PRIMARY_SHARDS and REPLICA_SHARDS
+if [ -f /usr/share/java/elasticsearch/config/shard_settings ]; then
```
richm: how about calling it `index_settings` - that way we can use it for many different things and don't have to rename it? - then you can get rid of the comment too

PR #1486, opened 2018-12-13
fix 1658632. Clean up SG indices

jcantrill:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1658632 by:
* Removing old SG indices
* Removing routing for SG index

Needs cherrypick to master

### DIFF
--- a/elasticsearch/init/0520-bz1658632-remove-old-sg-indices
+++ b/elasticsearch/init/0520-bz1658632-remove-old-sg-indices
@@ -1,0 +1,53 @@
+#!/bin/bash
+#
+# Copyright 2018 Red Hat, Inc. and/or its affiliates
+# and other contributors as indicated by the @author tags.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# init script to cleanup old searchguard indices
+# https://bugzilla.redhat.com/show_bug.cgi?id=1658632
+set -e
+
+if [ -n "${DEBUG:-}" ] ; then
+  set -x
+fi
+
+source "logging"
+source "$(dirname $0)/common"
+
+script=$(basename $0)
+
+info "Starting init script: ${script}"
+
+acknowledged(){
+  echo $1 | python -c "import json,sys;r=json.load(sys.stdin);print 'true' if 'acknowledged' in r and r['acknowledged'] else 'false'"
+}
+
+response=$(es_util --query='.searchguard.*' -XDELETE )
+if [ "true" != "$(es_acknowledged $response)" ] ; then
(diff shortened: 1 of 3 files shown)


elasticsearch/init/0520-bz1658632-remove-old-sg-indices
```diff
@@ -0,0 +1,66 @@
...
+  if k in obj.keys():
+    obj = obj[k]
+  else:
+    obj = "null"
+    break
+print obj
+'
+}
+acknowledged(){
+  echo $1 | python -c "import json,sys;r=json.load(sys.stdin);print 'true' if 'acknowledged' in r else 'false'"
```
ewolinetz: won't this always print true? I thought ES will have `acknowledged: true` or `acknowledged: false`

ewolinetz: won't this always print true? I thought ES will return `acknowledged: true` or `acknowledged: false`

jcantrill: hmm... i have not seen 'acknowledge' if there is an error... but I presume its possible to fail and have false.  I will correct it

elasticsearch/init/0520-bz1658632-remove-old-sg-indices
```diff
@@ -0,0 +1,66 @@
...
+acknowledged(){
+  echo $1 | python -c "import json,sys;r=json.load(sys.stdin);print 'true' if 'acknowledged' in r and r['acknowledged'] else 'false'"
+}
+
+response=$(es_util --query='.searchguard.*' -XDELETE )
+if [ "true" != "$(acknowledged $response)" ] ; then
+  error "Error deleting .searchguard.* indices: ${response}"
+fi
+
+settings=$(es_util --query='.searchguard/_settings')
```
ewolinetz: where is $settings used?

elasticsearch/init/0520-bz1658632-remove-old-sg-indices
```diff
@@ -0,0 +1,66 @@
...
+  set -x
+fi
+
+source "logging"
+
+script=$(basename $0)
+
+info "Starting init script: ${script}"
+
+routing_allocation(){
```
ewolinetz: nothing calls this?

test/0520-bz1658632-remove-old-sg-indices
```diff
@@ -0,0 +1,99 @@
...
+        mycmd=os::log::error
+    fi
+    $mycmd ${test_name} test finished at $( date )
+    # this will call declare_test_end, suite_end, etc.
+    os::test::junit::reconcile_output
+    exit $return_code
+}
+trap "cleanup" EXIT
+
+routing_allocation(){
```
ewolinetz: wouldn't we want to `source` the clean up script and use its functions to verify they work?

elasticsearch/init/0520-bz1658632-remove-old-sg-indices
jcantrill: correcting. got lost in the merge someplace

test/0520-bz1658632-remove-old-sg-indices
jcantrill: The test should exercise any of the function we create.  I would expect to not need to explicitly source them

ewolinetz: wouldn't we want to use the `routing_allocation()` from our clean up script rather than have it copied here as well though?

ewolinetz:
question regarding reusing function, otherwise lgtm

PR #1490, opened 2018-12-20
Dockerfile: remove python-setuptools

sosiouxme:
This package from base RHEL causes problems with container grading downstream; easiest to fix here by simply removing it.


jcantrill:
@sosiouxme cherrypick to 3.11? and master?

sosiouxme:
@jcantrill no need there. might need it in 3.9 at some point (feel free to cp it there). right now it's complaining about the pkg but not lowering the grade.

sosiouxme:
nevermind; this got solved in a better place.

PR #1499, opened 2019-01-25
[WIP] [3.11] get-machine-run-tests.sh

nhosoi:
latest image in 3.x env could cause the hang at "openshift_service_catalog install -
Wait for API Server rollout success".  Try the version first.

Adding "-e skip_sanity_checks=true" to logging installation to prevent this error in
Run variable sanity checks:
  "msg": "last_checked_host: localhost, last_checked_var: openshift_master_identity_providers;
  Found removed variables: openshift_node_labels is replaced by openshift_node_groups[<item>].labels; "

### DIFF
--- a/hack/get-machine-run-tests.sh
+++ b/hack/get-machine-run-tests.sh
@@ -559,12 +559,18 @@ pushd $OS_A_C_J_DIR > /dev/null
 # for some reason, the service-catalog image is not available
 # docker images|grep service-catalog is empty
 # so, pull the latest and tag it with ${OPENSHIFT_IMAGE_TAG:-\$( cat ./ORIGIN_IMAGE_TAG )}
+# Note: latest image in 3.x env could hang ansible at "openshift_service_catalog install -
+#       Wait for API Server rollout success"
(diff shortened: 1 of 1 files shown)


hack/get-machine-run-tests.sh
```diff
@@ -785,6 +791,7 @@ ANSIBLE_LOG_PATH=/tmp/ansible-logging.log ansible-playbook -vvv --become \
   -e openshift_logging_es_allow_external=${ES_ALLOW_EXTERNAL:-True} \
   -e openshift_logging_es_ops_allow_external=${ES_OPS_ALLOW_EXTERNAL:-True} \
   -e oreg_url='openshift/origin-\${component}:'"\${release_commit}" \
+  -e skip_sanity_checks=true \
```
richm: Is there a way to fix this?  If you skip all sanity checks, you could end up with a failed installation due to lack of resources - or, worst case, install seems to pass, but openshift and logging will have strange problems that are difficult to debug.

nhosoi:
Not applicable any more.  Closing...

PR #1503, opened 2019-01-30
fix race condition with oc exec

richm:
fix race condition with oc exec

### DIFF
--- a/hack/testing/check-EFK-running.sh
+++ b/hack/testing/check-EFK-running.sh
@@ -30,8 +30,8 @@ if [ "$USE_OPERATOR" = true ] ; then
 		oal_expected_deployments=( "kibana-app" "kibana-infra" )
 		oal_expected_cronjobs=( "curator-app" "curator-infra" )
 		oal_expected_routes=( "kibana-app" "kibana-infra" )
-		oal_expected_services=( kibana-app kibana-infra cluster-logging-operator elasticsearch-app \
-			elasticsearch-app-cluster elasticsearch-infra elasticsearch-infra-cluster elasticsearch-operator )
+		oal_expected_services=( kibana-app kibana-infra elasticsearch-app \
+			elasticsearch-app-cluster elasticsearch-infra elasticsearch-infra-cluster )
 		esdeploy=$( get_es_deployments elasticsearch-app )
 		if [ -z "$esdeploy" ] ; then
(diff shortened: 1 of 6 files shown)


richm:
needs https://github.com/openshift/cluster-logging-operator/pull/97 to merge before tests will pass

richm:
@jcantrill @ewolinetz @nhosoi This is finally ready to go - the way CI works now is quite a bit different - I'm sure you will have questions - but please review so we can get actual logging CI working again with 4.0

hack/testing/check-EFK-running.sh
```diff
@@ -30,8 +30,8 @@ if [ "$USE_OPERATOR" = true ] ; then
 		oal_expected_deployments=( "kibana-app" "kibana-infra" )
 		oal_expected_cronjobs=( "curator-app" "curator-infra" )
 		oal_expected_routes=( "kibana-app" "kibana-infra" )
-		oal_expected_services=( kibana-app kibana-infra cluster-logging-operator elasticsearch-app \
-			elasticsearch-app-cluster elasticsearch-infra elasticsearch-infra-cluster elasticsearch-operator )
+		oal_expected_services=( kibana-app kibana-infra elasticsearch-app \
```
jcantrill: are 'app' and 'infra' still expected given we don't support "split" deployments yet

richm: no - but - there are many other places in the test where we handle the split deployment - I'd like to merge this PR first, then go back and remove the split deployment handling from all of the tests, and fluentd configs, and elasticsearch configs, etc. etc.

hack/testing/templates/logging-ci-test-runner-template.yaml
```diff
@@ -12,6 +12,8 @@ objects:
     namespace: ${TEST_NAMESPACE_NAME}
   spec:
     terminationGracePeriodSeconds: 0
+    nodeSelector:
```
ewolinetz: we need to restrict the test pod to a specific node?

richm: yes
there are a lot of implicit assumptions in our CI tests that there is only 1 fluentd running.  In order to make that work, we restrict fluentd to run only on the node where kibana is running (primarily so fluentd can see the container log files written by kibana).  Which means the test pod also has to run on the same node (so it can see the fluentd position files, log files, buffer files, etc.).

PR #1508, opened 2019-02-08
add troubleshooting to cycle fluentd periodically

jcantrill:
Docs to fix https://bugzilla.redhat.com/show_bug.cgi?id=1664744

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -99,6 +99,113 @@ for p in $(oc get pods -l component=es -o jsonpath={.items[*].metadata.name}); d
 done
 ```
 
+## Fluentd
+### Fluentd is holding onto deleted journald files that have been rotated
+This [issue](https://bugzilla.redhat.com/show_bug.cgi?id=1664744) causes `/var/log` to fill up with file handles of 
+deleted journald files.  The only known solution at this time is to periodically cycle Fluentd.  An OKD `CronJob` can be added to perform this function, for example, by checking every 10 minutes the amount of available space to fluentd:
+```
+apiVersion: v1
+kind: Template
+metadata:
+  name: fluentd-reaper
+objects:
+- apiVersion: rbac.authorization.k8s.io/v1
+  kind: Role
+  metadata:
+    name: fluentd-reaper
+  rules:
+  - apiGroups:
+    - ""
+    resources:
+    - pods
+    verbs:
+    - delete
+  - apiGroups:
+    - ""
+    resources:
+    - pods/exec
+    verbs:
+    - create
+- apiVersion: rbac.authorization.k8s.io/v1
+  kind: RoleBinding
+  metadata:
+    name: fluentd-reaper
+  roleRef:
+    apiGroup: rbac.authorization.k8s.io
+    kind: Role
+    name: fluentd-reaper
+  subjects:
+  - kind: ServiceAccount
+    name: aggregated-logging-fluentd
+    namespace: ${LOGGING_NAMESPACE}
+- apiVersion: batch/v1beta1
+  kind: CronJob
+  metadata:
+    name: fluentd-reaper
+    labels:
+       provider: openshift
+       logging-infra: fluentd-reaper
+  spec:
+    schedule: "${REAP_SCHEDULE}"
+    jobTemplate:
+      spec:
+        template:
+          metadata:
+            labels:
+              provider: openshift
+              logging-infra: fluentd-reaper
+          spec:
+            serviceAccount: aggregated-logging-fluentd
+            serviceAccountName: aggregated-logging-fluentd
+            containers:
+            - env:
+              - name: REAP_THRESHOLD
+                value: "${REAP_THRESHOLD}"
+              name: cli
+              image: ${CLI_IMAGE}
+              command: ["/bin/bash", "-c"]
+              args:
+                - echo "Checking fluentd pods for space issues on /var/log...";
+                  pods=$(oc get pods -l component=fluentd -o jsonpath={.items[*].metadata.name});
+                  for p in $pods; do
+                    echo "Checking $p...";
+                    if ! $(oc get pod $p | grep Running >> /dev/null)  ; then
+                      echo "$p as its not in a Running state. Skipping...";
+                      continue;
+                    fi;
+                    space=$(oc exec -c fluentd-elasticsearch $p -- bash -c 'df --output=pcent /var/log | tail -1 | cut -d "%" -f1 | tr -d " "');
+                    echo "Capacity $space";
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.md
```diff
@@ -99,6 +99,108 @@ for p in $(oc get pods -l component=es -o jsonpath={.items[*].metadata.name}); d
...
+              args:
+                - if [ "$ENABLED" != "true" ]; then
+                    echo "Job is disabled; exiting early";
+                    exit 0;
+                  fi;
+                  echo "Checking fluentd pods for space issues on /var/log...";
+                  pods=$(oc get pods -l component=fluentd -o jsonpath={.items[*].metadata.name});
+                  for p in $pods; do
+                    echo "Checking $p...";
+                    space=$(oc exec -c fluentd-elasticsearch $p -- bash -c 'df -Ph /var/log | tail -1 | cut -s -d " " -f 12 | rev | cut -b 2- | rev');
```
pweil-: not sure if this is more readable but it removes the reversing `df --output=pcent /var/log | tail -1 | cut -d "%" -f1 | tr -d " "`

docs/troubleshooting.md
```diff
@@ -99,6 +99,108 @@ for p in $(oc get pods -l component=es -o jsonpath={.items[*].metadata.name}); d
...
+              provider: openshift
+              logging-infra: fluentd-reaper
+          spec:
+            serviceAccount: aggregated-logging-fluentd
+            serviceAccountName: aggregated-logging-fluentd
+            containers:
+            - env:
+              - name: ENABLED
+                value: "true"
+              - name: REAP_THRESHOLD
```
pweil-: probably should indicate the value is a percentage somehow

docs/troubleshooting.md
```diff
@@ -99,6 +99,108 @@ for p in $(oc get pods -l component=es -o jsonpath={.items[*].metadata.name}); d
...
+                  fi;
+                  echo "Checking fluentd pods for space issues on /var/log...";
+                  pods=$(oc get pods -l component=fluentd -o jsonpath={.items[*].metadata.name});
+                  for p in $pods; do
+                    echo "Checking $p...";
+                    space=$(oc exec -c fluentd-elasticsearch $p -- bash -c 'df -Ph /var/log | tail -1 | cut -s -d " " -f 12 | rev | cut -b 2- | rev');
+                    echo "Capacity $space";
+                    if [ $space -gt ${REAP_THRESHOLD} ] ; then 
+                      echo "Used capacity exceeds threshold. Deleting $pod";
+                      oc delete pod $p ;
```
pweil-: do you have to wait at all for stability or anything here or is it ok to just kill all the pods if they're all running out of space?

docs/troubleshooting.md
pweil-: Any way to make this more specific to our use case?  If /var/log is filling up in general it seems like we'll take down all pods even though we may not be the cause of it filling.  Maybe that doesn't matter since we'd run out of space anyway but may be a red herring for debugging.

docs/troubleshooting.md
jcantrill: modifying to REAP_THRESHOLD_PERCENTAGE

docs/troubleshooting.md
jcantrill: updated

docs/troubleshooting.md
jcantrill: Assuming the schedule is not a tight window, deleting the pod is sufficient.  I dont see any reason to wait for anything.  There is not sane shutdown to fluent anyway as best I can tell. I don't belive it responds to any SIGHUP

docs/troubleshooting.md
jcantrill: It would be nice if we could check the file descriptors but then we need to build `lsof` (?) into the image and then we could check for the deleted file handles (unless there is some other way).   I have general concerns with using disk space as the threshold; I can imagine a case where we have high log volume that is not being ingested fast enough into Elasticsearch.  It could be that we require a lot of disk to buffer (maybe exceeding the threshold) but that we gain nothing by cycling fluent because its not going to release any file handles that point to files of significant size.  This would put us into an undesired loop.

jcantrill:
@pweil- @richm The other concern I have with this approach of a general image is I saw cases where the script failed but the job was 'complete'.  Not sure how to make the pod exit as failed when the script fails

docs/troubleshooting.md
pweil-: If there are no adverse affects why not just cycle the pods every X minutes of uptime to clear open handles? Does that eliminate the hot loop?

pweil-:
> @pweil- @richm The other concern I have with this approach of a general image is I saw cases where the script failed but the job was 'complete'. Not sure how to make the pod exit as failed when the script fails

`set -euxo pipefail` might help you, `u` may be too much if you want to tolerate unset env vars

docs/troubleshooting.md
richm: I think SIGHUP just causes it to reload config. If it doesn't cause dropping open files, then it won't solve the problem.

docs/troubleshooting.md
```diff
@@ -99,6 +99,112 @@ for p in $(oc get pods -l component=es -o jsonpath={.items[*].metadata.name}); d
...
+          metadata:
+            labels:
+              provider: openshift
+              logging-infra: fluentd-reaper
+          spec:
+            serviceAccount: aggregated-logging-fluentd
+            serviceAccountName: aggregated-logging-fluentd
+            containers:
+            - env:
+              - name: ENABLED
```
jcantrill: @pweil- is this needed? cronjobs can be paused via some other formal fashion?

jcantrill:
> > @pweil- @richm The other concern I have with this approach of a general image is I saw cases where the script failed but the job was 'complete'. Not sure how to make the pod exit as failed when the script fails
> 
> `set -euxo pipefail` might help you, `u` may be too much if you want to tolerate unset env vars

Added but didnt seem to make any difference. Maybe set as a arg to bash?

docs/troubleshooting.md
pweil-:
```go
type CronJobSpec struct {
<snip>
        // This flag tells the controller to suspend subsequent executions, it does
	// not apply to already started executions.  Defaults to false.
	// +optional
	Suspend *bool `json:"suspend,omitempty" protobuf:"varint,4,opt,name=suspend"`
```

ewolinetz: remove this env var since its unnecessary

jcantrill: removed

PR #1509, opened 2019-02-11
Secure ES metrics endpoint with auth proxy

josefkarasek:
ES uses self-signed CA, which isn't trusted by external components,
namely Prometheus. To tackle this, we're adding a side-car proxy
in front of ES, that speak TLS to the outside and its certs are
signed by the openshift signing service. Communication between
the proxy and ES happens within the same network namespace.

### DIFF
--- a/elasticsearch/sgconfig/sg_roles_mapping.yml
+++ b/elasticsearch/sgconfig/sg_roles_mapping.yml
@@ -17,6 +17,7 @@ sg_role_curator:
 sg_role_admin:
   users:
     - 'CN=system.admin,OU=OpenShift,O=Logging'
+
 sg_role_prometheus:
   backendroles:
     - 'prometheus'


josefkarasek:
@jcantrill can you help me finish this? With the above configuration Searchguard is still trying to use cert based auth.

Request from the proxy with the following headers
```
X-Forwarded-For:[127.0.0.1]
X-Remote-User:[system:serviceaccount:openshift-monitoring:prometheus-k8s]
```
Causes
```
[2019-02-11T11:27:39,368][ERROR][c.f.s.h.SearchGuardHttpServerTransport] [elasticsearch-clientdatamaster-0-1] SSL Problem Received fatal alert: bad_certificate
javax.net.ssl.SSLException: Received fatal alert: bad_certificate
	at sun.security.ssl.Alerts.getSSLException(Alerts.java:208) ~[?:?]
        ...
```

josefkarasek:
Minimal changes at the end. No need for this PR.

PR #1514, opened 2019-02-12
[release-3.10] Update eventrouter submodule to fix CentOS image.

patrickdillon:
Update submodule after golang package was removed from CentOS 7.6


richm:
es didn't start - no indication why in logs - try again?  flake?
/retest

richm:
I think 3.10 branch is still broken

patrickdillon:
json-file and logging seem to fail when the Service Catalog does not install. When I look through the logs it looks like controller-manager is stuck in a crashloop because it cant mount the secret controllermanager-ssl. This seems to be a race condition. I'm going to try retesting over the weekend, when there is less demand on the CI node.

richm:
@jcantrill please force merge this PR - then I'll remove the eventrouter hacks from https://github.com/openshift/origin-aggregated-logging/pull/1536 and it should pass too - then we should have working release-3.10 ci

richm:
@jcantrill please force merge this PR - then I'll remove the eventrouter hacks from #1536 and it should pass too - then we should have working release-3.10 ci

PR #1515, opened 2019-02-13
add back tests that require user creation

richm:
wip


richm:
see if reverting https://github.com/openshift/origin-aggregated-logging/pull/1533 fixes acl/multi-tenancy tests

richm:
@jcantrill I had some errors in my test which was causing it to fail - doesn't look like I can reproduce the bz . . .

PR #1516, opened 2019-02-13
must now use apiVersion: template.openshift.io/v1 for Templates

richm:
must now use apiVersion: template.openshift.io/v1 for Templates
when was support for apiVersion: v1 dropped?
it was deprecated in 3.4
it was dropped with the k8s 1.12 rebase


richm:
we need this merged asap or ci will be broken yet again

richm:
I think ci is broken until https://github.com/openshift/release/pull/2845 merges - tests are now failing because we don't have the fedora base image for rsyslog :-(

PR #1517, opened 2019-02-13
Add cluster manage role to curator

pavolloffay:
Jaeger uses curator to delete indices, create templates and aliases.

The aliases and templates are cluster roles https://www.elastic.co/guide/en/shield/2.2/privileges-list.html

With the original curator configuration, I get 403:
```
// crate template
{"error":{"root_cause":[{"type":"security_exception","reason":"no permissions for [indices:admin/template/put] and User [name=CN=system.logging.curator,OU=OpenShift,O=Logging, roles=[]]"}],"type":"security_exception","reason":"no permissions for [indices:admin/template/put] and User [name=CN=system.logging.curator,OU=OpenShift,O=Logging, roles=[]]"},"status":403}

// crete alias
curator.exceptions.FailedExecution: Exception encountered.  Rerun with loglevel DEBUG and/or check Elasticsearch logs for more information. Exception: AuthorizationException(403, 'security_exception', 'no permissions for [indices:admin/aliases] and User [name=CN=system.logging.curator,OU=OpenShift,O=Logging, roles=[]]')
```

### DIFF
--- a/elasticsearch/sgconfig/sg_roles.yml
+++ b/elasticsearch/sgconfig/sg_roles.yml
@@ -35,6 +35,7 @@ sg_role_rsyslog:
 sg_role_curator:
   cluster:
     - CLUSTER_MONITOR
+    - MANAGE
   indices:
     '*':
       '*':


elasticsearch/sgconfig/sg_roles.yml
```diff
@@ -35,6 +35,7 @@ sg_role_rsyslog:
 sg_role_curator:
   cluster:
     - CLUSTER_MONITOR
+    - MANAGE
```
pavolloffay: We can do this or create a `jaeger_curator` with this role.

ewolinetz: I don't see a reason why not to add this here. It seems to me that we would want our curator to be able to manage as well
@lukas-vlcek ?

jcantrill: This give `MANAGE` rights to every index in the cluster.  Is this desired?

pavolloffay:
Could this be merged?

elasticsearch/sgconfig/sg_roles.yml
ewolinetz: just for Curator though, and it is tied to its SAN. If we want to have curator do rollover ourselves it would require this

richm:
rhel-images is just plain broken until we get the ruby-25 image added to the rhel base image registry:
```
error: could not resolve inputs: could not determine inputs for step [input:ruby-25]: could not resolve base image: imagestreamtags.image.openshift.io "4.0:ruby-25" not found
```

pavolloffay:
@jcantrill @richm can this be merged?

richm:
ci/prow/full-integ-aws is failing because of a problem with the cr:
```
time="2019-03-04T15:52:16Z" level=error msg="error syncing key (openshift-logging/elasticsearch): Failed to reconcile Elasticsearch deployment spec: Wrong RedundancyPolicy selected. Choose different RedundancyPolicy or add more nodes with data roles"
```
does the clo hack/cr.yaml need to be updated?

pavolloffay:
Is it a question for me? I don't think this change in SG causes the message you have posted.

richm:
@pavolloffay no, not a question for you

PR #1522, opened 2019-02-19
LOG-263. Document manual EventRouter deployment

jcantrill:
To resolve: https://jira.coreos.com/browse/LOG-263

### DIFF
--- a/docs/deploy-event-router.md
+++ b/docs/deploy-event-router.md
@@ -1,0 +1,111 @@
+# Deploying Event Router
+As of 4.0, the EventRouter is not deployable by the `cluster-logging-operator`. Creating a deployment using the following template
+will enable event collection equivalent to the same capability from 3.11 releases or earlier.
+Currently, all events are logged to `STDOUT`, collected by the logging collector and indexed to the `.operations` index when 
+the `EventRouter` is deployed into an operations namespace (e.g. `openshift-*`). Elevated permissions are required to create the cluster level roles and bindings referenced in the following template:
+
+```
+kind: Template
+apiVersion: v1
+metadata:
+  name: eventrouter-template
+  annotations:
+    description: "A pod forwarding kubernetes events to cluster logging stack."
+    tags: "events,EFK,logging, cluster-logging"
+objects:
+  - kind: ServiceAccount
+    apiVersion: v1
+    metadata:
+      name: cluster-logging-eventrouter
+      namespace: ${NAMESPACE}
+  - kind: ClusterRole
+    apiVersion: v1
+    metadata:
+      name: event-reader
+    rules:
+    - apiGroups: [""]
+      resources: ["events"]
+      verbs: ["get", "watch", "list"]
+  - kind: ClusterRoleBinding
+    apiVersion: v1
+    metadata:
+      name: event-reader-binding
+    subjects:
+    - kind: ServiceAccount
+      name: cluster-logging-eventrouter
+      namespace: ${NAMESPACE}
+    roleRef:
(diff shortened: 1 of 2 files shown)


jcantrill:
cc @mburke5678 . Not sure where this fits into the official documentation.

docs/deploy-event-router.md
```diff
@@ -0,0 +1,111 @@
...
+    - apiGroups: [""]
+      resources: ["events"]
+      verbs: ["get", "watch", "list"]
+  - kind: ClusterRoleBinding
+    apiVersion: v1
+    metadata:
+      name: event-reader-binding
+    subjects:
+    - kind: ServiceAccount
+      name: cluster-logging-eventrouter
```
ewolinetz: to follow the naming convention we have in CLO should can drop the "cluster-logging"?

docs/deploy-event-router.md
```diff
@@ -0,0 +1,111 @@
...
+  - name: IMAGE
+    displayName: Image
+    value: "quay.io/openshift/origin-logging-eventrouter:latest"
+  - name: MEMORY
+    displayName: Memory
+    value: "128Mi"
+  - name: CPU
+    displayName: CPU
+    value: "100m"
+  - name: NAMESPACE
```
ewolinetz: I'm not sure we should give users the ability to choose the namespace since they won't have the ability to change it when CLO performs this

mburke5678:
Added to end user docs:
https://jira.coreos.com/browse/OSDOCS-308

docs/deploy-event-router.md
jcantrill: The ONLY reason I left did not follow the convention was IF someone deployed to a namespace where it wasnt obvious what the event router did that it would be an indicator it was affiliated with cluster logging

docs/deploy-event-router.md
jcantrill: I'm indifferent here.  I don't THINK there should be a reason to move it elsewhere, but it goes inline with my comment regarding the SA name.  If we peg one I would be onboard with doing the same for the namespace.

nhosoi: Although not directly related to the topic, may I ask that eventrouter used to be deployed to the default namespace; why it's being changed?
https://github.com/openshift/openshift-ansible/tree/master/roles/openshift_logging_eventrouter
> eventrouter is deployed to default project, has a service account and its own role to read events

jcantrill: Originally this was because the events deployed to any other namespace but infra would cause the events to go to that index.  In 4.x, 'openshift-*' are all considered infra so the logs will end up in the operations namespace.  The correct solution should be to add a custom parser for events that sends them to an index associated with that namespace.

nhosoi: Thank you, @jcantrill!  I.e., to support eventrouter in the Collector Piplelines, the custom parser for events is going to be added?

docs/deploy-event-router.md
```diff
@@ -0,0 +1,111 @@
...
+        component: eventrouter
+        logging-infra: eventrouter
+        provider: openshift
+    spec:
+      selector:
+		matchLabels:
+          component: eventrouter
+          logging-infra: eventrouter
+          provider: openshift
+      replicas: "1"
```
mburke5678: @jcantrill 
Should this be `replicas: 1` withouth the quotes? I was getting a `unrecognized type: int32` error until I removed the quotes. A lucky guess based on https://bugzilla.redhat.com/show_bug.cgi?id=1595538#c19

PR #1526, opened 2019-02-20
fluent conf generator for collector pipes

jcantrill:
Initial thoughts about conf generation


fluentd/lib/fluentd-okd-config-generator/lib/fluentd-okd-config-generator/templates/store_elasticsearch.erb
```diff
@@ -0,0 +1,41 @@
+<store>
+    @type elasticsearch
+    @id <%= store_id(@target, @context) %>
+    host "<%= @target[:host] %>"
+    port "<%= @target[:port] %>"
+    scheme https
```
richm: So this store is only to be used with the internal cluster Elasticsearch, not an external customer supplied Elasticsearch.

jcantrill: Based on the input in the comment [1] and the description of in the design document, I don't see where this stanza restricts you to an internal cluster.  Assuming we have:
* a resolveable destination
* proper certificates to connect to the endpoints

is there more needed to route externally?

[1] https://github.com/openshift/origin-aggregated-logging/pull/1526/files#diff-a14b23e80be15decfe6c74cb2499ac83R10

richm: Yes - there are some customers who
- want to use http not https (so add the scheme to the endpoint e.g. use `https://host:port`)
- want to use username/password auth
- want to use token auth (provide custom http headers)
- want other stuff too

so - what we have now is fine for the internal cluster Elasticsearch, and probably good enough for some fraction of external cluster Elasticsearch use cases

jcantrill: > * want to use http not https (so add the scheme to the endpoint e.g. use `https://host:port`)
This can be inferred by the presence (or not) of secrets

> * want to use username/password auth
> * want to use token auth (provide custom http headers)
> * want other stuff too

I'm thinking of this purely from the perspective of the need to define something in a ClusterLogging object. I intentionally want to avoid providing customers with EVERY possible configuration as that would need to be surfaced to CL.  IMO, these are permutations we not desire to test and support; our team just doesn't have the coverage.  I believe we need to be extremely opinionated.

One case in point I was investigating yesterday is kafka.  The kafka plugin allows for quite a bit of config which is not common to any other plugins (e.g. topic, broker type).  I believe this is another place where we can establish a pattern like we always writing messages to a topic that is named the same as the source.

richm: ok - Then this config as you have it is fine because it solves our most immediate use cases

richm:
/test full-integ-aws
not sure if flake

richm:
not a flake

fluentd/Dockerfile
```diff
@@ -36,6 +36,11 @@ ADD vendored_gem_src/ ${HOME}/vendored_gem_src/
 ADD install-gems.sh *.patch.sh *.patch ${HOME}/vendored_gem_src/
 
 RUN cd ${HOME}/vendored_gem_src/ && ./install-gems.sh && cd / && rm -rf ${HOME}/vendored_gem_src/
+RUN mkdir -p /tmp/gems
+ADD lib/fluentd-okd-config-generator /tmp/gems/fluentd-okd-config-generator/
+RUN cd /tmp/gems/fluentd-okd-config-generator && \
```
richm: I would advise against creating and installing a gem for this.  Better to copy the files into `/etc/fluent/plugin` instead.

fluentd/configs.d/user/fluent.conf
```diff
@@ -24,12 +24,6 @@
...
-
-## matches
-  @include configs.d/openshift/output-pre-*.conf
-  @include configs.d/openshift/output-operations.conf
-  @include configs.d/openshift/output-applications.conf
-  # no post - applications.conf matches everything left
-##
+  @include configs.d/openshift/pipeline-output-labels.conf
 </label>
+@include configs.d/openshift/pipeline-output-endpoints.conf
```
richm: We don't use this file anymore

fluentd/lib/fluentd-okd-config-generator/.gitignore
```diff
@@ -0,0 +1,2 @@
+.vendor
+fluentd-okd-config-generator-*.gem
```
richm: don't use .gem file

fluentd/lib/fluentd-okd-config-generator/bin/fluentd-okd-config-generator
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env ruby
+
+require 'optparse'
+# require 'fluentd-okd-config-generator'
+require File.join(File.dirname(__FILE__), '..', 'lib/fluentd-okd-config-generator')
```
richm: use `require_relative` instead

fluentd/lib/fluentd-okd-config-generator/bin/fluentd-okd-config-generator
```diff
@@ -0,0 +1,136 @@
...
+  opts.on('-i'  , '--input-file=INPUTFILE', 'The full path to the input file') do |file|
+    options[:input_file] = file
+  end
+  opts.on('-o', '--output-file=OUTPUTFILE', 'The full path for a file that contains match statements only to route tags to labels (e.g. kubernetes.contains.** -> @LOGS_APPS') do |file|
+    options[:output_file] = file
+  end
+  opts.on('-t', '--targets-file=TARGETFILE', 'The full path for a file that contains output configurations from a given source to dest (e.g. @LOGS_APPS->elasticsearch') do |file|
+    options[:target_file] = file
+  end
+  opts.on('-m', '--max-targets-per-source=MAX_TARGETS', Integer, 'The maximum number of targets per source') do |max_targets|
```
richm: why do we need this?

fluentd/lib/fluentd-okd-config-generator/bin/fluentd-okd-config-generator
```diff
@@ -0,0 +1,136 @@
...
+}
+options = {
+    target_file: FluentdOKDConfigGenerator::InputValidator::STDOUT,
+    output_file: FluentdOKDConfigGenerator::InputValidator::STDOUT,
+    tags: {
+      'logs.app' => '**',
+      'logs.infra' => 'journal.** system.var.log** audit.log** default openshift **_openshift-*_** **_kube-*_**'
+    }
+}
+parser = OptionParser.new do |opts|
```
richm: formatting is inconsistent  - some places use 2 spaces for indentation, some use 4

fluentd/lib/fluentd-okd-config-generator/lib/fluentd-okd-config-generator/generator.rb
```diff
@@ -0,0 +1,141 @@
...
+                results.each do |conf|
+                    buffer << "\n"
+                    conf.each do | k, value|
+                        buffer << value.join('\n')
+                    end
+                end 
+            end
+        end
+        
+        def gen_output_label_match_confs(sources, tags)
```
richm: What does this function do?  Are you trying to differentiate which outputs should be the default/fallthrough/last options vs. those which are not?  @nhosoi how was this done in the rsyslog linux system role?

fluentd/Dockerfile
jcantrill: It's not a fluent plugin but a utility that is outputs conf files.  What are the reservations given we are vendoring everything and it lives only within our flunt image?

fluentd/configs.d/user/fluent.conf
jcantrill: Which file?  fluent.conf?  This is a path forward where the image can be tested without dependency on CLO.  I believe I may need to add a few files back in for the case when PIPELINE is disabled

fluentd/lib/fluentd-okd-config-generator/bin/fluentd-okd-config-generator
jcantrill: I have concerns of fluent's memory consumption in relation to the number of targets per source.  This is added as a safety until we can get some perf tests to determine the limitations of the feature.

fluentd/lib/fluentd-okd-config-generator/lib/fluentd-okd-config-generator/generator.rb
jcantrill: This generates the match configs which relabel tags to labels and are included at the end of the @INGRESS label.  The flow is diagramed in the design doc

fluentd/Dockerfile
richm: There's no reason to create a gem - it's just extra work - just use the ruby files directly

fluentd/configs.d/user/fluent.conf
richm: Did you add a test to deploy fluentd using `fluentd/configs.d/user/fluent.conf` without using CLO?  If not, then `fluentd/configs.d/user/fluent.conf` is unused.

fluentd/Dockerfile
richm: We also have other utilities that output files - we don't have gems for those.

fluentd/lib/fluentd-okd-config-generator/lib/fluentd-okd-config-generator/generator.rb
```diff
@@ -0,0 +1,141 @@
...
+            sources.sort! do |first,second|
+                case 
+                when first == 'logs.app'
+                    1
+                when second == 'logs.app'
+                    1
+                else
+                    0
+                end
+            end
```
richm: What happens if we decide that `logs.app` should _not_ be the last/default destination (the one with the `match **`)?  I think that we should have a formal way of specifying the ordering of the outputs rather than assuming `logs.app` will always be last.  I think linux-system-roles rsyslog has some way to denote this - @nhosoi ?

fluentd/lib/fluentd-okd-config-generator/lib/fluentd-okd-config-generator/generator.rb
```diff
@@ -0,0 +1,141 @@
...
+                    case target[:type]
+                    when 'elasticsearch'
+                        target[:store_id] = "#{source.split('.').join('_').downcase()}_elasticsearch"
+                        endpoint = target[:endpoint].split(':')
+                        target[:host] = endpoint.first
+                        target[:port] = endpoint.length == 1 ? "9200" : endpoint[1]
+                        target[:counter] = counters[:elasticsearch]
+                        counters[:elasticsearch] += 1
+                        target
+                    else
```
richm: Do we need to support secure_forward outputs as in 3.11?  Not for 4.2?

fluentd/lib/fluentd-okd-config-generator/lib/fluentd-okd-config-generator/input_loader.rb
```diff
@@ -0,0 +1,56 @@
...
+                logger.debug("Removing sources from input with no targets...")
+                case
+                when config.nil?
+                    logger.warn("Skipping source #{source} because the value is nil")
+                    true
+                when !config.is_a?(Hash)
+                    logger.warn("Skipping source #{source} because the value is not a Hash")
+                    true
+                when config[:targets].nil?
+                    logger.warn("Skipping source #{source} because targets is not a Hash")
```
richm: because no targets were specified

fluentd/lib/fluentd-okd-config-generator/lib/fluentd-okd-config-generator/templates/store_elasticsearch.erb
```diff
@@ -0,0 +1,47 @@
...
+    <%- end -%>
+    target_index_key viaq_index_name
+    id_key viaq_msg_id
+    remove_keys viaq_index_name
+    user fluentd
+    password changeme
+<% unless @target[:tls_key].nil? %> 
+    client_key "<%= @target[:tls_key] %>"
+    client_cert "<%= @target[:tls_cert] %>"
+    ca_file "<%= @target[:tls_cacert] %>"
```
richm: We shouldn't have two separate `unless @target[:tls_key].nil?` sections.
`ca_file` should be in the above

This doesn't matter if client cert auth is the only auth we support, but technically, `client_key` and `client_cert` should be in a separate section about authentication - that is, the user could use `https` with username/password auth or token auth.  But if the user is using client cert auth, they must use `https`.

fluentd/run.sh
```diff
@@ -113,35 +110,42 @@ export BUFFER_QUEUE_FULL_ACTION=${BUFFER_QUEUE_FULL_ACTION:-block}
...
+  - type: elasticsearch
+    endpoint: '$ES_HOST:$ES_PORT'
+    tls_key: '$ES_CLIENT_KEY'
+    tls_cert: '$ES_CLIENT_CERT'
+    tls_cacert: '$ES_CA'
+EOF
     fi
+
+    #use built-in fluent.conf
+    ln -sf $HOME/fluent-pipeline.conf /etc/fluent/fluent.conf
```
richm: what happens to the `/etc/fluent/fluent.conf` mounted from the configmap?

fluentd/lib/fluentd-okd-config-generator/lib/fluentd-okd-config-generator/generator.rb
nhosoi: > @nhosoi how was this done in the rsyslog linux system role?

This is an example of vars.yaml.  In this file, depending on the log collections, it switches the output.
https://github.com/linux-system-roles/logging/#varsyaml

Currently, there's only one output_role elasticsearch is defined.  But we could add other output roles there.
$ ls ~/linux-system-roles/logging/roles/rsyslog/roles/output_roles
elasticsearch/

PR #1527, opened 2019-02-22
fluentd - vendor in all gems as source

richm:
All fluentd gem files are now vendored into the source tree
as their expanded source (`gem unpack`).  The files are under
`fluentd/vendored_gem_src/$gemname` (could not use `vendor`
because that is gitignore'd because it is used by ruby libs
we include).

There is a new script - `hack/update-fluentd-vendor-gems.sh`
developers should periodically run this script to update the
vendored source.  The file `fluentd/Gemfile` is used by the
script to specify fluentd and the plugins we use, and can also
be used to lock down versions if needed.

When the image is built, the source is copied to the image.
The script `fluentd/install-gems.sh` is run inside the image
to install the gems from source.  There are various patches
that are applied to the source and gemspec files, then the
gem files are reconstructed `gem build $gemname.gemspec`,
then finally `gem install` is used to install the gems.


richm:
@smarterclayton @jcantrill @ewolinetz This uses the golang vendoring approach - vendoring in the source code allows us to easily see what has changed since the last time we vendored (as opposed to vendoring in the binary .gem files)
I have not yet merged the upstream and downstream Dockerfiles into a single one (which was one of the goals of this exercise)

sosiouxme:
LOL! Well I suppose that's one way.
(For the console nodejs modules I hacked up automation to resolve deps and insert them into dist-git lookaside to be installed in "offline mode", because devs really did not want to have to deal with it: https://github.com/openshift/ocp-build-data/blob/openshift-4.0/modifications/update-console-sources)

richm:
@smarterclayton @jcantrill before vendoring in source - git archive size is about 3.6MB.  After vendoring in everything for fluentd, including jemalloc, as source, git archive size is about 42.2MB.  If this is acceptable, I will proceed.

fluentd/Dockerfile.centos7
```diff
@@ -18,6 +18,8 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
     yum install -y --setopt=tsflags=nodocs \
       make \
       bc \
+      git \
+      libffi-devel \
```
jcantrill: We should probably convert this to a 2 stage Dockerfile so we don't install the build tools.  Maybe a followon PR

richm: ok

richm:
please review

richm:
flake?
```
[INFO] See if user loguser1-3449 can read /project.multi-tenancy-2.*
error: net/http: TLS handshake timeout
[ERROR] PID 47877: hack/testing/util.sh:200: `oc login --username=${1:-${LOG_ADMIN_USER:-admin}} --password=${2:-${LOG_ADMIN_PW:-admin}} > /dev/null` exited with status 1.
```
/retest

richm:
lots of flakes - trying to update to latest gems to see if that fixes things

richm:
adding hold until I get the CI issues fixed - then I will squash and rebase before merging

PR #1529, opened 2019-02-28
Bug 1684048 - Newly introduced log-rotation in fluentd not working

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1684048

Cause: The files that implement the new log rotation
functionality were not being copied to the correct
fluentd directory.
    
Consequence: Fluentd was not using log rotation and its log
files were not being rotated.

Fix: Change the container build to inspect the fluentd
gem to find out where to install the files.
    
Result: The files that implement log rotation are
copied to the correct directory for fluentd to use.

### DIFF
--- a/fluentd/Dockerfile.centos7
+++ b/fluentd/Dockerfile.centos7
@@ -61,8 +61,7 @@ ADD filter_k8s_meta_for_mux_client.rb /etc/fluent/plugin/
 ADD out_syslog_buffered.rb out_syslog.rb /etc/fluent/plugin/
 ADD parser_viaq_docker_audit.rb viaq_docker_audit.rb /etc/fluent/plugin/
 ADD run.sh generate_throttle_configs.rb generate_syslog_config.rb ${HOME}/
-ADD patch/logger/log.rb ${GEM_HOME}/gems/fluentd-0.12.43/lib/fluent/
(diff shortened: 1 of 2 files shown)


fluentd/Dockerfile.centos7
```diff
@@ -4,7 +4,7 @@ MAINTAINER OpenShift Development <dev@lists.openshift.redhat.com>
 
 ENV DATA_VERSION=1.6.0 \
     FLUENTD_VERSION=0.12.43 \
-    GEM_HOME=/opt/app-root/src \
+    GEM_HOME=/usr/share/gems \
```
nhosoi: Both in the 3.11 and master branch, it is set to /opt/app-root/src.  Are they also need to be fixed?
Dockerfile.centos7:    GEM_HOME=/opt/app-root/src \

richm: I don't know - 3.11 downstream is working as is - not sure why

richm:
nah - it's broken due to the difference in how different versions of docker build process ENV GEM_HOME  - I need to think about this some more

richm:
manually cherrypick to 3.10 - https://github.com/openshift/origin-aggregated-logging/pull/1544

PR #1534, opened 2019-03-01
LOG-281. Document custom certs for kibana route

jcantrill:
This PR provides documentation to resolve:  https://jira.coreos.com/browse/LOG-281


docs/kibana-custom-certs.md
```diff
@@ -0,0 +1,56 @@
...
+      caCertificate: |
+          ${CA_CERT}
+      destinationCACertificate: |
+          ${DEST_CA_CERT}
+      termination: reencrypt
+      insecureEdgeTerminationPolicy: Redirect
+    to:
+      kind: Service
+      name: ${SERVICE_NAME}
+parameters:
```
josefkarasek: I'm thinking, how to fail fast, in case the users screw things up.
I'd mark all these fields as `required`. Then, if eg `cacert` doesn't exist, `CA_CERT="$(cat cacert)` will be empty string and `oc process` will fail.

jcantrill: updated

PR #1535, opened 2019-03-04
[WIP] Fluentd run.sh fixes

nhosoi:
- fixing df option: use available disk size for calculating file buffer queue length.
- to check the file buffer storage type, do not use MUX_FILE_BUFFER_STORAGE_TYPE,
  which is not available. Instead, count the number of mount points and if there
  are multiple, consider the storage type hostmount.

### DIFF
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -83,7 +83,6 @@ if [ -z "${USE_MUX:-}" -o "${USE_MUX:-}" = "false" ] ; then
     else
         export USE_JOURNAL=false
     fi
-    unset MUX_FILE_BUFFER_STORAGE_TYPE
 else
     # mux requires USE_JOURNAL=true so that the k8s meta plugin will look
     # for CONTAINER_NAME instead of the kubernetes.var.log.containers.* tag
@@ -220,10 +219,13 @@ FILE_BUFFER_PATH=/var/lib/fluentd
 mkdir -p $FILE_BUFFER_PATH
 
 # Get the available disk size.
-DF_LIMIT=$(df -B1 $FILE_BUFFER_PATH | grep -v Filesystem | awk '{print $2}')
+DF_LIMIT=$(df -B1 $FILE_BUFFER_PATH --output=avail | tail -1)
 DF_LIMIT=${DF_LIMIT:-0}
-if [ "${MUX_FILE_BUFFER_STORAGE_TYPE:-}" = "hostmount" ]; then
-    # Use 1/4 of the disk space for hostmount.
+# Check how many mount points share the same filesystem (in case hostmount)
+MCNT=$( mount | grep $( mount | grep "$FILE_BUFFER_PATH" | awk '{print $1}' ) | wc -l )
+if [ $MCNT -gt 1 ]; then
+    # Use 1/4 of the disk space for the file buffer
+    # in case user specified FILE_BUFFER_LIMIT is too large.
     DF_LIMIT=$(expr $DF_LIMIT / 4) || :
 fi
 if [ $DF_LIMIT -eq 0 ]; then


fluentd/run.sh
```diff
@@ -220,10 +219,13 @@ FILE_BUFFER_PATH=/var/lib/fluentd
 mkdir -p $FILE_BUFFER_PATH
 
 # Get the available disk size.
-DF_LIMIT=$(df -B1 $FILE_BUFFER_PATH | grep -v Filesystem | awk '{print $2}')
+DF_LIMIT=$(df -B1 $FILE_BUFFER_PATH --output=avail | tail -1)
 DF_LIMIT=${DF_LIMIT:-0}
-if [ "${MUX_FILE_BUFFER_STORAGE_TYPE:-}" = "hostmount" ]; then
-    # Use 1/4 of the disk space for hostmount.
+# Check how many mount points share the same filesystem (in case hostmount)
+MCNT=$( mount | grep $( mount | grep "$FILE_BUFFER_PATH" | awk '{print $1}' ) | wc -l )
```
richm: Can you give an example of what the output looks like inside the fluentd pod?

nhosoi: Sorry, @richm, this is not what you asked for, but these mount points share the hostmount /dev/xvda2.
```
mount: /dev/xvda2 is already mounted or /var/lib/fluentd busy
       /dev/xvda2 is already mounted on /etc/docker
       /dev/xvda2 is already mounted on /var/log
       /dev/xvda2 is already mounted on /usr/share/zoneinfo/UTC
       /dev/xvda2 is already mounted on /dev/termination-log
       /dev/xvda2 is already mounted on /etc/docker-hostname
       /dev/xvda2 is already mounted on /etc/hosts
       /dev/xvda2 is already mounted on /etc/hostname
       /dev/xvda2 is already mounted on /etc/resolv.conf
       /dev/xvda2 is already mounted on /run/secrets
       /dev/xvda2 is already mounted on /var/lib/fluentd
       /dev/xvda2 is already mounted on /etc/sysconfig/docker
       /dev/xvda2 is already mounted on /var/lib/docker
       /dev/xvda2 is already mounted on /var/lib/docker/containers
       /dev/xvda2 is already mounted on /etc/origin/node
       /dev/xvda2 is already mounted on /etc/fluent/configs.d/user

```

richm: This seems fragile - e.g. the first line is going to give `mount` instead of `/dev/xvda2`.  Basically you want to know
- what is the device that `$FILE_BUFFER_PATH` is mounted on
- what other paths have mounted this device

I'm not sure, but maybe there is a better way than grepping the output of mount.

nhosoi: > This seems fragile - e.g. the first line is going to give `mount` instead of `/dev/xvda2`. Basically you want to know
> 
>     * what is the device that `$FILE_BUFFER_PATH` is mounted on
>     * what other paths have mounted this device
> 
> I'm not sure, but maybe there is a better way than grepping the output of mount.

I totally agree with you.  I also searched it if there's an api to return the storage type or any useful info to get the type, but no luck and ended up this plan B...  Ok, I'll continue searching it.

richm:
@nhosoi is there a bz for this?  Do we want to backport this?

nhosoi:
It turned out the calculation for queued_chunks_limit_size in run.sh has no impact on the buffer queue count.  This is the default config being used in my test.
```
queued_chunks_limit_size "#{ENV['BUFFER_QUEUE_LIMIT'] || '32' }"
chunk_limit_size "#{ENV['BUFFER_SIZE_LIMIT'] || '8m' }"
```
Stressing fluentd, the count of buffer queue file is increasing with no limit.
```
$ oc rsh fluentd-stlbm
sh-4.2# cd /var/lib/fluentd/buffer-output-es-config/
sh-4.2# ls -l *.log | wc -l
1632
sh-4.2# df -m .
Filesystem     1M-blocks  Used Available Use% Mounted on
/dev/xvda2        122569 17476    105094  15% /var/lib/fluentd
```
I found this issue was filed on Sep 16, 2018, but no comments were made since then.
https://github.com/fluent/fluentd/issues/2128

If queued_chunks_limit_size is just ignored, there's no need to "fix" run.sh...  Maybe, I should close this pr.

richm:
Is there a problem with fluentd running out of disk space that we need to solve?

nhosoi:
No, I don't think so.  I was rather worried we might be too conservative to limit the disk usage. :)

PR #1538, opened 2019-03-05
use marketplace to deploy logging

richm:
use marketplace to deploy logging

### DIFF
--- a/openshift/ci-operator/build-image/cr.yaml
+++ b/openshift/ci-operator/build-image/cr.yaml
@@ -1,0 +1,31 @@
+apiVersion: "logging.openshift.io/v1alpha1"
+kind: "ClusterLogging"
+metadata:
+  name: "example"
+spec:
+  managementState: "Managed"
+  logStore:
+    type: "elasticsearch"
+    elasticsearch:
+      nodeCount: 1
+      storage: {}
+      redundancyPolicy: "ZeroRedundancy"
+      resources:
+        limits:
+          cpu: 500m
+          memory: 4Gi
+        requests:
+          cpu: 500m
+          memory: 4Gi
+  visualization:
+    type: "kibana"
+    kibana:
+      replicas: 1
+  curation:
+    type: "curator"
+    curator:
+      schedule: "30 3 * * *"
+  collection:
+    logs:
+      type: "fluentd"
+      fluentd: {}
--- a/openshift/ci-operator/build-image/setup-logging-for-e2e.sh
+++ b/openshift/ci-operator/build-image/setup-logging-for-e2e.sh
@@ -6,28 +6,109 @@
 # https://github.com/openshift/release/blob/master/ci-operator/templates/cluster-launch-installer-src.yaml
 # where the script will sourced
 
-# it is expected the image was built with golang
-# but all of this operator stuff could go away if the ci test is changed to
-# deploy logging using OLM and a subscription
-gopath=$( go env GOPATH | cut -f1 -d: )
-if [ ! -d $gopath/src/github.com/openshift/elasticsearch-operator ] ; then
-    git clone https://github.com/${ESO_OPERATOR_REPO:-openshift}/elasticsearch-operator \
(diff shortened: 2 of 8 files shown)


openshift/ci-operator/build-image/cr.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: "logging.openshift.io/v1alpha1"
+kind: "ClusterLogging"
+metadata:
+  name: "example"
```
jcantrill: Do we need to explicitly define the namespace?

openshift/ci-operator/build-image/cr.yaml
```diff
@@ -0,0 +1,28 @@
...
+spec:
+  managementState: "Managed"
+  logStore:
+    type: "elasticsearch"
+    elasticsearch:
+      nodeCount: 1
+      storage: {}
+      redundancyPolicy: "ZeroRedundancy"
+      resources:
+        limits:
```
jcantrill: probably a copy and pace but wouldnt we rather this be 'request' only or peg request and limit to a known value

openshift/ci-operator/build-image/setup-logging-for-e2e.sh
```diff
@@ -6,28 +6,164 @@
...
+        sleep $interval
+    done
+    if [ $ii = $timeout ] ; then
+        return 1
     fi
+    return 0
+}
+
+# Create the openshift-logging namespace:
+cat > openshift-logging-namespace.yaml << EOF
```
jcantrill: Wouldn't it make more sense to define these in files and apply them directly instead of using the bash script?  I shortly will have to introduce a PR to redefine our `manifest` so we get into the Operator Marketplace.  Seems like there is an opportunity to re-use static files.

openshift/ci-operator/build-image/setup-logging-for-e2e.sh
```diff
@@ -6,28 +6,164 @@
...
+if ! wait_for_condition wait_func 300 > ${ARTIFACT_DIR}/test_output 2>&1 ; then
+    echo ERROR: one of or both of elasticsearch-operator and cluster-logging-operator pod not running
+    logging_err_exit
 fi
+
+# get the OLM pod
+olmpod=$( oc -n openshift-operator-lifecycle-manager get pods | awk '/^olm-operator-.* Running / {print $1}' )
+
+# disable the OLM so that we can change images in the cluster-logging-operator
+oc -n openshift-operator-lifecycle-manager scale --replicas=0 deploy/olm-operator
```
jcantrill: Why do we need to do this?  Is it because the `ci-operator` is not testing the `cluster-logging-operator` associated with the PR?

openshift/ci-operator/build-image/cr.yaml
richm: apparently not - not sure if it is because i do `oc project openshift-logging` first, or use `oc -n openshift-logging create -f cr.yaml`

openshift/ci-operator/build-image/cr.yaml
richm: I don't know, but it works like this - what should it look like?

openshift/ci-operator/build-image/setup-logging-for-e2e.sh
richm: Yes.   This test does not use ci-operator.

openshift/ci-operator/build-image/cr.yaml
jcantrill: both

openshift/ci-operator/build-image/cr.yaml
jcantrill: in 3.11 we pegged the request and limit to the same value.  It has something to do with scheduling and ensuring the pod got all the memory it was configuring in the run script.  something like:
```
resources:
  limits:
    memory: 16G
  requests:
    memory: 16G
```

richm:
Marketplace install finally passed - please take a look

Issue #1496, opened 2019-01-15
Cannot enable debug logging for fluentd


konradgaluszka:
Hi all,
I've noticed that fluentd installation in openshift image has **ids** declared in its plugins.
(https://github.com/openshift/origin-aggregated-logging/commit/de661200ef9de2e86101d5920b54f26af711527f).
I was trying to utilize this to enable debug logging in  elasticsearch plugin by adding my custom extra configuration (output-extra-es.conf) with something like this:
```
<store>
  @type elasticsearch
  @id elasticsearch-apps
  @log_level debug
</store>
```
But fluentd is not working as in my assumptions and therefore not merging store configuration from the  image  (in output-es-config.conf). 
Is there any way to enable debug logging only in this one plugin and not globally?
Is there any plan to add the log to output-es-config.conf with proper default level?

Regards

konradgaluszka:
@richm It's OpenShift v3.11.59, image version of fluentd is openshift3/ose-logging-fluentd:v3.11.59.
When I override es-output-config.conf file from /etc/fluent/configs.d/openshift/es-output-config.conf with my configmap and  add @log_level debug:

```
<store>
  @type elasticsearch
  @id elasticsearch-apps
  host "#{ENV['ES_HOST']}"
  port "#{ENV['ES_PORT']}"

 @log_level debug # <-- here
 
 scheme https
  ssl_version TLSv1_2
  target_index_key viaq_index_name
  id_key viaq_msg_id
  remove_keys viaq_index_name
  user fluentd
  password changeme

  client_key "#{ENV['ES_CLIENT_KEY']}"
  client_cert "#{ENV['ES_CLIENT_CERT']}"
  ca_file "#{ENV['ES_CA']}"

  type_name com.redhat.viaq.common
  retry_tag "retry_es"

  reload_connections "#{ENV['ES_RELOAD_CONNECTIONS'] || 'true'}"
  # https://github.com/uken/fluent-plugin-elasticsearch#reload-after
  reload_after "#{ENV['ES_RELOAD_AFTER'] || '200'}"
  # https://github.com/uken/fluent-plugin-elasticsearch#sniffer-class-name
  sniffer_class_name "#{ENV['ES_SNIFFER_CLASS_NAME'] || 'Fluent::ElasticsearchSimpleSniffer'}"
  reload_on_failure false
  flush_interval "#{ENV['ES_FLUSH_INTERVAL'] || '1s'}"
[...]

PR #1543, opened 2019-03-08
Set fluent es output plugin log level for es and es ops #1496

konradgaluszka:


fluentd/configs.d/openshift/output-es-ops-config.conf
```diff
@@ -1,6 +1,7 @@
     <store>
       @type elasticsearch
       @id elasticsearch-ops
+      @log_level "#{ENV['ES_OPS_LOG_LEVEL'] || 'warn'}"
```
richm:
```
@log_level "#{ENV['ES_OPS_LOG_LEVEL'] || ENV['ES_LOG_LEVEL'] || 'warn'}"
```
that way, you can just set `ES_LOG_LEVEL` and it will apply to both ops and non-ops, or you can set `ES_OPS_LOG_LEVEL` separately

richm:
/test full-integ-aws
@konradgaluszka don't worry - it isn't your PR - our CI system is having some problems - I will keep restarting the test

PR #1544, opened 2019-03-08
Bug 1684210 - Newly introduced log-rotation in fluentd not working

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1684210
Cause: The files that implement the new log rotation
functionality were not being copied to the correct
fluentd directory.

Consequence: Fluentd was not using log rotation and its log
files were not being rotated.

Fix: Change the container build to inspect the fluentd
gem to find out where to install the files.

Result: The files that implement log rotation are
copied to the correct directory for fluentd to use.

(cherry picked from commit 99d29508213fbb6f6642c75898caacc29edec5e4)


richm:
flake?
/test json-file

PR #1546, opened 2019-03-08
set default DHE ephemeral key size to 2048

richm:
set default DHE ephemeral key size to 2048
If you need to change the value e.g. to a smaller size
for old clients, use the environment variable
DHE_TMP_KEY_SIZE


richm:
https://github.com/openshift/origin-aggregated-logging/pull/1552
https://github.com/openshift/origin-aggregated-logging/pull/1553

PR #1554, opened 2019-03-11
Bug 1685243 - Allow MERGE_JSON_LOG=true for indexing of JSON payload fields

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1685243

add support for new fluentd viaq filter undefined fields
keeps the default values the same as what we have now, namely
- do not automatically convert undefined fields to string
- do not convert `.` to `_` in undefined field names
- no maximum number of fields

We don't want to break users who might be doing just fine
with the defaults, but we want to give users the ability
to turn on these knobs.

Cause: Using MERGE_JSON_LOG=true can create fields in the
record which will cause schema and syntax violations in
Elasticsearch.  It can also create too many fields for
Elasticsearch to handle without severe performance problems.

Consequence: Fluentd reports error 400 sending records to
Elasticsearch.  Elasticsearch performance degrades.

Fix: Allow users who experience these problems to tune their
Fluentd to accomodate their log record fields.

Result: Logs are ingested into Fluentd with no errors.
Elasticsearch performance does not degrade.

(cherry picked from commit f59ebef7e182d479d56cd1ea7c4958e2abeb8dc4)


richm:
viaq test flake?
/test logging

richm:
tests finally passed - please review

PR #1568, opened 2019-03-22
Fix bz 1688647 envvar subsitution in curator config files

josefkarasek:
Fix envvar subsitution in config files
PyYaml 5.1 regression doesnt resolve envvars in configuration files
Fix bz 1688647

See https://github.com/elastic/curator/issues/1368


josefkarasek:
@richm it looks like pyyaml 3.12 is the latest&stable pyyaml we have in rpm.
I know that dockerfile.centos works because of how pip builds dependency trees.
Let's see what yum will do.

curator/Dockerfile
```diff
@@ -22,7 +22,8 @@ RUN if [ -n "${LOCAL_REPO}" ] ; then \
      curl -s -o /etc/yum.repos.d/local.repo ${LOCAL_REPO} ; \
     fi
 
-RUN INSTALL_PKGS="elastic-curator-${CURATOR_VER} \
+RUN INSTALL_PKGS="python2-pyyaml-3.12-11 \
```
jcantrill: Do we need to specify the version if it is the latest one we have?

curator/README.md
```diff
@@ -100,7 +100,7 @@ If you want to be exact with curator, it is best to use `days` e.g. `delete: day
 [Curator issue](https://github.com/elastic/curator/issues/569)
 
 ### Compatibility with OpenShift 3.7
-In earlier releases admins could control the timezone and time when curator runs from the curator config. This configuration has been moved to openshift-ansible. The [timezone is not currently configurable](https://github.com/kubernetes/kubernetes/issues/47202), instead the timezone of OpenShift master node is used.
+In earlier releases admins could control the timezone and time when curator runs from the curator config. This configuration has been moved to openshift-ansible. The [timezone is not currently configurable](https://github.com/kubernetes/kubernetes/issues/47202), instead the timezone of the node where the pod runs is used.
```
jcantrill: Is it reasonable to remove this section entirely since this is for 4.x

curator/Dockerfile
richm: It depends.  For downstream, if this package is coming from a RHEL repository, and there are multiple versions of the package, then yes, we need to specify the version.  If this package is coming from a RHAOS compose, then no, because there can be only 1 package version per compose.

curator/README.md
richm: In general, I would rather leave in the references to older versions, although I suppose github users by now know to select the appropriate branch/tag for the release they are using.

jcantrill: I only highlight removing this section because we no longer rely on ansible and might as well remove the section which is no longer relevant in 4x.

josefkarasek: :+1: good way of thinking. Yes we should.

josefkarasek:
/hold

I'm struggling with building the rhel dockerfile

PR #1573, opened 2019-03-25
testing - fix marketplace api group and version

richm:
testing - fix marketplace api group and version
fix recent CI failures due to change of Marketplace APIs
(OperatorSource and CatalogSourceConfig) to
operators.coreos.com/v1 from marketplace.redhat.com/v1alpha1
error messages look like this:
```
error: unable to recognize "openshift/ci-operator/build-image/elasticsearch-catalogsourceconfig.yaml": no matches for kind "CatalogSourceConfig" in version "marketplace.redhat.com/v1alpha1"
```

### DIFF
--- a/hack/testing/entrypoint.sh
+++ b/hack/testing/entrypoint.sh
@@ -35,8 +35,8 @@ fi
 export LOGGING_NS
 
 # if using operators, turn off the managed state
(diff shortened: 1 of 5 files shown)


richm:
> May still fail though because clusterlogging needs to be named instance

Should I add that change to this one then?

PR #1575, opened 2019-03-29
make CI logging installation work with latest 4.1

richm:


richm:
now ready to review - should unblock CI

openshift/ci-operator/build-image/elasticsearch-subscription.yaml
```diff
@@ -7,5 +7,4 @@ spec:
   source: elasticsearch
   sourceNamespace: openshift-operators
   name: elasticsearch-operator
-  startingCSV: elasticsearch-operator.v0.0.1
   channel: preview
```
nhosoi: Is it ok having v1alpha1 intact?
`elasticsearch-subscription.yaml:apiVersion: operators.coreos.com/v1alpha1`

The same question for cluster-logging-subscription.yaml...
`cluster-logging-subscription.yaml:apiVersion: operators.coreos.com/v1alpha1`

richm: > Is it ok having v1alpha1 intact?
> `elasticsearch-subscription.yaml:apiVersion: operators.coreos.com/v1alpha1`
> 
> The same question for cluster-logging-subscription.yaml...
> `cluster-logging-subscription.yaml:apiVersion: operators.coreos.com/v1alpha1`

Yes.  For some reason, the subscription api is still v1alpha1

bparees: i would assume yes, i would have expected that in conjunction w/ the marketplace api changes to v1.  

@ecordell @jwforres can you weigh in?

richm: If so, we'll have to open another PR, if/when that happens . . .

PR #1566, opened 2019-03-21
Fix logging to console

josefkarasek:
When LOGGING_FILE_PATH=console, fluentd logs to file `$HOME/console`
instead to stdout/err

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1690232

PR #1576, opened 2019-03-29
[3.11] Fix logging to console

nhosoi:
When LOGGING_FILE_PATH=console, fluentd logs to file `$HOME/console`
instead to stdout/err

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1693035

Auto cherrypick failed; manually backported https://github.com/openshift/origin-aggregated-logging/pull/1566
(cherry picked from commit 3bb25f8bccfb755f943c571d14fdff31bf7829d1)

### DIFF
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -4,6 +4,7 @@ export MERGE_JSON_LOG=${MERGE_JSON_LOG:-true}
 CFG_DIR=/etc/fluent/configs.d
 ENABLE_PROMETHEUS_ENDPOINT=${ENABLE_PROMETHEUS_ENDPOINT:-"true"}
 OCP_OPERATIONS_PROJECTS=${OCP_OPERATIONS_PROJECTS:-"default openshift openshift- kube-"}
+LOGGING_FILE_PATH=${LOGGING_FILE_PATH:-"/var/log/fluentd/fluentd.log"}
 OCP_FLUENTD_TAGS=""
 for p in ${OCP_OPERATIONS_PROJECTS}; do
     if [[ "${p}" == *- ]] ; then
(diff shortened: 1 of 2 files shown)


nhosoi:
> nack - the implementation of logging in 4.x is completely different - in 4.x, we're using fluentd 1.x, which has built-in logging - but not in 3.x - I'm not sure what's wrong with 3.x - maybe I broke 3.x with [7874f95](https://github.com/openshift/origin-aggregated-logging/commit/7874f950ea0976699a142df4c990aff7eb439f0e) ? I didn't test the downstream code, but the fix worked with upstream . . .

Ok, thanks...  A bit puzzled at your comment "the fix worked with upstream".  Which "fix" you mentioned?  Instead of the patch #1566, am I supposed to use the "fix" in 3.x's?

richm:
https://github.com/openshift/origin-aggregated-logging/pull/1545 fixed a problem that the new log rotation code was broken - so either
- the fix for that PR broke logging to console
or
- logging to console has never worked correctly

either way, the fix for https://github.com/openshift/origin-aggregated-logging/pull/1566 is only applicable to 4.x, not 3.x

nhosoi:
I see.  Then...
"LOGGING_FILE_PATH=console" is already implemented in 3.x.
It works upstream.
But it failed QA.
What I need to do is figuring out why it does not work downstream.
^^^^^ is this correct?

richm:
> I see. Then...
> "LOGGING_FILE_PATH=console" is already implemented in 3.x.
> It works upstream.

I'm not sure if it works upstream.  I don't think we test it, and the default is logging to /var/log/fluentd/fluentd.log

> But it failed QA.
> What I need to do is figuring out why it does not work downstream.
> ^^^^^ is this correct?

Verify that it works upstream first - if not, then the fix is likely to be for both upstream and downstream.

nhosoi:
Ok, thanks for your comments, @richm !

fluentd/run.sh
```diff
@@ -16,16 +17,20 @@ for file in ${ocp_fluentd_files} ; do
...
 
-echo "============================="
-echo "Fluentd logs have been redirected to: $LOGGING_FILE_PATH"
-echo "If you want to print out the logs, use command:"
-echo "oc exec <pod_name> -- logs"
-echo "============================="
-
-if [ ! -d `dirname $LOGGING_FILE_PATH` ]; then
-  mkdir -p `dirname $LOGGING_FILE_PATH`
+loggingargs=""
```
richm: loggingargs is not used?

nhosoi:
> loggingargs is not used?

Oops. :p  Thanks!

nhosoi:
It works as expected.
```
$ oc logs $FLUENTD_POD
=============================
Fluentd logs have been redirected to: /var/log/fluentd/fluentd.log
If you want to print out the logs, use command:
oc exec <pod_name> -- logs
=============================

$ oc exec $FLUENTD_POD -- logs
umounts of dead containers will fail. Ignoring...
umount: /var/lib/docker/containers/1b5af6faf0f15961de3ff5899020f34f7ddfa00402aa910f0304f6a81ba66522/shm: not mounted
umount: /var/lib/docker/containers/1d307f6e42420d52faac1f94bd28fa13af4c22d6ca82d856512912bac7bb176e/shm: not mounted

$ oc set env ds/logging-fluentd LOGGING_FILE_PATH=console

$ oc  logs $NEW_FLUENTD_POD
umounts of dead containers will fail. Ignoring...
umount: /var/lib/docker/containers/1b5af6faf0f15961de3ff5899020f34f7ddfa00402aa910f0304f6a81ba66522/shm: not mounted
umount: /var/lib/docker/containers/1d307f6e42420d52faac1f94bd28fa13af4c22d6ca82d856512912bac7bb176e/shm: not mounted
```

PR #1586, opened 2019-04-05
better enable and disable of cluster-logging-operator

richm:


richm:
Looks like it got past the rsyslog build problem - please re-lgtm and hope we don't hit another flake . . .

hack/testing/util.sh
```diff
@@ -706,3 +706,31 @@ get_fluentd_cm_name() {
...
 
 fluentd_cm=${fluentd_cm:-$(get_fluentd_cm_name)}
+
+enable_cluster_logging_operator() {
+    if oc -n ${LOGGING_NS} get deploy cluster-logging-operator > /dev/null 2>&1 ; then
+        oc patch -n ${LOGGING_NS} clusterlogging instance --type=json --patch '[
+            {"op":"replace","path":"/spec/managementState","value":"Managed"}]'
+        clopod=$( oc -n ${LOGGING_NS} get pods | awk '/^cluster-logging-operator-.* Running / {print $1}' )
+        if [ -n "$clopod" ] ; then
+            oc delete --force pod $clopod
```
ewolinetz: why are we deleting the pod here but we don't do something similar for `disable_cluster_logging_operator`?

richm: in `disable_cluster_logging_operator` the scale down should delete any pods that are running.  But here, I want to make sure that there isn't a clo pod accidentally running when it isn't supposed to be.

richm:
hold until this is working consistently - I will squash before merging

PR #1594, opened 2019-04-10
add access control diagrams

jcantrill:
Add diagrams for access control


richm:
full-integ-aws died because it could not copy the test artifacts - flake?
/test full-integ-aws

PR #1595, opened 2019-04-10
debug recent test failures

richm:


richm:
It finally passed - please review so we can get archival working in CI tests

PR #1598, opened 2019-04-15
LOG-400 - rsyslog - support log rotation for rsyslog log files

nhosoi:
Adding utils/logs to rsyslog
Including cronie for logrotate
To run crond in the foreground mode, /etc/pam.d/crond must not enable pam_loginuid.so module.

### DIFF
--- a/rsyslog/Dockerfile
+++ b/rsyslog/Dockerfile
@@ -24,12 +24,14 @@ FROM registry.svc.ci.openshift.org/ocp/4.0:base
 
 COPY --from=builder /RPMS/ /RPMS/
 
-RUN yum -y install /RPMS/*.rpm && \
+RUN yum -y install /RPMS/*.rpm cronie && \
     rm -rf /RPMS && \
     yum clean all
 
 ADD install.sh /bin/install.sh
 ADD rsyslog.sh /bin/rsyslog.sh
 ADD uninstall.sh /bin/uninstall.sh
+COPY utils/** /usr/local/bin/
+RUN sed -i -e 's,\(session[ 	]*required[ 	]*pam_loginuid.so\),\#\1,' /etc/pam.d/crond
 
 CMD [ "/bin/rsyslog.sh" ]
--- a/rsyslog/utils/logs
+++ b/rsyslog/utils/logs
@@ -1,0 +1,43 @@
+#!/bin/bash -e
+set -euo pipefail
+
+log_dir=`dirname ${LOGGING_FILE_PATH:-/var/log/rsyslog/rsyslog.log}`
+cmd=cat
+
+helpMsg() {
+cat <<MSG
+ Usage: $0 [options]
+ Retrieve Rsyslog logs from the log directory using "LOGGING_FILE_PATH" environment variable.
+   Options:
+     -f                   Follow the log file in the rsyslog log directory.
+                          Default log file is "rsyslog.log".
+     -h --help               Prints out help.
+MSG
+}
+
+
+while (($#))
+do
+case $1 in
+    -f)
+      cmd="tail"
+      tail_args="-f"
+      break
+      ;;
+    --help|-h)
+      helpMsg
+      exit 0
+      ;;
+    *)
+      ;;
+  esac
+  shift
+done
+
+if [ $cmd = "cat" ] ; then
+    for log_file in $( ls $log_dir | sort -Vr ); do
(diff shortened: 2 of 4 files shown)


nhosoi:
Note: This is the OAL side of LOG-400.

nhosoi:
Question: Log rotation is primary implemented in CLO with https://github.com/openshift/cluster-logging-operator/pull/156, is README.md in OAL supposed to have the "changelog"?

jcantrill:
@nhosoi changed this PR to master-post-release.  We should sync master into master-post-release if there are changes you require that are not in the post release base

nhosoi:
@jcantrill, Thank you for changing to master-post-release.  I don't think there's any dependencies on the changes made in the master branch only.

rsyslog/Dockerfile
```diff
@@ -24,12 +24,14 @@ FROM registry.svc.ci.openshift.org/ocp/4.0:base
...
-RUN yum -y install /RPMS/*.rpm && \
+RUN yum -y install /RPMS/*.rpm cronie && \
     rm -rf /RPMS && \
     yum clean all
 
 ADD install.sh /bin/install.sh
 ADD rsyslog.sh /bin/rsyslog.sh
 ADD uninstall.sh /bin/uninstall.sh
+COPY utils/** /usr/local/bin/
+RUN sed -i -e 's,\(session[ 	]*required[ 	]*pam_loginuid.so\),\#\1,' /etc/pam.d/crond
```
richm: We need this when the container is used for cron/logrotate container, but does it have any effect when the container is used for rsyslog?

rsyslog/utils/logs
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash -e
+set -euo pipefail
+
+log_dir=`dirname ${LOGGING_FILE_PATH:-/var/log/rsyslog/rsyslog.log}`
+cmd=cat
+
+helpMsg() {
+cat <<MSG
+ Usage: $0 [options]
+ Retrieve Fluentd logs from the log directory using "LOGGING_FILE_PATH" environment variable.
```
richm: rsyslog

rsyslog/Dockerfile
nhosoi: Yes, that is correct.  How can I add it just to the logrotate container?

rsyslog/utils/logs
nhosoi: :p  Thanks!!

rsyslog/Dockerfile
richm: >  How can I add it just to the logrotate container?

Not sure, but I wouldn't worry about it unless it causes problems for the rsyslog container.

nhosoi: An interesting thing is to rsh into the container, you could do `"oc rsh $RSYSLOG_POD -c rsyslog"` for rsyslog, but it does not work for logrotate.  I.e., `"oc rsh $RSYSLOG_POD -c logrotate"` fails.  But kubectl works `"kubectl exec -it $RSYSLOG_POD -c logrotate /bin/sh"`.

Do you think it's normal?

richm: I don't know, but it doesn't matter as long as rsyslog works normally as an openshift pod.

nhosoi:
@richm , I have another question.  I'd like to add the description about the rsyslog logrotation to readme such as the environment variables and the way how to see the rsyslog logs.  Should it go to the readme in OAL or CLO?  The code is mainly in CLO except the log utility...

richm:
> @richm , I have another question. I'd like to add the description about the rsyslog logrotation to readme such as the environment variables and the way how to see the rsyslog logs. Should it go to the readme in OAL or CLO? The code is mainly in CLO except the log utility...

We should have an oal/rsyslog/README.md just like the one from fluentd https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/README.md - in fact, you could probably copy/paste some of that into rsyslog/README.md

richm:
/test full-integ-aws
looks like flakes - was hoping that no one would be using our aws resources during the holiday . . .

nhosoi:
> We should have an oal/rsyslog/README.md just like the one from fluentd https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/README.md - in fact, you could probably copy/paste some of that into rsyslog/README.md

Done.  Thanks!

PR #1606, opened 2019-04-23
update repo used with UBI

richm:
The yum repo used with UBI has moved.
See also https://github.com/openshift/cluster-logging-operator/pull/158


nhosoi:
I made similar changes in my local scripts although I was using 4.0 from the new location like this...
${GOPATH:-}/src/github.com/openshift/release/ci-operator/infra/openshift/release-controller/repos/ocp-4.0-default.repo

/lgtm

PR #1607, opened 2019-04-23
generate manifests for use in SRT tracking

richm:
For https://jira.coreos.com/browse/LOG-393
For fluentd, rsyslog, and elasticsearch, the images
where we have non-RPM content, we need to generate a
manifest for SRT to use, to track security issues in
the components we use.

For rsyslog, at source update time, when the
`hack/update-rsyslog-vendor-src.sh` script is run, it will
generate a `/contents` file containing the components, rpm
package name, version, and checksum.

For the other components with non-rpm content, this `/contents`
file will be created at image build time - the version, name,
etc. will be extracted from the jars/gems/whatever native
packaging format.

After the image has been build, the script `get-srt-manifest.sh`
will grab the image from the internal registry and generate
a manifest in SRT format using the image metadata and the
manifest.


richm:
please review
This pr does not cover elasticsearch nor kibana, but I would like to go ahead and get this PR merged so we can begin tracking fluentd (the main image in 4.1 we need tracking for) and rsyslog.

sosiouxme:
/approve
thanks :)

jasinner:
I got the back/build-image.sh to work locally. Where would I find the manifest? In the built containers?

richm:
> I got the back/build-image.sh to work locally. Where would I find the manifest? In the built containers?

We decided to use rh-manifest.txt which lives in git and is updated when the source files are updated.  Right now only fluentd and rsyslog will produce the rh-manifest.txt file.  rsyslog also has some golang sources - not sure if we need a manifest for those, or if it will be handled by the golang container automation.
And I'm still not sure what to do about elasticsearch, or if we can rely on existing Java container scanning.

richm:
hey what do you know - finally passed without a flake - miracles do happen - please review and `/lgtm`

elasticsearch/build-contents-manifest.sh
```diff
@@ -0,0 +1,204 @@
...
+    fi
+    if [ -z "$ver" ] ; then
+        ver=$( get_jar_ver $jarname )
+    fi
+    if [ -n "${jarver:-}" -a ! "${jarver:-}" = "${ver}" ] ; then
+        verbose ERROR: $jarname version $ver does not match version from jar $jarver
+    fi
+    local basename
+    case $jarname in
+    accessors-smart-*) basename=net.minidev-accessors-smart ;;
```
jcantrill: Speculating with no understanding if it is possible, but is this retrievable from the maven dependency goal?  Seems like we could miss something if we dont remember to update this list

richm: Just ignore this for now.  There should be a better way to extract Java information from the middleware team.

richm:
> My only comment would be that if Elasticsearch still shades some dependencies then I am not sure it would be picked up. We know Elasticsearch shaded some jars in the past, but I completely lost track if it is still the case after it switched to gradle build.
> 
> But even if ES (or any of its modules/plugins) still shades external jars then I do not think we can do much about it. For the context here you can read more about the [original reasoning behind shading](https://discuss.elastic.co/t/netty-transport-and-shaded-jar/10323/5) (not sure how much of that is still actual today).

Please ignore the Elasticsearch stuff in this commit - I guess I should just remove it.

richm:
I have removed the Elasticsearch stuff from this commit.  We'll have to figure out another way to address that, hopefully using tooling provided by the Java/middleware teams.

Please review and `/lgtm`

richm:
tests passed - please review - hopefully for the last time

richm:
https://github.com/openshift/origin-aggregated-logging/pull/1670

PR #1616, opened 2019-04-26
dev - logging deploy, test, undeploy and improvements

richm:
Split up the e2e script into separate scripts for deploying
logging and running the logging tests - the new scripts
`hack/deploy-logging-marketplace.sh` and `hack/test-logging.sh`

This also adds a script `hack/undeploy-logging-marketplace.sh`
so that devs can repeatedly deploy and undeploy logging.

The main entry point to all of these is `hack/get-cluster-run-tests.sh`
which can be run multiple times in multiple phases.

For example, if you use:

```
BUILD_IMAGES=false PUSH_IMAGES=false DEPLOY_LOGGING=false \
TEST_LOGGING=false hack/get-cluster-run-tests.sh
```

It will create and deploy an openshift cluster.  You can then
run `hack/get-cluster-run-tests.sh` to perform the other steps
like this, where `/tmp/openshift-cluster` is the temporary
directory created by the first phase containing the `auth`
and `installdir` directories:

```
DEPLOY_OPENSHIFT=false WORKDIR=/tmp/openshift-cluster \
hack/get-cluster-run-tests.sh
```

This will build the images locally, push them to the cluster,
deploy logging to the cluster, and run the logging tests.

If you just want to deploy logging to the cluster without
running the tests, use:

```
DEPLOY_OPENSHIFT=false WORKDIR=/tmp/openshift-cluster \
TEST_LOGGING=false hack/get-cluster-run-tests.sh
```


nhosoi:
/lgtm
Thanks soooo much for adding this feature, @richm !!

PR #1619, opened 2019-04-29
add rsyslog_exporter for prometheus

richm:
https://jira.coreos.com/browse/LOG-9

This adds the source for the soundcloud rsyslog_exporter
to the rsyslog image and builds it in the Dockerfile.
Requires https://github.com/openshift/cluster-logging-operator/pull/164 to configure it.


richm:
@bparees is this the right way to vendor in the rsyslog_exporter source code?

.gitignore
```diff
@@ -11,3 +11,4 @@ venv
 vendor
 !fluentd/vendored_gem_src/http_parser.rb/ext/ruby_http_parser/vendor
 !fluentd/vendored_gem_src/multi_json/lib/multi_json/vendor
+!rsyslog/go/src/github.com/soundcloud/rsyslog_exporter/vendor
```
bparees: why are we git ignoring vendored folders?  the idea here is that the repo should be entirely self-contained so i can just git clone it and build exactly what was committed to the repo, with no risk of dependencies having changed or gone missing.

richm: For ruby, the `vendor/` folder is used for the virtual env - 
```
cd fluentd/lib/generate_throttle_configs/
GEM_HOME=vendor bundle install
GEM_HOME=vendor bundle exec rake test
```
I suppose we could use a different name but the convention is to use `vendor/`.  @jcantrill @nhosoi are you ok with using a name other than `vendor/` for ruby bundle use?

richm:
closing in favor of https://github.com/openshift/origin-aggregated-logging/pull/1643

PR #1621, opened 2019-05-02
use openshift-operators-redhat for elasticsearch-operator

richm:
Change CI to create namespace openshift-operators-redhat and
operatorgroup.

FYI: The elasticsearch-operator needs to watch all namespaces
for elasticsearch objects.
https://github.com/operator-framework/community-operators/blob/master/community-operators/elasticsearch-operator/elasticsearch-operator.v4.1.0.clusterserviceversion.yaml#L90
This means we cannot list the targetNamespaces in the operatorgroup.
I could not find any documenation about operatorgroups, but I figured
out that if you do not set targetNamespaces in the operatorgroup it
will tell it to watch all namespaces:
`spec: {}`

### DIFF
--- a/hack/deploy-logging-marketplace.sh
+++ b/hack/deploy-logging-marketplace.sh
@@ -18,7 +18,7 @@ set -eux
 logging_err_exit() {
     oc get deploy >> ${ARTIFACT_DIR}/logging_err_exit.log 2>&1 || :
     oc get pods >> ${ARTIFACT_DIR}/logging_err_exit.log 2>&1 || :
-    oc -n openshift-operators get elasticsearch >> ${ARTIFACT_DIR}/logging_err_exit.log 2>&1 || :
+    oc -n $ESO_NS get elasticsearch >> ${ARTIFACT_DIR}/logging_err_exit.log 2>&1 || :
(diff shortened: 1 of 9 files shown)


mburke5678:
Adding new process to docs for 4.1
https://github.com/openshift/openshift-docs/pull/14656

mburke5678:
@richm @jcantrill 
Is this something that will be _fixed_ in 4.2 (or later)? So that folks can use the web console to install ESO?

ewolinetz:
@richm squash? LGTM

anpingli:
@richm  @jcantrill  @mburke5678  I don't think that is good idea to use  the namespace openshift-operators-redhat.   Duplicated EO may be deployed.    One is in openshift-operators namespace deployed via Webconsole. One is in openshift-operators-redhat deployed via CLI.   Why not use same name space?

bparees:
It cannot be deployed in openshift-operators because we can't allow
promethueus to scrape that namespace. If an admin installs it there it is
user error.


Ben Parees | OpenShift

richm:
ok - ready to go - assuming tests pass, no flakes

richm:
finally passed - please re-review and lgtm

PR #1625, opened 2019-05-05
debug ci failures

richm:


richm:
This should reduce some of the test output noise

PR #1628, opened 2019-05-06
no oc logs output for logging test pod

richm:
Do not send the output of the logging pod to stdout/stderr -
it cause problems with logging reading the output.
Make logging undeploy unlabel the nodes.
Try to fix flakes in fluentd-forward.
Make sure kibana runs on the logging-ci-test node if it gets restarted.


richm:
I think this PR is finally ready . . . if it actually passes without more flakes :|

richm:
Logging test suite test-multi-tenancy failed due to test flake:
```
[INFO] See if user loguser1-5748 is denied /project.default.*
F0509 16:34:23.282962   44709 helpers.go:119] error: net/http: TLS handshake timeout
```

richm:
finally - no flakes - please review so we can get this merged

PR #1629, opened 2019-05-06
Bug 1707538  - Fix typo of random_string that prevents forwarding to S…

nhosoi:
…YSLOG via UDP to fail

https://bugzilla.redhat.com/show_bug.cgi?id=1703904#c3
Although facility is supposed to have a numeric value (see man systemd.journal-fields)
or the corresponding string keys, systemd passes facility values which are not included
in the valid value set.  This patch resets the value to the default if such invalid
value is detected.

### DIFF
--- a/fluentd/out_syslog.rb
+++ b/fluentd/out_syslog.rb
@@ -72,14 +72,14 @@ module Fluent
           @packet.hostname = record['hostname'] || hostname
           @packet.severity = (record['level'].eql? 'warning')? 'warn' : record['level'] || @severity
           if @use_record && record.key?('systemd') && (record['systemd']).key?('u') && (record['systemd']['u']).key?('SYSLOG_FACILITY')
-            fval = record['systemd']['u']['SYSLOG_FACILITY'].to_i
-            if (1..23).include?(fval)
-              @packet.facility = fval
-            else
-              if record['systemd']['u']['SYSLOG_FACILITY'].eql? '0'
-                @packet.facility = 0
-              else
-                @packet.facility = record['systemd']['u']['SYSLOG_FACILITY'] || @facility
+            begin
+              @packet.facility = record['systemd']['u']['SYSLOG_FACILITY']
+            rescue
+              begin
+                @packet.facility = Integer(record['systemd']['u']['SYSLOG_FACILITY'])
+              rescue
+                log.warn "out:syslog: invalid facility value #{record['systemd']['u']['SYSLOG_FACILITY']}; reset to default #{@facility}"
+                @packet.facility = @facility
               end
             end
           elsif record.key?('_KERNEL_DEVICE')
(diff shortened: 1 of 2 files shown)


richm:
I think this entire code block could be simplified: https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/out_syslog.rb#L75-L84

```ruby
              begin
                @packet.facility = record['systemd']['u']['SYSLOG_FACILITY']
              rescue
                begin
                  @packet.facility = Integer(record['systemd']['u']['SYSLOG_FACILITY'])
                rescue
                  log.warn "out:syslog: invalid facility value #{record['systemd']['u']['SYSLOG_FACILITY']}; reset to default #{@facility}"
                  @packet.facility = @facility
                end
              end
```
The `@packet.facility = record['systemd']['u']['SYSLOG_FACILITY']` will raise `ArgumentError` if not an integer or string not valid.  The `@packet.facility = Integer(record['systemd']['u']['SYSLOG_FACILITY'])` will raise `ArgumentError` if the value is not an integer (raised by `Integer`), or if the value is an integer and is not in the range `0` to `23` (raised by `facility=`).  This should also correctly handle the case where `record['systemd']['u']['SYSLOG_FACILITY']` is the string value `'0'`, so we don't need a special case for that.

nhosoi:
> I think this entire code block could be simplified: https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/out_syslog.rb#L75-L84
> 
> ```ruby
>               begin
>                 @packet.facility = record['systemd']['u']['SYSLOG_FACILITY']
>               rescue
>                 begin
>                   @packet.facility = Integer(record['systemd']['u']['SYSLOG_FACILITY'])
>                 rescue
>                   log.warn "out:syslog: invalid facility value #{record['systemd']['u']['SYSLOG_FACILITY']}; reset to default #{@facility}"
>                   @packet.facility = @facility
>                 end
>               end
> ```
> 
> The `@packet.facility = record['systemd']['u']['SYSLOG_FACILITY']` will raise `ArgumentError` if not an integer or string not valid. The `@packet.facility = Integer(record['systemd']['u']['SYSLOG_FACILITY'])` will raise `ArgumentError` if the value is not an integer (raised by `Integer`), or if the value is an integer and is not in the range `0` to `23` (raised by `facility=`). This should also correctly handle the case where `record['systemd']['u']['SYSLOG_FACILITY']` is the string value `'0'`, so we don't need a special case for that.

Thank you, @richm.  Replaced the code with yours and pushed to master.

richm:
One more thing - if the remote syslog protocol is extensible and allows us to add custom fields, then the `SYSLOG_FACILITY` field should be added to the outgoing record with the invalid value. If the protocol is not extensible, then not applicable.

nhosoi:
> One more thing - if the remote syslog protocol is extensible and allows us to add custom fields, then the `SYSLOG_FACILITY` field should be added to the outgoing record with the invalid value. If the protocol is not extensible, then not applicable.

Could you give me a little more details?  If "the remote syslog protocol" means this implementation https://github.com/eric/syslog_protocol, I cannot see it could be extensible.  I.e., facility is strictly confined to [0..23] and its string representation.
https://github.com/eric/syslog_protocol/blob/master/lib/syslog_protocol/packet.rb#L26-L42

And the packet format is pretty much hardcoded.
https://github.com/eric/syslog_protocol/blob/master/lib/syslog_protocol/packet.rb#L14

Or could this implementation is just a small set of "remote syslog protocol" and it could be extensible?

PR #1630, opened 2019-05-06
Bug 1707176 - JOURNAL_READ_FROM_HEAD defaults to true

richm:
For the JOURNAL env vars, only use the value if it is set *and* the value is not an empty string.
Otherwise, if `JOURNAL_READ_FROM_HEAD` is set to an empty value, this
```
      read_from_head "#{ENV['JOURNAL_READ_FROM_HEAD'] || 'false'}"
```
will expand to this
```
     read_from_head
```
since `read_from_head` is a boolean parameter, this tells Fluentd to set the value to `true`, which is the opposite of the intention.

### DIFF
--- a/fluentd/configs.d/openshift/input-pre-systemd.conf
+++ b/fluentd/configs.d/openshift/input-pre-systemd.conf
@@ -2,7 +2,7 @@
   @type systemd
   @id systemd-input
   @label @INGRESS
-  path "#{ENV['JOURNAL_SOURCE'] || '/run/log/journal'}"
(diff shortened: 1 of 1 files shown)


richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1707176

richm:
I believe this is also the cause of some of our CI test flakes related to the `logger` operations log value not being found.  Many of our tests clear the journal cursor value expecting the systemd journal input plugin to read from tail.  But since it reads from head by default in this case, it takes too long and the tests time out.

PR #1645, opened 2019-05-13
Precreate indicies documentation

jcantrill:
Documentation for index precreate.  Pulling into 4.x will be a slight modification but ideally we could push it into 4.x operator for ES5 deployments

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -98,6 +98,144 @@ for p in $(oc get pods -l component=es -o jsonpath={.items[*].metadata.name}); d
   oc exec -c elasticsearch $ES_POD -- touch /opt/app-root/src/init_failures;  \
 done
 ```
+### Elasticsearch performance degrades around 00:00 UTC
+This situation is a result of Elasticsearch processing bulk index requests from all the nodes in the OKD cluster for
+logs for the new day.  Creation of a significant number of new indices simultaneously can cause bulk index processing 
+to slow down as it waits for all members of the cluster to become aware of each new index.  This
+is likely to happen when there are many active projects (namespaces) in a cluster, since each projects has its own index.
+```
+apiVersion: v1
+kind: Template
+metadata:
+  name: indices-precreate
+objects:
+- apiVersion: rbac.authorization.k8s.io/v1
+  kind: Role
+  metadata:
+    name: precreate-indices
+  rules:
+  - apiGroups:
+    - ""
+    resources:
+    - pods
+    verbs:
+    - list
+    - get
+  - apiGroups:
+    - ""
+    resources:
+    - pods/exec
+    verbs:
+    - create
+- apiVersion: rbac.authorization.k8s.io/v1
+  kind: RoleBinding
+  metadata:
+    name: precreate-indices
+  roleRef:
+    apiGroup: rbac.authorization.k8s.io
+    kind: Role
+    name: precreate-indices
+  subjects:
+  - kind: ServiceAccount
+    name: ${CRON_SERVICE_ACCOUNT}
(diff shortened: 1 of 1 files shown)


jcantrill:
@anpingli Could you sanity test this?

jcantrill:
cc @portante

docs/troubleshooting.md
```diff
@@ -98,6 +98,142 @@ for p in $(oc get pods -l component=es -o jsonpath={.items[*].metadata.name}); d
...
+              logging-infra: indices-precreate
+          spec:
+            serviceAccount: ${CRON_SERVICE_ACCOUNT}
+            serviceAccountName: ${CRON_SERVICE_ACCOUNT}
+            containers:
+            - name: cli
+              image: ${CLI_IMAGE}
+              command: ["/bin/bash", "-c"]
+              args:
+              - echo "Starting pre-create of indices for Cluster Logging Elasticsearch...";
```
portante: Seems like moving this script to a separate file would be easier to maintain, no?

docs/troubleshooting.md
```diff
@@ -98,6 +98,142 @@ for p in $(oc get pods -l component=es -o jsonpath={.items[*].metadata.name}); d
...
+                    oc exec -n ${LOGGING_NAMESPACE} -c elasticsearch $es_pod -- es_util --query=$idx -XPUT < /dev/null;
+                done < $TMPDIR/create.lis
+            restartPolicy: OnFailure
+parameters:
+- name: CLI_IMAGE
+  value: openshift/origin-cli:latest
+  description: "The image to use to execute the script"
+- name: CRON_SCHEDULE
+  value: "0 4,8,12,16,20 * * *"
+  description: "The schedule to to pre-create indices. Defaults to every 4 hours accept midnight"
```
portante: s/accept/except/

docs/troubleshooting.md
```diff
@@ -98,6 +98,142 @@ for p in $(oc get pods -l component=es -o jsonpath={.items[*].metadata.name}); d
   oc exec -c elasticsearch $ES_POD -- touch /opt/app-root/src/init_failures;  \
 done
 ```
+### Elasticsearch performance degrades around 00:00 UTC
+This situation is a result of Elasticsearch processing bulk index requests from all the nodes in the OKD cluster for
+logs for the new day.  Creation of a significant number of new indices simultaneously can consume significant resources when there are a large number of projects.  Periodically pre-creating the next day's indices can reduce the load on ElasticSearch when rolling over to the new day:
```
portante: s/ElasticSearch/Elasticsearch/

portante: s/new day:/new day./

portante: s/logs for the new day./logs for the new day where Elasticsearch auto-creates new indices for those logs./

portante: Consider re-writing:

    Creation of a significant number of new indices simultaneously
    can consume significant resources when there are a large
    number of projects.

To something like:

    Creation of a significant number of new indices simultaneously
    can cause bulk index processing to slow down as it waits for all
    members of the cluster to become aware of each new index.  This
    is likely to happen when there are many active projects (namespaces)
    in a cluster, since each projects has its own index.

docs/troubleshooting.md
jcantrill: It would, however, there is no easy way to provide that as a template which in this case is a 'one stop shop'. Ideally this is added into the product where it could then be moved into a script but its technically a feature enhancement so I'm certain it will live as a troubleshooting work around

jcantrill:
merging as this is a docs only change

PR #1649, opened 2019-05-16
LOG-400 - rsyslog - support log rotation for rsyslog log files

nhosoi:
- adding logrotate test cases

### DIFF
--- a/test/zzz-rsyslog.sh
+++ b/test/zzz-rsyslog.sh
@@ -36,6 +36,13 @@ fi
 # clear the journal
 clear_and_restart_journal
 
+restart_rsyslog_pod() {
+    local rpod=${1:-$( get_running_pod rsyslog )}
+    oc delete --force pod $rpod
+    os::cmd::try_until_failure "oc get pods $rpod > /dev/null 2>&1"
+    os::cmd::try_until_success "oc get pods 2> /dev/null | grep -q 'rsyslog.*Running'"
+}
+
 cleanup() {
     local return_code="$?"
     set +e
@@ -143,9 +150,7 @@ deploy_using_operators() {
     oc get cm rsyslog -o yaml | \
       sed -e 's/action(type="mmkubernetes"/action(type="mmkubernetes" annotation_match=["."]/' | \
       oc replace --force -f - 2>&1 | artifact_out
-    oc delete --force pod $rpod
-    os::cmd::try_until_failure "oc get pods $rpod > /dev/null 2>&1"
-    os::cmd::try_until_success "oc get pods 2> /dev/null | grep -q 'rsyslog.*Running'"
+    restart_rsyslog_pod $rpod
 }
 
 get_logmessage() {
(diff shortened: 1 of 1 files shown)


nhosoi:
@richm, in the previous run, all the CI tests passed except continuous-integration/travis-ci/pr.  (so, I tried rebasing with the master branch and see if it changes the behaviour, but it doesn't...)  According to the mail thread "Please Migrate Away From Travis", we are supposed to disable it?  If so, what is the best way to do that?  Removing .travis.yml?  Thanks!

richm:
we need to replace the travis CI with something based in api.ci.openshift.com, so let's do that first before we disable travis.  And let's not do that now, let's schedule the work (and hope that someone in some other group that is affected by the move away from travis CI comes up with an elegant solution that we can steal . . .)

If you need to retest the travis-ci/pr test, you can use `/test all` which should retest everything

nhosoi:
> we need to replace the travis CI with something based in api.ci.openshift.com, so let's do that first before we disable travis. And let's not do that now, let's schedule the work (and hope that someone in some other group that is affected by the move away from travis CI comes up with an elegant solution that we can steal . . .)
> 
> If you need to retest the travis-ci/pr test, you can use `/test all` which should retest everything

Ok.  Thanks!!

nhosoi:
Hi @richm, it happened again...  The rest of the CI tests passed, but it looks travis-ci does not return?  What should we do?

richm:
@nhosoi the fix to remove travis has merged - not sure how long it will take for that to go into production - looks like `/test all` didn't clear the travis state - try doing a rebase to see if that makes travis go away

test/zzz-rsyslog.sh
```diff
@@ -36,6 +36,13 @@ fi
 # clear the journal
 clear_and_restart_journal
 
+restart_rsyslog_pod() {
+    rpod=${1:-$( get_running_pod rsyslog )}
```
richm: `local rpod=...`

test/zzz-rsyslog.sh
```diff
@@ -235,3 +240,87 @@ rpod_ip="$( oc get pod ${rpod} -o jsonpath='{.status.podIP}' )"
...
+os::cmd::expect_success_and_text "oc logs $rpod -c rsyslog | grep 'oc exec <pod_name> -- logs'" "^oc exec .pod_name. -- logs$"
+os::cmd::expect_success "oc exec $rpod -c rsyslog -- ls /var/log/rsyslog/rsyslog.log > /dev/null 2>&1"
+logsize=$( oc exec $rpod -c rsyslog -- wc -c /var/log/rsyslog/rsyslog.log | awk '{print $1}' )
+
+# Check if logrotate works as expected
+if [ $logsize -gt 0 ]; then
+    # set max log file count and max log size
+    maxcount=3
+    maxsize=$( expr $logsize / $maxcount )
+    oc set env daemonset/rsyslog LOGGING_FILE_SIZE=$maxsize LOGGING_FILE_AGE=$maxcount
```
richm: this will restart rsyslog - you should try_until_failure for `oc get pod $rpod`, then try_until_success for `oc get pods | grep 'rsyslog.*Running'` or something like that

richm: never mind - I see you do this below

test/zzz-rsyslog.sh
nhosoi: Thank you, @richm!   Fixed.

PR #1656, opened 2019-06-06
rsyslog ci - fix rotation check flake

richm:
Give the rotation size check a little extra padding


nhosoi:
/lgtm
(Thanks  a lot, @richm !!)

PR #1657, opened 2019-06-06
Updating Kibana from 5 to 6, using vendored src

ewolinetz:
To resolve https://jira.coreos.com/browse/LOG-383

Currently we need the version of `node` to be bumped to `v10.15.2`

From Kibana when started otherwise: `Kibana does not support the current Node.js version v10.10.0. Please use Node.js v10.15.2`


kibana/Dockerfile
```diff
@@ -1,55 +1,50 @@
-FROM rhscl/nodejs-6-rhel7
+FROM rhscl/nodejs-10-rhel7 AS builder
```
richm: Use the ubi base image since you aren't using nodejs in the builder phase - I guess there isn't anything to build?

kibana/Dockerfile
```diff
@@ -1,55 +1,50 @@
...
+
+ENV KIBANA_VERSION=6.8.0 \
+    KIBANA_DOWNLOAD_URI=https://artifacts.elastic.co/downloads/kibana/
```
richm: internal builds won't be able to pull from `https://artifacts.elastic.co`

ewolinetz: what should we do here then, download and add it to the repo?

richm: For fluentd and rsyslog, we are downloading the source code, adding the expanded source code to git, and building the container from that expanded source code.  We are also planning to do the same with with Elasticsearch.  I suppose we could add the kibana tarball .tar.gz file to git, but that is generally frowned upon, and is probably against openshift practice - @bparees

jcantrill: Wouldn't we also expand the source in the same manor as the other components if we are going to go the vendoring approach?

bparees: adding binaries to our git repos is definitely against our practices.

ewolinetz:
```
Kibana does not support the current Node.js version v10.16.3. Please use Node.js v10.15.2.
running 'bin/kibana --optimize' failed with exit code 1
```
:roll_eyes:

ewolinetz:
not sure if the tests will pass since kibana requires a matching version of ES
@jcantrill @richm PTAL

kibana/Dockerfile
```diff
@@ -1,55 +1,44 @@
-FROM rhscl/nodejs-6-rhel7
+FROM rhoar-nodejs/nodejs-10
 
 MAINTAINER OpenShift Development <dev@lists.openshift.redhat.com>
 
 EXPOSE 5601
 
-ENV ELASTICSEARCH_URL=https://logging-es:9200 \
+ENV ELASTICSEARCH_URL=https://elasticsearch:9200 \
```
jcantrill: Do we need the full URL so global proxy works correctly?

kibana/kibana.yml
```diff
@@ -6,19 +6,26 @@
...
 # The maximum payload size in bytes for incoming server requests.
 #server.maxPayloadBytes: 1048576
 
 # The Kibana server's name.  This is used for display purposes.
 #server.name: "your-hostname"
 
-# The URL of the Elasticsearch instance to use for all your queries.
-#elasticsearch.url: "https://localhost:9200"
+# The URLs of the Elasticsearch instances to use for all your queries.
+elasticsearch.hosts: ["https://elasticsearch:9200"]
```
jcantrill: Same here?  Do we need the full url to support proxy?

kibana/Dockerfile
ewolinetz: i dont think we need to worry about that, but it couldn't hurt.. will update

ewolinetz:
the image fails to build currently because CI is replacing the Dockerfile `FROM` so `node` is no longer available. I'm currently investigating why this is done and how I am expected to work around it.

jcantrill:
> the image fails to build currently because CI is replacing the Dockerfile `FROM` so `node` is no longer available. I'm currently investigating why this is done and how I am expected to work around it.

This is done by the release process to ensure the base image is a rhel based

ewolinetz:
> This is done by the release process to ensure the base image is a rhel based

So then what can I do to get around this? it should be a RHEL based image since its coming from `rhscl`

jcantrill:
> > This is done by the release process to ensure the base image is a rhel based
> 
> So then what can I do to get around this? it should be a RHEL based image since its coming from `rhscl`

I believe we simply need to contact ART and update the config file they use to replace the FROM

ewolinetz:
> I believe we simply need to contact ART and update the config file they use to replace the FROM

@jcantrill after talking with Tim, I think we need to update something here https://github.com/openshift/release/blob/master/ci-operator/config/openshift/origin-aggregated-logging/openshift-origin-aggregated-logging-release-4.4.yaml

kibana/Dockerfile
```diff
@@ -1,55 +1,44 @@
-FROM rhscl/nodejs-6-rhel7
```
ewolinetz: looks like `rhscl/nodejs-10-rhel7` is now available on registry.redhat.com... i'll try to switch to using that

jcantrill:
> > I believe we simply need to contact ART and update the config file they use to replace the FROM
> 
> @jcantrill after talking with Tim, I think we need to update something here https://github.com/openshift/release/blob/master/ci-operator/config/openshift/origin-aggregated-logging/openshift-origin-aggregated-logging-release-4.4.yaml

I believe we need to dup https://github.com/openshift/release/blob/master/ci-operator/config/openshift/origin-aggregated-logging/openshift-origin-aggregated-logging-master.yaml  and make the file suffix match the branch we desire

ewolinetz:
will require https://github.com/openshift/release/pull/6267 before tests will start passing

jcantrill:
Merging to unblock es6 work

PR #1658, opened 2019-06-07
allow get-clusters-run-tests.sh to deploy official 4.x

richm:


hack/deploy-logging.sh
```diff
@@ -122,15 +125,17 @@ EXTERNAL_REGISTRY=${EXTERNAL_REGISTRY:-registry.svc.ci.openshift.org}
...
-elif [ "${USE_IMAGE_STREAM:-false}" = true ] ; then
-    USE_CUSTOM_IMAGES=${USE_CUSTOM_IMAGES:-true}
-elif [ "${USE_CLO_LATEST_IMAGE:-false}" = true -o "${USE_EO_LATEST_IMAGE:-false}" = true ] ; then
-    USE_CUSTOM_IMAGES=${USE_CUSTOM_IMAGES:-true}
-else
-    # default to false
-    USE_CUSTOM_IMAGES=${USE_CUSTOM_IMAGES:-false}
+if [ -z "${USE_CUSTOM_IMAGES:-}" ] ; then
+    if [ -n "${OPENSHIFT_BUILD_NAMESPACE:-}" -a -n "${IMAGE_FORMAT:-}" ] ; then
+        USE_CUSTOM_IMAGES=${USE_CUSTOM_IMAGES:-true}
```
nhosoi: It does not matter, but we can have just `USE_CUSTOM_IMAGES=true` since USE_CUSTOM_IMAGES is guaranteed empty?

richm: not sure what you mean

richm:
needs https://github.com/openshift/origin-aggregated-logging/pull/1661

hack/deploy-logging.sh
ewolinetz: this is basically the same logic as we had before, but now we short circuit it if `USE_CUSTOM_IMAGES` is already set.

hack/deploy-logging.sh
```diff
@@ -140,20 +145,20 @@ if [ ! -d $ARTIFACT_DIR ] ; then
...
 if oc get project $LOGGING_NS > /dev/null 2>&1 ; then
     echo using existing project $LOGGING_NS
 else
     oc create -f $TEST_OBJ_DIR/openshift-logging-namespace.yaml
 fi
 
+if [ "${USE_OLM:-false}" = true ] ; then
+    deploy_logging_using_olm
+else
+    ESO_NS=openshift-logging
```
ewolinetz: so that this matches what we'd see in a typical cluster deployment can we make this `openshift-operators-redhat` instead?

hack/deploy-logging.sh
richm: still not sure what is the suggested change to this code

hack/deploy-logging.sh
richm: then it won't match the cluster-logging-operator `make deploy`

hack/deploy-logging.sh
nhosoi: Sorry for my slow response...  I just thought since USE_CUSTOM_IMAGES is always empty/unset when it comes to the assignment line, we could just set to true (without '${USE_CUSTOM_IMAGES:-' and '}'), which looks simpler.  But as it does not change the behaviour, it is not an issue at all.
```
if [ -z "${USE_CUSTOM_IMAGES:-}" ] ; then
    if [ -n "${OPENSHIFT_BUILD_NAMESPACE:-}" -a -n "${IMAGE_FORMAT:-}" ] ; then
        USE_CUSTOM_IMAGES=true
        ^^^^^^^^^^^^^^^^^^^^^^
```

richm:
flake unrelated to journald/json-parsing

/test full-integ-aws

PR #1660, opened 2019-06-08
[release-3.11] Bug 1717397 - [RFE] Add feature in fluentd to send logs over UDP to a…

nhosoi:
…n external syslog server.

Introducing REMOTE_SYSLOG_TYPE environment variable to switch between syslog_buffered (TCP)
and syslog (UDP).


nhosoi:
Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1717397

ewolinetz:
I don't think this should be labeled as a bug. Looking at that bz it was opened as a feature request.

PR #1661, opened 2019-06-10
not sure why kibana is not being deployed now

richm:
There is not enough resources available to restart Kibana, and we do not need to restart it for CI, so make sure not to restart it.
Add more debugging.


lukas-vlcek:
@richm Do I get it correctly that in the end you DO NOT apply the linux label on all nodes but reduce resources requested by ES nodes instead? If that is the case I suggest updating the title of this issue as well. Otherwise
/lgtm

richm:
I think this PR is no longer needed - the problem was the underpowered ci machines - closing

PR #1671, opened 2019-06-20
LOG-372 - rsyslog - support improved undefined field handling

nhosoi:
Adding mmexternal/undefined_field/undefined_field.go

### DIFF
--- a/rsyslog/Dockerfile
+++ b/rsyslog/Dockerfile
@@ -22,12 +22,15 @@ RUN cd /vendored_src && \
 
 ADD go/ /go/
 RUN cd /go/src/github.com/soundcloud/rsyslog_exporter ; GOPATH=/go go build
+ADD undefined_field/ /undefined_field/
+RUN cd /undefined_field/ ; go build
 
 FROM registry.svc.ci.openshift.org/ocp/4.2:base
 
 COPY --from=builder /RPMS/ /RPMS/
 COPY --from=builder /contents /contents
 COPY --from=builder /go/src/github.com/soundcloud/rsyslog_exporter/rsyslog_exporter /usr/local/bin/
+COPY --from=builder /undefined_field/ /usr/local/bin/
 
 RUN yum -y install /RPMS/*.rpm cronie && \
     rm -rf /RPMS && \
--- a/rsyslog/undefined_field/undefined_field.go
+++ b/rsyslog/undefined_field/undefined_field.go
@@ -54,6 +54,9 @@
  *    - Debug flag used in undefined_field as well as the config file calling the plugin.
  *      default to "false".
  *
+ * Note: the default values are defined in cluster-logging-operator/files/rsyslog/rsyslog.sh.
+ *       The startup script generates the config file undefined.json explaned next.
+ *
  * undefined_field expects a config file undefined.json in /var/lib/rsyslog.pod.
  * example undefined.json with default settings
  * {
@@ -75,27 +78,30 @@ package main
 import (
 	"encoding/json"
 	"bufio"
-	"strconv"
 	"strings"
 	"fmt"
+	"io/ioutil"
 	"os"
 )
 
 const (
-	initial_undefined_debug            = false
-	initial_merge_json_log             = false
-	initial_use_undefined              = false
-	initial_undefined_name             = "undefined"
-	initial_undefined_to_string        = false
-	initial_default_keep_fields        = "CEE,time,@timestamp,aushape,ci_job,collectd,docker,fedora-ci,file,foreman,geoip,hostname,ipaddr4,ipaddr6,kubernetes,level,message,namespace_name,namespace_uuid,offset,openstack,ovirt,pid,pipeline_metadata,rsyslog,service,systemd,tags,testcase,tlog,viaq_msg_id"
-	initial_extra_keep_fields          = ""
-	initial_keep_empty_fields          = ""
-	initial_undefined_max_num_fields   = -1
-	initial_undefined_dot_replace_char = "UNUSED"
 	initial_logging_file_path          = "/var/log/rsyslog/rsyslog.log"
 	undefined_config                   = "/var/lib/rsyslog.pod/undefined.json"
 	noChanges                          = "{}"
 )
+
+type UndefinedConfig struct {
+  Debug bool `json:"UNDEFINED_DEBUG"`
+  Merge_json_log bool `json:"MERGE_JSON_LOG"`
+  Use_undefined bool `json:"CDM_USE_UNDEFINED"`
+  Undefined_to_string bool `json:"CDM_UNDEFINED_TO_STRING"`
+  Default_keep_fields string `json:"CDM_DEFAULT_KEEP_FIELDS"`
+  Extra_keep_fields string `json:"CDM_EXTRA_KEEP_FIELDS"`
+  Undefined_name string `json:"CDM_UNDEFINED_NAME"`
+  Keep_empty_fields string `json:"CDM_KEEP_EMPTY_FIELDS"`
+  Undefined_dot_replace_char string `json:"CDM_UNDEFINED_DOT_REPLACE_CHAR"`
+  Undefined_max_num_fields int64 `json:"CDM_UNDEFINED_MAX_NUM_FIELDS"`
+}
 
 var (
 	undefined_debug bool
@@ -107,6 +113,7 @@ var (
 	undefined_to_string bool
 	undefined_dot_replace_char string
 	undefined_max_num_fields int64
+	undefined_cur_num_fields int64
 	logfile *os.File
 	noaction		  = false
 	replacer		  = &strings.Replacer{}
@@ -132,92 +139,26 @@ func onInit() {
 		panic(fmt.Errorf("Could not open file [%s]: [%v]", logging_file_path, err))
 	}
 
-	// initializing
-	undefined_debug = initial_undefined_debug
-	merge_json_log = initial_merge_json_log
-	use_undefined = initial_use_undefined
-	default_keep_fields := initial_default_keep_fields
-	extra_keep_fields := initial_extra_keep_fields
-	undefined_name = initial_undefined_name
-	tmp_keep_empty_fields := initial_keep_empty_fields
-	undefined_to_string = initial_undefined_to_string
-	undefined_dot_replace_char = initial_undefined_dot_replace_char
-	undefined_max_num_fields = initial_undefined_max_num_fields
-
+	var undefined_config_obj UndefinedConfig
+	var default_keep_fields string
+	var extra_keep_fields string
+	var tmp_keep_empty_fields string
 	if config, err := os.Open(undefined_config); err == nil {
-		reader := bufio.NewReader(config)
-		scanner := bufio.NewScanner(reader)
-		scanner.Split(bufio.ScanLines)
-		for scanner.Scan() {
-			rawStr := scanner.Text()
-			// skip comments
-			if strings.HasPrefix(rawStr, "#") {
-				continue
-			}
-			configMap := make(map[string]interface{})
-			if err := json.Unmarshal([]byte(rawStr), &configMap); err != nil {
-				fmt.Fprintln(logfile, "ERROR: Failed to parse config file [%s]: [%v]", undefined_config, err)
-				panic(fmt.Errorf("Failed to parse config file [%s]: [%v]", undefined_config, err))
-			}
-
-			if val, ok := getMapStringValue(configMap, "UNDEFINED_DEBUG"); ok {
-				if rval, err := strconv.ParseBool(val); err == nil {
-					undefined_debug = rval
-				} else {
-					fmt.Fprintln(logfile, "ERROR: Invalid UNDEFINED_DEBUG value [%s]: [%v]", val, err)
-					panic(fmt.Errorf("Invalid UNDEFINED_DEBUG value [%s]: [%v]", val, err))
-				}
-			}
-			if val, ok := getMapStringValue(configMap, "MERGE_JSON_LOG"); ok {
-				if rval, err := strconv.ParseBool(val); err == nil {
-					merge_json_log = rval
-				} else {
-					fmt.Fprintln(logfile, "ERROR: Invalid MERGE_JSON_LOG value [%s]: [%v]", val, err)
-					panic(fmt.Errorf("Invalid MERGE_JSON_LOG value [%s]: [%v]", val, err))
-				}
-			}
-			if val, ok := getMapStringValue(configMap, "CDM_USE_UNDEFINED"); ok {
-				if rval, err := strconv.ParseBool(val); err == nil {
-					use_undefined = rval
-				} else {
-					fmt.Fprintln(logfile, "ERROR: Invalid CDM_USE_UNDEFINED value [%s]: [%v]", val, err)
-					panic(fmt.Errorf("Invalid CDM_USE_UNDEFINED value [%s]: [%v]", val, err))
-				}
-			}
-			if val, ok := getMapStringValue(configMap, "CDM_DEFAULT_KEEP_FIELDS"); ok {
-				default_keep_fields = val
-			}
-			if val, ok := getMapStringValue(configMap, "CDM_EXTRA_KEEP_FIELDS"); ok {
-				extra_keep_fields = val
-			}
-			if val, ok := getMapStringValue(configMap, "CDM_UNDEFINED_NAME"); ok {
-				undefined_name = val
-			}
-			if val, ok := getMapStringValue(configMap, "CDM_KEEP_EMPTY_FIELDS"); ok {
-				tmp_keep_empty_fields = val
-			}
-			if val, ok := getMapStringValue(configMap, "CDM_UNDEFINED_TO_STRING"); ok {
-				if rval, err := strconv.ParseBool(val); err == nil {
-					undefined_to_string = rval
-				} else {
-					fmt.Fprintln(logfile, "ERROR: Invalid CDM_UNDEFINED_TO_STRING value [%s]: [%v]", val, err)
-					panic(fmt.Errorf("Invalid CDM_UNDEFINED_TO_STRING value [%s]: [%v]", val, err))
-				}
-			}
-			if val, ok := getMapStringValue(configMap, "CDM_UNDEFINED_DOT_REPLACE_CHAR"); ok {
-				undefined_dot_replace_char = val
-			}
-			if val, ok := getMapStringValue(configMap, "CDM_UNDEFINED_MAX_NUM_FIELDS"); ok {
-				var rval64 int64
-				if rval64, err = strconv.ParseInt(val, 10, 64); err == nil {
-					undefined_max_num_fields = rval64
-				} else {
-					fmt.Fprintln(logfile, "ERROR: Invalid CDM_UNDEFINED_MAX_NUM_FIELDS value [%s]: [%v]", val, err)
-					panic(fmt.Errorf("Invalid CDM_UNDEFINED_MAX_NUM_FIELDS value [%s]: [%v]", val, err))
-				}
-			}
-		}
 		defer config.Close()
+
+		config_read, _ := ioutil.ReadAll(config)
+		json.Unmarshal(config_read, &undefined_config_obj)
+
+		undefined_debug = undefined_config_obj.Debug
+		merge_json_log = undefined_config_obj.Merge_json_log
+		use_undefined = undefined_config_obj.Use_undefined
+		default_keep_fields = undefined_config_obj.Default_keep_fields
+		extra_keep_fields = undefined_config_obj.Extra_keep_fields
+		undefined_name = undefined_config_obj.Undefined_name
+		tmp_keep_empty_fields = undefined_config_obj.Keep_empty_fields
+		undefined_to_string = undefined_config_obj.Undefined_to_string
+		undefined_dot_replace_char = undefined_config_obj.Undefined_dot_replace_char
+		undefined_max_num_fields = undefined_config_obj.Undefined_max_num_fields
 	} else if !strings.Contains(err.Error(), "no such file or directory") {
 		fmt.Fprintln(logfile, "ERROR: Could not open config file [%s]: [%v]", undefined_config, err)
 		panic(fmt.Errorf("Could not open config file [%s]: [%v]", undefined_config, err))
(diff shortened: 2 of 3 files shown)


rsyslog/go/src/github.com/mmexternal/undefined_field/undefined_field.go
```diff
@@ -0,0 +1,239 @@
+package main
```
nhosoi: @richm , is it ok to place the plugin in this path `rsyslog/go/src/github.com/mmexternal/undefined_field`?   Or maybe I'd better map to the rsyslog source tree structure, something like `rsyslog/go/src/github.com/rsyslog/plugins/external/undefined_field/undefined_field.go`?

And what "package" name I'm supposed to use?  Or it does not matter since it's just an independent plugin?

nhosoi:
@richm , I haven't implemented MERGE_JSON_LOG yet.  To do so, I'm supposed to treat the value of message specially, i.e., not to parse it?  I'd think json.Unmarshal parses the value of message into json if the value is a json string.

This patch has `undefined_to_string` option, which converts the undefined value into the json string.  Do you think if I add `message_to_string`, it covers the requirement for MERGE_JSON_LOG=false?

rsyslog/go/src/github.com/mmexternal/undefined_field/undefined_field.go
richm: > @richm , is it ok to place the plugin in this path `rsyslog/go/src/github.com/mmexternal/undefined_field`? Or maybe I'd better map to the rsyslog source tree structure, something like `rsyslog/go/src/github.com/rsyslog/plugins/external/undefined_field/undefined_field.go`?

Well, the path is supposed to represent the source of the code - so the path `go/src/github.com/mmexternal/undefined_field` would represent `https://github.com/mmexternal/undefined_field` which does not exist.  Which means it would logically go under `go/src/github.com/openshift/origin-aggregated-logging/rsyslog`, but that is redundant.

Therefore, I think the code should live in the `rsyslog` directory.

> 
> And what "package" name I'm supposed to use? Or it does not matter since it's just an independent plugin?

https://golang.org/doc/code.html#PackageNames
"Executable commands must always use package `main`."

This is a standalone executable command, so, `main` is good.

richm:
> @richm , I haven't implemented MERGE_JSON_LOG yet.

It is already implemented in the rscript config: https://github.com/openshift/cluster-logging-operator/blob/master/files/rsyslog/65-viaq-formatting.conf#L12 and https://github.com/openshift/cluster-logging-operator/blob/master/files/rsyslog/65-viaq-formatting.conf#L34

> To do so, I'm supposed to treat the value of message specially, i.e., not to parse it? I'd think json.Unmarshal parses the value of message into json if the value is a json string.
> 
> This patch has `undefined_to_string` option, which converts the undefined value into the json string. Do you think if I add `message_to_string`, it covers the requirement for MERGE_JSON_LOG=false?

I think the calls to `mmexternal` should go in the same if clauses where `if (strlen(`echo $MERGE_JSON_LOG`) > 0) and (`echo $MERGE_JSON_LOG` == "true") then {`

rsyslog/Dockerfile
```diff
@@ -22,12 +22,14 @@ RUN cd /vendored_src && \
 
 ADD go/ /go/
 RUN cd /go/src/github.com/soundcloud/rsyslog_exporter ; GOPATH=/go go build
+RUN cd /go/src/github.com/mmexternal/undefined_field ; GOPATH=/go go build
```
richm: hmm - then put the code in the `rsyslog/undefined_field` sub-directory
```
ADD undefined_field/ /undefined_field/
RUN cd /undefined_field/ ; go build
```
not sure what to use for `GOPATH` here - not sure it matters.

rsyslog/go/src/github.com/mmexternal/undefined_field/undefined_field.go
```diff
@@ -0,0 +1,239 @@
...
+)
+
+func onInit() {
+	// opening the rsyslog log file
+	logging_file_path := initial_logging_file_path
+	if eval := os.Getenv("LOGGING_FILE_PATH"); eval != "" {
+		logging_file_path = eval
+	}
+	if f, err := os.OpenFile(logging_file_path, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644); err == nil {
+		logfile = f
```
richm: Is this how `mmexternal` and `omprog` modules are supposed to log?

rsyslog/go/src/github.com/mmexternal/undefined_field/undefined_field.go
nhosoi: Thank you, @richm!  Let me adjust the path to rsyslog including the Dockerfile below.

rsyslog/go/src/github.com/mmexternal/undefined_field/undefined_field.go
```diff
@@ -0,0 +1,239 @@
...
+	keep_fields []string
+	keep_empty_fields []string
+	undefined_name string
+	undefined_to_string bool
+	undefined_dot_replace_char string
+	undefined_max_num_fields int64
+	logfile *os.File
+	noaction          = false
+	replacer          = &strings.Replacer{}
+	basic_keep_fields = "_STREAM_ID,_SYSTEMD_INVOCATION_ID"
```
richm: we really should add these to https://github.com/ViaQ/elasticsearch-templates/blob/master/namespaces/systemd.yml#L43 , then add them to https://github.com/openshift/cluster-logging-operator/blob/master/files/rsyslog/65-viaq-formatting.conf#L91 . . . but ok for now

rsyslog/go/src/github.com/mmexternal/undefined_field/undefined_field.go
```diff
@@ -0,0 +1,239 @@
...
+func onInit() {
+	// opening the rsyslog log file
+	logging_file_path := initial_logging_file_path
+	if eval := os.Getenv("LOGGING_FILE_PATH"); eval != "" {
+		logging_file_path = eval
+	}
+	if f, err := os.OpenFile(logging_file_path, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644); err == nil {
+		logfile = f
+	} else {
+		panic(err.Error())
```
richm: `panic(fmt.Errorf("Could not open file [%s]: [%v]", logging_file_path, err))`

rsyslog/go/src/github.com/mmexternal/undefined_field/undefined_field.go
```diff
@@ -0,0 +1,239 @@
...
+	extra_keep_fields := initial_extra_keep_fields
+	undefined_name = initial_undefined_name
+	tmp_keep_empty_fields := initial_keep_empty_fields
+	undefined_to_string = initial_undefined_to_string
+	undefined_dot_replace_char = initial_undefined_dot_replace_char
+	undefined_max_num_fields = initial_undefined_max_num_fields
+
+	config, err := os.Open("/var/lib/rsyslog.pod/undefined.yaml")
+	if err != nil {
+	  panic(err.Error())
```
richm: see above

rsyslog/go/src/github.com/mmexternal/undefined_field/undefined_field.go
```diff
@@ -0,0 +1,239 @@
...
+	use_undefined = initial_use_undefined
+	default_keep_fields := initial_default_keep_fields
+	extra_keep_fields := initial_extra_keep_fields
+	undefined_name = initial_undefined_name
+	tmp_keep_empty_fields := initial_keep_empty_fields
+	undefined_to_string = initial_undefined_to_string
+	undefined_dot_replace_char = initial_undefined_dot_replace_char
+	undefined_max_num_fields = initial_undefined_max_num_fields
+
+	config, err := os.Open("/var/lib/rsyslog.pod/undefined.yaml")
```
richm: the go way to do this is to combine the assignment and the err check in a single statement
```go
if config, err := os.Open("/var/lib/rsyslog.pod/undefined.yaml"); err != nil {
    panic(fmt.Errorf("Could not open file [%s]: [%v]", "/var/lib/rsyslog.pod/undefined.yaml", err))
}
```

rsyslog/go/src/github.com/mmexternal/undefined_field/undefined_field.go
```diff
@@ -0,0 +1,239 @@
...
+	if err != nil {
+	  panic(err.Error())
+	}
+	defer config.Close()
+
+	reader := bufio.NewReader(config)
+	scanner := bufio.NewScanner(reader)
+	scanner.Split(bufio.ScanLines)
+	for scanner.Scan() {
```
richm: I would encourage you to write out the config file in `yaml` format rather than `key: value` format.  Then you can use the built-in yaml parser rather than having to write your own parser.

rsyslog/go/src/github.com/mmexternal/undefined_field/undefined_field.go
```diff
@@ -0,0 +1,239 @@
...
+				undefined_max_num_fields = rval64
+			}
+		}
+	}
+
+	if !use_undefined && tmp_keep_empty_fields == "" && !undefined_to_string && undefined_dot_replace_char == "UNUSED" {
+		noaction = true
+	}
+	if undefined_max_num_fields == -1 {
+		undefined_max_num_fields = 9223372036854775807
```
richm: use something like https://stackoverflow.com/a/6878625
```go
undefined_max_num_fields = int(^uint(0) >> 1)
```

rsyslog/go/src/github.com/mmexternal/undefined_field/undefined_field.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	if use_undefined {
+		tmp_basic := strings.Split(basic_keep_fields, ",")
+		tmp_default := strings.Split(default_keep_fields, ",")
+		tmp_extra := strings.Split(extra_keep_fields, ",")
+		keep_fields = append(tmp_basic, tmp_default...)
+		keep_fields = append(keep_fields, tmp_extra...)
```
richm: in general it is better to make `keep_fields` a `map[string]string` so you can use `O(1)` hash lookups and avoid having to use the `O(n)` function `contains` - same with `keep_empty_fields`.

rsyslog/go/src/github.com/mmexternal/undefined_field/undefined_field.go
```diff
@@ -0,0 +1,239 @@
...
+	if *copyTop == nil {
+		*copyTop = cp
+	}
+	for key, value := range jsonMap {
+		if undefined_dot_replace_char != "UNUSED" {
+			// replace '.' with specified char (e.g., '_')
+			key = replacer.Replace(key)
+		}
+		// skip empty or not?
+		valuemap, ok := value.(map[string]interface{})
```
richm: is it possible that `value` here is not a `map[string]interface{}`?

rsyslog/go/src/github.com/mmexternal/undefined_field/undefined_field.go
richm: I see - there is no built-in yaml parser - a quick check of our operator code shows that `"github.com/ghodss/yaml"` is used, but that would have to be vendored in - so you could either 

- keep the format the way you have it
- write out the file in json format and use the built-in json parser
- use yaml and vendor in `"github.com/ghodss/yaml"`

rsyslog/go/src/github.com/mmexternal/undefined_field/undefined_field.go
```diff
@@ -0,0 +1,239 @@
...
+		if noaction {
+			fmt.Println(rawStr)
+			continue
+		}
+		jsonStr, err := strconv.Unquote(string(rawStr))
+		if err != nil {
+			jsonStr = rawStr
+		}
+		if err := json.Unmarshal([]byte(jsonStr), &jsonMap); err != nil {
+			fmt.Println(logfile, "json.Unmarshal failed (", err, "): ", rawStr)
```
richm: `fmt.Fprintln`

rsyslog/go/src/github.com/mmexternal/undefined_field/undefined_field.go
```diff
@@ -0,0 +1,239 @@
...
+		}
+		// use_undefined and key is not in keep_fields and less than unmdefined_max_num_fields?
+		if domove && use_undefined && !contains(keep_fields, key) && undefined_max_num_fields > 0 {
+			tmpcp := *copyTop
+			if tmpcp[undefined_name] == nil {
+				subcp := make(map[string]interface{})
+				tmpcp[undefined_name] = subcp
+			}
+			if ok {
+				rval := replaceDotAndSkipEmpty(copyTop, valuemap, false)
```
richm: why do you need to recurse here?  we are only concerned with the top level fields - are you trying combine both the undefined field handling, and checking to see if the nested structure is empty?  If so, I think it would be better to have a separate `isempty` or `delempty` function like https://github.com/ViaQ/fluent-plugin-viaq_data_model/blob/master/lib/fluent/plugin/filter_viaq_data_model.rb#L232

richm: > I think you need to change the undefined_dot_replace_char check to only check the top level fields

you also want to make sure you use the original `key` for testing `keep_empty_fields` and `keep_fields` - the new `key` should only be used for `cp[newkey] = rval`, and only for top level undefined fields - that is, if `key` is a defined field (in `keep_fields`), or `key` is a recursively nested field, we should not do the replacement on it.  Maybe the code already does that, but I can't tell.

rsyslog/go/src/github.com/mmexternal/undefined_field/undefined_field.go
nhosoi: Is it ok to allow the plugin to work without the undefined.yaml file with this warning in the log file?
`fmt.Fprintln(logfile, "WARNING: Could not open file [%s]: [%v]; Using the default configurations.", undefined_config, err)`
It might be a bit selfish, but the yaml file is generated by CLO pr https://github.com/openshift/cluster-logging-operator/pull/202, which depends upon this pr.  I'd like to make this pr having no dependency...

rsyslog/go/src/github.com/mmexternal/undefined_field/undefined_field.go
nhosoi: Ok.  I'd go for "write out the file in json format and use the built-in json parser", which looks most light weight...

rsyslog/go/src/github.com/mmexternal/undefined_field/undefined_field.go
nhosoi: :p Thanks!

rsyslog/go/src/github.com/mmexternal/undefined_field/undefined_field.go
richm: so if the file is opened successfully, ok
if the file cannot be opened, because it does not exist, ok - no log is necessary
if the file cannot be opened, for some reason other than "does not exist", then panic with an error

nhosoi:
> It is already implemented in the rscript config: https://github.com/openshift/cluster-logging-operator/blob/master/files/rsyslog/65-viaq-formatting.conf#L12 and https://github.com/openshift/cluster-logging-operator/blob/master/files/rsyslog/65-viaq-formatting.conf#L34
> I think the calls to `mmexternal` should go in the same if clauses where `` if (strlen(``echo $MERGE_JSON_LOG``) > 0) and (``echo $MERGE_JSON_LOG`` == "true") then { ``

@richm , I'm thinking to call mmexternal(undefined_field) after 65-viaq-formatting.conf.  (https://github.com/openshift/cluster-logging-operator/pull/202/files#diff-f352aec93a80474539bd17b44867dac0).  That means, it's already processed correctly and undefined_field plugin can just accept the result?

richm:
> > It is already implemented in the rscript config: https://github.com/openshift/cluster-logging-operator/blob/master/files/rsyslog/65-viaq-formatting.conf#L12 and https://github.com/openshift/cluster-logging-operator/blob/master/files/rsyslog/65-viaq-formatting.conf#L34
> > I think the calls to `mmexternal` should go in the same if clauses where ` if (strlen(`echo $MERGE_JSON_LOG`) > 0) and (`echo $MERGE_JSON_LOG`== "true") then {`
> 
> @richm , I'm thinking to call mmexternal(undefined_field) after 65-viaq-formatting.conf. (https://github.com/openshift/cluster-logging-operator/pull/202/files#diff-f352aec93a80474539bd17b44867dac0). That means, it's already processed correctly and undefined_field plugin can just accept the result?

Ok. I was thinking that you should only do the mmexternal if `MERGE_JSON_LOG=true`, but looking at our current code we do the undefined field processing regardless https://github.com/ViaQ/fluent-plugin-viaq_data_model/blob/master/lib/fluent/plugin/filter_viaq_data_model.rb#L508 - I suppose if `MERGE_JSON_LOG=false` then there should be no undefined fields . . . still, it may be an optimization for rsyslog if we do not use mmexternal if `MERGE_JSON_LOG` is not defined or `false`

rsyslog/undefined_field/undefined_field.go
```diff
@@ -0,0 +1,392 @@
...
+ *
+ * How to use undefined_field
+ * 1) compile - go build undefined_field.go
+ * 2) rsyslog config example to call undefined_field
+ *    module(load="mmexternal")
+ *    action(type="mmexternal" binary="/usr/local/bin/undefined_field" interface.input="fulljson")
+ *    if (strlen($!openshift_logging_all) > 0) then {
+ *        set $.openshift_logging_all = $!openshift_logging_all;
+ *          unset $!;
+ *          set $! = $.openshift_logging_all;
```
richm: so `unset $!;` then `set $!` works?  I thought it wasn't working?  Or was it `reset $! = $.openshift_logging_all;` that wasn't working?

rsyslog/undefined_field/undefined_field.go
```diff
@@ -0,0 +1,392 @@
...
+	use_undefined = initial_use_undefined
+	default_keep_fields := initial_default_keep_fields
+	extra_keep_fields := initial_extra_keep_fields
+	undefined_name = initial_undefined_name
+	tmp_keep_empty_fields := initial_keep_empty_fields
+	undefined_to_string = initial_undefined_to_string
+	undefined_dot_replace_char = initial_undefined_dot_replace_char
+	undefined_max_num_fields = initial_undefined_max_num_fields
+
+	if config, err := os.Open(undefined_config); err == nil {
```
richm: I think this is ok for now, but fyi the simpler way would have been to define a config struct and use the built-in json to struct parsing e.g.
```go
type UndefinedConfig struct {
  bool Debug `json:"UNDEFINED_DEBUG"`
  bool Merge_json_log `json:"MERGE_JSON_LOG"`
  bool Use_undefined `json:"CDM_USE_UNDEFINED"`
  string Default_keep_fields `json:"CDM_DEFAULT_KEEP_FIELDS"`
  ...
  int Undefined_max_num_fields `json:"CDM_UNDEFINED_MAX_NUM_FIELDS"`
}
...
var UndefinedConfig undefined_config_obj
// not sure how to convert the file object config to an array of bytes
json.Unmarshal([]byte(config.read()), &undefined_config_obj)
// or maybe dec = json.NewDecoder(config)
// dec.Decode(&undefined_config_obj)
fmt.Println("undefined_debug %v", undefined_config_obj.Debug)
```
see https://blog.golang.org/json-and-go and https://www.sohamkamani.com/blog/2017/10/18/parsing-json-in-golang/

rsyslog/undefined_field/undefined_field.go
```diff
@@ -0,0 +1,392 @@
...
+				var rval64 int64
+				if rval64, err = strconv.ParseInt(val, 10, 64); err == nil {
+					undefined_max_num_fields = rval64
+				} else {
+					fmt.Fprintln(logfile, "ERROR: Invalid CDM_UNDEFINED_MAX_NUM_FIELDS value [%s]: [%v]", val, err)
+					panic(fmt.Errorf("Invalid CDM_UNDEFINED_MAX_NUM_FIELDS value [%s]: [%v]", val, err))
+				}
+			}
+		}
+		defer config.Close()
```
richm: this should be the first statement after the `open`

rsyslog/undefined_field/undefined_field.go
```diff
@@ -0,0 +1,392 @@
...
+		if topPropLevel && merge_json_log && undefined_dot_replace_char != "UNUSED" {
+			// replace '.' with specified char (e.g., '_')
+			key = replacer.Replace(origkey)
+			if key != origkey {
+				replace_me = true
+			}
+		}
+		// skip empty or not?
+		valuemap, ismap := value.(map[string]interface{})
+		valuearraymap, isarraymap := value.([]map[string]interface{})
```
richm: I think this should be `value.([]interface{})`

rsyslog/undefined_field/undefined_field.go
nhosoi: > so `unset $!;` then `set $!` works? I thought it wasn't working? 

It works...  I thought the same - "not working", but maybe there was some other bugs that confused me...

> Or was it `reset $! = $.openshift_logging_all;` that wasn't working?

I believe so.  but let me double check it.

rsyslog/undefined_field/undefined_field.go
nhosoi: Ok.  Let me apply your patch and rerun the test...  Thanks!

rsyslog/undefined_field/undefined_field.go
nhosoi: :) all right.

rsyslog/undefined_field/undefined_field.go
nhosoi: Ok...

rsyslog/undefined_field/undefined_field.go
nhosoi: > >     Or was it reset $! = $.openshift_logging_all; that wasn't working?
> I believe so. but let me double check it.

I double checked `reset $! = $.openshift_logging_all; ` adds $.openshift_logging_all on top of $! which I don't want to do.

rsyslog/undefined_field/undefined_field.go
```diff
@@ -317,6 +266,24 @@ func replaceDotMoveUndefined(jsonMap map[string]interface{}, topPropLevel bool)
 	return cp, replace_me, has_undefined
 }
 
+func replaceDotMoveUndefinedArray(inputs []interface{}, topPropLevel bool) []interface{} {
```
richm: `topPropLevel bool` is unused?

nhosoi: Indeed...  I thought "go build" tells me all the unused vars...  Obviously, args are out of the service. :p  Removing it.  Thanks!

PR #1673, opened 2019-06-21
Bug 1722898: Logging data from all projects are stored to .orphaned indexes with Elasticsearch

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1722898
Cause: Fluentd is unable to correctly determine the docker log
driver.  It thinks the log driver is journald when it is json-file.
Fluentd then looks for the `CONTAINER_NAME` field in the record to
hold the kubernetes metadata and it is not present.

Consequence: Fluentd is not able to add kubernetes metadata to
records.  Records go to the .orphaned index.  Fluentd spews lots
of errors like this:
`[error]: record cannot use elasticsearch index name type project_full: record is missing kubernetes field`

Fix: Fluentd should not rely on reading the docker configuration file
to determine if the record contains kubernetes metadata.  It should
look at both the record tag and the record data and use whatever
kubernetes metadata it finds there.

Result: Fluentd can correctly add kubernetes metadata and assign
records to the correct indices no matter which log driver docker
is using.

Records read from files under `/var/log/containers/*.log` will have
a fluentd tag like `kubernetes.var.log.containers.**`.  This applies
both to CRI-O and docker file logs.  Kubernetes records read from
journald with `CONTAINER_NAME` will have a tag like
`journal.kubernetes.**`.  There is no CRI-O journald log driver yet,
and it is not clear how those records will be represented, but
hopefully they will follow the same `CONTAINER_NAME` convention, in
which case they will Just Work.

### DIFF
--- a/fluentd/configs.d/openshift/filter-k8s-meta.conf
+++ b/fluentd/configs.d/openshift/filter-k8s-meta.conf
@@ -7,9 +7,16 @@
   ssl_partial_chain "#{ENV['SSL_PARTIAL_CHAIN'] || 'true'}"
 </filter>
 
-<filter kubernetes.**>
+<filter kubernetes.journal.**>
   @type parse_json_field
   merge_json_log "#{ENV['MERGE_JSON_LOG'] || 'false'}"
   preserve_json_log "#{ENV['PRESERVE_JSON_LOG'] || 'true'}"
-  json_fields "#{ENV['JSON_FIELDS'] || (ENV['USE_JOURNAL'] == 'true' ? 'MESSAGE,log' : 'log,MESSAGE')}"
+  json_fields "#{ENV['JSON_FIELDS'] || 'MESSAGE,log'}"
+</filter>
+
+<filter kubernetes.var.log.containers.**>
+  @type parse_json_field
+  merge_json_log "#{ENV['MERGE_JSON_LOG'] || 'false'}"
+  preserve_json_log "#{ENV['PRESERVE_JSON_LOG'] || 'true'}"
+  json_fields "#{ENV['JSON_FIELDS'] || 'log,MESSAGE'}"
 </filter>
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -60,35 +60,12 @@ issue_deprecation_warnings() {
     : # none at the moment
 }
 
-docker_uses_journal() {
-    # need to be able to handle cases like
-    # OPTIONS='--log-driver=json-file ....' # or use --log-driver=journald
-    # if "log-driver" is set in /etc/docker/daemon.json, assume that it is
-    # authoritative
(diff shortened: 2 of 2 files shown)


fluentd/configs.d/openshift/filter-k8s-meta.conf
```diff
@@ -7,9 +7,18 @@
...
+  json_fields "#{ENV['JSON_FIELDS'] || (ENV['USE_JOURNAL'] == 'false' ? 'log,MESSAGE' : 'MESSAGE,log')}"
+</filter>
+
+<filter kubernetes.var.log.containers.**>
   @type parse_json_field
   merge_json_log "#{ENV['MERGE_JSON_LOG'] || 'false'}"
   preserve_json_log "#{ENV['PRESERVE_JSON_LOG'] || 'true'}"
-  json_fields "#{ENV['JSON_FIELDS'] || (ENV['USE_JOURNAL'] == 'true' ? 'MESSAGE,log' : 'log,MESSAGE')}"
+  # use log as the field to parse for json unless USE_JOURNAL is explicitly set to true
+  json_fields "#{ENV['JSON_FIELDS'] || (ENV['USE_JOURNAL'] == 'true' ? 'MESSAGE,log' : 'log,MESSAGE' )}"
```
ewolinetz: is the use journal check still necessary here? do we expect to have a `/var/log/containers/*` entry that would use anything other than `'log,MESSAGE'`?

richm: It doesn't make much sense.  I'll get rid of it.

fluentd/configs.d/openshift/filter-k8s-meta.conf
```diff
@@ -7,9 +7,18 @@
   ssl_partial_chain "#{ENV['SSL_PARTIAL_CHAIN'] || 'true'}"
 </filter>
 
-<filter kubernetes.**>
+<filter kubernetes.journal.**>
+  @type parse_json_field
+  merge_json_log "#{ENV['MERGE_JSON_LOG'] || 'false'}"
+  preserve_json_log "#{ENV['PRESERVE_JSON_LOG'] || 'true'}"
+  # use MESSAGE as the field to parse for json unless USE_JOURNAL is explicitly set to false
```
ewolinetz: comment doesn't seem to apply anymore

richm:
Hmm - we can't backport this to 3.10 and earlier
- 3.10 and earlier still use fluent-plugin-kubernetes_metadata_filter-1.0.3
- 3.10 and earlier do not use the separate merge_json_log code

Not sure what we can do, except to suggest to customers to make sure /etc/sysconfig/docker and/or /etc/docker/daemon.json specify the log driver in a way that fluentd can parse . . .

PR #1679, opened 2019-06-24
[WIP] add xpack to es5 to be used when migrating to es6 per es docs

ewolinetz:
As part of moving to es6 we want to verify that the indices and templates that we currently have are valid before migrating the cluster. There was no way to install this into an already running cluster due to lack of permissions in the container with `oc exec`

The intention is that the elasticsearch/clusterlogging operator would call this endpoint before beginning an upgrade.

Ref: https://www.elastic.co/guide/en/elastic-stack/6.8/upgrading-elastic-stack.html#xpack-stack-upgrade


lukas-vlcek:
@ewolinetz Are you sure it is legal to include and re-distribute x-pack even when it is disabled? Will it work correctly when SG is installed first?

ewolinetz:
@lukas-vlcek i'm not sure we can distribute it. after talking with rich we were going to go through the breaking changes docs and write checks based on that instead.

lukas-vlcek:
Yes, going through the release notes is one thing to do, the other is to run series of real cluster upgrades and keep an eye on ES log where deprecation logs are printed.

PR #1680, opened 2019-06-25
Bug 1722380: Logging data from all projects are stored to .orphaned indexes with Elasticsearch

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1722380
Cause: Fluentd is unable to correctly determine the docker log
driver.  It thinks the log driver is journald when it is json-file.
Fluentd then looks for the `CONTAINER_NAME` field in the record to
hold the kubernetes metadata and it is not present.

Consequence: Fluentd is not able to add kubernetes metadata to
records.  Records go to the .orphaned index.  Fluentd spews lots
of errors like this:
```
[error]: record cannot use elasticsearch index name type project_full: record is missing kubernetes field
```

Fix: Fluentd should not rely on reading the docker configuration file
to determine if the record contains kubernetes metadata.  It should
look at both the record tag and the record data and use whatever
kubernetes metadata it finds there.

Result: Fluentd can correctly add kubernetes metadata and assign
records to the correct indices no matter which log driver docker
is using.

Records read from files under `/var/log/containers/*.log` will have
a fluentd tag like `kubernetes.var.log.containers.**`.  This applies
both to CRI-O and docker file logs.  Kubernetes records read from
journald with `CONTAINER_NAME` will have a tag like
`journal.kubernetes.**`.  There is no CRI-O journald log driver yet,
and it is not clear how those records will be represented, but
hopefully they will follow the same `CONTAINER_NAME` convention, in
which case they will Just Work.

(cherry picked from commit 33011c5efdd9c7a4927e012aa6543ebf1558926e)
manual cherrypick of https://github.com/openshift/origin-aggregated-logging/pull/1678

### DIFF
--- a/fluentd/configs.d/openshift/filter-k8s-meta.conf
+++ b/fluentd/configs.d/openshift/filter-k8s-meta.conf
@@ -5,13 +5,20 @@
   watch "#{ENV['K8S_METADATA_WATCH'] || 'false'}"
   bearer_token_file /var/run/secrets/kubernetes.io/serviceaccount/token
   ca_file /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
-  use_journal "#{ENV['USE_JOURNAL'] || 'false'}"
+  use_journal "#{ENV['USE_JOURNAL'] || 'nil'}"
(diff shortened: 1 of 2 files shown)


richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1722380

richm:
tests are passing now - please review - see the `nil` trick

PR #1685, opened 2019-07-09
LOG-372 - rsyslog - support improved undefined field handling

nhosoi:
Adding test cases to zzz-rsyslog.sh

This PR depends on https://github.com/openshift/cluster-logging-operator/pull/202.


test/zzz-rsyslog.sh
```diff
@@ -274,7 +275,7 @@ if [ $logsize -gt 0 ]; then
     oc exec $rpod -c rsyslog -- ls -l /var/log/rsyslog/ | artifact_out
     oc exec $rpod -c rsyslog -- cat /var/log/rsyslog/rsyslog.log > $ARTIFACT_DIR/rsyslog-before-rotation.log
     # this will restart the rsyslog pod
-    oc set env daemonset/rsyslog LOGGING_FILE_SIZE=$maxsize LOGGING_FILE_AGE=$maxcount
+    oc set env $rsyslog_ds LOGGING_FILE_SIZE=$maxsize LOGGING_FILE_AGE=$maxcount
```
richm:
```
/go/src/github.com/openshift/origin-aggregated-logging/test/zzz-rsyslog.sh: line 278: rsyslog_ds: unbound variable
```

richm: `$rsyslog_ds` does not have the correct value:
the zzz-rsyslog test failed:
```
F0709 18:21:33.913427  108255 helpers.go:119] Error from server (NotFound): daemonsets.extensions "logging-rsyslog" not found
```
should be `rsyslog` not `logging-rsyslog`

hack/testing/util.sh
```diff
@@ -784,6 +798,16 @@ get_fluentd_cm_name() {
 
 fluentd_cm=${fluentd_cm:-$(get_fluentd_cm_name)}
 
+get_rsyslog_ds_name() {
```
richm: I think you should just hardcode this to be `ds/rsyslog`

test/zzz-rsyslog.sh
nhosoi: I can revert it.  But why this does not work?  I 

+get_rsyslog_ds_name() {
+    if oc -n ${LOGGING_NS} get daemonset rsyslog -o name > /dev/null 2>&1 ; then
+        echo daemonset/rsyslog
+    else
+        echo daemonset/logging-rsyslog
+    fi
+}
+
+rsyslog_ds=${rsyslog_ds:-$(get_rsyslog_ds_name)}

richm: because `daemonset rsyslog` does not exist when `get_rsyslog_ds_name` is called?

nhosoi: I see.  That's possible...  Thanks.  I'm going to change the code as you suggested.

> I think you should just hardcode this to be ds/rsyslog

richm:
```
/go/src/github.com/openshift/origin-aggregated-logging/test/zzz-rsyslog.sh: line 367: $3: unbound variable
```

test/zzz-rsyslog.sh
```diff
@@ -320,9 +321,122 @@ else
...
+# add our test data to the rsyslog pipeline via a volume mount
+oc set volumes $rsyslog_ds --add --name=rsyslog-undefined-test \
+    -t hostPath -m /etc/rsyslog.d/66-debug.conf \
+    --path $cfg 2>&1 | artifact_out
+
+restart_rsyslog_pod $rpod
+rpod=$( get_running_pod rsyslog )
+
+artifact_log "1. default values"
+os::cmd::expect_success_and_text "oc oc exec $rpod -c rsyslog -- grep "use_undefined:" /var/log/rsyslog/rsyslog.log | tail -n 1 | | awk '{print $3}'" "false"
```
richm: no `$3` here?

richm: hmm - looks like some other copy/paste errors.

richm:
```
os::cmd::expect_success_and_text "oc exec $rpod -c rsyslog -- grep 'use_undefined:' /var/log/rsyslog/rsyslog.log | tail -n 1 | | awk '{print $3}'" "false"
```

test/zzz-rsyslog.sh
```diff
@@ -320,9 +321,122 @@ else
...
+os::cmd::expect_success_and_text "oc oc exec $rpod -c rsyslog -- grep "undefined_name:" /var/log/rsyslog/rsyslog.log | tail -n 1 | | awk '{print $3}'" "undefined"
```
richm: see above

test/zzz-rsyslog.sh
nhosoi: Thanks.  Yeah, the lines were broken... :(
For some reason, my attempt to run the CI from get-cluster-run-tests fail in the earlier section.  So, I decided to debug here.

```
177 sleep 10
178 wait_for_fluentd_to_catch_up get_logmessage get_logmessage2 
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

test/zzz-rsyslog.sh
```diff
@@ -320,9 +321,122 @@ else
...
+oc set volumes $rsyslog_ds --add --name=rsyslog-undefined-test \
+    -t hostPath -m /etc/rsyslog.d/66-debug.conf \
+    --path $cfg 2>&1 | artifact_out
+
+restart_rsyslog_pod $rpod
+rpod=$( get_running_pod rsyslog )
+
+artifact_log "1. default values"
+os::cmd::expect_success_and_text "oc oc exec $rpod -c rsyslog -- grep 'use_undefined:' /var/log/rsyslog/rsyslog.log | tail -n 1 | | awk '{print \$3}'" "false"
+os::cmd::expect_success_and_text "oc oc exec $rpod -c rsyslog -- grep 'undefined_name:' /var/log/rsyslog/rsyslog.log | tail -n 1 | | awk '{print \$3}'" "undefined"
```
richm: these lines still have the double `oc oc` and the double `| |` which are causing the test failure:
```
/go/src/github.com/openshift/origin-aggregated-logging/hack/lib/cmd.sh: eval: line 337: syntax error near unexpected token `|'
/go/src/github.com/openshift/origin-aggregated-logging/hack/lib/cmd.sh: eval: line 337: `oc oc exec rsyslog-n4c29 -c rsyslog -- grep 'use_undefined:' /var/log/rsyslog/rsyslog.log | tail -n 1 | | awk '{print $3}''
```

nhosoi: There were soooo many errors in each line... :(

richm:
flake?
/test full-integ-aws

richm:
@nhosoi ok to remove the work-in-progress label?

richm:
does not appear to be a flake 
```
FAILURE after 599.759s: hack/testing/util.sh:531: executing 'curl_es elasticsearch /.operations.*/_count -X POST -d '{"query":{"bool":{"filter":{"match_phrase":{"message":"GET /e3b0287c620b46628a58006cbcb81c0c 404 "}},"must":{"term":{"kubernetes.container_name":"kibana"}}}}}' | get_count_from_json' expecting any result and text '1'; re-trying every 0.2s until completion or 600.000s: the command timed out
```
This is consistently failing in the same place.  Did something change in the rsyslog pipeline which causes application logs to be dropped?

hack/testing/util.sh
```diff
@@ -717,13 +717,15 @@ get_mux_pod_log() {
...
-    if oc exec $pod -- logs 2>&1 ; then
+    oc logs $pod -c rsyslog 2>&1
+    if oc exec $pod -c rsyslog -- logs 2>&1 ; then
         : # done
     elif oal_sudo test -f $logfile ; then
         # can't read from the pod directly - see if we can get the log file
         oal_sudo cat $logfile
     fi
+    oc logs $pod -c logrotate 2>&1
+    oc exec $pod -c logrotate -- logs 2>&1 || :
```
richm: does `logs` get both the /var/log/rsyslog/logrotate.log and /var/lib/rsyslog.pod/logrotate.log?

test/zzz-rsyslog.sh
```diff
@@ -63,6 +63,9 @@ cleanup() {
         start_fluentd true 2>&1 | artifact_out
     else
         rpod=$( get_running_pod rsyslog )
+        oc logs $rpod -c rsyslog > $ARTIFACT_DIR/cleanup.rsyslog.log 2>&1 || :
+        oc exec $rpod -c rsyslog -- logs >> $ARTIFACT_DIR/cleanup.rsyslog.log 2>&1 || :
+        oc exec $rpod -c logrotate -- /usr/bin/cat /var/log/rsyslog/logrotate.log > $ARTIFACT_DIR/cleanup.logrotate.log 2>&1 || :
```
richm: you should use `get_rsyslog_pod_log` here

hack/testing/util.sh
```diff
@@ -583,7 +583,6 @@ function wait_for_fluentd_to_catch_up() {
             os::log::error ops record for "$uuid_es_ops" not found in journal
         fi
         os::log::error here is the starting fluentd journal cursor
-        echo $journalstartcursor
```
richm: remove the line above it too

test/zzz-rsyslog.sh
```diff
@@ -178,8 +178,8 @@ fi
 $deployfunc
 
 # MERGE_JASON_LOG has to be true?
-oc set env $rsyslog_ds MERGE_JSON_LOG=true
-rpod=$( get_running_pod rsyslog )
+#oc set env $rsyslog_ds MERGE_JSON_LOG=true
+#rpod=$( get_running_pod rsyslog )
```
richm: yeah - I don't see any processing that requires `MERGE_JSON_LOG=true`

hack/testing/util.sh
nhosoi: Currently, it doesn't.  (and actually, "oc exec $pod -c logrotate -- logs" does not work as expected. It just reads /var/log/rsyslog/rsyslog.log...)  I'm going to fix rsyslog/utils/logs to cover these logrotate logs...

test/zzz-rsyslog.sh
nhosoi: Yeah, get_rsyslog_pod_log is called via get_all_logging_pod_logs at line 50.  I added them (line 66-68) to get them manually. :)  Once I could make zzz-rsyslog.sh work, I'm deleting these...
```
47 cleanup() {
 ...
 50     get_all_logging_pod_logs
```

hack/testing/util.sh
nhosoi: Ok.

test/zzz-rsyslog.sh
nhosoi: This [1] (and equivalent $!log) is the only places MERGE_JSON_LOG has an impact.  And my CLO PR disabled MERGE_JSON_LOG, which looked the biggest change other than mmexternal.  

I verified even if mmexternal is disabled by setting USE_MMEXTERNAL=false, zzz-rsyslog.sh failed so I was trying MERGE_JSON_LOG next.  But before running the test with MERGE_JSON_LOG=true, I found the "unbound variable" error, so I'm testing the fix now.  Sorry, very slow move...

[1] - snippet of 65-viaq-formatting.conf
```
 11     set $!pipeline_metadata!collector!original_raw_message = $!MESSAGE;
 12     if (strlen(`echo $MERGE_JSON_LOG`) > 0) and (`echo $MERGE_JSON_LOG` == "true") then {
 13         action(type="mmnormalize" ruleBase="/etc/rsyslog.d/parse_json.rulebase" variable="$!MESSAGE")
 14     }
```

richm:
I think I've figured out why rsyslog is broken - https://github.com/richm/cluster-logging-operator/tree/rsyslog-cleanup

nhosoi:
> I think I've figured out why rsyslog is broken - https://github.com/richm/cluster-logging-operator/tree/rsyslog-cleanup
> 
> specifically - [richm/cluster-logging-operator@57d2d70#diff-f352aec93a80474539bd17b44867dac0R8](https://github.com/richm/cluster-logging-operator/commit/57d2d7004430cf3c3cd7056139e2c4bb20b86103#diff-f352aec93a80474539bd17b44867dac0R8)

Oh, my...  I thought the location of the module load was the least harmful change I made...  I put it there at the very end of my implementing work since I thought it's cleaner...  But I broke the code with it...   :(

Let me apply your patch (except the USE_MMEXTERNAL setting).  Thank you soooo much, Rich!

nhosoi:
> Let me apply your patch (except the USE_MMEXTERNAL setting).

Or maybe, could you please open a PR with  https://github.com/richm/cluster-logging-operator/tree/rsyslog-cleanup (with enabling USE_MMEXTERNAL)?  I will set lgtm immediately.

richm:
you might want to steal this change - https://github.com/openshift/origin-aggregated-logging/pull/1689/files#diff-0144f31f058e341bb59f3d9838899a29L125 - so that you are using the latest cluster-logging-operator from the ocp repo - the origin one with the fix might not be available for several hours

richm:
ok - looks like you can use registry.svc.ci.openshift.org/origin/4.2:cluster-logging-operator now - it has the 67-mmexternal.conf fix

nhosoi:
> ok - looks like you can use registry.svc.ci.openshift.org/origin/4.2:cluster-logging-operator now - it has the 67-mmexternal.conf fix

Thanks, Rich!  This round of CI has a code dumping rsyslog configmap to artifacts dir.  I was thinking to add it if the dump does not show the latest config.  So, I don't have to worry it. :)

nhosoi:
Thank you for debugging the PR, Rich!

I could cherry-pick "no parentheses in if test" and "undefined_field output should be quiet unless undefined_debug", but cherry-picking "Merge branch 'master' into undefined_external" fails.  I also tried rebase against the latest master, but it fails with conflicts in hack/testing/util.sh, too.  Could you tell me what's the cleanest way to sync the source tree with yours?

I noticed one thing in hack/testing/util.sh...  I forgot to delete the previous get_rsyslog_pod_log at line 713.
https://github.com/nhosoi/origin-aggregated-logging/blob/undefined_external/hack/testing/util.sh#L713

This line 732 one is the right one.
https://github.com/nhosoi/origin-aggregated-logging/blob/undefined_external/hack/testing/util.sh#L732

If you could remove line 713 get_rsyslog_pod_log, I'd appreciate it.

nhosoi:
Recently, this line keeps failing due to the grep failure - [1], which command line used to work...
`377 os::cmd::expect_success_and_text "oc exec $rpod -c rsyslog -- grep 'use_undefined:' /var/log/rsyslog/rsyslog.log | tail -n 1 |     awk '{print \$3}'" "false"`
But /var/log/rsyslog/rsyslog.log should have this line "mmexternal: use_undefined:  false" according to the rsyslog.log copy in the artifacts.  It might be just a timing issue.  Can we add this `try_until_success`?

```
diff --git a/test/zzz-rsyslog.sh b/test/zzz-rsyslog.sh
index e901dc7f..6ae92366 100755
--- a/test/zzz-rsyslog.sh
+++ b/test/zzz-rsyslog.sh
@@ -367,11 +367,11 @@ rpod=$( get_running_pod rsyslog )
 oc exec $rpod -c rsyslog -- ls -l /etc/rsyslog.d/66-debug.conf | artifact_out
 oc exec $rpod -c rsyslog -- cat /etc/rsyslog.d/66-debug.conf | artifact_out
 
-# DEBUG DEBUG DEBUG
+# UNDEFINED_DEBUG needs to be enabled for checking configs in the following tests.
 oc set env $rsyslog_ds UNDEFINED_DEBUG=true
 os::cmd::try_until_failure "oc get pods $rpod > /dev/null 2>&1"
 os::cmd::try_until_success "oc get pods 2> /dev/null | grep -q 'rsyslog.*Running'"
-# DEBUG DEBUG DEBUG
+os::cmd::try_until_success "oc exec $rpod -c rsyslog -- grep 'use_undefined:' /var/log/rsyslog/rsyslog.log"
```



[1]
Running test/zzz-rsyslog.sh:377: executing 'oc exec rsyslog-b8xdd -c rsyslog -- grep 'use_undefined:' /var/log/rsyslog/rsyslog.log | tail -n 1 | awk '{print $3}'' expecting success and text 'false'...
FAILURE after 0.367s: test/zzz-rsyslog.sh:377: executing 'oc exec rsyslog-b8xdd -c rsyslog -- grep 'use_undefined:' /var/log/rsyslog/rsyslog.log | tail -n 1 | awk '{print $3}'' expecting success and text 'false': the command returned the wrong error code; the output content test failed
There was no output from the command.

nhosoi:
@richm, never mind about my previous comments.  I could just fetch and checkout this PR with your changes. :p  Thanks!

nhosoi:
@richm, this is the line causing the CI failure - [1]  I'm setting string "True".  I tried the value without double-quotes, then set failed.  (but I'm retrying the test one more time from now...)
https://github.com/nhosoi/origin-aggregated-logging/blob/undefined_external/test/zzz-rsyslog.sh#L349

```
[1]
FAILURE after 0.039s: test/zzz-rsyslog.sh:383: executing 'cat /tmp/artifacts/scripts/test-zzz-rsyslog/artifacts/zzz-rsyslog-record.json | python /go/src/github.com/openshift/origin-aggregated-logging/hack/testing/test-viaq-data-model.py test1' expecting success: the command returned the wrong error code
Error: input field [undefined12] expected value [True] does not match actual value [True]
```

test/zzz-rsyslog.sh
```diff
@@ -346,11 +346,11 @@ cat > $cfgdir/66-debug.conf << EOF
 # DEBUGGING
 set \$!undefined1 = "undefined1";
 set \$!undefined11 = 1111;
-set \$!undefined12 = "True";
+set \$!undefined12 = True;
```
richm: https://www.rsyslog.com/doc/v8-stable/rainerscript/data_types.html
```
RainerScript is a typeless language. That doesn’t imply you don’t need to care about types. Of course, expressions like “A” + “B” will not return a valid result, as you can’t really add two letters (to concatenate them, use the concatenation operator &).  However, all type conversions are automatically done by the script interpreter when there is need to do so.
```
also https://www.rsyslog.com/doc/v8-stable/rainerscript/variable_property_types.html

So I don't think you can set a JSON boolean value in rainierscript.  I also don't see how you can set an array valued object - just strings and numbers, and JSON hash via using the `$!hash!value` construction.

This begs the question of how the JSON parser in rsyslog works - how does it internally represent JSON boolean and JSON array values?

I think for testing purposes we'll need to figure out how to pass in JSON booleans and arrays.  I think we may have to have rsyslog parse a file, or use the `parse_json` function if it is available in this version of rsyslog.

nhosoi: Looks promising.  I see `doFunc_parse_json` in origin-aggregated-logging/rsyslog/vendored_src/rsyslog/rsyslog/grammar/rainerscript.c.  I'll give it a try...

test/zzz-rsyslog.sh
```diff
@@ -320,9 +322,120 @@ else
...
+
+artifact_log "************************"
+artifact_log "Testing undefined fields"
+artifact_log "************************"
+
+artifact_log "0. Preparing test data"
+cfgdir=$( mktemp -d )
+oc extract cm/rsyslog --to=$cfgdir
+cat > $cfgdir/66-debug.conf << EOF
+set $.ret = parse_json("{\"undefined1\":\"undefined1\",\"undefined11\":1111,\"undefined12\":\"True\",\"empty1\":\"\",\"undefined2\":{\"undefined2\":\"undefined2\",\"\":\"\",\"undefined22\":2222,\"undefined23\":\"False\"},\"undefined3\":{\"emptyvalue\":\"\"},\"undefined4\":\"undefined4\",\"undefined5\":\"undefined5\",\"undefined.6\":\"undefined6\"}", "\$!");
```
richm: I'm running https://github.com/rsyslog/rsyslog/blob/master/tests/rscript_parse_json.sh#L11 - I modified it to add a boolean and an array like this:
```
 local4.* {
-       set $.ret = parse_json("{ \"c1\":\"data\" }", "\$!parsed");
+       set $.ret = parse_json("{ \"c1\":\"data\",\"bool\":True,\"ary\":[1,2,3] }", "\$!parsed");
        action(type="omfile" file=`echo $RSYSLOG_OUT_LOG` template="outfmt")
```
This is the output
```
{ "parsed": { "c1": "data", "bool": true, "ary": [ 1, 2, 3 ] } }
```
So it looks like you can use `\"undefined\":True` as the boolean value.

Perhaps the problem is when rsyslog passes the data to the mmexternal undefined_field program?  I'm not sure what problem you are trying to solve.

nhosoi: Thanks, Rich.  When I set the following test data, then I got the errors [1].  So, first I suspected the boolean, but it'd be ok.  Then, \"\":\"\" is not acceptable?
`set $.ret = parse_json("{\"undefined1\":\"undefined1\",\"undefined11\":1111,\"undefined12\":True,\"empty1\":\"\",\"undefined2\":{\"undefined2\":\"undefined2\",\"\":\"\",\"undefined22\":2222,\"undefined23\":False},\"undefined3\":{\"emptyvalue\":\"\"},\"undefined4\":\"undefined4\",\"undefined5\":\"undefined5\",\"undefined.6\":\"undefined6\"}", "\$!");`
```
[1]
Jul 15 05:56:46 rsyslog-v25vh rsyslogd-2207: error during parsing file /etc/rsyslog.d/66-debug.conf, on or before line 1: invalid character '"' in expression - is there an invalid escape sequence somewhere? [v8.37.0-9.el7 try http://www.rsyslog.com/e/2207 ]
Jul 15 05:56:46 rsyslog-v25vh rsyslogd-2207: error during parsing file /etc/rsyslog.d/66-debug.conf, on or before line 1: invalid character '"' in expression - is there an invalid escape sequence somewhere? [v8.37.0-9.el7 try http://www.rsyslog.com/e/2207 ]
Jul 15 05:56:46 rsyslog-v25vh rsyslogd: [origin software="rsyslogd" swVersion="8.37.0-9.el7" x-pid="14" x-info="http://www.rsyslog.com"] start
Jul 15 05:56:46 rsyslog-v25vh rsyslogd-2441: internal error:  getJSONRootAndMutex; invalid indicator char {(7b) [v8.37.0-9.el7 try http://www.rsyslog.com/e/2441 ]
Jul 15 05:56:46 rsyslog-v25vh rsyslogd-2441: internal error:  getJSONRootAndMutex; invalid indicator char {(7b) [v8.37.0-9.el7 try http://www.rsyslog.com/e/2441 ]
```

richm: >  Then, "":"" is not acceptable?

I don't know.  I don't know what it is complaining about

richm: This is the output in the artifact out file when running the test:
```
set $.ret = parse_json("{\"undefined1\":\"undefined1\",\"undefined11\":1111,\"undefined12\":True,\"empty1\":\"\",\"undefined2\":{\"undefined2\":\"undefined2\",\"\":\"\",\"undefined22\":2222,\"undefined23\":\"False\"},\"undefined3\":{\"emptyvalue\":\"\"},\"undefined4\":\"undefined4\",\"undefined5\":\"undefined5\",\"undefined.6\":\"undefined6\"}", "$!");
```
* I suggest to use single quotes for the string passed to `parse_json` to avoid have to escape so many double quotes https://www.rsyslog.com/doc/v8-stable/rainerscript/constant_strings.html
* the final `"$!"` looks wrong - according to https://www.rsyslog.com/doc/v8-stable/rainerscript/functions/rs-parse_json.html it should look like this - `"\$!"` - the shell here document is eating one of the escapes `\` - try using single quotes for this too:
```
cat > $cfgdir/66-debug.conf << EOF
set \$.ret = parse_json('{"undefined1":"undefined1","undefined11":1111,"undefined12":True,"empty1":"","undefined2":{"undefined2":"undefined2","":"","undefined22":2222,"undefined23":"False"},"undefined3":{"emptyvalue":""},"undefined4":"undefined4","undefined5":"undefined5","undefined.6":"undefined6"}', '\$!');
EOF
```
that should expand to 
```
set $. = .... ,  '$!');
```

richm: it looks like using `$.` in a bash here document is preserved, I guess because `$.` isn't a legal variable in shell, but `$!` is _not_ preserved, because `$!` has a special meaning in bash.

nhosoi: Thanks, Rich.  I replaced the 66-debug.conf generation code with your suggestion.

richm:
removing `noaction` helps.  This program should always have an "action" - it should always remove empty fields.  One of the problems when using `noaction` was that the field `"":""` was being passed through to Elasticsearch which was returning an error 400 - syntax error - cannot have an empty field name.  Once I removed `noaction`, there was one other problem - I was seeing fields like `"undefined3":{}` being sent - that is, it was not removing the empty field if the field value was an empty map.  I fixed that too: https://github.com/richm/origin-aggregated-logging/commit/4da2e6b571fe0703dbfb4d0cacfc232072b5d87d

nhosoi:
> removing `noaction` helps. This program should always have an "action" - it should always remove empty fields. One of the problems when using `noaction` was that the field `"":""` was being passed through to Elasticsearch which was returning an error 400 - syntax error - cannot have an empty field name. Once I removed `noaction`, there was one other problem - I was seeing fields like `"undefined3":{}` being sent - that is, it was not removing the empty field if the field value was an empty map. I fixed that too: [richm@4da2e6b](https://github.com/richm/origin-aggregated-logging/commit/4da2e6b571fe0703dbfb4d0cacfc232072b5d87d)

Thanks a lot, Rich!  Yeah, my first intention was avoiding to execute undefined_field as much as possible could help performance degradation if any...  Currently, 67-mmexternal.conf in CLO calls parse_json_skip_empy if SKIP_EMPTY is true.  We may want to enable it if all the undefined field related settings are off but only empty fields to be processed.  That way, we could avoid exec undefined_field such cases.

nhosoi:
Sorry, found a bug...  The current code is not supporting the integer/float value... :(

hack/testing/test-viaq-data-model.py
```diff
@@ -142,7 +142,7 @@ def empty(thing):
         if xx not in dd['_source']:
             print "Error: input does not have the field [%s]" % xx
             sys.exit(1)
-        if not match[xx] == dd['_source'][xx]:
+        if not match[xx] == dd['_source'][xx] or not match[xx].decode('utf-8') == dd['_source'][xx.decode('utf-8')]:
```
richm: I doubt that this is the problem . . .

nhosoi: Ok.  According to this output [1], certainly `stream` is an issue.  Do you think once stream is dropped, the comparison returns SUCCESS?

[1] - http://pastebin.test.redhat.com/781238

richm: > Do you think once stream is dropped, the comparison returns SUCCESS?

Yes

nhosoi: All right.  Let me undo the change...

nhosoi:
This PR depends on https://github.com/openshift/cluster-logging-operator/pull/214.

nhosoi:
This PR has a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1730207
Bug 1730207 - Too many "No Action Needed for xxxxxxxxx" logs in rsyslog pod.

nhosoi:
Hi @richm,

Since this morning, this test has started failing:
https://github.com/nhosoi/origin-aggregated-logging/blob/undefined_external/test/zzz-rsyslog.sh#L224

This is the exact failure in the output https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_origin-aggregated-logging/1685/pull-ci-openshift-origin-aggregated-logging-master-full-integ-aws/1091/build-log.txt.
```
Running test/zzz-rsyslog.sh:224: executing 'diff /tmp/artifacts/scripts/test-zzz-rsyslog/artifacts/zzz-rsyslog-expected-nsannotations.json /tmp/artifacts/scripts/test-zzz-rsyslog/artifacts/zzz-rsyslog-actual-nsannotations.json' expecting success...
FAILURE after 0.012s: test/zzz-rsyslog.sh:224: executing 'diff /tmp/artifacts/scripts/test-zzz-rsyslog/artifacts/zzz-rsyslog-expected-nsannotations.json /tmp/artifacts/scripts/test-zzz-rsyslog/artifacts/zzz-rsyslog-actual-nsannotations.json' expecting success: the command returned the wrong error code
2d1
<   "openshift_io/node-selector": "",
```
Indeed, zzz-rsyslog-expected-nsannotations.json [1] and actual namespace_annotations [2] do not match.
```
    [1]
    {
      "openshift_io/node-selector": "",
      "openshift_io/sa_scc_mcs": "s0:c22,c19",
      "openshift_io/sa_scc_supplemental-groups": "1000500000/10000",
      "openshift_io/sa_scc_uid-range": "1000500000/10000"
    }

    [2]
    "namespace_annotations": {
         "openshift_io/sa_scc_mcs": "s0:c22,c19",
         "openshift_io/sa_scc_supplemental-groups": "1000500000/10000",
         "openshift_io/sa_scc_uid-range": "1000500000/10000"
    },
```

Do you have any idea what caused this mismatch?

richm:
`"openshift_io/node-selector": ""` has an empty value.  Are we stripping fields with empty values by default?  If not, then possibly a bug in mmkubernetes.
This may be a newly added field in openshift, which is why we haven't seen this problem before, and/or we are now stripping fields with empty values that we didn't used to do.

rsyslog/undefined_field/undefined_field.go
```diff
@@ -205,6 +205,7 @@ func processUndefinedAndMaxNumFields(input map[string]interface{}) (string, map[
 			}
 			undefMap[field] = val
 			count--
+			if count < 0 { break }
```
richm: There may be more undefined fields.  For example, if the `cfg.UndefinedMaxNumFields` is 2, and there are 4 undefined fields in the record, we want `undefMap` to contain all 4 of them, so that when we convert `undefMap` to `undefString`, it will contain all of the undefined fields.

nhosoi: > There may be more undefined fields. For example, if the `cfg.UndefinedMaxNumFields` is 2, and there are 4 undefined fields in the record, we want `undefMap` to contain all 4 of them, so that when we convert `undefMap` to `undefString`, it will contain all of the undefined fields.

@richm, I might still misunderstand the definition of processUndefinedAndMaxNumFields, but I still think the function should return 2 undefString and remove 4 fields from input (in your example).  I've updated the patch according to it and it passes the unit test.  Can I have your thought?

richm: it should work like this: https://github.com/ViaQ/fluent-plugin-viaq_data_model
```
If undefined_max_num_fields is a number greater than -1, and if the number of undefined fields is greater than this number, all of the undefined fields will be converted to their JSON string representation and stored in the undefined_name named field.
...
undefined_max_num_fields - integer - default -1

    If the number of undefined fields exceeds the value of undefined_max_num_fields, then convert the hash of undefined fields to its JSON string representation, and store the values in the undefined_name field - see below
    Use a value of -1 if you want to have an unlimited number of undefined fields (not recommended)
    Using undefined_max_num_fields implies that you want to use undefined_name as the name of the field to store the value, even if use_undefined is not set - if you want to use a different field name than "undefined" then set undefined_name

```
We have to find all of the undefined fields so we can convert all of them to string.

nhosoi:
> `"openshift_io/node-selector": ""` has an empty value. Are we stripping fields with empty values by default? If not, then possibly a bug in mmkubernetes.
> This may be a newly added field in openshift, which is why we haven't seen this problem before, and/or we are now stripping fields with empty values that we didn't used to do.

All right.  An empty value for `openshift_io/node-selector` has this important meaning - [1].  That is, we have to include `openshift_io/node-selector` in CDM_KEEP_EMPTY_FIELDS, by default, and we have no chance to use SKIP_EMPTY... :(

[1] - https://docs.openshift.com/enterprise/3.1/admin_guide/managing_projects.html
> "openshift_io/node-selector": "" has an empty value. Are we stripping fields with empty values by default? If not, then possibly a bug in mmkubernetes.
> This may be a newly added field in openshift, which is why we haven't seen this problem before, and/or we are now stripping fields with empty values that we didn't used to do.
>

richm:
Looks like not a flake this time.  It gets most of the way through the rsyslog test, then
```
[ERROR] PID 107076: test/zzz-rsyslog.sh:616: `return $rc` exited with status 1.
```
I think it failed here - https://github.com/openshift/origin-aggregated-logging/blob/05267ecbe8ccc42595eb1ceee5fb7eb3d56f1def/test/zzz-rsyslog.sh#L397

nhosoi:
> Looks like not a flake this time. It gets most of the way through the rsyslog test, then
> 
> ```
> [ERROR] PID 107076: test/zzz-rsyslog.sh:616: `return $rc` exited with status 1.
> ```
> 
> I think it failed here -
> 
> https://github.com/openshift/origin-aggregated-logging/blob/05267ecbe8ccc42595eb1ceee5fb7eb3d56f1def/test/zzz-rsyslog.sh#L397

Yes, I'm looking into it.  Thanks!

nhosoi:
Actually, I have one question.  This is the message in [1] - search result from the undefined test case "1. default value".
`"message": "{\"type\":\"response\",\"@timestamp\":\"2019-07-23T21:54:26Z\",\"tags\":[],\"pid\":223,\"method\":\"get\",\"statusCode\":404,\"req\":{\"url\":\"/ab0ee6b7ada6442e9547810ff6b483d0\",\"method\":\"get\",\"headers\":{\"user-agent\":\"curl/7.29.0\",\"host\":\"localhost:5601\",\"accept\":\"*/*\"},\"remoteAddress\":\"127.0.0.1\",\"userAgent\":\"127.0.0.1\"},\"res\":{\"statusCode\":404,\"responseTime\":1,\"contentLength\":9},\"message\":\"GET /ab0ee6b7ada6442e9547810ff6b483d0 404 1ms - 9.0B\"}",`
Is this the expected message value?  I was somehow the value is supposed to be just `"GET /ab0ee6b7ada6442e9547810ff6b483d0 404 1ms - 9.0B"`.  But I could be wrong...

[1] - https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_origin-aggregated-logging/1685/pull-ci-openshift-origin-aggregated-logging-master-full-integ-aws/1109/artifacts/full-integ-aws/scripts/test-zzz-rsyslog/artifacts/zzz-rsyslog-record.json

richm:
> Is this the expected message value? I was somehow the value is supposed to be just "GET /ab0ee6b7ada6442e9547810ff6b483d0 404 1ms - 9.0B". But I could be wrong...

Looks like MERGE_JSON_LOG is false?  https://github.com/openshift/cluster-logging-operator/blob/master/files/rsyslog/65-viaq-formatting.conf#L31

test/zzz-rsyslog.sh
```diff
@@ -347,6 +347,8 @@ oc set env $rsyslog_ds LOGGING_FILE_PATH-
 start_rsyslog
 rpod=$( get_running_pod rsyslog )
 
+clear_and_restart_journal
```
richm: this doesn't really help with the new containerized test - if the problem is that the journal is so large that it takes more than 10 minutes for rsyslog to start at the beginning of the journal and read through all of the records to get to the latest one, then the best solution is to use `start_rsyslog cleanfirst` which will remove the imjournal state file and any queued records.  The test uses `RSYSLOG_JOURNAL_READ_FROM_TAIL=on RSYSLOG_FILE_READ_FROM_TAIL=on` https://github.com/openshift/origin-aggregated-logging/pull/1685/files#diff-17f5db5ab4de5254d79187c056ba91b4R154 - which means if there is no imjournal state file, and no queued records, rsyslog will start reading from the end of the journal, which means it should be extremely fast to read the very latest records

nhosoi: > this doesn't really help with the new containerized test - if the problem is that the journal is so large that it takes more than 10 minutes for rsyslog to start at the beginning of the journal and read through all of the records to get to the latest one, then the best solution is to use `start_rsyslog cleanfirst` which will remove the imjournal state file and any queued records. The test uses `RSYSLOG_JOURNAL_READ_FROM_TAIL=on RSYSLOG_FILE_READ_FROM_TAIL=on` https://github.com/openshift/origin-aggregated-logging/pull/1685/files#diff-17f5db5ab4de5254d79187c056ba91b4R154 - which means if there is no imjournal state file, and no queued records, rsyslog will start reading from the end of the journal, which means it should be extremely fast to read the very latest records

Thanks!  I'll apply your suggestions.
The good news is rsyslog tests themselves have passed except hack/testing/util.sh:521: wait_for_ops_log_message...
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_origin-aggregated-logging/1685/pull-ci-openshift-origin-aggregated-logging-master-full-integ-aws/1112/artifacts/full-integ-aws/scripts/test-zzz-rsyslog/artifacts/zzz-rsyslog.sh-artifacts.txt

richm: > application logs are found in journalctl output

???

nhosoi: > > application logs are found in journalctl output
> 
> ???

This is the function ERROR is reported.  (exit code is 141) - [1]  I thought the problem is in journalctl.  But it seems like an bash pipe issue.  Do you think adding `trap "" PIPE` to zzz-rsyslog.sh as in [2] helps?
https://github.com/openshift/origin-aggregated-logging/blob/master/hack/testing/util.sh#L469-L475
```
[1]
[..] [ERROR] PID 130938: hack/testing/util.sh:476: `awk -v "es_ops=SYSLOG_IDENTIFIER=$uuid_es_ops" -v es_ops_out=$ARTIFACT_DIR/es_ops_out.txt '
[..] [ERROR]             BEGIN{RS="";FS="\n"};
[..] [ERROR]             $0 ~ es_ops {print > es_ops_out; exit 0}'` exited with status 141.
[..] [INFO] 		Stack Trace: 
[..] [INFO] 		  1: hack/testing/util.sh:476: `awk -v "es_ops=SYSLOG_IDENTIFIER=$uuid_es_ops" -v es_ops_out=$ARTIFACT_DIR/es_ops_out.txt '
[..] [INFO]             BEGIN{RS="";FS="\n"};
[..] [INFO]             $0 ~ es_ops {print > es_ops_out; exit 0}'`
[..] [INFO] 		  2: hack/testing/util.sh:521: wait_for_ops_log_message
[..] [INFO] 		  3: test/zzz-rsyslog.sh:383: wait_for_fluentd_to_catch_up
[..] [INFO]   Exiting with code 141.
```
```
[2]
--- a/test/zzz-rsyslog.sh
+++ b/test/zzz-rsyslog.sh
@@ -88,6 +88,7 @@ cleanup() {
     exit $return_code
 }
 trap "cleanup" EXIT
+trap "" PIPE 
```

richm: I'm not sure - the `[ERROR] PID 130938: hack/testing/util.sh:476: `awk -v ...` is benign - I think yes, it is related to the pipe exiting prematurely - not sure if adding `trap "" PIPE` will fix it, or if that will also mask other pipe errors which we do want to be reported.  I never could figure out how to suppress that output, I think it has something to do with the fact that it is run in the background.
At any rate, I don't think it is an error.  Check the badly[1] named `es_ops_out.txt` file in the artifacts directory to see if the record added by `logger` could be found in the journal before the test was completed.
[1] it is a bad name because it implies that it is the output from elasticsearch, when in reality it is the record that should have been added to the es ops

hack/testing/test-viaq-data-model.py
```diff
@@ -119,10 +119,7 @@ def empty(thing):
               "undefined3": {"emptyvalue": ""}
     },
     "test7": {"myname":
-              {"undefined1": "undefined1",
-               "undefined11": 1111,
-               "undefined12": True,
-              },
+              "{\"undefined.6\":\"undefined6\",\"undefined1\":\"undefined1\",\"undefined11\":1111,\"undefined12\":true,\"undefined2\":{\"\":\"\",\"undefined2\":\"undefined2\",\"undefined22\":2222,\"undefined23\":false}}",
```
richm: This is going to be tricky - when the hash is converted back to a string, there is no guarantee of the order - so sometimes you will get the string `"{\"undefined.6\":\"undefined6\",\"undefined1\":\"undefined1\",...}"` and sometimes you will get the string `"{\"undefined1\":\"undefined1\",\"undefined.6\":\"undefined6\",...}"`

What I did in the go test for undefined to string is to convert the strings back to maps (which would be dicts in python), then compare the map (or dict) objects.

hack/testing/test-viaq-data-model.py
```diff
@@ -25,6 +25,16 @@ def empty(thing):
             return True
     return False
 
+def dictCompare(str1, str2):
+    s1_obj = json.loads(s1)
+    s2_obj = json.loads(s2)
```
richm:
```python
    return s1_obj == s2_obj
```
comparison of `dict` in python is recursive https://docs.python.org/3/reference/expressions.html#comparisons
what it doesn't say is that if the value is a `dict`, it will do a deep comparison of those:
```
>>> bar = {"sub":{7:8,"ary":[1,2,3,4],"subsub":{"a":"b"}},5:6, 3:4, 1:2}
>>> foo = {1:2, 5:6, 3:4, "sub":{"subsub":{"a":"b"},"ary":[1,2,3,4],7:8}}
>>> foo == bar
True
```

hack/testing/test-viaq-data-model.py
```diff
@@ -25,6 +25,14 @@ def empty(thing):
...
+    if s1_obj == s2_obj:
```
richm: `s1_obj == s2_obj` evaluates to either `True` or `False`, so you can just do
```python
    return s1_obj == s2_obj
```

richm: or even simpler
```python
def dictCompare(str1, str2):
    return json.loads(s1) == json.loads(s2)
```

rsyslog/utils/logs
```diff
@@ -42,10 +42,12 @@ done
 
 if [ $cmd = "cat" ] ; then
     if [ "$include_logrotation" = "true" ] ; then
-        echo "===== rsyslog logs ====="
-        for log_file in $( ls $log_dir/rsyslog.* | sort -Vr ); do
-            cat $log_file
-        done
+        if ls $pod_dir/rsyslog.* > /dev/null 2>&1 ; then
```
richm: this is to fix the bz?

nhosoi: Yes, I asked QE to open a bug by mistake...  I thought the bug was already in the master branch, but it was only in this PR.  I added the logs utility improvement which was needed for debugging.

hack/testing/test-viaq-data-model.py
nhosoi: Thanks!  Done.

Since the CI tests have passed at last (the last piece was the Elasitcsearch type...  If you switch the type of undefined field from json to json string or vice versa, the same field name, e.g., "undefined" cannot be used...  I guess if not yet, we want to add the note to doc.), I'd like to squash the commits.  There are plenty of your contributions and comments.  Is it ok to put them into one commit?

nhosoi:
This PR is fixing the following bugs, as well.
https://bugzilla.redhat.com/show_bug.cgi?id=1730207
https://bugzilla.redhat.com/show_bug.cgi?id=1733011
https://bugzilla.redhat.com/show_bug.cgi?id=1730211

PR #1689, opened 2019-07-10
deploy es with storage and multi-nodes for upgrade testing

richm:


richm:
This PR is ready - ptal

richm:
> lgtm. only question I would ask is 32G disk large enough for the upgrade?

I don't know.  I figure I would start small to save resources, then increase as needed.

> Do we have concerns of infra logs filling up that disk before we complete the test?

I don't know, but we can increase it later if so.

PR #1691, opened 2019-07-10
Bug 1728856: Fix prometheus metrics visibility

jcantrill:
This PR fixes https://bugzilla.redhat.com/show_bug.cgi?id=1728856 and #1686 by:

* Providing configuration to the backend authenticator to properly evaluate SARs

PR #1695, opened 2019-07-11
Bug 1729297: Add authentication config to properly check SAR for addi…

jcantrill:
…tional roles

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1729297 for #1691

### DIFF
--- a/elasticsearch/sgconfig/sg_config.yml
+++ b/elasticsearch/sgconfig/sg_config.yml
@@ -13,6 +13,21 @@ searchguard:
         http_authenticator:
           challenge: false
           type: io.fabric8.elasticsearch.plugin.auth.OpenShiftTokenAuthentication
+          config:
+             subjectAccessReviews:
+               prometheus:
+                 verb: get
+                 resource: /metrics
+               jaeger:
+                 verb: get
+                 resource: jaeger
+                 namespace: set-in-run-script
+                 resourceAPIGroup: elasticsearch.jaegertracing.io
+               elasticsearch-operator:
+                 verb: '*'
+                 resource: '*'
+                 namespace: '*'
+                 resourceAPIGroup: logging.openshift.io
         authentication_backend:
           type: io.fabric8.elasticsearch.plugin.auth.OpenShiftTokenAuthentication
           config:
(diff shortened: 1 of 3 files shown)


richm:
should this be marked as wip because of the inclusion of the searchguard zip file?

test/verify-es-metrics-access.sh
```diff
@@ -0,0 +1,67 @@
...
+fi
+
+service_ip=$(oc -n ${LOGGING_NS} get service elasticsearch-metrics -o jsonpath={.spec.clusterIP})
+
+os::log::info Checking ${UNAUTHORIZED_SA} ability to read metrics through metrics service
+os::cmd::expect_success_and_text "curl -kv https://${service_ip}:60000/_prometheus/metrics -H\"Authorization: Bearer $(oc serviceaccounts get-token $UNAUTHORIZED_SA)\" -w '%{response_code}\n'" '403$'
+
+os::log::info Checking ${AUTHORIZED_SA} ability to read metrics
+if os::cmd::expect_success_and_text "curl -kv https://${service_ip}:60000/_prometheus/metrics -H\"Authorization: Bearer $(oc serviceaccounts get-token $AUTHORIZED_SA)\" -w '%{response_code}\n'" '200$' ; then
+    os::log::info "Received data from metrics endpoint"
```
lukas-vlcek: May be we could investigate the response body and do some [simple regex check that the data is in expected format](https://github.com/vvanholl/elasticsearch-prometheus-exporter/blob/master/src/test/resources/rest-api-spec/test/resthandler/20_00_metrics.yml#L18-L23) so that we know the response is not empty for example (despite 200 code which actually comes from the proxy and not ES directly).

elasticsearch/sgconfig/sg_config.yml
```diff
@@ -13,6 +13,21 @@ searchguard:
...
+             subjectAccessReviews:
+               prometheus:
+                 verb: get
+                 resource: /metrics
+               jaeger:
+                 verb: get
+                 resource: jaeger
+                 namespace: set-in-run-script
+                 resourceAPIGroup: elasticsearch.jaegertracing.io
+               elasticsearch-operator:
```
ewolinetz: shouldn't this already be in here? is a rebase needed?

jcantrill: It exists for the authentication_backend but not the http_authenticator.  The class is both but is instantiated multiple times

PR #1702, opened 2019-07-19
Bug 1715370: Some orphaned logs sent to .operations.* index

nhosoi:
by mistake.

Adding a check to fluentd/run.sh if bearer_token_file (/var/run/secrets/kubernetes.io/serviceaccount/token)
for the kubernetes api server exists and is not empty.


nhosoi:
> Note that for the bugzilla checking bot, the bugzilla format for the title should be `Bug xxx:`

Fixed.  Thanks!

PR #1704, opened 2019-07-23
imagePullPolicy Always - ensure admin user

richm:
When running `hack/get-cluster-run-tests.sh` - set env var
`LOGGING_IMAGE_PULL_POLICY=Always` in order to make sure you
are using your latest images.
Or add this to your `~/.config/get-cluster-run-tests.sh`

Also, it is possible to do an `oc login` to another cluster with
your newly minted `kubeconfig` and change the current context.
This fix ensures that you are using an `admin` context when the
script needs it.

### DIFF
--- a/hack/build-images.sh
+++ b/hack/build-images.sh
@@ -173,7 +173,7 @@ function login_to_registry() {
     fi
     if [ -z "$token" ] ; then
       echo ERROR: could not determine token to use to login to "$1"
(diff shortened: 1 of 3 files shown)


richm:
@VladMasarik this should also fix the strange issue you ran into

VladMasarik:
Nice, I will try it out soon.

PR #1707, opened 2019-07-24
upgrade - add support for upgrading a deployment

richm:
In the CI environment, `hack/deploy-logging.sh` will be first
run with `LOGGING_IMAGE_STREAM=stable-initial` which will use
the latest built images for the branch.  That is, this will
deploy logging version `N`.  Then, the deploy will be run
again with `LOGGING_DEPLOY_MODE=upgrade`, which will change
the logging images to use `LOGGING_IMAGE_STREAM=stable` which are
the images just built in the pipeline for the PR (the proposed
`N+1` version).  Then, the full CI tests will be run.
In a dev environment, if you have built and pushed logging images,
you can achieve the same thing by first deploying with
`USE_CUSTOM_IMAGES=true OPENSHIFT_BUILD_NAMESPACE=does-not-exist`
which will use the latest pushed images from
`registry.svc.ci.openshift.org/origin/4.2`.  Then, run deploy
again with `USE_CUSTOM_IMAGES=true LOGGING_DEPLOY_MODE=upgrade`
The script will look for the images built and pushed to the
`openshift` namespace and upgrade the logging components to
use those new images.

### DIFF
--- a/hack/deploy-logging.sh
+++ b/hack/deploy-logging.sh
@@ -85,6 +85,70 @@ switch_to_admin_user() {
     fi
 }
 
+# what numeric version does master correspond to?
+MASTER_VERSION=${MASTER_VERSION:-4.2}
+# what namespace to use for operator images?
+EXTERNAL_REGISTRY=${EXTERNAL_REGISTRY:-registry.svc.ci.openshift.org}
+EXT_REG_IMAGE_NS=${EXT_REG_IMAGE_NS:-origin}
+# for dev purposes, image builds will typically be pushed to this namespace
+OPENSHIFT_BUILD_NAMESPACE=${OPENSHIFT_BUILD_NAMESPACE:-openshift}
+
+construct_image_name() {
+    local component="$1"
+    local tagsuffix="${2:-latest}"
+    # if running in CI environment, IMAGE_FORMAT will look like this:
+    # IMAGE_FORMAT=registry.svc.ci.openshift.org/ci-op-xxx/stable:${component}
+    # stable is the imagestream containing the images built for this PR, or
+    # otherwise the most recent image
+    if [ -n "${IMAGE_FORMAT:-}" ] ; then
+        if [ -n "${LOGGING_IMAGE_STREAM:-}" ] ; then
+            local match=/stable:
+            local replace="/${LOGGING_IMAGE_STREAM}:"
+            IMAGE_FORMAT=${IMAGE_FORMAT/$match/$replace}
+        fi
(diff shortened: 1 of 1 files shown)


richm:
@jcantrill any idea why the unit test image build is failing?

richm:
finally passed - please review

hack/deploy-logging.sh
```diff
@@ -127,31 +191,18 @@ deploy_logging_using_clo_make() {
...
-    if [ -n "${IMAGE_FORMAT:-}" ] ; then
-        EO_IMAGE=$( echo "$IMAGE_FORMAT" | sed 's/\${component}/elasticsearch-operator/' )
-        CLO_IMAGE=$( echo "$IMAGE_FORMAT" | sed 's/\${component}/cluster-logging-operator/' )
-    else
-        EO_IMAGE=$EXTERNAL_REGISTRY/$EXT_REG_IMAGE_NS/$MASTER_VERSION:elasticsearch-operator
-        CLO_IMAGE=$EXTERNAL_REGISTRY/$EXT_REG_IMAGE_NS/$MASTER_VERSION:cluster-logging-operator
-    fi
-
+    EO_IMAGE=$( construct_image_name elasticsearch-operator master )
+    CLO_IMAGE=$( construct_image_name cluster-logging-operator master )
```
nhosoi: tagsuffix="master" ?

richm: good catch

richm:
this time, it looks like something killed zzz-rsyslog.sh in the middle of the last test - there is no error message, no stack trace, nothing - just straight to the cleanup - maybe the test ran too long and was killed by some sort of ci test monitor?
/test full-integ-aws

nhosoi:
> this time, it looks like something killed zzz-rsyslog.sh in the middle of the last test - there is no error message, no stack trace, nothing - just straight to the cleanup - maybe the test ran too long and was killed by some sort of ci test monitor?
> /test full-integ-aws

Yes, I was also looking it and had the same impression.  Should we split the rsyslog tests into multiple files?

richm:
> Yes, I was also looking it and had the same impression. Should we split the rsyslog tests into multiple files?

If this happens again, we may have to turn on full `set -xv` level tracing . . .

richm:
this time - multi-tenancy failed, and zzz-rsyslog passed :-(

/retest

richm:
rsyslog test is broken - needs https://github.com/openshift/origin-aggregated-logging/pull/1719 before this PR will pass

PR #1708, opened 2019-07-25
LOG-425. Add run for testing

jcantrill:
@richm Added test script per suggestion so that "build" is separate from "test" and it can be reflected as pass/fail in the list of PR jobs


richm:
Looks like you picked up some vim .swp files in your last commit . . .

jcantrill:
> Looks like you picked up some vim .swp files in your last commit . . .

removed

PR #1712, opened 2019-07-29
Bug 1734080: Add jaeger index templates

pavolloffay:

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1734080
Jaeger jira: https://issues.jboss.org/browse/TRACING-629


richm:
nack - new fields should be added by submitting a PR against https://github.com/ViaQ/elasticsearch-templates/ - you will want to add a new namespace to https://github.com/ViaQ/elasticsearch-templates/tree/master/namespaces - then the logging team will generate the index templates and patterns from the namespace file, cut a new release, and push it into all of the appropriate branches of origin-aggregated-logging

pavolloffay:
@richm do you mean I should translate `jaeger-span.json` and `jaeger-service.json` which are in this PR  to `jaeger.yaml` and add it to https://github.com/ViaQ/elasticsearch-templates/tree/master/namespaces? 

We have to make sure that the index template matches our upstream versions in https://github.com/jaegertracing/jaeger/tree/master/plugin/storage/es/mappings. Not sure if translating them to your format and then back will be easy to maintain.

pavolloffay:
Does that yaml format support `dynamic_templates` or `_default_` https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/es/mappings/jaeger-span.json#L15?

richm:
> @richm do you mean I should translate `jaeger-span.json` and `jaeger-service.json` which are in this PR to `jaeger.yaml` and add it to https://github.com/ViaQ/elasticsearch-templates/tree/master/namespaces?

yes
 
> We have to make sure that the index template matches our upstream versions in https://github.com/jaegertracing/jaeger/tree/master/plugin/storage/es/mappings. Not sure if translating them to your format and then back will be easy to maintain.

The viaq upstream has the logic which knows how to generate the proper elasticsearch index templates and kibana index patterns for all supported ES versions.  If your upstream has that same ability, to generate all of those files for all supported ES versions, then we could accept the .json files directly into origin-aggregated-logging.  Including @lukas-vlcek

richm:
> Does that yaml format support `dynamic_templates` or `_default_` https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/es/mappings/jaeger-span.json#L15?

I think so, but if not, we should add it - @lukas-vlcek

pavolloffay:
> If your upstream has that same ability, to generate all of those files for all supported ES versions, then we could accept the .json files directly into origin-aggregated-logging. 

The index mapping in this PR is tested against ES 5.x and 6.x we are planning to have a version for 7.x very soon.

richm:
hmm - on the other hand, maybe we don't need the jaeger information in viaq, so I think we can accept this PR as-is.  Does jaeger use Kibana?  If so, you'll also need index pattern files like https://github.com/openshift/origin-aggregated-logging/tree/master/elasticsearch/index_patterns

pavolloffay:
Some community members do use jaeger with Kibana, however it's not part of our product, we do not deploy Kibana in service mesh product or stadalone so it's not needed in this image.

PR #1717, opened 2019-08-01
Bug 1735530: rsyslog undefined field handling - fluentd parity

nhosoi:
Fixing the cases:
 - CDM_UNDEFINED_TO_STRING=true && CDM_USE_UNDEFINED=true (BAD_1 in bz)
 - CDM_USE_UNDEFINED=false && CDM_UNDEFINED_MAX_NUM_FIELDS=<NUM> (BAD_2 in bz)

### DIFF
--- a/rsyslog/undefined_field/undefined_field.go
+++ b/rsyslog/undefined_field/undefined_field.go
@@ -143,7 +143,7 @@ func onInit() {
 	} else {
 		checkMaxNumFields = true
 	}
-	if cfg.UseUndefined {
+	if cfg.UseUndefined || checkMaxNumFields {
 		tmpDefault := strings.Split(cfg.DefaultKeepFields, ",")
 		tmpExtra := strings.Split(cfg.ExtraKeepFields, ",")
 		keepFields = make(map[string]string)


rsyslog/undefined_field/undefined_field.go
```diff
@@ -214,7 +214,7 @@ func processUndefinedAndMaxNumFields(input map[string]interface{}) (string, map[
 		for field := range undefMap {
 			delete(input, field)
 		}
-		if (checkMaxNumFields && count < 0) {
+		if (checkMaxNumFields && count < 0) || (cfg.UseUndefined && cfg.UndefinedToString) {
```
richm: This isn't the code that does the undefined-to-string processing for individual fields - it does the undefined-to-string processing for all of the undefined fields as a single JSON string of the undefined map value.

nhosoi: I'm trying to solve this problem [1].  I think this configuration is supposed to do what you described...  Please note that _CDM_UNDEFINED_TO_STRING=true && CDM_USE_UNDEFINED=false_ case is working just fine [2].   Well, after my opening bz1735530, I found what I thought broken were all working as expected, and rather some other cases were not... :p

https://bugzilla.redhat.com/show_bug.cgi?id=1735530#c1
[1] - BAD_1) CDM_UNDEFINED_TO_STRING with CDM_USE_UNDEFINED=true
[2] - GOOD_3) CDM_UNDEFINED_TO_STRING

richm: ok - I updated the bz

nhosoi: Thanks, @richm.  I reverted the change.

PR #1720, opened 2019-08-07
Bug 1730945: Rsyslog pod always restart when set `ES_REBIND_INTERVAL…

nhosoi:
…` to a small value.

omelasticsearch - force to close connection by setting CURLOPT_FORBID_REUSE to 1
                  when the count of operation reaches rebindinterval.

### DIFF
--- a/rsyslog/vendored_src/rsyslog/rsyslog/plugins/omelasticsearch/omelasticsearch.c
+++ b/rsyslog/vendored_src/rsyslog/rsyslog/plugins/omelasticsearch/omelasticsearch.c
@@ -1482,6 +1482,12 @@ curlPost(wrkrInstanceData_t *pWrkrData, uchar *message, int msglen, uchar **tpls
 		/* by default, reuse existing connections */
 		curl_easy_setopt(curl, CURLOPT_FRESH_CONNECT, 0);
 	}
+	if ((pWrkrData->pData->rebindInterval > -1) &&
+		(pWrkrData->nOperations == pWrkrData->pData->rebindInterval)) {
+		curl_easy_setopt(curl, CURLOPT_FORBID_REUSE, 1);
+	} else {
+		curl_easy_setopt(curl, CURLOPT_FORBID_REUSE, 0);
+	}
 
 	if(pWrkrData->pData->numServers > 1) {
 		/* needs to be called to support ES HA feature */


rsyslog/vendored_src/rsyslog/rsyslog/plugins/omelasticsearch/omelasticsearch.c
```diff
@@ -1482,6 +1482,12 @@ curlPost(wrkrInstanceData_t *pWrkrData, uchar *message, int msglen, uchar **tpls
 		/* by default, reuse existing connections */
 		curl_easy_setopt(curl, CURLOPT_FRESH_CONNECT, 0);
 	}
+	if ((pWrkrData->pData->rebindInterval > -1) &&
+		(pWrkrData->nOperations == pWrkrData->pData->rebindInterval)) {
+		curl_easy_setopt(curl, CURLOPT_FORBID_REUSE, 1);
+	} else {
+		curl_easy_setopt(curl, CURLOPT_FORBID_REUSE, 0);
+	}
```
richm: Is this just for testing purposes, or does this really solve the problem?
Why not put the `curl_easy_setopt(curl, CURLOPT_FORBID_REUSE, 1);` setting just after the `curl_easy_setopt(curl, CURLOPT_FRESH_CONNECT, 1);` setting and the `curl_easy_setopt(curl, CURLOPT_FORBID_REUSE, 0);` just after the `curl_easy_setopt(curl, CURLOPT_FRESH_CONNECT, 0);` setting?

richm:
this had the rsyslog es auth flake:
```
{"@timestamp":"2019-08-07T02:08:42.173157+00:00", "host":"rsyslog-rxhkf", "level":"warning", "facility":"syslog", "syslog-tag":"rsyslogd-3000:", "source":"rsyslogd-3000", "procid":"-", "msgid":"-", "message":"omelasticsearch: could not parse JSON result [v8.37.0-9.el7]"}
```
```
[2019-08-07T02:08:30,467][WARN ][c.f.s.a.BackendRegistry  ] Authentication finally failed for null
```
/retest

rsyslog/vendored_src/rsyslog/rsyslog/plugins/omelasticsearch/omelasticsearch.c
nhosoi: ???  The patch works as expected, I believe.  Following is a snippet of Curl_done.  If reuse_forbid is set to true, Curl_disconnect is called.  So, I think we have to set it when we want to close the connection...

5648   if((data->set.reuse_forbid && !(conn->ntlm.state == NTLMSTATE_TYPE2 ||
5649                                   conn->proxyntlm.state == NTLMSTATE_TYPE2))
5650      || conn->bits.close || premature ||
5651      (-1 == conn->connection_id)) {
5652     CURLcode res2 = Curl_disconnect(conn, premature); /* close connection */
5653 
5654     /* If we had an error already, make sure we return that one. But
5655        if we got a new error, return that. */
5656     if(!result && res2)
5657       result = res2;
5658   }

nhosoi:
> this had the rsyslog es auth flake:
Is that the one you mentioned that when a pod was started, the secret was not ready?

rsyslog/vendored_src/rsyslog/rsyslog/plugins/omelasticsearch/omelasticsearch.c
richm: ok - please move `curl_easy_setopt(curl, CURLOPT_FORBID_REUSE, 1);` into the `if` clause with `curl_easy_setopt(curl, CURLOPT_FRESH_CONNECT, 1);`, and move `curl_easy_setopt(curl, CURLOPT_FORBID_REUSE, 0);` to the `else` clause with `curl_easy_setopt(curl, CURLOPT_FRESH_CONNECT, 0);`

richm:
> > this had the rsyslog es auth flake:
> 
> Is that the one you mentioned that when a pod was started, the secret was not ready?

Yes

rsyslog/vendored_src/rsyslog/rsyslog/plugins/omelasticsearch/omelasticsearch.c
nhosoi: It may not be a big deal, but the closing connection timing would be changed by your proposal a little bit.  I wanted to close it before opening a new one in the previous cycle.  That keeps the count of ESTABLISHED connection always 1.  But if we move curl_easy_setopt(curl, CURLOPT_FORBID_REUSE, 1) next to curl_easy_setopt(curl, CURLOPT_FRESH_CONNECT, 1), you could have 2 connections for a short time.  If it's ok, I'll update the PR.

```
$ while true; do lsof -i -P -n |grep rsyslog; sleep 5; echo ""; done
rsyslogd  28431 nhosoi   58u  IPv4 3891135      0t0  TCP 10.195.132.86:38212->34.209.7.166:443 (ESTABLISHED)

rsyslogd  28431 nhosoi   58u  IPv4 3891135      0t0  TCP 10.195.132.86:38212->34.209.7.166:443 (ESTABLISHED)
rsyslogd  28431 nhosoi   60u  IPv4 3893709      0t0  TCP 10.195.132.86:38224->34.209.7.166:443 (ESTABLISHED)
```

richm: hmm - ok - I think it is better to close before opening a new one - if your patch already does that, then let's leave it as it is

richm:
Since this affects the core rsyslog code, we need to do the following:
file a PR upstream against rsyslog
file a BZ against downstream RHEL8 rsyslog
the BZ does not affect RHEL7

richm:
/lgtm

We'll have to remember to "revert" this patch once the patch lands in RHEL8 and we resync

nhosoi:
> We'll have to remember to "revert" this patch once the patch lands in RHEL8 and we resync

When we migrate to RHEL8, what happens to rsyslog/vendored_src?  Do we delete the directory and just use RHEL8 rsyslog packages, replace the current vendored_src/* with the ones from RHEL8, or merge the current branch with the RHEL8 branch?

If we do merge, shall I add the change as a patch (adding bz1730945.patch to the rsyslog dir and the spec file) to make it easier to drop when switching to RHEL8?

richm:
> > We'll have to remember to "revert" this patch once the patch lands in RHEL8 and we resync
> 
> When we migrate to RHEL8, what happens to rsyslog/vendored_src? Do we delete the directory and just use RHEL8 rsyslog packages, replace the current vendored_src/* with the ones from RHEL8, or merge the current branch with the RHEL8 branch?

If by "migrate to RHEL8" you mean "use RHEL8 as the base os for the containers instead of RHEL7", then it depends on what version of RHEL8.  I don't believe all of the functionality we require will be in rsyslog until RHEL 8.2.  Until then, we have to vendor the source - https://github.com/openshift/origin-aggregated-logging/blob/master/HACKING.md#updating-sources-for-rsyslog and https://github.com/openshift/origin-aggregated-logging/blob/master/hack/update-rsyslog-vendor-src.sh

> 
> If we do merge, shall I add the change as a patch (adding bz1730945.patch to the rsyslog dir and the spec file) to make it easier to drop when switching to RHEL8?

As long as running hack/update-rsyslog-vendor-src.sh doesn't wipe out the patch.  You can also commit the patch as a regular commit to the master branch.  When we rebase rsyslog, we should get a git error like a merge error or a rebase conflict, then we'll know we have to resolve or remove the patch we added.

nhosoi:
> Since this affects the core rsyslog code, we need to do the following:
> file a PR upstream against rsyslog

Filed an upstream PR:
https://github.com/rsyslog/rsyslog/pull/3801

PR #1721, opened 2019-08-08
Bug 1735530: rsyslog undefined field handling - fluentd parity

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1735530
We have to do undefined field checking in all cases, not only
when `CDM_USE_UNDEFINED=true` or when checking for max num fields.
I also refactored the code somewhat so that we could easily test
combinations of `CDM_USE_UNDEFINED=true` and max num field checking
with `CDM_UNDEFINED_TO_STRING=true` and `CDM_UNDEFINED_DOT_REPLACE_CHAR`
and empty field processing.

### DIFF
--- a/rsyslog/undefined_field/undefined_field.go
+++ b/rsyslog/undefined_field/undefined_field.go
@@ -143,16 +143,14 @@ func onInit() {
 	} else {
 		checkMaxNumFields = true
 	}
-	if cfg.UseUndefined || checkMaxNumFields {
-		tmpDefault := strings.Split(cfg.DefaultKeepFields, ",")
-		tmpExtra := strings.Split(cfg.ExtraKeepFields, ",")
-		keepFields = make(map[string]string)
-		for _, str := range tmpDefault {
-			keepFields[str] = str
-		}
-		for _, str := range tmpExtra {
-			keepFields[str] = str
-		}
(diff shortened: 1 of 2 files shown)


richm:
please review

rsyslog/undefined_field/undefined_field.go
```diff
@@ -389,6 +387,51 @@ func processUndefinedAndEmpty(input map[string]interface{}, hasDefinedFields, ha
...
+		}
+		fmt.Println(noChanges)
+		return changes
+	}
+	if len(undefString) > 0 || undefMap != nil {
+		changes = true
+	}
+	if undefMap != nil {
+		if processUndefinedAndEmpty(undefMap, false, true) {
+			changes = true
```
nhosoi: No harm at all, but changes is already set to true.  It confused me for a moment if there's a case to reset changes to false by any chance... :)  I guess no need to check the return value from processUndefinedAndEmpty...?  (same for line 416)

richm: ok - yes, I changed 410 to ignore the return value since it is redundant

but 416 (now line 414) - I think I still need to set `changes = true` at line 415 - I think `changes` can be false at line 414

nhosoi: Ah, you are right!  Sorry for the noise...

/lgtm

PR #1725, opened 2019-08-14
read_from_head=false to speed up tests, cut down flakes

richm:
Try to speed up tests, and cut down on flakes, by using
JSON_FILE_READ_FROM_HEAD=false JOURNAL_READ_FROM_HEAD=false
some tests erase the pos files, which means by default fluentd
has to read from the beginning of every single file for those
tests - some of our times for test records to show up in
Elasticsearch are now too long - maybe this will help cut
down those times.

### DIFF
--- a/test/eventrouter.sh
+++ b/test/eventrouter.sh
@@ -17,14 +17,9 @@ cleanup() {
     local return_code="$?"
     set +e
     fpod=$( get_running_pod fluentd )
-    oc label node --all logging-infra-fluentd- 2>&1 | artifact_out
-    os::cmd::try_until_text "oc get daemonset logging-fluentd -o jsonpath='{ .status.numberReady }'" "0" $FLUENTD_WAIT_TIME
-    if [ -n "${fpod:-}" ] ; then
-        os::cmd::try_until_failure "oc get pod $fpod > /dev/null 2>&1" $FLUENTD_WAIT_TIME
-    fi
-    oc set env ds/logging-fluentd $muxmode
-    oc label node --all logging-infra-fluentd=true 2>&1 | artifact_out
-    os::cmd::try_until_text "oc get pods -l component=fluentd" "^logging-fluentd-.* Running " $FLUENTD_WAIT_TIME
+    stop_fluentd $fpod 2>&1 | artifact_out
+    oc set env ds/logging-fluentd $muxmode TRANSFORM_EVENTS- 2>&1 | artifact_out
+    start_fluentd true 2>&1 | artifact_out
     # this will call declare_test_end, suite_end, etc.
     os::test::junit::reconcile_output
     exit $return_code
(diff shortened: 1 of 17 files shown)


richm:
so one thing I notice from problematic tests:
* they use mux
* they repeatedly test, reconfigure, and restart fluentd and/or mux
* the response time is very long and increases dramatically

For example, in the viaq test - the first time it generates an app log and queries it in ES takes 40(!) seconds.  The next time is 80, then 120, then boom.  It doesn't seem as if `read_from_head=false` is helping.  Also, when the test fails, the log says that the record could not be found in the container log file.  One thing we could do is examine the pipeline_metadata for the records for the tests that are slow but still pass, to see when fluentd sees the log vs. when mux sees the log.

test/eventrouter.sh
```diff
@@ -75,20 +68,17 @@ else
...
         prev_event_count=$( curl_es $esopssvc /.operations.*/_count?q=kubernetes.event.verb:* | get_count_from_json )
 
         # MUX_CLIENT_MODE: minimal
-        oc label node --all logging-infra-fluentd- 2>&1 | artifact_out
-        os::cmd::try_until_text "oc get daemonset logging-fluentd -o jsonpath='{ .status.numberReady }'" "0" $FLUENTD_WAIT_TIME
+        stop_fluentd 2>&1 | artifact_out
         oc set env ds/logging-fluentd MUX_CLIENT_MODE=minimal 2>&1 | artifact_out
-        oc label node --all logging-infra-fluentd=true 2>&1 | artifact_out
-        os::cmd::try_until_text "oc get pods -l component=fluentd" "^logging-fluentd-.* Running " $FLUENTD_WAIT_TIME
+        start_fluentd true 2>&1 | artifact_out
```
nhosoi: Considering the CI test failure, we may not want to have "true" here?  (maybe line 73, too?)

richm: I'm still trying to figure out all of the different failures - still working my way through the json-file test errors - not sure why eventrouter works in the json-file test and fails in the journald test - in general using `start_fluentd true` should be safe, and should be the best way to start fluentd for a test

richm: @nhosoi I have a question about this commit https://github.com/openshift/origin-aggregated-logging/commit/6ff3498c91b53c59937864bf335ea550b573a184 - it added support for remote-syslog test - but it also changed forward-fluentd.sh so that it _does not_ test the two output case: https://github.com/openshift/origin-aggregated-logging/commit/6ff3498c91b53c59937864bf335ea550b573a184#diff-8e2d11cf561cc73246c98381a5ea43f9R240
I don't suppose you remember why that was added . . .

richm: I think I managed to fix it - needed to add the genid filter for the two output case

richm:
passed - please review

test/eventrouter.sh
nhosoi: Thanks, Rich.  Yeah, that's what I wanted to do... (adding to return OK... ):

PR #1726, opened 2019-08-15
Bug 1741585: remote-syslog error when the record has level=unknown

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1741585
We recently switched most container logs to use log level=unknown.  This causes problems for the remote-syslog plugin as `unknown` isn't in the list of levels/severities:
https://github.com/eric/syslog_protocol/blob/master/lib/syslog_protocol/common.rb#L58

```
2019-08-15 01:00:20 +0000 [warn]: emit transaction failed: error_class=ArgumentError error="'unknown' is not a designated severity" location="/opt/app-root/src/gems/syslog_protocol-0.9.2/lib/syslog_protocol/packet.rb:72:in `severity='" tag="output_ops_tag"
2019-08-15 01:00:20 +0000 [warn]: /opt/app-root/src/gems/syslog_protocol-0.9.2/lib/syslog_protocol/packet.rb:72:in `severity='
```

The fix is to use the configured `@severity` if level is unknown, which is
the same as what happens if level is not set.

### DIFF
--- a/fluentd/out_syslog.rb
+++ b/fluentd/out_syslog.rb
@@ -70,7 +70,14 @@ module Fluent
       es.each {|time,record|
         if @use_record
           @packet.hostname = record['hostname'] || hostname
-          @packet.severity = (record['level'].eql? 'warning')? 'warn' : record['level'] || @severity
+          case record['level']
+          when 'warning'
+            @packet.severity = 'warn'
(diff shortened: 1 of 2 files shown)


richm:
please review

PR #1731, opened 2019-08-19
Update to master

jcantrill:
FF to master for 4.3 work

cc @openshift/aosqe-intservice


richm:
I think the test failure is not a flake - it is consistently failing in the rsyslog test - perhaps there is a missing commit?

PR #1739, opened 2019-08-28
minor change to trigger merge and rebuild

richm:


richm:
grr - persistent problem with eventrouter . . .

PR #1741, opened 2019-08-31
disable CVO and OLM for testing

richm:
For testing, we have to disable CVO, then OLM, then we can disable/enable CLO and EO as needed.


richm:
I don't know if anyone is around - please review and `/lgtm` so we can unblock CI

richm:
and - if anyone knows how to disable OLM, please let me know - https://github.com/openshift/origin-aggregated-logging/blob/master/hack/test-logging.sh#L78 doesn't work now, something restarts OLM - @shawn-hurley

jcantrill:
> and - if anyone knows how to disable OLM, please let me know - https://github.com/openshift/origin-aggregated-logging/blob/master/hack/test-logging.sh#L78 doesn't work now, something restarts OLM - @shawn-hurley

To follow up on my slack message:  Disabling the cluster-version-operator will disable changes to OLM

richm:
please review

PR #1747, opened 2019-09-10
Bug 1739558: rsyslog underperforming compared to fluentd

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1739558
imfile: followrename - continue to follow renamed files

Some applications do rotation by renaming the file, then continuing to
write to it for a short time, before recreating the original file. See
IMFILE for the gory details.

Introduce a new imfile parameter `followrename` (boolean) to handle this
case.  If this parameter is set, when inotify detects a rename event
(IN_MOVED_FROM), it will keep reading from the file until it sees that a
new file has been created with the same name.  It will then create a new
state file with state saved from the old state file e.g. prevLineSegment,
prevMsgSegment, and unlink the old file from the active list. the
moved_from file from the active list, and create a new active list entry
for the moved_to file, preserving the state file from the moved_from file. 
This also does some code refactoring in order to reuse some of the
functions for adding a new act_obj for the renamed file. Use
STREAMTYPE_FILE_SINGLE for monitored files instead of 
STREAMTYPE_FILE_MONITOR so that the stream module won't attempt to reopen
the file if it detects the filename has a different inode number.

Enabling valgrind testing on imfile revealed a memory leak which is fixed
in this PR.

Changed some "ignored" error messages to use LOG_INFO - if they are really
not that serious, they should not be printed by default.

Fixed some spelling errors.

(cherry picked from commit 678b7dc4fb9a74309781f260bb8ef8134ee3e6d2)

merge in latest rsyslog 8.1.z

Add a performance test
that runs a pod using logging-load-driver to write and verify logs for
records per second, skipped records, and duplicate records.

### DIFF
--- a/hack/logging-load-driver/README.md
+++ b/hack/logging-load-driver/README.md
@@ -1,0 +1,1 @@
+Copied from https://github.com/ViaQ/logging-load-driver
--- a/hack/logging-load-driver/loader
+++ b/hack/logging-load-driver/loader
@@ -1,0 +1,267 @@
+#!/usr/bin/env python
+#
+# ViaQ logging load generator
+#
+# Copyright 2018 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from __future__ import print_function
+
+import sys
+import time
+import uuid
+import random
+import string
+import logging
+import logging.handlers
+import argparse
+
+
+DEFAULT_STDDEV = 32
+
+
+def msgsizgen_fixed(stats, mean, stddev = None):
+    '''Generate a series of "fixed" sizes, always yielding the given
+       "mean" as the value.
+
+       The stddev value is always ignored.
+    '''
+    while True:
+        yield stats.nextseq(mean)
+
+
+def msgsizgen_gaussian(stats, mean, stddev = DEFAULT_STDDEV):
+    '''Generate a series of random numbers with a Guassian distribution,
+       where the mean and standard deviation are given arguments.
+
+       The mean is a required parameter, while the stddev will default to
+       32.
+
+       The generator will yield numbers continuously.
+    '''
+    randomizer = random.Random()
+    while True:
+        yield stats.nextseq(int(round(randomizer.gauss(mean, stddev))))
+
+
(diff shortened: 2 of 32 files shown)


richm:
doesn't work - fails when crio/kubelet compresses rotated files :-(

nhosoi:
> doesn't work - fails when crio/kubelet compresses rotated files :-(

That means rsyslog could not catch up the crio log writing speed?  When rotated, one log is just renamed to `logname`.log.`date`.  When the next log file is rotated, the `logname`.log.`date` would be compressed...

richm:
> > doesn't work - fails when crio/kubelet compresses rotated files :-(
> 
> That means rsyslog could not catch up the crio log writing speed? When rotated, one log is just renamed to `logname`.log.`date`. When the next log file is rotated, the `logname`.log.`date` would be compressed...

I don't think so - the symptom I see is that rsyslog skips a lot of records when this happens.  I think it has to do with the way rsyslog stores state files using the inode number, and the fact that when rsyslog rotates the second time, it reuses the inode number of the original file.  What may be happening is that when rsyslog rotates the file a second time, it somehow reuses the existing state file for that inode number, and skips over the records it thinks were already read.

portante:
Not sure this helps, but I used https://sequencediagram.org/ to create two sequence diagrams that *might* explain the good vs bad scenarios for rsyslog (cut and paste them into the dialog box to auto generate the diagrams):

**Good**:
```
title rsyslog file rotation (good)

cri-o->log-file-0:Write 0
cri-o->log-file-0:Write 1
log-file-0->rsyslog:Read
cri-o->log-file-0:Write 2
cri-o->log-file-0:Write 3
cri-o->log-file-0:Write 4
kubelet->log-file-0:Renamed to log-file-0.YYYY-MM-DD
kubelet->cri-o:kill -HUP
cri-o->log-file-0:Close
cri-o->log-file-1:Create
cri-o->log-file-1:Write
kernel->rsyslog:inotify event for rename of log-file-0
log-file-0->rsyslog:Read
rsyslog->log-file-0:Close
kernel->rsyslog:inotify event for create of log-file-1
rsyslog->log-file-1:Open
log-file-1->rsyslog:Read
```

**Bad**:
```
title rsyslog file rotation (bad)

cri-o->log-file-0:Write 0
cri-o->log-file-0:Write 1
log-file-0->rsyslog:Read
cri-o->log-file-0:Write 2
cri-o->log-file-0:Write 3
kubelet->log-file-0:Renamed to log-file-0.YYYY-MM-DD
kernel->rsyslog:inotify event for rename of log-file-0
log-file-0->rsyslog:Read
cri-o->log-file-0:Write 4
kubelet->cri-o:kill -HUP
rsyslog->log-file-0:Close
cri-o->log-file-0:Close
cri-o->log-file-1:Create
kernel->rsyslog:inotify event for create of log-file-1
rsyslog->log-file-1:Open
cri-o->log-file-1:Write
log-file-1->rsyslog:Read
```

The "Bad" scenario is bad because (theorizing) the `rsyslog` receives the `inotify` event for the log file rename and reads from the rotated log file before `cri-o` writes its last buffer to the rotated file.

`rsyslog` would have to keep the rotated log file open for a period of time long enough for `cri-o` to receive its `HUP` signal, writes its last buffer, and close the file.

PR #1753, opened 2019-09-13
bug 1752162: Drop rsyslog support from 4.1

richm:
We never supported rsyslog in 4.1, and it causes test flakes trying
to merge 4.1 PRs.  Would rather just drop it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ build-images:
 .PHONY: build-images
 
 test:
-	EXCLUDE_SUITE="upgrade" hack/testing/entrypoint.sh
+	EXCLUDE_SUITE="upgrade|zzz-rsyslog" hack/testing/entrypoint.sh
 .PHONY: test
 
 test-upgrade:


richm:
hmm - I thought I could get around the `Needs approved, bugzilla/valid-bug, cherry-pick-approved` workflow since this isn't a direct cherrypick.

@bparees I would like to get an exception, and merge this PR ASAP.  We didn't support rsyslog in 4.1, and the rsyslog test is a frequent source of test flakes, which delays merging important 4.1 PRs.  Or, if someone can just directly `Merge pull request`, it would help get 4.1 PRs merged and 4.1 bugs fixed faster.

bparees:
this is a makefile only change, does not introduce any 4.1.z release stability concerns, so i'm circumventing the merge process (forcing @richm to create a dummy BZ just to make the bot happy seems like a waste of time) and merging this manually.  We are within the merge window for 4.1.z.

@knobunc since i think you're patch manager this week, fyi.

@richm I am going to wait until I see full green on all the jobs and then i will manually merge.

bparees:
> hmm - I thought I could get around the Needs approved, bugzilla/valid-bug, cherry-pick-approved workflow since this isn't a direct cherrypick.

fyi @richm you should definitely try to circumvent the process.  Although it's called "cherry-pick-approved" it's really more like "z-stream-approved" (ie someone has agreed this is ok to merge into the z-stream because we've verified the change in master already and the change is needed/low-risk/etc, and we are in a period where we are accepting changes into the z-stream).  It really doesn't have to do w/ whether it's a direct cherry-pick or not (though most things going into the z-streams will be cherry-picks)

bparees:
ok, i went ahead and created the BZs to make the process happy.  @knobunc i'm going to apply the cherry-pick approved label myself since i'm not sure when you're next going to look at these PRs and i don't want to block the team over the weekend.  I believe it now meets all the requirements for backporting.

bparees:
@richm @jcantrill were you aware that e2e-aws is configured as "optional" on this repo?  so it will not block merges?  That might make sense since i don't think anything you merge here can actually affect e2e-aws(there are no logging tests in e2e-aws i'm guessing?), but then it would beg the question of why you're even running it on PRs to this repo....

richm:
> @richm @jcantrill were you aware that e2e-aws is configured as "optional" on this repo? so it will not block merges? That might make sense since i don't think anything you merge here can actually affect e2e-aws(there are no logging tests in e2e-aws i'm guessing?), but then it would beg the question of why you're even running it on PRs to this repo....

I think at some point we were under the impression that the e2e-aws test was required for all github.com/openshift/ repos.  Then, sometime later, it was blocking PRs because it was too flaky, so we made it optional.  I think based on what you've said recently, we can get rid of it altogether.

bparees:
> I think at some point we were under the impression that the e2e-aws test was required for all github.com/openshift/ repos. 

yeah, that message/requirement probably didn't take into account that some repos literally have no bearing on e2e-aws.

of course you can also add a test to e2e-aws that *does* deploy logging, in case you want some certainty that other people(ie platform, olm, etc) do not break logging.  But it would be tricky since getting e2e-aws to actually test *your* PRs would be difficult, as you well know (getting it to deploy logging from your PR, as part of the e2e-aws job).

PR #1757, opened 2019-09-19
curator/Dockerfile: allow latest python-elasticsearch

sosiouxme:
I rebuilt the downstream with a bumped epoch, hoping that it will fix our grading problem.
The version ceiling excludes it, though. We can just remove that now.

### DIFF
--- a/curator/Dockerfile
+++ b/curator/Dockerfile
@@ -23,7 +23,7 @@ RUN if [ -n "${LOCAL_REPO}" ] ; then \
     fi
 
 RUN yum install -y --setopt=skip_missing_names_on_install=False --setopt=tsflags=nodocs \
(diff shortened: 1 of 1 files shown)


curator/Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN if [ -n "${LOCAL_REPO}" ] ; then \
     fi
 
 RUN yum install -y --setopt=skip_missing_names_on_install=False --setopt=tsflags=nodocs \
-        elastic-curator-${CURATOR_VER} python2-ruamel-yaml "python-elasticsearch < 5.5" \
```
richm: This syntax `"python-elasticsearch < 5.5"` works with `yum install`?

sosiouxme: yep! has for months :) but we can get rid of it now

PR #1758, opened 2019-09-25
Bug 1753936: fluentd couldn't forward logs to rsyslog out:syslog: inv…

richm:
…alid facility value DEVICE

https://bugzilla.redhat.com/show_bug.cgi?id=1753936
Do not print the error at the `warn` level - use `debug` instead
This also re-enables the remote-syslog CI test, by rewriting to
use a custom fluentd pod syslog listener instead of using rsyslog
on the node.  It also gets rid of the mux stuff from the test.


richm:
multi-tenancy flake
/retest

PR #1761, opened 2019-09-30
Bug 1757180: ACL documents are _always_ reseeded

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1757180
Fix logic/syntax errors in script.
Use curl `--output /dev/null` to throw away the output we don't
care about rather than trying to parse it with `cut`.

Cause: There were some logic errors and a bash syntax error in
the seed acls script.

Consequence: ACL documents were always being reseeded.

Fix: Correct the logic and syntax errors in the script.  Use
curl `--output /dev/null` to throw away the output we don't
care about rather than trying to parse it with `cut`.

Result: ACL documents are only reseeded if they need to be, or if
user specifies `FORCE_SEED_ACL=true`.

### DIFF
--- a/elasticsearch/init/0010-seed-acl
+++ b/elasticsearch/init/0010-seed-acl
@@ -17,16 +17,16 @@
 
 source "logging"
 
-force=${FORCE_SEED_ACL:-''}
+force=${FORCE_SEED_ACL:-"true"}
 config=${ES_CONF}/elasticsearch.yml
 index=$(python -c "import yaml; print yaml.load(open('${config}'))['searchguard']['config_index_name']" | xargs -i sh -c "echo {}")
 
-if [ -n "${force}" ] ; then
+if [ "${force}" = "false" ] ; then
   docs=(roles rolesmapping actiongroups internalusers config)
   for d in $docs; do
-    resp=$(es_util --query=${index}/$d/0 -w %{response_code} | cut -d '}' -f2)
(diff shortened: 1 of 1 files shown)


richm:
image builds of fluentd are broken until https://github.com/openshift/origin-aggregated-logging/pull/1763 is merged

richm:
tests finally passed - please review @jcantrill I think this will still work as it does not - by default, reseed everything - only way to avoid that is to set `FORCE_SEED_ACL=false` - then it will not reseed unless there is a missing acl doc

richm:
please review

richm:
please review

richm:
centos is down?
```
[openshift/origin-logging-fluentd] Could not retrieve mirrorlist http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=os&infra=container error was
[openshift/origin-logging-fluentd] 14: curl#7 - "Failed to connect to 2001:4178:5:200::10: Network is unreachable"
```
/retest

richm:
hmm - maybe not - try again
/retest

richm:
install flake
/retest

richm:
install flake
/retest

richm:
not sure if this is a flake - failed to restart node during origin install - maybe need to increase default timeout?
/retest

jcantrill:
Merging as 3.11 ci is broken

PR #1762, opened 2019-10-01
Add filter for linux auditd logs and remove docker ones

JAORMX:
The linux audit logs parsing used to be exclusive to docker, this now changes the
input to fit more general audit logs from auditd.

This PR also removes the unused docker audit configurations and
related files.


fluentd/configs.d/openshift/input-pre-audit-log.conf
```diff
@@ -5,5 +5,7 @@
   path "#{ENV['AUDIT_FILE'] || '/var/log/audit/audit.log'}"
```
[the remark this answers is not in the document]
JAORMX: that's why I'm leaving this with the WIP prefix. But you're right; we definitely don't want this to be enabled by default. It's only a provisional measure for me to be able to test this in the meantime.

fluentd/configs.d/openshift/filter-pre-k8s-audit.conf
```diff
@@ -0,0 +1,6 @@
+<filter audit.log>
+  @type record_transformer
+  <record>
+    time ${time}
```
richm: I think you want the audit log parser to add either the `time` or the `@timestamp` field

JAORMX: isn't that what this configuration does?

the audit log parser is meant for auditd logs, not kubernetes ones :/

fluentd/configs.d/openshift/input-pre-k8s-audit-log.conf
```diff
@@ -0,0 +1,16 @@
...
+  @type tail
+  @id k8s-audit-input
+  @label @INGRESS
+  @log_level debug 
+  path "#{ENV['K8S_AUDIT_FILE'] || '/var/log/kube-apiserver/audit.log'}"
+  pos_file "#{ENV['K8S_AUDIT_POS_FILE'] || '/var/log/kube-apiserver/audit.log.pos'}"
+  tag audit.log
+  <parse>
+    @type json
+    time_key requestReceivedTimestamp
```
richm: this means the record will contain a field called `requestReceivedTimestamp`?  I think you will need a record transformer to rename this to `time` or `@timestamp` - looks like the format is already rfc3339

The problem with record_transformer plugins is that they are very, very slow and will cause performance problems - it will probably be necessary to extend the viaq data model filter plugin to make `src_time_name` is a list, then add `requestReceivedTimestamp` to that list

fluentd/configs.d/openshift/filter-pre-k8s-audit.conf
richm: I'm confused - are you using the tag `audit.log` for both auditd logs and kubernetes audit logs?

fluentd/configs.d/openshift/input-pre-k8s-audit-log.conf
```diff
@@ -0,0 +1,16 @@
+<source>
+  @type tail
+  @id k8s-audit-input
+  @label @INGRESS
+  @log_level debug 
+  path "#{ENV['K8S_AUDIT_FILE'] || '/var/log/kube-apiserver/audit.log'}"
+  pos_file "#{ENV['K8S_AUDIT_POS_FILE'] || '/var/log/kube-apiserver/audit.log.pos'}"
+  tag audit.log
```
richm: We're going to need a separate tag for this - I suggest `audit.k8s` - this also means changes elsewhere in the pipeline - we might be able to change many of the existing uses of `audit.log` to denote operations logs, to use `audit.**` instead which means "consider any log tagged with audit.** to be an operations log"

fluentd/configs.d/openshift/filter-pre-k8s-audit.conf
richm: ok - try this
```
   @type record_transformer
   enable_ruby true
   <record>
     time "${record['requestReceivedTimestamp'] || record['time']}"
   </record>
```
that way it should work with either kind of record tagged with `audit.log`

fluentd/parser_viaq_host_audit.rb
```diff
@@ -0,0 +1,27 @@
+require 'fluent/parser'
```
jcantrill: You could move this to the fluentd/lib dir and add tests... package it like a proper gem..  we have unit tests that run all the tests in  the lib dir

JAORMX: ack; will do.

fluentd/configs.d/openshift/input-pre-k8s-audit-log.conf
JAORMX: This has been moved to the cluster-logging-operator: https://github.com/jcantrill/cluster-logging-operator/pull/1

JAORMX: would definitely still use some feedback there :)

fluentd/configs.d/openshift/filter-pre-k8s-audit.conf
JAORMX: Thanks! I took the feedback and added to the CLO PR https://github.com/jcantrill/cluster-logging-operator/pull/1

PR #1763, opened 2019-10-01
fix build dependencies for fluentd

richm:


richm:
One of the consequences of this PR is that, from now on, we will have to manually monitor when changes to gems occur, and make sure those changes are compatible with fluentd 0.12 and ruby 2.0, and then update the gem version in the Dockerfile.centos7, then update the downstream RPM.

PR #1770, opened 2019-10-18
Bug 1756920: fluentd pods doesn't process kubernetes events

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1756920
Use the new feature of the viaq plugin to only do the
event transform for records with a specific tag.


richm:
multi-tenancy flake
/retest

richm:
can't cherrypick until https://github.com/openshift/origin-aggregated-logging/pull/1768 merges

PR #1771, opened 2019-10-24
use docker login instead of podman login

richm:
apparently we can no longer use `podman login` to login to the
internal OpenShift registry - we have to use `docker login`.
@nhosoi @jcantrill @ewolinetz ptal

### DIFF
--- a/hack/build-images.sh
+++ b/hack/build-images.sh
@@ -182,7 +182,7 @@ function login_to_registry() {
       username=kubeadmin
     fi
   fi
-  podman login --tls-verify=false -u "$username" -p "$token" "$1" > /dev/null
+  docker login -u "$username" -p "$token" "$1" > /dev/null
 }
 
 function push_image() {


hack/build-images.sh
```diff
@@ -182,7 +182,7 @@ function login_to_registry() {
       username=kubeadmin
     fi
   fi
-  podman login --tls-verify=false -u "$username" -p "$token" "$1" > /dev/null
+  docker login -u "$username" -p "$token" "$1" > /dev/null
```
JAORMX: why not? I've been using this (even today)

richm: What version of podman, docker, and oc are you using?

richm: see https://github.com/openshift/cluster-logging-operator/pull/253

JAORMX: .... I built my own. So that might not be very helpful info. Lets disregard my comment in the meantime in that case.

JAORMX: So... I was following other steps, not necessarily using this script. imagebuilder seems to require docker anyway.

richm:
please `/lgtm`

PR #1774, opened 2019-10-31
Added information related to viaq_index_name field

mijicherian:


richm:
> @lukas-vlcek this info is generated in a viaq repo is it not?

yes - please make the change in https://github.com/ViaQ/elasticsearch-templates/blob/master/namespaces/_default_.yml

We will then generate the index settings, index patterns and documentation for all supported ocp releases and es versions

PR #1776, opened 2019-11-01
Take into account .audit index in curator

JAORMX:


JAORMX:
@richm @jcantrill I'm still unsure if this covers the curator. I'm unable to find where the default configuration comes from, as it seems to be all commented out in the ConfigMap that's generated by default.

richm:
still something going on with ci tests - fluentd logs say that fluentd cannot connect to elasticsearch

richm:
multi-tenancy flake
/retest

PR #1779, opened 2019-11-05
Bug 1765646: Fluentd single pod logging performance regression 3.x

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1765646
Allows setting the fluentd in_tail plugin
`refresh_interval` parameter via a new environment variable
`CONTAINER_LOGS_REFRESH_INTERVAL`.  The new default value is 5 (seconds).
The old default value was 60 seconds, which could lead to missing
rotated files.
For example, to use a value of 60 seconds:
`oc set env ds/logging-fluentd CONTAINER_LOGS_REFRESH_INTERVAL=60`
This also adds support for setting the `rotate_wait` parameter via
a new environment variable `CONTAINER_LOGS_ROTATE_WAIT`.  The new
default value is `5` which helps when container logs are rotated
quickly.

### DIFF
--- a/fluentd/lib/generate_throttle_configs/lib/generate_throttle_configs.rb
+++ b/fluentd/lib/generate_throttle_configs/lib/generate_throttle_configs.rb
@@ -64,6 +64,14 @@ def get_all_throttle_files()
   return Dir.glob('/var/log/es-container-*.log.pos')
 end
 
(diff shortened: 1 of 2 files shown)


richm:
see also https://github.com/openshift/openshift-ansible/pull/11999

PR #1780, opened 2019-11-06
improvements to hack/build-images.sh

richm:
1) do not use hack/lib functions
2) use `imagebuilder` directly, as in clo and eo
3) update location of repo files for UBI builds
4) allow to specify custom tag instead of `latest`


richm:
aws flake
/retest

PR #1792, opened 2019-11-27
Elasticsearch 6 image

richm:
first pass at ES6 image


richm:
I need to rebase and push master-es6x branch - for now, ignore everything not under `elasticsearch/`

elasticsearch/Dockerfile
```diff
@@ -1,33 +1,43 @@
...
+FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 
 MAINTAINER OpenShift Development <dev@lists.openshift.redhat.com>
 
 EXPOSE 9200
 EXPOSE 9300
 USER 0
 
-ENV ES_CONF=/etc/elasticsearch/ \
+ENV ES_PATH_CONF=/etc/elasticsearch/ \
```
ewolinetz: nit but i'd suggest `ES_CONF_PATH` instead

elasticsearch/utils/es_util_env
```diff
@@ -6,3 +6,18 @@ fi
 
 ES_BASE=${ES_BASE:-https://localhost:9200}
 curl_get='curl -s -X GET --cacert /etc/elasticsearch/secret/admin-ca --cert /etc/elasticsearch/secret/admin-cert --key /etc/elasticsearch/secret/admin-key'
+
+
+get_security_conf_index() {
```
ewolinetz: this function is here and also in `utils/logging` ?

elasticsearch/Dockerfile
richm: the elasticsearch scripts use `ES_PATH_CONF` - can't use anything else

elasticsearch/utils/es_util_env
richm: yes - so - I don't know if it is safe to source `utils/logging` from scripts that use `es_util_env` and vice versa - looks like an opportunity for some refactoring here

elasticsearch/utils/es_acl
```diff
@@ -70,7 +70,7 @@ if [ ${cmd} == "get" ] ; then
   fi
 fi
 
-config=${ES_CONF}/elasticsearch.yml
+config=${ES_PATH_CONF}/elasticsearch.yml
 index=$(python -c "import yaml; print yaml.load(open('${config}'))['searchguard']['config_index_name']" | xargs -i sh -c "echo {}")
```
jcantrill: missing config key change here

elasticsearch/utils/es_util_env
```diff
@@ -6,3 +6,18 @@ fi
...
+  local es_config_file=${ES_PATH_CONF}/elasticsearch.yml
+  local index=$( python -c 'import sys,yaml
+hsh = yaml.load(open(sys.argv[1]))
+if "searchguard" in hsh and "config_index_name" in hsh["searchguard"]:
```
jcantrill: i suspect we don't need this since without changes we dynamically init the security index and we can use the OD default which i believe is actually 'security'

richm: We need this for now, until everything is using es6 and es6 components

richm:
With these fixes I am able to deploy a 4.4 cluster with logging using es6 - no blocking errors in the es logs.

richm:
closing - this PR is replaced by https://github.com/openshift/origin-aggregated-logging/pull/1801

PR #1794, opened 2019-12-03
[release-4.1] fix the way we look up ci images

richm:


bparees:
Note to patch-manager: waiving BZ requirement as this is not shipping code.

PR #1796, opened 2019-12-06
fix recent CI failures

richm:
Try harder to get the correct version of components/images to use.  Use the version number from the CLO and EO manifests directory as the version of the CLO and EO image to pull.  If that is not available, get the version number from the clusterversion.  Note that this PR allows us to use a different version of the CLO and EO, which is useful in the current situation where the EO has been bumped to 4.4 but not the CLO.

### DIFF
--- a/hack/deploy-logging.sh
+++ b/hack/deploy-logging.sh
@@ -160,8 +160,18 @@ switch_to_admin_user() {
     fi
 }
 
+get_cluster_version_maj_min() {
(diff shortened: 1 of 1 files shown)


richm:
hmm - access control related tests failed - flake?
/retest

richm:
installer flake
/retest

richm:
tests look good - please review to unblock oal tests

PR #1801, opened 2019-12-11
Elasticsearch 6 image

richm:
Build logging elasticsearch6 image from oss zip distribution, including
OS4ES and the prometheus plugin
Convert scripts, configs, etc. to run with es6
Use OD4ES instead of searchguard - scripts, configs, etc.
use securityadmin.sh add get_security_conf_index rename acl files to remove
sg_ prefix use unified install scripts minimize difference between upstream
and downstream Dockerfiles
use content-type application/json for all util commands

Have to use Content-type:application/json for all POST/PUT commands with
data now.  To make this easier to use, I refactored the utility and init
scripts to make sure they all use es_util instead of using curl directly,
as well as some other refactoring.

I fixed some of the permissions to make sure the init scripts could modify
files in place.


richm:
requires https://github.com/openshift/elasticsearch-operator/pull/209 (afaik there is not yet a feature-es6x branch in elasticsearch-operator)

jcantrill:
Merging to unblock es6 work

PR #1804, opened 2019-12-13
use operators.coreos.com/v1; handle multiple manifests versions

richm:
fix ci - operatorgroups are now v1 not alpha
there may be multiple versions of manifests - pick the latest one


richm:
didn't know the `/manifests/` dir could have multiple versions :-(

please re-review

PR #1809, opened 2019-12-23
fix ci for 4.3 - use release-4.3 branch for eo and clo

richm:


richm:
note to reviewer - test fix only - does not need bz

richm:
```
+++ curl -s 'https://api.github.com/repos/openshift/elasticsearch-operator/contents/hack/lib/test?ref=release-4.3'
+++ jq -r '.[] | .name + " " + .type + " " + .download_url'
+++ read name type download_url
jq: error (at <stdin>:4): Cannot index string with string "download_url"
```
I'm thinking this is a github api rate limit - might need to redo the way we download these files.  The reason I didn't want to use the `archive` api was that the tarballs downloaded could be very large, with all of the vendored stuff.

The best way to do this would be to somehow find the correct eo and clo images to use, then somehow extract the manifest files directly from the image . . .

/retest

richm:
this should unblock ci for https://github.com/openshift/origin-aggregated-logging/pull/1808

bparees:
/lgtm

overriding BZ requirement as this is non-shipping code that is verified by passing CI jobs.

PR #1811, opened 2020-01-02
fix #1262. remove umount logic from fluent run.sh

### LINKED ISSUE #1262: Remove unmount logic from fluend
We see a lot of the following in the fluent logs
```
umount: /var/lib/docker/containers/ec7c381febb53c13dd783940010d562f2aebaadef9f0b55d99d76c1628d9de44/shm: not mounted
umount: /var/lib/docker/containers/ee28ce7ebe4c8c59d12981a44a4d250f00bdfd002c8d0c36a3a9b8ad22f3261c/shm: not mounted
umount: /var/lib/docker/containers/efe0327286e0a47c5068747f8345ad15e718effcaa4106282614268f98d7cb7d/shm: not mounted
umount: /var/lib/docker/containers/f1efa1c4a045265bffa2d183cb27312b747d76f6962b1e09f9bff77d93bcc467/shm: not mounted
umount: /var/lib/docker/containers/f2938245cae1661b4dc46bd149ca1f13f1a9abf2286c4818c5fae35597873a24/shm: not mounted
umount: /var/lib/docker/containers/f307ed2dee303b490b36fdef82fdb10546ceb679c5df7a2c9b859531ded4101b/shm: not mounted
umount: /var/lib/docker/containers/fe70cc01c814827f0e8524994fa35a3f58513ed8e15d4543341316886c99160c/shm: not mounted

```
This was added because of an issue with pods terminating and either fluent or someother process holding onto file handles which I believe has since been resolved for several releases.

@richm @mrunalp  is this something you recall?  Can we remove it from the run script in fluentd?

jcantrill:
fix #1262. Remove umount logic from fluent run.sh

### DIFF
--- a/fluentd/run.sh
+++ b/fluentd/run.sh
@@ -200,18 +200,6 @@ else
     rm -f $CFG_DIR/openshift/input-pre-debug.conf
 fi
 
-# bug https://bugzilla.redhat.com/show_bug.cgi?id=1437952
-# pods unable to be terminated because fluentd has them busy
-if [ -d /var/lib/docker/containers ] ; then
-    # If oci-umount is fixed, we can remove this.
-    if [ -n "${VERBOSE:-}" ] ; then
-        echo "umounts of dead containers will fail. Ignoring..."
-        umount /var/lib/docker/containers/*/shm || :
-    else
-        umount /var/lib/docker/containers/*/shm > /dev/null 2>&1 || :
-    fi
-fi
-
 if [[ "${USE_REMOTE_SYSLOG:-}" = "true" ]] ; then
     # The symlink is a workaround for https://github.com/openshift/origin-aggregated-logging/issues/604
     found=


ewolinetz:
are we sure the original issue has been resolved? i feel like it would be dangerous to remove this if fluentd still held on to rolled over container log files..

jcantrill:
> are we sure the original issue has been resolved? i feel like it would be dangerous to remove this if fluentd still held on to rolled over container log files..

@ewolinetz There was a bug associated in the runtime that was addressed when we originally added this code that was fixed.  That was at least 3 releases ago. I believe.  @richm Do you remember if this was related specifically to docker or journal?

richm:
It was related to oci - so 1) not sure if it has really been fixed in current 3.11 releases 2) not sure if it affects cri-o/4.x at all

jcantrill:
> are we sure the original issue has been resolved? i feel like it would be dangerous to remove this if fluentd still held on to rolled over container log files..

@ewolinetz from the containers team:

```
Peter Hunt  10:16 AM
to be honest, I wouldn't expect the condition to pass anyway as we don't put our containers in /var/lib/docker/containers

10:17
so you should definitely be safe removing it, this code path won't touch cri-o containers
```

jcantrill:
Closing as run script was moved into the CLO

PR #1816, opened 2020-01-21
Fixing Kibana 6 and ES 6 working together

ewolinetz:


ewolinetz:
currently trying to troubleshoot the kibana6 probe... it gets a `401` currently

elasticsearch/sgconfig/config.yml
```diff
@@ -12,39 +12,12 @@ opendistro_security:
         order: 0
         http_authenticator:
           challenge: false
-          type: io.fabric8.elasticsearch.plugin.auth.OpenShiftTokenAuthentication
+          type: proxy
```
jcantrill: Since you are making the change, you might as well go and update the config to what we need https://github.com/opendistro-for-elasticsearch/security/pull/85  The DLS isnt in place so I expect it to function the same as `proxy`

ewolinetz: `extended_proxy` ?

jcantrill: > `extended_proxy` ?

dash https://github.com/opendistro-for-elasticsearch/security/pull/85/files#diff-062271cb239849b80aa8fb7d9b55487cR217

elasticsearch/sgconfig/config.yml
```diff
@@ -12,39 +12,12 @@ opendistro_security:
...
-                 resource: jaeger
-                 namespace: set-in-run-script
-                 resourceAPIGroup: elasticsearch.jaegertracing.io
-               elasticsearch-operator:
-                 verb: '*'
-                 resource: '*'
-                 namespace: '*'
-                 resourceAPIGroup: logging.openshift.io
+            user_header: 'x-forwarded-user'
+            roles_header: 'x-forwarded-roles'
```
jcantrill: please also add in https://github.com/opendistro-for-elasticsearch/security/pull/85/files#diff-062271cb239849b80aa8fb7d9b55487cR217 with a value that will be provided by the 'elasticsearch-proxy' so we are ready for DLS.

ewolinetz:
@jcantrill i still need to figure out the readiness probe fix so we don't need to keep retesting this currently. it is good to see that the upgrade test passed meaning we're able to start everything up now

jcantrill:
Merging.  Failed test is likely caused form LF as I was unable to manually run the test either.

jcantrill:
Merging manually as failed test is unrelated to these changes and likely from something broken by LogForwarding

PR #1817, opened 2020-01-22
Fix es6 build

jcantrill:
This PR:
* Modifies elasticsearch scripts and dockerfiles to rely on ES_PATH_CONF for dynamic generated things (e.g. truststores)
* removes obsoleted tests

@ewolinetz This will likely not pass tests until I can merge the dependent operator changes to resolve configuration

The images need a config created by https://github.com/openshift/elasticsearch-operator/pull/226


ewolinetz:
/hold
i want to investigate this a bit more since i don't see this same issue

elasticsearch/probe/readiness.sh
```diff
@@ -21,7 +21,7 @@ export NSS_SDB_USE_CACHE=no
 # TODO: try re-use code from ./run.sh
 ES_REST_BASEURL=https://localhost:9200
 EXPECTED_RESPONSE_CODE=200
-secret_dir=/etc/elasticsearch/secret
```
ewolinetz: if we change this then the configmap will be incorrect https://github.com/openshift/elasticsearch-operator/blob/feature-es6x/pkg/k8shandler/configuration_tmpl.go#L38

elasticsearch/utils/logging
```diff
@@ -17,7 +17,7 @@ esac
 
 # the deployment mounts the secrets at this location - not necessarily the same
 # as $ES_PATH_CONF
-secret_dir=/etc/elasticsearch/secret
```
ewolinetz: if we change this then the configmap will be incorrect https://github.com/openshift/elasticsearch-operator/blob/feature-es6x/pkg/k8shandler/configuration_tmpl.go#L38

jcantrill:
Closing as unable to replicate original issue

PR #1824, opened 2020-01-31
Bug 1791231: Use hostname for utils in support of ipv6

jcantrill:
To additionally resolve https://bugzilla.redhat.com/show_bug.cgi?id=1791231  because localhost on ipv6 did not allow the various util scripts to work properly

### DIFF
--- a/elasticsearch/utils/logging
+++ b/elasticsearch/utils/logging
@@ -26,7 +26,7 @@ REPLICA_SHARDS=${REPLICA_SHARDS:-0}
 
 RETRY_COUNT=${RETRY_COUNT:-300}		# how many times
 RETRY_INTERVAL=${RETRY_INTERVAL:-1}	# how often (in sec)
-ES_REST_BASEURL=${ES_REST_BASEURL:-https://localhost:9200}
+ES_REST_BASEURL=${ES_REST_BASEURL:-https://$HOSTNAME:9200}
 LOG_FILE=${LOG_FILE:-elasticsearch_connect_log.txt}
 retry=$RETRY_COUNT
 max_time=$(( RETRY_COUNT * RETRY_INTERVAL ))	# should be integer


alanconway:
FYI in Go and Java you can use the empty string "" as the host in Listen calls (NULL in C/C++), that 
means listen on all interfaces, for all protocols. May be more robust than $HOSTNAME, which can  be mis-configured so the local resolver won't resolve it, and may not return all interfaces on a multi-homed host.

jcantrill:
> FYI in Go and Java you can use the empty string "" as the host in Listen calls (NULL in C/C++), that
> means listen on all interfaces, for all protocols. May be more robust than $HOSTNAME, which can be mis-configured so the local resolver won't resolve it, and may not return all interfaces on a multi-homed host.

What does that look like in this specific PR with curl?

alanconway:
In Go on the server side it's 
    l, err := net.Listen("tcp", ":6000") // all networks, all protocols

On the client side, the following *should* be equivalent: both should use the loopback addr, and may pick either ipv4 or v6 if both are available. A client can try both if both are enabled but lots of production clients ignore all but the first address returned by the resolver.

    c, err := net.Conn("tcp", ":6000")
    c, err := net.Conn("tcp", "localhost:6000")

Empty host won't work with curl, http://:foo is officially an invalid URL.  You're supposed to be able to use localhost if you don't know which of v4/v6 is available.

In the logs on this bug I saw servers listening on "[::]:60000" rather than ":60000" - that means they are *only* listening on v6. A theory: the client uses "localhost", it picks v4, so can't connect.

I think the right solution is to make sure all servers by default listen for all protocols i.e. 
    net.Listen("tcp", ":60000") 
Then it doesn't matter what the clients pick, and it will work with mixed v4/v6 clients. 
If my theory is right, `curl http://localhost:6000/...` will work with such servers.

jcantrill:
Closing for now as it seems there is something else to be addressed

Issue #1547, opened 2019-03-11
fluentd remote - syslog over TLS


sghosh151:
Would it be possible to upgrade the Fluentd output to remote syslog server to support TLS connections instead of raw  UDP.

eg: [https://github.com/dlackty/fluent-plugin-remote_syslog](https://github.com/dlackty/fluent-plugin-remote_syslog)

richm:
https://github.com/openshift/origin-aggregated-logging/commit/d54579359357b0b79d5d67d8316cae3eca91ee75

sghosh151:
Is patch for just TCP or supporting TLS as well?

richm:
Yes, sorry, no TLS support

sghosh151:
Any thoughts on pattern for forwarding over TLS?

richm:
> Any thoughts on pattern for forwarding over TLS?

The only way to do that now is to use secure_forward.  We don't currently have any plans to switch from using `fluent-plugin-remote-syslog` to `fluent-plugin-remote_syslog` or some other syslog plugin with TLS support - we haven't had any Red Hat customers asking for such a feature.  Patches welcome, but there is a lot of work that would need to be done to the ruby script that generates the config - https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/generate_syslog_config.rb

PR #1826, opened 2020-02-05
Add fluent-plugin-remote_syslog

ahadas:
This PR adds [fluent-plugin-remote_syslog ](https://github.com/dlackty/fluent-plugin-remote_syslog) to the plugins for fluentd. As opposed to the current plugin we use for interacting with an external syslog server, this one supports both secure connections (see #1547 ) and contains the implementation of UDP, TCP and TCP+TLS within a single script.

### DIFF
--- a/fluentd/Dockerfile
+++ b/fluentd/Dockerfile
@@ -77,7 +77,7 @@ COPY --from=builder /contents /contents
 
 RUN mkdir -p /etc/fluent/plugin
 ADD configs.d/ /etc/fluent/configs.d/
-ADD out_syslog_buffered.rb out_syslog.rb out_rawtcp.rb /etc/fluent/plugin/
+ADD out_syslog_buffered.rb out_syslog.rb out_rawtcp.rb out_remote_syslog.rb /etc/fluent/plugin/
 ADD run.sh generate_syslog_config.rb ${HOME}/
 ADD lib/generate_throttle_configs/lib/*.rb ${HOME}/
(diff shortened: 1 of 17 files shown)


jcantrill:
@ahadas can we remove the previous one in favor of this one?  Is it a drop in replacement?

ahadas:
> @ahadas can we remove the previous one in favor of this one? Is it a drop in replacement?

@jcantrill I think we can't, at least not at the moment, because although the new one seems to cover the functionality of the previous one from a high level perspective (transport protocols, properties), it most probably doesn't offer all the things achieved by the fixes/enhancements that were done by us for the previous plugin (as @richm noted elsewhere).

I think we should also keep the previous one for the time being and start working on getting a feature parity between the two in terms of the aforementioned changes. we can expose the overlapping parts of the new plugins, as non-default, to (hopefully) get feedback on their use compared to those the previous one, alongside TLS (and UDP?) support that defaults to the new plugin. does it make sense?

richm:
As a start, you could alter the test https://github.com/openshift/origin-aggregated-logging/blob/master/test/remote-syslog.sh to make it work with the new plugin, and see if it can pass this test.  I don't know if this test will cover all of the features/parameters of the old remote syslog plugin, but at least it is a start.

ahadas:
@richm , right, that's a good idea. I'd like to do this as a second phase where the first phase is to get this new plugin in more or less as it is (except for the second commit in this PR that removes the validation of the hostname to enable basic flows) and provide support for TLS on top of it.

PR #1827, opened 2020-02-07
elasticsearch/sgconfig: Added a role with DLS query to support multit…

vimalk78:
elasticsearch/sgconfig: Added a role with DLS query to support multitenancy

fixes https://issues.redhat.com/browse/LOG-600

### DIFF
--- a/elasticsearch/sgconfig/roles.yml
+++ b/elasticsearch/sgconfig/roles.yml
@@ -161,3 +161,21 @@ sg_project_operations:
         - indices:admin/mappings/fields/get*
         - indices:admin/validate/query*
         - indices:admin/get*
+
+# To support multi-tenancy. User's access to indices is restricted to indices belonging to the user's projects, enforced by DLS.    
+project_user:    
+  readonly: true    
+  cluster:
+    - CLUSTER_COMPOSITE_OPS_RO
+  indices:    
+    app:    
+      '*':    
+        - READ    
+        - indices:data/read/search
+      _dls_: "{\"bool\":{\"filter\":{\"script\":{\"script\":{\"lang\":\"painless\",\"params\":{\"param1\":\"${attr.proxy.namespace}\"},\"source\":\"String namespace = doc['kubernetes.namespace_name.keyword'][0];StringTokenizer st = new StringTokenizer(params.param1,\\\",\\\");while (st.hasMoreTokens()){if (st.nextToken().equalsIgnoreCase(namespace)){return true;}}return false;\"}}}}}"
+    logs.app:    
+      '*':    
+        - READ    
+        - indices:data/read/search
+      _dls_: "{\"bool\":{\"filter\":{\"script\":{\"script\":{\"lang\":\"painless\",\"params\":{\"param1\":\"${attr.proxy.namespace}\"},\"source\":\"String namespace = doc['kubernetes.namespace_name.keyword'][0];StringTokenizer st = new StringTokenizer(params.param1,\\\",\\\");while (st.hasMoreTokens()){if (st.nextToken().equalsIgnoreCase(namespace)){return true;}}return false;\"}}}}}"
+
--- a/elasticsearch/sgconfig/roles.yml
+++ b/elasticsearch/sgconfig/roles.yml
@@ -163,19 +163,17 @@ sg_project_operations:
         - indices:admin/get*
 
 # To support multi-tenancy. User's access to indices is restricted to indices belonging to the user's projects, enforced by DLS.    
-project_user:    
-  readonly: true    
+project_user:
+  readonly: true
   cluster:
     - CLUSTER_COMPOSITE_OPS_RO
-  indices:    
-    app:    
-      '*':    
-        - READ    
-        - indices:data/read/search
+  indices:
+    app:
+      '*':
+        - READ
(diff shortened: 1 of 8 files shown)


elasticsearch/sgconfig/roles.yml
```diff
@@ -161,3 +161,17 @@ sg_project_operations:
         - indices:admin/mappings/fields/get*
         - indices:admin/validate/query*
         - indices:admin/get*
+
+# To support multi-tenancy. User's access to indices is restricted to indices belonging to the user's projects, enforced by DLS.    
+project_user:    
+  readonly: true
```
jcantrill: I think we will also need these cluster permissions: https://github.com/opendistro-for-elasticsearch/security/blob/v0.10.0.4/securityconfig/roles.yml#L51-L52

vimalk78: added

jcantrill:
/hold
Let's see if we can get the tests re-enabled

test/multi_tenancy.sh
```diff
@@ -32,26 +35,29 @@ delete_users=""
...
-    fi
-    for proj in multi-tenancy-1 multi-tenancy-2 multi-tenancy-3 ; do
-        oc delete project $proj 2>&1 | artifact_out
-        os::cmd::try_until_failure "oc get project $proj" 2>&1 | artifact_out
-    done
-    # this will call declare_test_end, suite_end, etc.
-    os::test::junit::reconcile_output
+	if [ "${DO_CLEANUP:-true}" == "true" ] ; then
+		mkdir -p $ARTIFACT_DIR/$test_name
+		oc -n $LOGGING_NS get configmap fluentd -o jsonpath={.data} --ignore-not-found > $ARTIFACT_DIR/$test_name/fluent-configmap.log ||:
```
jcantrill: Nit more then anything but do you need fluent config in context of this test?  Not a blocker for me

test/multi_tenancy.sh
```diff
@@ -74,51 +80,60 @@ function create_user_and_assign_to_projects() {
 }
 
 function add_message_to_index() {
-    # project is $1
-    # message is $2
-    # espod is $3
+
+    local namespace=$1
     local project_uuid=$( oc get project $1 -o jsonpath='{ .metadata.uid }' )
     local index="project.$1.$project_uuid.$(date -u +'%Y.%m.%d')"
```
jcantrill: Since you are making this change in context of the new data model i would go ahead and structure the setup accordingly instead of using the legacy format:
* data lives in an index like (recalling from memory): `logs.apps-000001`
* index is aliased like `logs.apps`

Or maybe it doesnt matter since we are relying on the alias and OpenDistro will secure the actual index.  Eitherway the index name likely does not need to be as complicated now as it was previously

test/multi_tenancy.sh
```diff
@@ -74,51 +80,60 @@ function create_user_and_assign_to_projects() {
...
+
+    local xff="-H X-Forwarded-For:127.0.0.1"
+    local xocpns='-H X-Ocp-Namespace:'"$namespace"
+    local contenttype="-H Content-Type:application/json"
+
+    local payload="{\"log\":\"log message 1\",\"stream\":\"stderr\",\"time\":\"2014-09-25T21:15:03.499185026Z\",\"kubernetes\":{\"namespace_name\":\"$namespace\",\"pod_name\":\"synthetic-logger-0.25lps-pod\"},\"docker\":{\"container_id\":\"container123\"}}"
+    local alias="{\"actions\":[{\"add\":{\"index\":\"$index\",\"alias\":\"app\"}}]}"
+
+    # add index
+    os::log::debug $( curl_es_pod "$pod" "/$index/multi-tenancy-test/" -XPOST $xff $xocpns $contenttype -d "$payload" | python -mjson.tool 2>&1 )
```
jcantrill: Change the doc type here to `_doc` since doc types are going away in future releases of ES and '_doc' is the default one

test/multi_tenancy.sh
```diff
@@ -74,51 +80,60 @@ function create_user_and_assign_to_projects() {
...
+    local xfuser='-H x-forwarded-user:'"$user"
+    local xocpns='-H x-ocp-namespace:'"$user_project_list"
+    local xfroles='-H x-forwarded-roles:project_user'
+    
+    nrecs=$( curl_es_pod_with_token $espod "/$alias/_count" $test_token -XGET $xfuser $xocpns $xfroles | get_count_from_json )
+    expected=${1:-$user_project_num}
+    if ! os::cmd::expect_success "test $nrecs = $expected" ; then
+        os::log::error $user cannot access alias /$alias
+        curl_es_pod_with_token $espod "/$alias/_count" $test_token -XGET $xfuser $xocpns $xfroles | python -mjson.tool
+        oc exec -c elasticsearch $espod -- es_acl get --doc=roles
```
jcantrill: Consider writing these to a file in the artifact dir

test/multi_tenancy.sh
vimalk78: changed to elasticsearch configmap

test/multi_tenancy.sh
vimalk78: simplified the index name

test/multi_tenancy.sh
vimalk78: done

test/multi_tenancy.sh
vimalk78: done

PR #1828, opened 2020-02-10
enable smoke test for es6

jcantrill:
This PR makes a first pass to re-enable a smoke test to do basic es6 testing

### DIFF
--- a/hack/deploy-logging.sh
+++ b/hack/deploy-logging.sh
@@ -32,6 +32,12 @@
 
 set -eux
 
+#HACK to temp override upgrade job
+if [ -z ${OPENSHIFT_BUILD_NAMESPACE:-''} -a -z ${IMAGE_FORMAT:-''} -a "${USE_CUSTOM_IMAGES:-false}" == "true" ] ; then
+    echo "Exiting gracefully as this run is assumed to be for upgrade"
+    exit 0
+fi
+
 logging_err_exit() {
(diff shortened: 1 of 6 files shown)


ewolinetz:
what about modifying and using `hack/testing/check-EFK-running.sh` ?

hack/deploy-logging.sh
```diff
@@ -235,24 +239,50 @@ update_images_in_clo_yaml() {
...
+    echo INFO: Expecting $expectedcollectors collector to start
     # we expect $nodeCount elasticsearch pods
+    es_ready="false"
+    fluent_ready="false"
+    kibana_ready="false"
     wait_func() {
-        local actuales=$( oc get pods -l component=elasticsearch 2> /dev/null | grep -c 'elasticsearch.* 2/2 .*Running' )
+        result=0
+        local actuales=$( oc -n ${LOGGING_NS} get pods -l component=elasticsearch 2> /dev/null | grep -c 'elasticsearch.* 2/2 .*Running' )
         if [ $expectedes -ne ${actuales:-0} ] ; then
```
periklis: `$expectedes` seems unset before usage in this condition.

jcantrill: This value is set leading up to the call to this function.  It's not obvious.  I believe this is something to be revisited as we refactor.

PR #1829, opened 2020-02-11
Increase timeout for waiting for port forwarding

vimalk78:


ewolinetz:
@vimalk78 is this still necessary?

vimalk78:
> @vimalk78 is this still necessary?

Many a times i have faced this issue that port forwading actually succeeds but script gives up early. increasing this timeout will save time, as the only other option is to run the script again from beginning.

PR #1830, opened 2020-02-14
Adding tribe

RiRa12621:
In order to allow customers in a very locked down OSD environment to access indexed documents via the we want to introduce tribe nodes.
While the concept was not originally planned for this from an upstream perspective, it does give us the functionality that we need: an API to query elasticsearch without control over the clustser itself

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,11 @@ proxy is required that runs in front of Kibana.
 Curator allows the admin to remove old indices from Elasticsearch on a per-project
 basis.
 
+### Tribe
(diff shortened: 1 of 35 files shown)


jcantrill:
@RiRa12621 we are no longer merging features into 3.11, please rebase this change on master.  Also, I'm not certain why there needs to be an entire 'tribe' subdirectory for Elasticsearch but I would imagine there is a place that such changes would be integrated into the existing scripts and images.

PR #1831, opened 2020-02-14
Bug 1794085:  [fluentd] 3.11 remote syslog message is incomplete/truncated

syedriko:
This fix makes maximum size of syslog messages sent by Fluentd configurable,
for both the syslog (UDP) and syslog_buffered (TCP) plugins.
Previously, the maximum message size was hardcoded to 1024 bytes. This fix
keeps that limit as the default.

A new environment variable is now recognized:
* REMOTE_SYSLOG_MAX_SIZE, which specifies the maximum size of syslog messages, in bytes.


jcantrill:
@bparees the equivalent 4.x changes will be superseded by LF. Please approve me

bparees:
overriding dependent-bug requirement based on https://github.com/openshift/origin-aggregated-logging/pull/1831#issuecomment-586478484

PR #1832, opened 2020-02-18
Drop xtrace verbosity from deploy and test scripts

periklis:


jcantrill:
@alanconway are you suggesting we should have a practice that is different then what's here.  This is the pattern used mostly by all our scripts.

periklis:
@alanconway PTAL

PR #1840, opened 2020-02-27
Updated FROM image reference

vimalk78:
Added registry name `registry.access.redhat.com` to the FROM image reference for local building of image


periklis:
We don't need this if we land #1892

vimalk78:
closing

PR #1842, opened 2020-02-28
fix operator deployment to deploy static resources

jcantrill:
Deploy static resources that OLM will not

cc @periklis @vimalk78


jcantrill:
@periklis i think we will need to backport a similar change to 4.4 since i think your update is moving there.. correct?

periklis:
@jcantrill Yes we need to cherry-pick this on for `release-4.4`

periklis:
This needs to be amended after merging openshift/cluster-logging-operator/pull/389 because the static roles for `openshift-config-managed` will be droped

PR #1844, opened 2020-03-04
Modify Elasticsearch Dockerfile to mitigate issue with buildah

jcantrill:
This PR modifies elasticsearch Dockerfile to mitigate:
* https://bugzilla.redhat.com/show_bug.cgi?id=1810069

where buildah is now used and not expanding vars properly

cc @vimalk78

### DIFF
--- a/elasticsearch/Dockerfile
+++ b/elasticsearch/Dockerfile
@@ -45,7 +45,7 @@ ADD index_patterns/ ${ES_HOME}/index_patterns/
 ADD init/ ${ES_HOME}/init/
 ADD kibana_ui_objects/ ${ES_HOME}/kibana_ui_objects/
 ADD probe/ ${ES_HOME}/probe/
-ADD init.sh run.sh prep-install.${RELEASE_STREAM} install.sh ${HOME}/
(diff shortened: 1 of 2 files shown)


elasticsearch/Dockerfile
```diff
@@ -45,7 +45,7 @@ ADD index_patterns/ ${ES_HOME}/index_patterns/
 ADD init/ ${ES_HOME}/init/
 ADD kibana_ui_objects/ ${ES_HOME}/kibana_ui_objects/
 ADD probe/ ${ES_HOME}/probe/
-ADD init.sh run.sh prep-install.${RELEASE_STREAM} install.sh ${HOME}/
+ADD init.sh run.sh prep-install* install.sh ${HOME}/
```
periklis: Do we actually use the `prep-install.prod` file at all? If not, can we simplify the ADD directive to just copy `prep-install.origin`?

jcantrill: > Do we actually use the `prep-install.prod` file at all? If not, can we simplify the ADD directive to just copy `prep-install.origin`?

Yes.  they get used differently depending if you are building in CI or vi ART.  It's an attempt to provide normalization between the two build options.

PR #1852, opened 2020-03-16
Restrict access to ES from localhost only

vimalk78:
`trustedProxies` is regex pattern which is matched against the remote address in the http request. 

configure `trustedProxies` to match only localhost ipv4 and ipv6 addesses

### DIFF
--- a/elasticsearch/sgconfig/config.yml
+++ b/elasticsearch/sgconfig/config.yml
@@ -4,8 +4,8 @@ opendistro_security:
       xff:
         enabled: true
         remoteIpHeader: 'x-forwarded-for'
+        internalProxies: '0\:0\:0\:0\:0\:0\:0\:1|127\.0\.0\.1'
         trustedProxies: '.*'
-        internalProxies: '.*'
     authc:
       openshift_domain:
         enabled: true


elasticsearch/sgconfig/config.yml
```diff
@@ -4,7 +4,7 @@ opendistro_security:
       xff:
         enabled: true
         remoteIpHeader: 'x-forwarded-for'
-        trustedProxies: '.*'
+        trustedProxies: '0\:0\:0\:0\:0\:0\:0\:1|127\.0\.0\.1'
```
periklis: Are you sure this works also for address like `::1`? In addition, IPv6 provides also link-local address, e.g. ~`fe::80...`. Should these be also trusted?

elasticsearch/sgconfig/config.yml
```diff
@@ -4,7 +4,7 @@ opendistro_security:
...
         internalProxies: '.*'
```
jcantrill: Remind me what 'internalProxies' means?  Does this need to be modified too?

vimalk78:
```
opendistro_security:
  dynamic:
    http:
      xff:
        enabled: true
        remoteIpHeader: 'x-forwarded-for'
        trustedProxies: '<regex-pattern>'
        internalProxies: '<regex-pattern>'
```
The remote address (InetAddress) from proxy request is matched against `internalProxies`.
`final String originalRemoteAddr = ((InetSocketAddress)request.getRemoteAddress()).getAddress().getHostAddress();`

If match fails, OPENDISTRO_SECURITY_XFF_DONE is false.
If `internalProxies` match succeeds, then the values in `X-Forwarded-For` header are matched against `trustedProxies`.
The matching is done in reverse order. Header values that match are put in `X-Forwarded-By` header, and values that do not match remain in `X-Forwarded-For` header. The `OPENDISTRO_SECURITY_XFF_DONE` is set to true as long as there is at least one value. 

`HTTPProxyAuthenticator` only checks if `OPENDISTRO_SECURITY_XFF_DONE` is true.
So we can use `internalProxies` to enforce the check that acess to ES is frol localhost

vimalk78:
i had configured `trustedProxies` instead of `internaProxies`.

corrected it. hence the update to PR

vimalk78:
duplicate of https://github.com/openshift/origin-aggregated-logging/pull/1855 
closing

PR #1853, opened 2020-03-16
Add fluentd-plugin-kafka to fluentd image

periklis:
This PR vendors the [fluentd-plugin-kafka](https://github.com/fluent/fluent-plugin-kafka) into the fluentd image.


fluentd/Gemfile
```diff
@@ -5,6 +5,7 @@ gem 'elasticsearch'
 gem 'fluent-plugin-kubernetes_metadata_filter'
 gem 'fluent-plugin-concat'
 gem 'fluent-plugin-elasticsearch'
+gem 'fluent-plugin-kafka'
```
jcantrill: @alanconway do you think we should start locking down versions to be repeatable here and not be "surprised" by updates.

periklis: @jcantrill For this purpose i've added 767c29f5b2f2f05982884da73d3be50dd25e98fa to expedite this PR.

fluentd/Gemfile
```diff
@@ -5,6 +5,7 @@ gem 'elasticsearch'
 gem 'fluent-plugin-kubernetes_metadata_filter'
 gem 'fluent-plugin-concat'
 gem 'fluent-plugin-elasticsearch'
+gem 'fluent-plugin-kafka', ENV['FLUENTD_PLUGIN_KAFKA_VERSION']
```
jcantrill: I might recommend the opposite change here and set the version in the Gemfile and not Dockerfile.  I don't see the advantage of it existing in the Dockerfile at all for this or fluentd unless there is some dependency there.

periklis:
@jcantrill I've removed the extra ENV check for `fluentd-plugin-kafka`. In addition i've split the PR in three more obvious commits to enable readability.

periklis:
Please ignore 109d3ce for. Depends on: #1858

PR #1855, opened 2020-03-20
Fixed ES access via es-proxy

vimalk78:
Restrict access to ES from localhost only
Modified multi-tenancy test to test from kibana pod
Fixed a bug in DLS painless query

### DIFF
--- a/elasticsearch/sgconfig/config.yml
+++ b/elasticsearch/sgconfig/config.yml
@@ -4,8 +4,8 @@ opendistro_security:
       xff:
         enabled: true
         remoteIpHeader: 'x-forwarded-for'
+        internalProxies: '0:0:0:0:0:0:0:1|127\.0\.0\.1'
         trustedProxies: '.*'
-        internalProxies: '.*'
     authc:
       openshift_domain:
         enabled: true
--- a/hack/testing/util.sh
+++ b/hack/testing/util.sh
@@ -262,7 +262,6 @@ curl_es_from_kibana() {
        --cert "${secret_dir}cert" \
        --key "${secret_dir}key" \
        -H "Authorization: Bearer $test_token" \
-       -H "X-Forwarded-For: 127.0.0.1" \
        "https://${eshost}:9200${endpoint}"
 }
 
@@ -325,11 +324,22 @@ function curl_es_pod_with_token() {
     shift; shift; shift;
     local args=( "${@:-}" )
     local authtoken="-H Authorization: Bearer $test_token"
-    local xff="-H X-Forwarded-For:127.0.0.1"
     oc -n $LOGGING_NS exec -c elasticsearch "${pod}" -- curl --silent --insecure "${args[@]}" \
                              $authtoken \
-                             $xff \
                              "https://localhost:9200${endpoint}"
+}
+
+function curl_es_pod_from_kibana_with_token() {
+    local pod="$1"
+    local eshost="$2"
+    local endpoint="$3"
+    local test_token="$4"
+    shift; shift; shift; shift;
+    local args=( "${@:-}" )
+    local authtoken="-H Authorization: Bearer $test_token"
+    oc -n $LOGGING_NS exec -c kibana "${pod}" -- curl --silent --insecure "${args[@]}" \
+                             $authtoken \
+                             "https://${eshost}:9200${endpoint}"
 }
 
 function curl_es_pod_with_username_password() {
@@ -378,14 +388,27 @@ function curl_es_pod_with_token_and_input() {
     local args=( "${@:-}" )
     local authtoken="-H Authorization: Bearer $test_token"
     local contenttype="-H Content-type:application/json"
-    local xff="-H X-Forwarded-For:127.0.0.1"
 
     oc -n $LOGGING_NS exec -c elasticsearch -i "${pod}" -- curl --silent --insecure "${args[@]}" \
-                             $xff         \
                              $authtoken   \
                              $contenttype \
                              "https://localhost:9200${endpoint}"
 }
+
+function curl_es_pod_from_kibana_with_token_and_input() {
+    local pod="$1"
+    local eshost="$2"
+    local endpoint="$3"
+    local test_token="$4"
+    shift; shift; shift; shift;
+    local args=( "${@:-}" )
+    local contenttype="-H Content-type:application/json"
+    oc -n $LOGGING_NS exec -c kibana -i "${pod}" -- curl --silent --insecure "${args[@]}" \
+                            -H "Authorization: Bearer $test_token" \
+                             $contenttype \
+                             "https://${eshost}:9200${endpoint}"
+}
+
 
 function curl_es_with_token_and_input() {
     local svc_name="$1"
--- a/test/multi_tenancy.sh
+++ b/test/multi_tenancy.sh
@@ -51,6 +51,8 @@ function cleanup_es() {
         oc delete user $user 2>&1 | artifact_out
     done
     if [ -n "${espod:-}" ] ; then
+        oc -n $LOGGING_NS logs -c elasticsearch $espod > $ARTIFACT_DIR/es.log
+        oc -n $LOGGING_NS exec -c elasticsearch $espod -- logs >> $ARTIFACT_DIR/es.log
         curl_es_pod $espod /logs-app-* -XDELETE 2>&1 | artifact_out
     fi
     for proj in $PROJECTS ; do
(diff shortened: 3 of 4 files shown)


hack/testing/util.sh
```diff
@@ -325,13 +324,24 @@ function curl_es_pod_with_token() {
...
-                             $xff \
                              "https://localhost:9200${endpoint}"
 }
 
+function curl_es_pod_from_kibana_with_token() {
+    local pod="$1"
+    local eshost="$2"
+    local endpoint="$3"
+    local test_token="$4"
+    shift; shift; shift; shift;
```
periklis: I should be possible to do `shift 4` here, isn't?

hack/testing/util.sh
```diff
@@ -325,13 +324,24 @@ function curl_es_pod_with_token() {
...
+    local args=( "${@:-}" )
+    local authtoken="-H Authorization: Bearer $test_token"
+    oc -n $LOGGING_NS exec -c kibana "${pod}" -- curl --silent --insecure "${args[@]}" \
+                             $authtoken \
```
periklis: I would skip the variable here for `-H "Authorization: Bearer ${test_token}"`. It enhance readability of the curl command as if you were typing and reading in the CLI.

hack/testing/util.sh
```diff
@@ -378,15 +388,28 @@ function curl_es_pod_with_token_and_input() {
     local args=( "${@:-}" )
     local authtoken="-H Authorization: Bearer $test_token"
     local contenttype="-H Content-type:application/json"
-    local xff="-H X-Forwarded-For:127.0.0.1"
 
     oc -n $LOGGING_NS exec -c elasticsearch -i "${pod}" -- curl --silent --insecure "${args[@]}" \
-                             $xff         \
                              $authtoken   \
                              $contenttype \
```
periklis: Same here, put them in-line to enhance readability.

hack/testing/util.sh
```diff
@@ -378,15 +388,28 @@ function curl_es_pod_with_token_and_input() {
...
+    local pod="$1"
+    local eshost="$2"
+    local endpoint="$3"
+    local test_token="$4"
+    shift; shift; shift; shift;
+    local args=( "${@:-}" )
+    local contenttype="-H Content-type:application/json"
+    oc -n $LOGGING_NS exec -c kibana -i "${pod}" -- curl --silent --insecure "${args[@]}" \
+                            -H "Authorization: Bearer $test_token" \
+                             $contenttype \
```
periklis: And here.

test/access_control.sh
```diff
@@ -3,26 +3,20 @@
 # test access control
 source "$(dirname "${BASH_SOURCE[0]}" )/../hack/lib/init.sh"
 source "${OS_O_A_L_DIR}/hack/testing/util.sh"
+current_dir=$(dirname "${BASH_SOURCE[0]}" )
+repo_dir=${repodir:-"$current_dir/../"}
+test_name=$(basename $0)
```
periklis: It would enable re-usability and improve maintenance if `current_dir` moves to the the top. In addition keeping to `${BASH_SOURCE[0]}` instead of flipping with `$0` makes the script more robust when sourced.

test/access_control.sh
```diff
@@ -51,19 +65,30 @@ function cleanup() {
...
 trap cleanup EXIT
 
+if [ "${DO_SETUP:-true}" == "true" ] ; then
+	os::log::info "Deploying cluster logging..."
+	${repo_dir}/hack/deploy-logging.sh
+fi
+
+espod=$( get_es_pod es )
+esopspod=$( get_es_pod es-ops )
+esopspod=${esopspod:-$espod}
```
periklis: Is `es-ops` still used in ES6?

test/access_control.sh
```diff
@@ -51,19 +65,30 @@ function cleanup() {
...
+es_svc=$( get_es_svc es )
+es_ops_svc=$( get_es_svc es-ops )
+es_ops_svc=${es_ops_svc:-$es_svc}
```
periklis: Same here. Do you need the svc for es-ops?

test/access_control.sh
```diff
@@ -132,85 +162,98 @@ function test_user_has_proper_access() {
...
+        "run_curl_cmd 200 curl_es_pod_with_token $esopspod '/.operations.*/_count' $test_token -XGET $xff $xfuser $xocpns $xfroles| get_count_from_json" \
         "^0\$"
 
     esopshost=$( get_es_svc es-ops )
     if [ "$espod" = "$esopspod" ] ; then
         esopshost=$( get_es_svc es )
     fi
+
+    # accessing from kibana pod will result in 200 with access to 0 docs
     test_message="$user has improper access to .operations.* indices from kibana"
```
periklis: This and the test messages below still refer to the old `operations.*` indices.

test/access_control.sh
```diff
@@ -132,85 +162,98 @@ function test_user_has_proper_access() {
...
     os::cmd::expect_success_and_text \
-        "run_curl_cmd 403 curl_es_from_kibana $kpod $esopshost '/.operations.*/_count' $test_token | get_count_from_json" \
+        "run_curl_cmd 200 curl_es_from_kibana $kpod $esopshost '/.operations.*/_count' $test_token | get_count_from_json" \
         "^0\$"
 
+    # accessing directly from es pod without setting xff header will result in 401 error
     test_message="$user has improper access to .operations.* indices from es with no token"
     os::cmd::expect_success_and_text \
         "run_curl_cmd 401 curl_es_pod_with_token $esopspod '/.operations.*/_count' '' | get_count_from_json" \
         "^0\$"
```
periklis: This test and the next test in L225 assert exactly the same, don't they?

test/access_control.sh
```diff
@@ -248,8 +293,8 @@ create_user_and_assign_to_projects $LOG_USER2 $LOG_PW2 access-control-2 access-c
 oc login --username=system:admin > /dev/null
 oc project ${LOGGING_NS} > /dev/null
 
-test_user_has_proper_access $LOG_NORMAL_USER $LOG_NORMAL_PW access-control-1 access-control-2 -- access-control-3
-test_user_has_proper_access $LOG_USER2 $LOG_PW2 access-control-2 access-control-3 -- access-control-1
+test_user_has_proper_access $LOG_NORMAL_USER $LOG_NORMAL_PW access-control-1 access-control-2 access-control-3
+test_user_has_proper_access $LOG_USER2 $LOG_PW2 access-control-2 access-control-3 access-control-1
```
periklis: What do `access-control-{1,2,3}` mean? These access for user access control would benefit from proper semantic names.

test/multi_tenancy.sh
```diff
@@ -175,6 +180,69 @@ function test_user_has_proper_access() {
...
+    
+    #nrecs=$( curl_es_from_kibana $kpod $eshost "/$alias/_count" $test_token -XGET | get_count_from_json )
+    #expected=${1:-$user_project_num}
+    #if ! os::cmd::expect_success "test $nrecs = $expected" ; then
+    #    os::log::error $user cannot access alias /$alias
+    #    curl_es_from_kibana $kpod $eshost "/$alias/_count" $test_token -XGET | python -mjson.tool
+    #    oc exec -c elasticsearch $espod -- es_acl get --doc=roles
+    #    oc exec -c elasticsearch $espod -- es_acl get --doc=rolesmapping
+    #    exit 1
+    #fi
```
periklis: Can these commented block be dropped? Or is this a mistake?

test/multi_tenancy.sh
```diff
@@ -175,6 +180,69 @@ function test_user_has_proper_access() {
     fi
 }
 
+## This function tests acces to ES index/alias from kibana pod.
```
periklis: nit. `acces(s)`

test/multi_tenancy.sh
```diff
@@ -175,6 +180,69 @@ function test_user_has_proper_access() {
...
+    #    os::log::error $user cannot access alias /$alias
+    #    curl_es_from_kibana $kpod $eshost "/$alias/_count" $test_token -XGET | python -mjson.tool
+    #    oc exec -c elasticsearch $espod -- es_acl get --doc=roles
+    #    oc exec -c elasticsearch $espod -- es_acl get --doc=rolesmapping
+    #    exit 1
+    #fi
+
+    # test user has access for msearch for multiple indices
+    os::log::info See if user $user can _msearch alias "/$alias"
+    nrecs=$( { echo '{"index":'\"${alias}\"'}'; echo '{"size":0,"query":{"match_all":{}}}'; } | \
```
periklis: nit. Readbility improved unlocked if put in multiple lines, e.g. L216-219

test/multi_tenancy.sh
```diff
@@ -175,6 +180,69 @@ function test_user_has_proper_access() {
...
+        os::log::error $user cannot access "/$alias" indices with _msearch
+        {
+            echo '{"index":'\"${alias}\"'}';
+            echo '{"size":0,"query":{"match_all" : {}}}';
+        } | curl_es_pod_from_kibana_with_token_and_input $kpod $eshost "/_msearch" $test_token -XGET --data-binary @- 
+        exit 1
+    fi
+
+    # verify normal user has no access to default indices
+    os::log::info See if user $user is denied /project.default.*
```
periklis: What is the purpose to use the old index here? Same applies to L234-243

test/access_control.sh
jcantrill: No.   anything "ops" related can be removed

test/access_control.sh
```diff
@@ -33,6 +27,26 @@ function check_es_acls() {
...
+		oc delete ns/openshift-logging ns/openshift-operators-redhat --force --grace-period=0
+		for item in "ns/openshift-logging" "ns/openshift-operators-redhat"; do
+			os::cmd::try_until_failure "oc get ${item}" "$((1 * $minute))"
+		done
+	fi
+    exit $return_code
+}
+
+
+function cleanup_es() {
```
jcantrill: Do we really need this step?  Most of our tests are moving to simply removing ES entirely and creating anew

test/access_control.sh
```diff
@@ -132,85 +162,98 @@ function test_user_has_proper_access() {
...
+
+    # no token
+    os::log::info Checking access providing no username or token
+    test_message="invalid access from es with no token"
+    os::cmd::expect_success "run_curl_cmd 401 curl_es_pod_with_token $espod '/app/_count' ''"
+
+    test_message="invalid access from kibana with no token"
+    os::cmd::expect_success "run_curl_cmd 200 curl_es_from_kibana $kpod $eshost '/app/_count' ''"
 
     test_message="$user has improper access to .operations.* indices from es"
```
jcantrill: operations has been replaced by "infra-00001"

test/access_control.sh
```diff
@@ -132,85 +162,98 @@ function test_user_has_proper_access() {
...
+    test_message="invalid access from kibana with BOGUS token"
+    os::cmd::expect_success "run_curl_cmd 500 curl_es_from_kibana $kpod $eshost '/app/_count' BOGUS"
+
+    # no token
+    os::log::info Checking access providing no username or token
+    test_message="invalid access from es with no token"
+    os::cmd::expect_success "run_curl_cmd 401 curl_es_pod_with_token $espod '/app/_count' ''"
+
+    test_message="invalid access from kibana with no token"
+    os::cmd::expect_success "run_curl_cmd 200 curl_es_from_kibana $kpod $eshost '/app/_count' ''"
```
jcantrill: Any invalid response should give us something like a 401 or 403 I would think

elasticsearch/sgconfig/config.yml
```diff
@@ -4,8 +4,8 @@ opendistro_security:
       xff:
         enabled: true
         remoteIpHeader: 'x-forwarded-for'
+        internalProxies: '0\:0\:0\:0\:0\:0\:0\:1|127\.0\.0\.1'
```
ewolinetz: why are we escaping ':' ?

test/multi_tenancy.sh
jcantrill: I believe these are here to dump the access docs for when we dynamically were generating them

hack/testing/util.sh
```diff
@@ -378,15 +388,28 @@ function curl_es_pod_with_token_and_input() {
     local args=( "${@:-}" )
     local authtoken="-H Authorization: Bearer $test_token"
     local contenttype="-H Content-type:application/json"
-    local xff="-H X-Forwarded-For:127.0.0.1"
```
ewolinetz: why are we removing this if we test by curling from within the elasticsearch container? does the proxy somehow pick this request up?

test/access_control.sh
ewolinetz: +1 we can remove this -- we haven't used it in 4.x at all

test/access_control.sh
```diff
@@ -119,9 +151,7 @@ function run_curl_cmd() {
 function test_user_has_proper_access() {
     local user=$1; shift
     local pw=$1; shift
-    # rest - indices to which access should be granted, followed by --,
-    # followed by indices to which access should not be granted
-    local expected=1
+    local expected=2
```
ewolinetz: why do we still set `expected` ? it doesn't look like we still use this variable

test/multi_tenancy.sh
```diff
@@ -126,12 +129,14 @@ function test_user_has_proper_access() {
     local xfuser='-H x-forwarded-user:'"$user"
     local xocpns='-H x-ocp-namespace:'"$user_project_list"
     local xfroles='-H x-forwarded-roles:project_user'
+    local xff='-H x-forwarded-for:localhost'
```
ewolinetz: does this get expanded automatically? or will the config reject it since it expects `::1|127.0.0.1`?

hack/testing/util.sh
vimalk78: the calling test will set the header. This was added with multi-tenancy test case when es-proxy was not there. But now we have es-proxy, and multi-tenancy test case will go through es-proxy, we shouldnt add it from utils function.

test/access_control.sh
vimalk78: this helps to save time when i test locally.

test/access_control.sh
vimalk78: if removing "ops" pod here, can we remove checks which use ops pod also? 
e.g. `test_message="$user has improper access to .operations.* indices from es"` etc

test/access_control.sh
vimalk78: I wanted to discuss this. This testcase sends empty auth token, and from proxy logs i see that the user identified is kibana user. This is strange. This is why we get 200 rather than 40x.

```
time="2020-03-22T05:25:06Z" level=info msg="Serving request: &{GET /app/_count HTTP/1.1 1 1 map[User-Agent:[curl/7.29.0] Accept:[*/*] Testheader:[vimal] Authorization:[Bearer]] {} <nil> 0 [] false elasticsearch:9200 map[] map[] <nil> map[] 10.128.0.160:51642 /app/_count 0xc42047aa50 <nil> <nil> 0xc420449900}"
time="2020-03-22T05:25:06Z" level=info msg="Handling request \"authorization\""
time="2020-03-22T05:25:06Z" level=info msg="Request: &{GET /app/_count HTTP/1.1 1 1 map[User-Agent:[curl/7.29.0] X-Forwarded-User:[CN=system.logging.kibana,OU=OpenShift,O=Logging] Accept:[*/*] Testheader:[vimal] X-Forwarded-For:[localhost]] {} <nil> 0 [] false elasticsearch:9200 map[] map[] <nil> map[] 10.128.0.160:51642 /app/_count 0xc42047aa50 <nil> <nil> 0xc420449900}"
```

(I have modified proxy  to print complete incoming request, and added TestHeader from test case running locally to identify the request)

test/multi_tenancy.sh
vimalk78: the header gets compared against the regex `.*` , and is used in OD to resolve the address by creating instance of `InetSocketAddress`. So the value passed here must be usable as hostname parameter of `InetSocketAddress`.

elasticsearch/sgconfig/config.yml
vimalk78: updated.

test/access_control.sh
vimalk78: there is a difference of `xff` header. the reason for failure is different.

test/multi_tenancy.sh
vimalk78: updated

test/multi_tenancy.sh
vimalk78: need to clarify

test/multi_tenancy.sh
vimalk78: it was a mistake, uncommented. thanks!

test/multi_tenancy.sh
vimalk78: updated

test/access_control.sh
vimalk78: users are assigned different projects, and access is checked depending on the which project a user has

test/access_control.sh
vimalk78: need to clarify

test/access_control.sh
vimalk78: deleted

test/access_control.sh
jcantrill: It may be because it matches the certificates which are valid for the middleware server

elasticsearch/sgconfig/config.yml
```diff
@@ -4,8 +4,8 @@ opendistro_security:
       xff:
         enabled: true
         remoteIpHeader: 'x-forwarded-for'
+        internalProxies: '0:0:0:0:0:0:0:1|127\.0\.0\.1'
```
ewolinetz, line 7: shouldn't this be 'localhost'?
https://github.com/openshift/elasticsearch-proxy/commit/afa4328776afa3c253781217168a2c82482e7983

vimalk78: This is compared against the socket's hostAddress

https://github.com/opendistro-for-elasticsearch/security/blob/703253ed58e4471d5573e9bf950637797c534b55/src/main/java/com/amazon/opendistroforelasticsearch/security/http/RemoteIpDetector.java#L132

PR #1858, opened 2020-03-23
[fluentd] Pin gem versions

periklis:
### Motivation
Pins current dependencies as mentioned in `rh-manifest.txt`.


hack/update-fluentd-vendor-gems.sh
```diff
@@ -68,17 +68,10 @@ while read gemname gemver ; do
...
-        fi
-        if foundname=$( brew search package --exact $pkgname ) && [ "$foundname" = "$pkgname" ] ; then
-            : # package exists in brew
-        else
-            pkgname=$gemname
-        fi
+    if [ $gemname = fluentd ] ; then
+        pkgname=fluentd
+    else
+        pkgname="rubygem-${gemname}"
```
jcantrill: do we need this?  This I think is a holdover of gems being packaged in RPMs which would be unnecessary if we are vendoring and no longer relying on brew

periklis: OMG, i thought this is a hindsight of someone using brew on a macOS machine.

periklis: I fixed this script for local dev machines, where brew is missing.

fluentd/Gemfile
```diff
@@ -1,14 +1,14 @@
 gem 'fluentd', ENV['FLUENTD_VERSION']
```
jcantrill: please remove the env var here and from the Dockerfiles and we'll peg the version in the Gemfile too

periklis: AFAICS, `ENV['FLUENTD_VERSION']` is used heavily in the source code. I hesitate to delete this. We should keep the version pin in the dockerfile. WDYT?

PR #1864, opened 2020-03-28
[DO_NOT_MERGE] do not merge - intentional breaking change

bparees:


bparees:
```
elasticsearch_test.go:382: timed out waiting for Deployment elasticsearch-cdm-wo41ch2g-1: timed out waiting for the condition
```

failed as expected (elasticsearch deployment will never go ready since the init script was intentionally broken)

PR #1868, opened 2020-04-06
Bug 1825955: Updating Kibana to use opendistro security plugin multitenancy

ewolinetz:
Logging in to Kibana showed that we were missing READ permissions accessing the `.kibana` index
Note: I'm not sure if this will need to change when we move to multitenancy via Kibana as the index will likely be different.


ewolinetz:
we also seem to be missing index patterns for users... i'll add it onto here

ewolinetz:
@jcantrill @vimalk78 
we still don't have index patterns created for users yet... that might need to happen with the multitenant work... we also will need to figure out how to limit users to their kibana indices. `CRUD` is so they can create things like visualizations in their namespace.. but it also lets them delete indexpatterns i believe.

However, with an index pattern manually created i can see logs each user is only allowed to see with the following changes.

jcantrill:
> @jcantrill @vimalk78
> we still don't have index patterns created for users yet... that might need to happen with the multitenant work... 

We originally discussed this being static, atleast for admins.  @lukas-vlcek is there a way insert a document when an index is created?

> we also will need to figure out how to limit users to their kibana indices. `CRUD` is so they can create things like visualizations in their namespace.. but it also lets them delete indexpatterns i believe.

If user's are limited to their own tenant, then this isn't an issue.  If all user's are using a single tenant then I don't see how we can offer the ability to create visualizations.  For starters we may need to limit them to only querying for logs and possibly even identifying how to make that happen for them.

> 
> However, with an index pattern manually created i can see logs each user is only allowed to see with the following changes.

elasticsearch/sgconfig/roles.yml
```diff
@@ -168,13 +135,20 @@ project_user:
   readonly: true
   cluster:
     - CLUSTER_COMPOSITE_OPS_RO
+    - indices:data/write/bulk  #required for being able to let index mappings update... is this required still with multitenancy?
   indices:
     app:
       '*':
         - READ
-      _dls_: "{\"bool\":{\"filter\":{\"script\":{\"script\":{\"lang\":\"painless\",\"params\":{\"param1\":\"${attr.proxy.ns}\"},\"source\":\"String namespace = doc['kubernetes.namespace_name.keyword'][0];StringTokenizer st = new StringTokenizer(params.param1,\\\",\\\");while (st.hasMoreTokens()){if (st.nextToken().equalsIgnoreCase(namespace)){return true;}}return false;\"}}}}}"
-    logs-app:
```
vimalk78, line 176: multi-tenancy test is failing because the `logs-app` alias is removed from index patterns.

The test adds `logs-app` alias to the created index, and not both `app` and `logs-app`. So if possible can you please modify the test case to add both alias? [here](https://github.com/openshift/origin-aggregated-logging/blob/master/test/multi_tenancy.sh#L111)
Thx!!

@jcantrill

vimalk78: or let me know i can send a tiny PR for this change ahead of this PR merging

ewolinetz: @vimalk78 i will fix the tests -- we don't need both 'app' and 'logs-app' for it to work in kibana.

ewolinetz:
seeing an issue where plugin status gets stuck with the following after a kibana restart:
`plugin:opendistro_security@6.8.1 | Setting up index template.`

need to investigate further

ewolinetz:
> seeing an issue where plugin status gets stuck with the following after a kibana restart

I believe this is stemming from the fact that the kibana server is being assigned the `project_user` role and it lacks necessary permissions...

ewolinetz:
need to rebase after https://github.com/openshift/origin-aggregated-logging/pull/1879 merges

ewolinetz:
smoke testing seems to be failing due to lack of being able to deploy CLO

PR #1869, opened 2020-04-07
[kibana] Fix config options override per env variable

periklis:
### Description
This PR addresses a hindsight in the `kibana/utils` script that is responsible to override the kibana configuration options from environment variables. However, this script was broken per se because it appended the updates at the end of the yaml file resulting in yaml parser exceptions.

### Background
To address kibana NS fixes for E2E testing on random NS: openshift/elasticsearch-operator#305

### DIFF
--- a/kibana/utils
+++ b/kibana/utils
@@ -57,21 +57,6 @@ update_config_from_env_vars() {
       server.ssl.certificate
       server.ssl.enabled
       server.ssl.key
-      xpack.apm.enabled
-      xpack.ccr.enabled
-      xpack.graph.enabled
-      xpack.grokdebugger.enabled
-      xpack.ilm.enabled
-      xpack.infra.enabled
-      xpack.infra.sources.default.logAlias
-      xpack.maps.enabled
-      xpack.ml.enabled
-      xpack.monitoring.enabled
-      xpack.remote_clusters.enabled
-      xpack.reporting.enabled
-      xpack.searchprofiler.enabled
-      xpack.spaces.enabled
-      xpack.uptime.enabled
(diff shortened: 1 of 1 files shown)


kibana/utils
```diff
@@ -85,7 +85,7 @@ update_config_from_env_vars() {
         if [ -n "${DEBUG:-}" ] ; then
           echo "updating ${conf_file} - '${kibana_var}: ${value}'"
         fi
-        echo "${kibana_var}: ${value}" >> ${conf_file}
+        sed -i "s,${kibana_var}:.*,${kibana_var}:\ ${value}," "${conf_file}"
```
jcantrill, line 88: This only will work if the supported var is actually defined in the config file.  What if it is not?

periklis: Good catch we need to add this logic too. However it would never have worked with the old append, because it destroy the very last config option if any available or amend a comment line.

ewolinetz:
We shouldn't be supporting any `x-pack` features since we use OpenDistro...
lets remove the following from the allowed vars
```
      xpack.apm.enabled
      xpack.ccr.enabled
      xpack.graph.enabled
      xpack.grokdebugger.enabled
      xpack.ilm.enabled
      xpack.infra.enabled
      xpack.infra.sources.default.logAlias
      xpack.maps.enabled
      xpack.ml.enabled
      xpack.monitoring.enabled
      xpack.remote_clusters.enabled
      xpack.reporting.enabled
      xpack.searchprofiler.enabled
      xpack.spaces.enabled
      xpack.uptime.enabled
```

PR #1877, opened 2020-04-15
Bug 1822686: Require python2-pip in vendoring and removing python-thrift

jcantrill:
Additional change to support https://issues.redhat.com/browse/LOG-690 for downstream


jcantrill:
retest

PR #1879, opened 2020-04-16
Bug 1826096: Fixing failing image tests

ewolinetz:


bparees:
@ewolinetz this is going to need to be backported to our other branches, right?  since the repo is broken everywhere.  might as well create a BZ.

ewolinetz:
these changes are also contained within https://github.com/openshift/origin-aggregated-logging/pull/1868

PR #1880, opened 2020-04-17
Bug 1825275: Modify Dockerfile to mitigate issue with buildah

jcantrill:
https://bugzilla.redhat.com/show_bug.cgi?id=1825275


russellb:
@jcantrill This doesn't seem to refer to the right bug.  This bug only describes COPY not working with relative paths.  There must be another bug that refers to the variable expansion issue?  Can you look at getting the right bug in place?

It would also be very helpful to have more information in the commit message.  "Modify Dockerfile to mitigate issue with buildah" doesn't really provide any information about what the problem was and why this change is the correct way to resolve it.

PR #1881, opened 2020-04-17
Revert "Merge pull request #1862 from periklis/fluentd/fix-centos7-build"

periklis:
This reverts commit 83bcc676757ed6f451847ec32d85b519701d8a96, reversing changes made to 797ed84f8a00a06c3330369607d8030c7040ca8b.

cc @lukas-vlcek

### DIFF
--- a/fluentd/Dockerfile.centos7
+++ b/fluentd/Dockerfile.centos7
@@ -14,13 +14,6 @@ ENV DATA_VERSION=1.6.0 \
 # iproute needed for ip command to get ip addresses
 # autoconf redhat-rpm-config for building jemalloc
 USER 0
-
-#
-# Replace community build service repo with actuall SCLO mirror
-#
-RUN yum-config-manager --disable cbs.centos.org_repos_sclo7-rh-ruby25-rh-candidate_x86_64_os_ && \
-    yum-config-manager --add-repo http://mirror.centos.org/centos/7/sclo/x86_64/rh/rh-ruby25/
-
 RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
     yum install -y --setopt=tsflags=nodocs \
       make \


ewolinetz:
@periklis i noticed that this needed to be removed/commented out for fluentd to build correctly... why do we think we need to revert this?

periklis:
> @periklis i noticed that this needed to be removed/commented out for fluentd to build correctly [1]... why do we think we need to revert this?
> 
> [1] #1879

When #1862 was introduced we had for a long time broken builds because the centos7 image relied on a CBS mirror, which was broken. Thus, the PR introduced a mirror change for the same SCLO packages. However, it seems the CBS is back and the other mirror is gone.

ewolinetz:
@periklis ah, i misread what this change was... thanks!
/lgtm

jcantrill:
> > @periklis i noticed that this needed to be removed/commented out for fluentd to build correctly [1]... why do we think we need to revert this?
> > [1] #1879
> 
> When #1862 was introduced we had for a long time broken builds because the centos7 image relied on a CBS mirror, which was broken. Thus, the PR introduced a mirror change for the same SCLO packages. However, it seems the CBS is back and the other mirror is gone.

To add to this, I'm not certain why we ever added this change.  The image itself is an SCL image so there should have never been a need to enable a repo to install ruby 25.

bparees:
@periklis can we close this in deference to https://github.com/openshift/origin-aggregated-logging/pull/1868 ?

periklis:
@bparees Sure, if @ewolinetz could just delete the comments, i have no problem: https://github.com/openshift/origin-aggregated-logging/pull/1868/commits/c6c32db2cd12aa90bd83cb48fa50cfc1a12681f9#diff-7fb10660751a8f79a8556265ef5caa95R21

ewolinetz:
@periklis done

PR #1882, opened 2020-04-21
Fluent plugins to support rfc5424 params

vimalk78:
- added appname, procid, msgid to syslog plugin
 - modified parse_json_field filter to support replacing parsed json object in place
 - added method to map facility/severity string values
 - temporarily removed length restriction on syslog tag
 - moved following plugins from vendored_gem_src to lib
   - fluent-plugin-remote_syslog
   - remote_syslog_sender
   - syslog_protocol


fluentd/lib/filter_parse_json_field/lib/filter_parse_json_field.rb
```diff
@@ -39,17 +40,22 @@ def configure(conf)
     end
 
     def filter_stream(tag, es)
-      return es unless @merge_json_log
-
+      return es unless @merge_json_log or @replace_json_log
```
jcantrill, line 43: use `||` instead of `or`

fluentd/lib/filter_parse_json_field/lib/filter_parse_json_field.rb
```diff
@@ -39,17 +40,22 @@ def configure(conf)
...
-        new_es.add(time, record)
-      }
-
+      if @merge_json_log
+        es.each { |time, record|
+          record = do_merge_json_log(record)
+          new_es.add(time, record)
+        }
+      end
+      if @replace_json_log
```
jcantrill, line 51: IDoes this do what you expect:  ` new_es.add(time, record)` puts the record into the stream.  This means if you have both merge and replace enabled you will apply the logic separately to the record and add it to the stream twice.  My expectation is that we would do whatever logic and then at the end put the message into the stream.

fluentd/lib/filter_parse_json_field/lib/filter_parse_json_field.rb
```diff
@@ -72,5 +78,23 @@ def do_merge_json_log(record)
       record
     end
 
+    def do_replace_json_log(record)
+      json_fields.each do |merge_json_log_key|
+        if record.has_key?(merge_json_log_key)
+          value = record[merge_json_log_key].strip
+          if value[0].eql?('{') && value[-1].eql?('}')
```
jcantrill, line 85: We should prefer `starts_with?` and `ends_with?` https://ruby-doc.org/core-2.5.1/String.html#method-i-start_with-3F

fluentd/vendored_gem_src/fluent-plugin-remote_syslog/lib/fluent/plugin/out_remote_syslog.rb
```diff
@@ -93,13 +98,35 @@ def write(chunk)
...
         facility = extract_placeholders(@facility, chunk.metadata)
+        facility = sanitize_facility(facility)
         severity = extract_placeholders(@severity, chunk.metadata)
+        severity = sanitize_severity(severity)
         program = extract_placeholders(@program, chunk.metadata)
         hostname = extract_placeholders(@hostname, chunk.metadata)
 
-        packet_options = {facility: facility, severity: severity, program: program}
+        packet_options = {}
+        if facility == nil || facility.length == 0
```
jcantrill, line 108: `if facility.nil? || facility.strip.empty?`  or even consider something like 
```
packet_options = {facility: facility, severity: severity, program: program}
 packet_options[:facility] = 'user' if  packet_options[:facility].nil? ||  packet_options[:facility].strip.empty?
```

fluentd/vendored_gem_src/fluent-plugin-remote_syslog/lib/fluent/plugin/out_remote_syslog.rb
```diff
@@ -93,13 +98,35 @@ def write(chunk)
...
+          packet_options[:severity] = 'notice'
+        else
+          packet_options[:severity] = severity 
+        end
+        packet_options[:program] = program
+
         packet_options[:hostname] = hostname unless hostname.empty?
         packet_options[:rfc] = @rfc
+        if @rfc == :rfc5424
+          appname = extract_placeholders(@appname, chunk.metadata) unless @appname == nil || @appname.length == 0
```
jcantrill, line 123: `nil?`

fluentd/vendored_gem_src/syslog_protocol/lib/syslog_protocol/packet.rb
```diff
@@ -7,6 +7,7 @@ class Packet
 
     def initialize
       @rfc = :rfc3164
+      @appname = 'fluentd' ##TODO(vimalk78) change it to NILVALUE
```
jcantrill, line 10: does this need to be addressed now?

fluentd/vendored_gem_src/syslog_protocol/lib/syslog_protocol/packet.rb
```diff
@@ -68,7 +68,8 @@ def tag=(t)
       unless t && t.is_a?(String) && t.length > 0
         raise ArgumentError, "Tag must not be omitted"
       end
-      if t.length > 32
+      #if t.length > 32
+      if t.length > 500
```
jcantrill, line 72: Does it make sense to add replace this with a constant someplace which is more descriptive?

fluentd/vendored_gem_src/syslog_protocol/lib/syslog_protocol/packet.rb
```diff
@@ -112,6 +113,55 @@ def hostname=(h)
       @hostname = h
     end
 
+    def appname=(a)
+      unless a and a.is_a? String and a.length > 0
```
jcantrill, line 117: prefer `&&` over `and`

fluentd/vendored_gem_src/syslog_protocol/lib/syslog_protocol/packet.rb
```diff
@@ -112,6 +113,55 @@ def hostname=(h)
...
+        raise ArgumentError.new("Appname may not be omitted")
+      end
+      #if a.length > 48
+      if a.length > 500
+        raise ArgumentError, "Appname must not be longer than 48 characters"
```
jcantrill, line 122: match app length validation to comparison

fluentd/vendored_gem_src/syslog_protocol/lib/syslog_protocol/packet.rb
```diff
@@ -112,6 +113,55 @@ def hostname=(h)
...
+      end
+      if a =~ /\s/
```
jcantrill, line 124: Not certain if the regex captures all whatespace but may be more descriptive to use `include?` https://ruby-doc.org/core-2.5.1/String.html#method-i-include-3F

vimalk78: The next regex after this checks for allowed values as per rfc. So this already excludes space (`0x20`).
Perhaps `/\s/` regex's intention is not to check for all whitespace characters but check for a common case of `' '` character. This check is actually not needed.

```
      if a =~ /[^\x21-\x7E]/      
        raise ArgumentError.new("Appname may only contain ASCII characters 33-126")      
      end      
```

fluentd/vendored_gem_src/syslog_protocol/lib/syslog_protocol/packet.rb
vimalk78: This is se to value `'fluentd'` because the tests in CLO master repo checks for this.

fluentd/vendored_gem_src/fluent-plugin-remote_syslog/lib/fluent/plugin/out_remote_syslog.rb
vimalk78: updated

fluentd/lib/filter_parse_json_field/lib/filter_parse_json_field.rb
vimalk78: changed it to `if..elsif..` logic. `@replace_json_log` will not run if `@merge_json_log` is enabled.

vimalk78:
```
level=error msg="Error: Unable to find matching route for Route Table (rtb-0cd1b57b3f0fb99b2) and destination CIDR block (0.0.0.0/0)."
level=error
level=error msg="  on ../tmp/openshift-install-125822389/vpc/vpc-public.tf line 34, in resource \"aws_route\" \"igw_route\":"
level=error msg="  34: resource \"aws_route\" \"igw_route\" {"
```
/retest

PR #1890, opened 2020-05-06
Bug 1809498: Limit shards for security index to fix upgrade to 4.5

jcantrill:
This PR

* fixes the number of the security index replicas based on cluster nodes so as to not block upgrade where it is unable to land a shard

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1809498

### DIFF
--- a/elasticsearch/utils/es_seed_acl
+++ b/elasticsearch/utils/es_seed_acl
@@ -6,6 +6,13 @@ ES_CLUSTER_HOST=${ES_CLUSTER_HOST:-"localhost"}
 ES_CLUSTER_PORT=${ES_CLUSTER_PORT:-9300}
 MAX_WAIT_SEED=${MAX_WAIT_SEED:-$((60*60*24*7))} #one week
 
+NODES=$(es_util --query=_cat/nodes | wc -l)
+if [ $NODES -ge 2 ] ; then
+  SECURITY_SHARDS=1
+else
+  SECURITY_SHARDS=0
+fi
+
 info "Seeding the security ACL index.  Will wait up to $MAX_WAIT_SEED seconds."
 
 function sgadmin {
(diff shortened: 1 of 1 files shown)


elasticsearch/utils/es_seed_acl
```diff
@@ -6,6 +6,13 @@ ES_CLUSTER_HOST=${ES_CLUSTER_HOST:-"localhost"}
 ES_CLUSTER_PORT=${ES_CLUSTER_PORT:-9300}
 MAX_WAIT_SEED=${MAX_WAIT_SEED:-$((60*60*24*7))} #one week
 
+NODES=$(es_util --query=_cat/nodes | wc -l)
```
ewolinetz, line 9: will this work how we anticipate? isn't there a chance we haven't yet joined the cluster at this point in time during the node's initial initialization?

jcantrill: my quick test for an "error" condition would produce a value of zero meaning we would have zero replicas.  Given we run this N times for each node it should settle on a value of 1 for 3 or more nodes.  Ideally the operator will control this properly when it inits the cluster

elasticsearch/utils/es_seed_acl
```diff
@@ -6,6 +6,13 @@ ES_CLUSTER_HOST=${ES_CLUSTER_HOST:-"localhost"}
...
+if [ $NODES -ge 3 ] ; then
```
ewolinetz, line 10: this should probably be `2`, right?

jcantrill: I thought we said 3 but one replica for a two node cluster would be valid, though is it not bad practice to have a 2 node cluster because we would have 2 masters?

ewolinetz: thats for a forming a quorum of masters, completely different than having redundancy of shards

PR #1892, opened 2020-05-10
Migrate image building to a single UBI7-based Dockerfile

periklis:
This PR addresses the issue of maintaining two Dockerfiles per component, one based on internal hosted rhel7 base images and one based on public available centos7 base images. The selected base image is UBI7 (based on RHEL7) or available variants for `ruby-2.5`, `nodejs-10` and `python-36`. In addition UBI7 images are public available.

**WARNING:** This PR needs a change in openshift/release to address the base imagestream availability for E2E tests.

**Additional information**
- The only exception to use a single dockerfile is the elasticsearch image. It is currently not possible to instrument the build with different args accross local builder, CI and ART. Nevertheless both Dockerfiles are based upon ubi7
- The fluentd image includes an optimized installation for jemalloc.


curator/Dockerfile
```diff
@@ -1,43 +1,52 @@
...
-    ls /usr/local/bin/curator
+RUN pip install $(ls . | grep -v curator) -q --no-index --find-links . && \
+    pip install elasticsearch-curator* --no-index -q && \
+    rm -rf $HOME/vendor
 
 WORKDIR ${HOME}
 USER 1001
 CMD ["sh", "run.sh"]
+
+LABEL io.k8s.description="Curator elasticsearch container for elasticsearch deletion/archival" \
```
jcantrill, line 43: To take advantage of the layering wouldnt we prefer to move this higher up?

elasticsearch/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM openshift/ose-base:ubi7
+FROM ubi7:latest
```
jcantrill, line 1: same question here about what does ART/CI use?

elasticsearch/Dockerfile.origin
```diff
@@ -0,0 +1,72 @@
+FROM ubi7:latest
```
jcantrill, line 1: If we intend to remove the centos image what is the advantage of having Dockerfile.origin in lieu of the centos one?

curator/Dockerfile
```diff
@@ -1,43 +1,52 @@
-FROM rhel7:7-released
+FROM ubi7/python-36:latest
```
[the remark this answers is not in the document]
syedriko: ART's using openshift/ose-base:ubi7

elasticsearch/Dockerfile
syedriko: openshift/ose-base:ubi7

elasticsearch/Dockerfile.origin
periklis: The `Dockerfile.centos7` names are just symbolic links to satisfy the trampoline merges with openshift/release/pull/9136. Need to be removed after merging the later. Elasticsearch is the only project we can't use a single Dockerfile due to the ES plugin downloads from brew in the production Dockerfile. I renamed the developer Dockerfile to `Dockerfile.origin` because thi matches the prepare install scripts `prep-install.prod` vs. `prep-install.origin`.

fluentd/Dockerfile
```diff
@@ -34,73 +25,35 @@ ARG TEST_REPO
...
-  INSTALL_PKGS="hostname \
-                bc \
-                iproute" && \
-  SCL_PKGS="rh-ruby25 rh-ruby25-runtime rh-ruby25-ruby rh-ruby25-rubygem-openssl \
-            rh-ruby25-rubygem-json rh-ruby25-ruby-libs rh-ruby25-rubygems \
-            rh-ruby25-ruby-irb" && \
-  yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS $SCL_PKGS && \
-  rpm -V $INSTALL_PKGS $SCL_PKGS && \
-  yum clean all
+ADD jemalloc-5.2.1-2.el8.x86_64.rpm jemalloc-5.2.1-2.el8.x86_64.rpm
```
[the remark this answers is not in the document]
periklis: This is an addition to speedup the fluentd build. We can keep this out from this PR.

curator/Dockerfile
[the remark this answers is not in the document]
periklis: Isn't that the same? Just only a different registry/namespace and tagging strategy?

curator/Dockerfile
```diff
@@ -1,43 +1,52 @@
...
 ENV HOME=/opt/app-root/src \
     ES_HOST=localhost \
     ES_PORT=9200 \
     ES_CA=/etc/curator/keys/ca \
-    ES_CLIENT_CERT=/etc/curator/keys/cert \
-    ES_CLIENT_KEY=/etc/curator/keys/key \
-    CURATOR_CONF_LOCATION=/etc/curator/settings/config.yaml \
-    CURATOR_CONF_FILE=/etc/curator/settings/curator5.yaml \
-    CURATOR_ACTIONS_FILE=/etc/curator/settings/actions.yaml \
+    ES_CLIENT_CERT=/opt/app-root/etc/curator/keys/cert \
```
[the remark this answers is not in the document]
periklis: Indeed this is a mistake.

fluentd/Dockerfile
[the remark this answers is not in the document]
periklis: Removed for now. We should add this in a separate PR if needed

curator/Dockerfile
periklis: Can you elaborate how layering is affected here? I can't really catch your point.

curator/Dockerfile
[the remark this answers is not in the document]
syedriko: I don't know. I'm assuming it's different until shown otherwise.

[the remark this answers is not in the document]
sosiouxme: `openshift/ose-base:ubi7` is for all practical purposes the same as ubi7 - it may occasionally have updates but otherwise same contents.

It certainly doesn't include python36.

curator/Dockerfile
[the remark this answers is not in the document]
sosiouxme: The minimal required set will be set by ART and OSBS. Ones like this that you supply, though, it's up to you whether you want them in the image or whether they still serve any purpose (for instance displaying in OLM?). All labels are shown in the container catalog on the portal, and sometimes it's confusing seeing labels set by the base image.

curator/Dockerfile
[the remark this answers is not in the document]
periklis: Yes the UBI7 flavors for ruby-25, python-36 and nodejs-10 need to be added in a follow-up in the ocp-build-data `streams.yaml`

elasticsearch/Dockerfile
periklis: As ART answered above the images are the same.

periklis:
@jcantrill PTAL a look again and resolve/mark anything missing to move this forward. Merging this enables pending follow ups PRs: openshift/release#9136 and openshift/ocp-build-data#522

curator/Dockerfile
jcantrill: one layer builds upon the next so less likely to change layers should come before more likely ones meaning labels should almost be first

curator/Dockerfile
```diff
@@ -16,28 +16,38 @@ ENV HOME=/opt/app-root/src \
...
+    rm -rf $HOME/vendor
 
 WORKDIR ${HOME}
 USER 1001
 CMD ["sh", "run.sh"]
+
+LABEL io.k8s.description="Curator elasticsearch container for elasticsearch deletion/archival" \
+    io.k8s.display-name="Curator ${CURATOR_VER}" \
+    io.openshift.tags="logging,efk,elasticsearch,curator" \
+    com.redhat.component="logging-curator${CURATOR_VER}-container" \
```
jcantrill, line 47: @sosiouxme is it correct to say this label and all those after it can be removed?

elasticsearch/Dockerfile
```diff
@@ -57,3 +57,14 @@ RUN ${HOME}/install.sh && rm -rf /artifacts
 WORKDIR ${HOME}
 USER 1000
 CMD ["sh", "/opt/app-root/src/run.sh"]
+
+LABEL io.k8s.description="Elasticsearch container for storing and aggregating logs" \
+    io.k8s.display-name="Elasticsearch ${ES_VER}" \
+    io.openshift.tags="logging,efk,elasticsearch" \
+    com.redhat.component="logging-elastiscearch6-container" \
```
jcantrill, line 64: @sosiouxme same here?

elasticsearch/Dockerfile.origin
jcantrill: I think the naming change make sense if we were to keep it but if the intention is to remove it in future the change is not required and just lends to add noise to the PR

periklis: FYI, my mistake yesterday, the `Dockerfile.origin` refers to the Dockerfile used by our upstream CI here. AFAIK there is a way without friction for local development and upstream at the same time to access the private koji repos via `fetch-artifacts-koji.yaml`. Therefore we need to keep the two Dockerfile distinct for elastisearch until we figure out how to build elasticsearch incl. all deps and plugins through vendoring on our own. 

In summary:
- `Dockerfile` is the target for our downstream CI used by ART to build images using `fetch-artifacts-koji.yaml`
- `Dockerfile.origin` is the target for our local development and upstream CI and replaces the old centos7 baked image.
- Both files are identical except of the parameters `MAVEN_REPO_URL`, `PROMETHEUS_EXPORTER_VER` and `OPENDISTRO_VER`

I suggest the state as is for now to land the UBI 7 streamlining and re-evaluate a single-Dockerfile strategy for elasticsearch at a later stage.

periklis:
@jcantrill Regarding labels, according to ART team the following labels are provided by the downstream CI:
- com.redhat.*
- version
- release
- name

Everything else can stay in our Dockerfiles but any duplication with ocp-build-data results in ocp-build-data taking precedence over our Dockerfiles, e.g.:

[Fluentd Dockerfile](https://github.com/openshift/origin-aggregated-logging/pull/1892/commits/6c1e5b4ef62634b753657498aed0133698f56fa0#diff-0a8d3a0f4d1455bc2bc2a083ab0a0a4bR5) vs. [ocp-build-data/logging-fluentd](https://github.com/openshift/ocp-build-data/blob/openshift-4.6/images/logging-fluentd.yml#L19)

Since both place are the same, I suggest we delete the labels in our Dockerfiles and keep them only in the ocp-build-data repo? WDYT?

(See https://coreos.slack.com/archives/CB95J6R4N/p1592914800037400)

fluentd/source.jemalloc
```diff
@@ -1,3 +1,3 @@
-JEMALLOC_VER=5.2.0
-JEMALLOC_SOURCE=https://github.com/jemalloc/jemalloc/releases/download/${JEMALLOC_VER}/jemalloc-${JEMALLOC_VER}.tar.bz2
-JEMALLOC_SHA512SUM=e3be4d534770126caf10f2684aed9fe4ba1422dd47625fe50343cfb750f26eff869fcc7d1e30a96dd6c73f6614c4bbcd560fd24fc26b55ac731c43e60fd05234
+JEMALLOC_VER=5.2.1
```
igor-karpukhin, line 1: `JEMALLOC_VER` is unused?

periklis: Good catch I can fix this in a follow-up PR no deal breaker for now.

PR #1893, opened 2020-05-13
Bug 1836450: Restricting which es node can seed -- limiting to just elected master

ewolinetz:
Updating the elasticsearch seeding pattern to only let the elected master do the seeding.
Also fixed the readiness probe to match this pattern.

### DIFF
--- a/elasticsearch/Dockerfile
+++ b/elasticsearch/Dockerfile
@@ -30,7 +30,8 @@ RUN packages="java-${JAVA_VER}-openjdk-headless \
               hostname \
               openssl \
               zip \
-              unzip" && \
+              unzip \
+              jq" && \
     yum install -y --setopt=tsflags=nodocs ${packages} && \
     rpm -V ${packages} && \
     yum clean all
--- a/elasticsearch/Dockerfile.centos7
+++ b/elasticsearch/Dockerfile.centos7
@@ -25,12 +25,15 @@ ENV ES_PATH_CONF=/etc/elasticsearch/ \
 
 ARG MAVEN_REPO_URL=https://repo1.maven.org/maven2/
 
-RUN packages="java-${JAVA_VER}-openjdk-headless \
+RUN yum install -y --setopt=tsflags=nodocs --nogpgcheck \
+        https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm && \
+    packages="java-${JAVA_VER}-openjdk-headless \
               PyYAML  \
               hostname \
               openssl \
               zip \
-              unzip" && \
+              unzip \
+              jq" && \
     yum install -y --setopt=tsflags=nodocs ${packages} && \
     rpm -V ${packages} && \
     yum clean all
--- a/elasticsearch/init.sh
+++ b/elasticsearch/init.sh
@@ -18,19 +18,27 @@
 source "logging"
 
 wait_for_port_open
-failed=0
 
-touch ${HOME}/init_running
-pushd "${ES_HOME}/init"
-  files=$(ls --hide common | sort)
-  for init_file in ${files}; do
-    if [ -f "${init_file}" ] ; then
-      ./"${init_file}" || { failed=1 ; echo "failed init: ${init_file}" >> ${HOME}/init_failures ; }
-    fi
-  done
-popd
+# find out who the master of the cluster is
+master_node=$(get_master_node)
 
-rm -f ${HOME}/init_running
-if [ $failed -eq 0 ]; then
-  touch ${HOME}/init_complete
+# if we're the master -- do the init
+if [[ "$master_node" == "$DC_NAME" ]]; then
+
+  failed=0
+
+  touch ${HOME}/init_running
+  pushd "${ES_HOME}/init"
+    files=$(ls --hide common | sort)
+    for init_file in ${files}; do
+      if [ -f "${init_file}" ] ; then
+        ./"${init_file}" || { failed=1 ; echo "failed init: ${init_file}" >> ${HOME}/init_failures ; }
+      fi
+    done
+  popd
+
+  if [ $failed -eq 1 ]; then
+    show_failures
+  fi
+
 fi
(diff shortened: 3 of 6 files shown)


ewolinetz:
WIP because i need to double check the es6 security index paths

elasticsearch/Dockerfile.centos7
```diff
@@ -25,12 +25,15 @@ ENV ES_PATH_CONF=/etc/elasticsearch/ \
...
-RUN packages="java-${JAVA_VER}-openjdk-headless \
+RUN yum install -y --setopt=tsflags=nodocs --nogpgcheck \
+        https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm && \
+    packages="java-${JAVA_VER}-openjdk-headless \
               PyYAML  \
               hostname \
               openssl \
               zip \
-              unzip" && \
+              unzip \
```
jcantrill, line 35: Need to confirm jq is available for in the past it was not

elasticsearch/init.sh
```diff
@@ -18,19 +18,27 @@
...
-    fi
-  done
-popd
-
-rm -f ${HOME}/init_running
-if [ $failed -eq 0 ]; then
-  touch ${HOME}/init_complete
+
+# find out who the master of the cluster is
+master_node=$(get_master_node)
```
jcantrill, line 23: Do we have any concerns about the outcome here?  Is there possibly a delay between the running of the script and the election of the master?

elasticsearch/init/0010-seed-acl
```diff
@@ -23,9 +23,8 @@ index=$( get_security_conf_index )
 if [ -n "${force}" ] ; then
   docs=(roles rolesmapping actiongroups internalusers config)
   for d in $docs; do
-    resp=$(es_util --query=${index}/$d/0 -w %{response_code} | cut -d '}' -f2)
+    if [ check_index_exists "${index}/security/$d" -ne 0 ]; then
```
jcantrill, line 26: Presumably the doctype changes in future.  Any chance they made it easy for us to migrate in the 6.x version of the security code?

elasticsearch/probe/readiness.sh
```diff
@@ -59,5 +61,19 @@ function show_failures() {
...
+for d in $docs
+do
+    check_index_exists "${SECURITY_INDEX}/security/${d}"
+done
+
+for template_file in ${ES_HOME}/index_templates/*.json
+do
+    template=`basename $template_file`
+    check_index_exists "_template/${template}"
+done
```
jcantrill, line 79: So readiness no longer depends on file exists but a exit 0 of this script which is:  security docs exist, templates exist

elasticsearch/Dockerfile.centos7
ewolinetz: i had to add the epel to get it for centos

elasticsearch/init.sh
ewolinetz: not sure... since this is after `wait_for_port_open` i would guess its a small window.. but I could always add logic in that function to ensure we have a value first

elasticsearch/probe/readiness.sh
ewolinetz: yeah, we were creating files based on if we finished seeding... as long as its available to the cluster we should consider a node ready (and we can get to it)

jcantrill:
/hold
pending understanding if we will have `jq`

elasticsearch/init/0010-seed-acl
ewolinetz: I didn't see anything that let us configure the `type` they used

elasticsearch/init.sh
ewolinetz: i just saw this on my cluster... 
```
jq: error: CShWUf/0 is not defined at <top-level>, line 1:
.nodes.apRFr1-CShWUf-_ZIQOWbw.name              
jq: error: _ZIQOWbw/0 is not defined at <top-level>, line 1:
.nodes.apRFr1-CShWUf-_ZIQOWbw.name                     
jq: 2 compile errors
(23) Failed writing body
```
I'll try to add some hardening

ewolinetz:
failing test dependent on https://github.com/openshift/elasticsearch-operator/pull/355

elasticsearch/utils/logging
```diff
@@ -69,6 +69,42 @@ get_kibana_index_name() {
...
+#      "name" : "elasticsearch-cdm-h5el9dsq-2",
+#      "ephemeral_id" : "zWCXff-eQwi2YLMpZ9qbqA",
+#      "transport_address" : "10.131.0.17:9300",
+#      "attributes" : { }
+#    }
+#  }
+#}
+get_master_node() {
+    local master=$(es_util --query="_cluster/state/master_node?pretty" | jq -r .master_node)
+    echo $(es_util --query="_cluster/state/nodes?pretty" | jq -r .nodes[\"${master}\"].name)
```
ewolinetz, line 98: this can be condensed into a single ES curl
`_cluster/state/master_node,nodes?pretty`

elasticsearch/Dockerfile
```diff
@@ -30,7 +30,8 @@ RUN packages="java-${JAVA_VER}-openjdk-headless \
               hostname \
               openssl \
               zip \
-              unzip" && \
+              unzip \
+              jq" && \
```
lukas-vlcek, line 34: I thought we were not recommended to pull in `jq`?

ewolinetz: not that i know of... ART has built it for other images

jcantrill: > I thought we were not recommended to pull in `jq`?

I believe this was only because it was not available but has since been productized

elasticsearch/utils/logging
lukas-vlcek: Or even easier using `_cat/master?h=id`.
This gives you just the ID of the master node. Or use [`help`](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/cat.html#help) to list all available columns.

See: https://www.elastic.co/guide/en/elasticsearch/reference/6.8/cat-master.html

PR #1897, opened 2020-05-19
Bug 1840274: Add script to let fluentd check ES version

ewolinetz:
This is part of addressing the concern where CLO upgrades before EO does.
With this script we can create an init container for Fluentd that will call this script (it will cause the main container to wait until the correct ES version is found) which can be passed via a command arg

Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1840274

### DIFF
--- a/fluentd/wait_for_es_version.rb
+++ b/fluentd/wait_for_es_version.rb
@@ -1,0 +1,38 @@
+#! /usr/bin/env ruby
+
+require 'net/https'
+require 'json'
+
+if ARGV.length != 2
+  abort "Usage: wait_for_es_version.rb <min_es_version> <es_url>"
+end
+
+min_version = ARGV[0]
+es_url = ARGV[1]
+
+uri = URI(es_url)
+
+cert_dir = "/etc/fluent/keys/"
+
+fluentd_ca_file = File.join(cert_dir, "ca-bundle.crt")
+
+client_file = File.join(cert_dir, "tls.crt")
+client_cert = OpenSSL::X509::Certificate.new File.read client_file
+
+key_file = File.join(cert_dir, "tls.key")
+client_key = OpenSSL::PKey::RSA.new File.read key_file
+
+
+Net::HTTP.start(uri.host, uri.port, :use_ssl => true, :ca_file => fluentd_ca_file, :cert => client_cert, :key => client_key) do |http|
+  request = Net::HTTP::Get.new uri
+  response = http.request request
+
+  hash = JSON.parse(response.body)
+
+  version = hash["version"]["number"]
+
+  if version.to_f < min_version.to_f
+    warn "Elasticsearch is currently version: #{version} - Expecting it to be at least: #{min_version}"
+    exit 1
+  end
+end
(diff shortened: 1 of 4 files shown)


jcantrill:
> This is part of addressing the concern where CLO upgrades before EO does.
> With this script we can create an init container for Fluentd that will call this script (it will cause the main container to wait until the correct ES version is found) which can be passed via a command arg

We need to consider this approach more carefully.  Checking for version does not guarantee success in all cases just because of a certain version.  In our case for instance, index management fails if and index already exists witha '-write' suffix; this isn't tied to server version

ewolinetz:
@jcantrill no, but having ES6 does mean that it is using the config where it blocks auto creation of the `*-write` index at which point once IM has created the alias fluentd logs will flow as normal.

fluentd/wait_for_es_version.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
jcantrill, line 1: Given you already have ruby available, consider writing the entire script as ruby instead of shuffling back and forth to bash https://ruby-doc.org/stdlib-2.4.1/libdoc/net/http/rdoc/Net/HTTP.html

ewolinetz: my thought was it was easy to do the curl in bash vs implementing configuring and using a http client in ruby

fluentd/wait_for_es_version.sh
```diff
@@ -0,0 +1,3 @@
+#! /bin/bash
+
+ruby wait_for_es_version.rb $@
```
ewolinetz, line 3: we need to do this because of how ruby is added to the image. this will be the entry point for the init container

fluentd/wait_for_es_version.rb
```diff
@@ -0,0 +1,38 @@
...
+Net::HTTP.start(uri.host, uri.port, :use_ssl => true, :ca_file => fluentd_ca_file, :cert => client_cert, :key => client_key) do |http|
+  request = Net::HTTP::Get.new uri
+  response = http.request request
+
+  hash = JSON.parse(response.body)
+
+  version = hash["version"]["number"]
+
+  if version.to_f < min_version.to_f
+    message = ["Elasticsearch is currently version: ", version, ". Expecting it to be at least: ", min_version]
```
jcantrill, line 34: nit.. fyi.. you can do: `"Elasticsearch is currently version: #{version}. Expecting it to be at least: #{min_version}"`

fluentd/wait_for_es_version.rb
```diff
@@ -0,0 +1,38 @@
...
+require 'net/https'
+require 'json'
+
+if ARGV.length != 1
+  abort "Usage: wait_for_es_version.rb <min_es_version>"
+end
+
+min_version = ARGV[0]
+
+uri = URI('https://elasticsearch.openshift-logging.svc:9200')
```
jcantrill, line 12: any reason we need to maybe pass this in as an arg?

ewolinetz: i was wondering this as well... it wouldn't be difficult to add it in. my initial thought was this was going to be specific for a CLO deployed fluentd/ES...

fluentd/wait_for_es_version.sh
jcantrill: I think you can also do this with `#!/usr/bin/env ruby`

PR #1900, opened 2020-05-23
Bug 1835903: Add guard script for kibana to wait for index migration

periklis:
This PR adds an helper utility for init containers to check if the required alias-to-index mapping is matching before starting kibana >=6.

To address: https://bugzilla.redhat.com/show_bug.cgi?id=1835903


periklis:
This an obsolete step. Kibana 5 to 6  migration works fine without this extra guard.

PR #1902, opened 2020-05-26
Bug 1837867: Partial fix. Remove dsl restriction for admin_user

jcantrill:
This is a partial fix to remove the DSL role restriction for an admin_user.  It appears, however, this role is combined with the project_user role and DSL is still applied

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1837867

### DIFF
--- a/elasticsearch/sgconfig/roles_mapping.yml
+++ b/elasticsearch/sgconfig/roles_mapping.yml
@@ -38,5 +38,5 @@ admin_user:
 
 project_user:
   readonly: true    
-  users:    
-    - '*'
+  backendroles:
+    - 'project_user'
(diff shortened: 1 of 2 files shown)


elasticsearch/sgconfig/roles_mapping.yml
```diff
@@ -38,5 +38,5 @@ admin_user:
 
 project_user:
   readonly: true    
-  users:    
-    - '*'
\ No newline at end of file
+  backendroles:
```
ewolinetz, line 41: not sure we need to do this based on https://github.com/openshift/elasticsearch-proxy/pull/39#issuecomment-634960510

PR #1903, opened 2020-05-27
Adding Makefile targets for building and deploying each component

ewolinetz:
To simplify building and deploying individual component images when testing fixes locally vs original make target that builds and pushes every image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,10 @@
 OPERATOR_LOGGING_IMAGE_STREAM?=stable
+
+FLUENTD_IMAGE?="openshift/origin-logging-fluentd"
+KIBANA_IMAGE?="openshift/origin-logging-kibana6"
+ELASTICSEARCH_IMAGE?="openshift/origin-logging-elasticsearch6"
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,4 +1,10 @@
 OPERATOR_LOGGING_IMAGE_STREAM?=stable
+
+FLUENTD_IMAGE?="openshift/origin-logging-fluentd"
+KIBANA_IMAGE?="openshift/origin-logging-kibana6"
+ELASTICSEARCH_IMAGE?="openshift/origin-logging-elasticsearch6"
+CURATOR_IMAGE?="openshift/origin-logging-curator5"
+
 # Build the Docker images for Origin Aggregated Logging
 build-images:
```
ewolinetz, line 9: i'm not sure if this is used by CI anywhere, need to investigate that further before removing it

ewolinetz:
@bparees can we bypass the need for a bugzilla for this? it is limited to just changes on the Makefile for local development

bparees:
Bz overridden, dev changes only.

PR #1904, opened 2020-05-28
Bug 1833486: Mapping "template" field becomes "index_patterns"

lukas-vlcek:
After we updated common data model to ES6x we also need to
update the rest of index templates files that are not generated
from ViaQ/elasticsearch-templates.


elasticsearch/index_templates/common.settings.kibana.template.json
```diff
@@ -4,5 +4,5 @@
     "index.number_of_replicas": $REPLICA_SHARDS,
     "index.number_of_shards": $PRIMARY_SHARDS
   },
-  "template": ".kibana*"
+  "index_patterns": ".kibana*"
```
ewolinetz, line 7: does this have any relation to the kibana term "index pattern"? or is it just overloaded at this point...?

elasticsearch/index_templates/common.settings.operations.template.json
```diff
@@ -7,5 +7,8 @@
     "index.translog.flush_threshold_size": "256mb",
     "index.unassigned.node_left.delayed_timeout": "2m"
   },
-  "template": ".operations*"
+  "index_patterns": [
+    "infra-*",
+    "audit.infra-*"
```
ewolinetz, line 12: @jcantrill isn't `audit.infra` subsumed by `infra` ?

ewolinetz:
overall :+1: just a few questions

elasticsearch/index_templates/common.settings.kibana.template.json
lukas-vlcek: It is overloaded atm.
They changed the field name from `template` to `index-patterns` which is probably better name as it now accepts also list of strings (not only a single string value like `template` did).

See:
- 5.6 https://www.elastic.co/guide/en/elasticsearch/reference/5.6/indices-templates.html
- 6.x https://www.elastic.co/guide/en/elasticsearch/reference/6.8/indices-templates.html (although a single string value is still accepted)
- 7.x https://www.elastic.co/guide/en/elasticsearch/reference/7.8/indices-templates-v1.html#put-index-template-v1-api-request-body (in 7.x the docs is precise in terms of allowed data types)

elasticsearch/index_templates/common.settings.operations.template.json
lukas-vlcek: According to https://github.com/openshift/enhancements/blob/master/enhancements/cluster-logging/cluster-logging-es-rollover-data-design.md#data-model there will be `infra-*` and `audit-*` logs but no `audit.infra-*` which means I need to change it. But I also need to change it in our data model and release a new version of index templates and create another PR.

jcantrill: There are currently only 3 patterns we support in index managment:  infra*, app*, audit*

PR #1906, opened 2020-05-29
[WIP] Bug 1841832: Adding logic to check if we are newer than the master for seeding acl

ewolinetz:
Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1841832

### DIFF
--- a/elasticsearch/init.sh
+++ b/elasticsearch/init.sh
@@ -22,8 +22,8 @@ wait_for_port_open
 # find out who the master of the cluster is
 master_node=$(get_master_node)
 
-# if we're the master -- do the init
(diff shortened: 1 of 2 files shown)


ewolinetz:
/hold
waiting to update with bz to track against

ewolinetz:
/hold
i want to do a bit more local testing first

ewolinetz:
looks like we need to force seeding in the case where we have a mixed version cluster

ewolinetz:
@alanconway after doing some manual testing this didn't resolve the issue we saw all the time. I think the plan may be to scrap this and instead do a full cluster restart as part of es 5x -> 6x simply because there seems to be an oddness going from searchguard to opendistro

PR #1911, opened 2020-06-01
Bug 1842439: Vendor ruamel.yaml

jcantrill:
Vendro Vendor ruamel.yaml

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1842439

4.5 backport of https://github.com/openshift/origin-aggregated-logging/pull/1910


curator/Dockerfile.centos7
```diff
@@ -35,8 +35,8 @@ RUN mkdir -p $(dirname "$CURATOR_CONF_LOCATION") && \
     chgrp -R 0 ${HOME} && \
     chmod -R g=u ${HOME}
 WORKDIR ${HOME}/vendor
-RUN pip3 install -q $(ls . | grep -v curator) --no-index --find-links . && \
-    pip3 install -q --no-index elasticsearch-curator*  && \
+RUN pip3 install $(ls . | grep -v curator) -q --no-index --find-links . && \
+    pip3 install --no-index elasticsearch-curator* -q && \
```
periklis, line 39: Just in case I missed something. In #1910 the order is flipped, first the package name pattern and then the arguments. Is this relevant?

jcantrill: This is a straight cherrypick of #1910 .  It appears the only diff is the order between upstream and downstream.  I tested both builds to confirm everything builds and runs

PR #1912, opened 2020-06-02
Bug 1842357: Limit role project_user to access only owned kibana idx

periklis:
To address https://bugzilla.redhat.com/show_bug.cgi?id=1842357

Needs backport: release-4.5

### DIFF
--- a/elasticsearch/sgconfig/roles.yml
+++ b/elasticsearch/sgconfig/roles.yml
@@ -142,7 +142,7 @@ project_user:
       '*':
         - READ
       _dls_: "{\"bool\":{\"filter\":{\"script\":{\"script\":{\"lang\":\"painless\",\"params\":{\"param1\":\"${attr.proxy.ns}\"},\"source\":\"String namespace = doc['kubernetes.namespace_name'][0];StringTokenizer st = new StringTokenizer(params.param1,\\\",\\\");while (st.hasMoreTokens()){if (st.nextToken().equalsIgnoreCase(namespace)){return true;}}return false;\"}}}}}"
-    '?kibana_*':
+    '?kibana_*_${user_name}':
       '*':
         - CRUD
 


elasticsearch/sgconfig/roles.yml
```diff
@@ -142,7 +142,7 @@ project_user:
       '*':
         - READ
       _dls_: "{\"bool\":{\"filter\":{\"script\":{\"script\":{\"lang\":\"painless\",\"params\":{\"param1\":\"${attr.proxy.ns}\"},\"source\":\"String namespace = doc['kubernetes.namespace_name'][0];StringTokenizer st = new StringTokenizer(params.param1,\\\",\\\");while (st.hasMoreTokens()){if (st.nextToken().equalsIgnoreCase(namespace)){return true;}}return false;\"}}}}}"
-    '?kibana_*':
+    '?kibana_*_${user_name}':
```
jcantrill, line 145: @vimalk78 @ewolinetz  I don't think this is a valid substitution nor does this result in a valid index name.  It was my impression the index name somehow hashed the username

jcantrill:
@periklis this will also need a valid 4.5 clone in order to resolve against 4.5?

elasticsearch/sgconfig/roles.yml
periklis: @jcantrill Actually this works as expected. In addition according the OpenDistro docs, the user name is not hashed: https://opendistro.github.io/for-elasticsearch-docs/docs/security-access-control/multi-tenancy/#manage-kibana-indices

ewolinetz: the multitenant indices are `.kibana_<user_hash>_<user_name>`

jcantrill: @ewolinetz @periklis what about permissions for the tenants? Does that need to be incorporated as well?

periklis: you mean `.kibana_<user has>_<tenant_name>` ?

ewolinetz: project users only have their private tenant... so tenant and user name should be the same here.. 
but to be more correct it should be the tenant name rather than user name.

periklis: @ewolinetz Can you provide a reference that prooves that `tenant_name` is equal `user_name`?

ewolinetz: @periklis the docs are vague about this [1], i would recommend logging into a kibana6 instance and check the indices in kibana (there is an `indices` command you can use if you `oc exec -c elasticsearch <es_pod>`)

[1] https://docs.search-guard.com/latest/kibana-multi-tenancy

periklis: @ewolinetz & @jcantrill Basically i tested this PR by created two non-admin users and checked the listing of indices under `Management -> Create Index pattern -> Include system indices`. I will check the `.kibana_<user hash>_<tenant_name>` configuration but the `.kibana_<user hash>_<user name>` matched the listing well enough.

periklis: @ewolinetz & @jcantrill Tested this again with `.kibana_<user hash>_<tenant_name>` and it seems that only `user_name` is replaced by OpenDistro.

PR #1927, opened 2020-06-18
Bug 1848696: Disable bootstrap.system_call_filter on non-amd64 architectures

yselkowitz:


yselkowitz:
CI has passed, anyone to review and merge?

elasticsearch/install-es.sh
```diff
@@ -25,6 +25,7 @@ mkdir -p ${ES_PATH_CONF}/scripts
 chmod 777 ${ES_PATH_CONF}/scripts
 install -m 660 config/* ${ES_PATH_CONF}
 popd
+test $(arch) = x86_64 || echo "bootstrap.system_call_filter: false" >> ${ES_PATH_CONF}/elasticsearch.yml
```
jcantrill, line 28: This config is insignificant in the world of using operators since this is managed by the CLO and changes are clobbered if they differ from what it provides. The change would have to be added here https://github.com/openshift/elasticsearch-operator/blob/master/pkg/k8shandler/configuration_tmpl.go#L3  I think you will need to set an env var that can be consumed in the config like: https://github.com/openshift/elasticsearch-operator/blob/master/pkg/k8shandler/configuration_tmpl.go#L5

yselkowitz: Is there any point in fixing this here anyway?  And would this make more sense then:

https://github.com/multi-arch/elasticsearch-operator/commit/6aa962f138edb92b0343d9c18f51b5b89e064e37

PR #1936, opened 2020-07-07
Bug 1835396: Disable http/2 for kubernetes-client

jcantrill:
This PR bumps the openshift-elasticsearch-plugin dependency to pull in a change to disable the http/2 client.  This is to resolve an issue on an IBM cluster where the client was not able to communicate with the api server

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1835396


elasticsearch/fetch-artifacts-koji.yaml
```diff
@@ -1,3 +1,3 @@
 - nvr: org.elasticsearch.plugin.prometheus-elasticsearch-prometheus-exporter-5.6.16.0_redhat_1-1
```
syedriko, line 1: Should we bump this to 5.6.16.3 as well?

PR #1938, opened 2020-07-10
fix update-fluentd-vendor-gems.sh and rely on Gemfile.lock

jcantrill:
This PR:

* fixes the update-fluentd-vendor-gems.sh script that was broken as part of the ubi changes
* relies on Gemfile.lock version to produce the gems that should be vendored as this is how ruby would normally lock down a ruby app dependencies


fluentd/Gemfile
```diff
@@ -1,17 +1,18 @@
...
+gem 'fluent-plugin-concat'
+gem 'fluent-plugin-elasticsearch'
+gem 'fluent-plugin-kafka'
+gem 'fluent-plugin-multi-format-parser'
+gem 'fluent-plugin-record-modifier'
+gem 'fluent-plugin-rewrite-tag-filter'
+gem 'fluent-plugin-systemd'
+gem 'fluent-plugin-remote-syslog'
+gem 'fluent-plugin-prometheus'
+gem 'fluent-plugin-splunk-hec'
```
periklis, line 16: Why does this PR drop the version constraints initially introduced by #1858?

periklis: IC, we rely on Gemfilelock.

jcantrill: This was done in order to produce a Gemfile.lock that was consistent with 4.4  while testing.  The intention is to revert fluentd in https://github.com/openshift/origin-aggregated-logging/pull/1939 so I will look at applying those changes back in  #1939

jcantrill: Actually ran into the following when defining versions which im not certain about the resolution:
```
Resolving dependencies...
Bundler could not find compatible versions for gem "activesupport":
  In Gemfile:
    fluent-plugin-splunk-hec (= 1.2.1) was resolved to 1.2.1, which depends on
      openid_connect (~> 1.1.8) was resolved to 1.1.8, which depends on
        activemodel was resolved to 5.2.4.3, which depends on
          activesupport (= 5.2.4.3)

    fluent-plugin-splunk-hec (= 1.2.1) was resolved to 1.2.1, which depends on
      activesupport (~> 5.2)
```

PR #1941, opened 2020-07-16
Bug 1850849: disable kibana metrics to address CVEs

ewolinetz:
Addresses:
* https://bugzilla.redhat.com/show_bug.cgi?id=1850849
* https://bugzilla.redhat.com/show_bug.cgi?id=1850850

### DIFF
--- a/kibana/kibana.yml
+++ b/kibana/kibana.yml
@@ -120,3 +120,7 @@ opendistro_security.multitenancy.enabled: true
 opendistro_security.allow_client_certificates: true
 opendistro_security.multitenancy.tenants.enable_global: false
 opendistro_security.multitenancy.tenants.enable_private: true
+
+# addresses https://access.redhat.com/security/cve/CVE-2020-7013
+# addresses https://access.redhat.com/security/cve/CVE-2020-7015
+metrics.enabled: false


hack/logging-dump.sh
```diff
@@ -194,17 +195,34 @@ check_collector() {
...
+  pods=$(oc -n $namespace get pods -l name=cluster-logging-operator -o jsonpath={.items[*].metadata.name})
+  for pod in $pods
+  do
+    get_env $pod $clo_folder $namespace "Dockerfile-*operator*"
+    get_pod_logs $pod $clo_folder $namespace
+  done
+  oc -n $namespace get deployment cluster-logging-operator -o yaml > $clo_folder/deployment
+  csv_name="$(oc -n $namespace get csv -o name | grep 'cluster-logging-operator')"
+  oc -n $namespace get "${csv_name}" -o yaml > "${clo_folder}/csv"
+}
```
periklis, line 211: This is really really cool! Can we expand this on the catalog and olm operator logs? Maybe also get the describe of the subscription and the install plan too. They are super useful if the CRDs are broken or similar bugs raise up.

hack/logging-dump.sh
```diff
@@ -251,9 +273,14 @@ check_kibana() {
...
     check_kibana_connectivity $pod
+    get_kibana_cr
   done
 }
 
+get_elasticsearch_cr() {
+  oc get -n ${NAMESPACE} elasticsearch elasticsearch -o yaml > $es_folder>cr
+}
+
 get_elasticsearch_status() {
```
periklis, line 284: For ES6 we need to expand this on the index management pod logs too.

ewolinetz:
oops, accidentally added in an unrelated file :laughing:

hack/logging-dump.sh
ewolinetz: whoops, this file shouldn't be part of this pr!

PR #1946, opened 2020-07-20
expanding on hack/logging-dump.sh to also include CRs and CSVs

ewolinetz:


ewolinetz:
i thought the csv contains most of that information already...

out of curiosity, what does knowing the subscription buy us for troubleshooting?

periklis:
> i thought the csv contains most of that information already...
> 

Unfortunately, the CSV contains everything if it can be applied. The install plan maybe broken, e.g. because minKubeVersion does not apply, then you don't get a CSV created.

> out of curiosity, what does knowing the subscription buy us for troubleshooting?

The subscription tells us which channel is used and what operator-registry image.

ewolinetz:
from what i've seen locally if `minKubeVersion` doesn't match that just stops the deployment from being created, the csv will still be out there.
if i've tried to `make deploy` but the kube version is wrong i can `oc edit` the csv to fix that and then i see it is able to complete the operator installation.

periklis:
> from what i've seen locally if `minKubeVersion` doesn't match that just stops the deployment from being created, the csv will still be out there.
> if i've tried to `make deploy` but the kube version is wrong i can `oc edit` the csv to fix that and then i see it is able to complete the operator installation.

This was also my perception of how things work with OLM. However, the experience with the `minKubeVersion` blocker on CI shows that you cannot detect this issue only via CSV, i.e. the install-plan did not even install the CSV to detect this issue.

> 
> just to clarify -- my intent with this expansion first and foremost was to help with gathering data from customer cases where they already have logging installed.

ACK, I understood the intention. I am just proposing to gather also information about the whole process of operator installation that in extend is the same on operator upgrades (Subscription->Install-Plan->CSV+CRD+RBAC+Deployment+etc.)

ewolinetz:
> Subscription->Install-Plan

can you give me the commands you use to grab these? i've never needed to interact with those before

periklis:
> > Subscription->Install-Plan
> 
> can you give me the commands you use to grab these? i've never needed to interact with those before

I am citing the command from an answer I gave lately in a chat thread:

If the symptom is "no CSV" installed then it's OLM installation procedure. To debug in general the whole process I can recommend to take a look on:
In general the installation process works like this (explained more visually in our Openshift docs though):
1. Operator installation starts by creating a subscription for the registry's catalog. First hint to check for broken requirements is : `oc get subscription -o yaml`
2. The subscription is translated in an install plan that tells OLM what resources to install, i.e. CSV, RBAC, CRDs, ServiceAccount. So the next hint is: `oc get installplan -o yaml`
3. If the install plan was applied correctly our should see the installed resources, ie. CSV, rbac, CRDs, ServiceAccount, Deployment in your target namespace: So next hint is: `oc get csv -o yaml`
4. If the pods don't start, you should check the logs of your operator, e.g `oc logs elasticsearch-operator-...`
5. Finally, checking the logs of the operator-framework can also give hints on what is broken:
- Logs for catalog-operator: `oc -n openshift-operator-lifecycle-manager logs catalog-operator-...`
- Logs for olm-operator: `oc -n openshift-operator-lifecycle-manager logs olm-operator-...
`

ewolinetz:
@periklis PTAL, I added the installer information to be collected as well (however it will need to be specified explicitly given we don't expect customers to need to use it when gathering for bzs)

PR #1887, opened 2020-04-28
Bug 1795897: Move curator to python 3, remove python-thrift

jcantrill:
Ref:

https://bugzilla.redhat.com/show_bug.cgi?id=1795897

This PR is 4.4 backports of:

https://github.com/openshift/origin-aggregated-logging/pull/1866
https://github.com/openshift/origin-aggregated-logging/pull/1886
https://github.com/openshift/origin-aggregated-logging/pull/1901

PR #1948, opened 2020-07-23
Bug 1795897: curator: base on py36

sosiouxme:
https://github.com/openshift/origin-aggregated-logging/pull/1887 got some but not all changes from master... Dockerfile should look pretty much like https://github.com/openshift/origin-aggregated-logging/blob/master/curator/Dockerfile

(existing changes fail because ubi7 does not include python3 or pip3; the alternate path would be to put back the step for installing those)


ewolinetz:
/hold
still having a discussion

PR #1949, opened 2020-07-23
Bug 1795897: curator/Dockerfile: don't be picky about where pip installs curator

sosiouxme:
Apparently in this environment pip doesn't put curator in `/usr/local/bin` - note that the `ls` check was removed in master. This checks that it's in the path somewhere and leaves it at that.


ewolinetz:
/lgtm
thanks @sosiouxme

PR #1951, opened 2020-07-24
fixing smoke test for release-4.5

ewolinetz:
Hardcoding the CLO and EO branch versions to release-4.5

PR #1954, opened 2020-07-28
adding logic for restricting operator branches

ewolinetz:
manual cherry-pick of https://github.com/openshift/origin-aggregated-logging/pull/1951

### DIFF
--- a/hack/deploy-logging.sh
+++ b/hack/deploy-logging.sh
@@ -168,14 +168,23 @@ get_cluster_version_maj_min() {
     fi
 }
 
+MASTER_RELEASE_VERSION=${MASTER_RELEASE_VERSION:-4.6}
 get_cluster_version_maj_min
 # what numeric version does master correspond to?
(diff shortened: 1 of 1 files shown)


hack/deploy-logging.sh
```diff
@@ -168,15 +168,24 @@ get_cluster_version_maj_min() {
     fi
 }
 
+MASTER_RELEASE_VERSION=4.6
```
jcantrill, line 171: I thought we moved away from needing to hardcode this here.  Isn't there a VAR that comes from CI that tells us the release stream?  @bparees   would it make more sense to place this in the Makefile?

bparees: i'm not aware of one....at best there might be one that tells you the branch (e.g. release-4.6).

as for putting it in the makefile or here.....  it appears to be a default here, that could be overridden from the makefile (or other invoker of the script)?

ewolinetz: i'll update it so we could override it in here

jcantrill:
overrding bz label as this address CI improvements

PR #1958, opened 2020-08-05
Bug 1860793: Vendor ruamel.yaml

jcantrill:
Vendor ruamel.yaml

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1860793

Backport of https://github.com/openshift/origin-aggregated-logging/pull/1911


jcantrill:
@sosiouxme please review.  We've had so many changes to curator that don't build in downstream I would like to ensure this will build properly based on your findings

curator/run.sh
```diff
@@ -7,6 +7,7 @@ fi
 if [ "$CURATOR_LOG_LEVEL" = WARN ] ; then
     export CURATOR_LOG_LEVEL="WARNING"
 fi
+PYTHON_BIN=${PYTHON_BIN:-python3}
```
ewolinetz, line 10: does it normally run with python 3?

jcantrill: There is a descrepency between the upstream and downstream images where I believe when I tested the scratch builds this was the only way to resolve

sdodson:
/hold
I did a quick review but I'm not familiar with this and would appreciate someone else looking over the code changes before this merges.

PR #1961, opened 2020-08-06
removing jq from ES images

ewolinetz:
follow up from https://github.com/openshift/origin-aggregated-logging/pull/1893#discussion_r430913185

also fixes `hack/build-component-images.sh`

### DIFF
--- a/elasticsearch/utils/logging
+++ b/elasticsearch/utils/logging
@@ -69,33 +69,8 @@ get_kibana_index_name() {
     echo .kibana.$(get_hash "$1")
 }
 
-#es_util --query="_cluster/state/master_node?pretty"
-# {
-#  "cluster_name" : "elasticsearch",
-#  "master_node" : "xD9yX9NTRBCqM8YA0Nym6w"
-#}
-#
-#es_util --query="_cluster/state/nodes?pretty"
-# {
-#  "cluster_name" : "elasticsearch",
-#  "nodes" : {
-#    "xD9yX9NTRBCqM8YA0Nym6w" : {
-#      "name" : "elasticsearch-cdm-h5el9dsq-1",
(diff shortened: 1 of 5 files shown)


ewolinetz:
fyi @vimalk78 @syedriko

elasticsearch/utils/logging
```diff
@@ -69,33 +69,8 @@ get_kibana_index_name() {
...
-#      "ephemeral_id" : "zWCXff-eQwi2YLMpZ9qbqA",
-#      "transport_address" : "10.131.0.17:9300",
-#      "attributes" : { }
-#    }
-#  }
-#}
 get_master_node() {
-    local master=$(es_util --query="_cluster/state/master_node?pretty" | jq -r .master_node)
-    echo $(es_util --query="_cluster/state/nodes?pretty" | jq -r .nodes[\"${master}\"].name)
+    echo $(es_util --query="_cat/master?h=node")
```
vimalk78, line 73: we can replace the `jq` with equivalent python ones
```
local master_node=$(es_util --query="_cluster/state/master_node?pretty" | python -c "import json,sys;r=json.load(sys.stdin);print r['master_node']")
echo $(es_util --query="_cluster/state/nodes?pretty" |  python -c "import json,sys;r=json.load(sys.stdin);print r['nodes'][\"$master_node\"]['name']"
```
if decide to add this, then please dont remove the commented example above, helps to understand how the command output looks like

lukas-vlcek: I think using `_cat` API is a lot better than parsing JSON.
BTW, if needed `_cat` API can also return the result formatted in JSON. See https://www.elastic.co/guide/en/elasticsearch/reference/6.8/cat.html#_response_as_text_json_smile_yaml_or_cbor

PR #1962, opened 2020-08-11
LOG-814: creating kibana rh-manifest.txt file

ewolinetz:
generated using `kibana/module_list.sh`

Addresses: https://issues.redhat.com/browse/LOG-866

/cc @jcantrill


jcantrill:
overriding bz valid label to address gap in security manifest

PR #1971, opened 2020-08-26
Bug 1847275: Allow cluster-admin to retrieve metrics

jcantrill:
This PR adds the admin_reader backend role to retrieve metrics

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1847275

### DIFF
--- a/elasticsearch/sgconfig/roles_mapping.yml
+++ b/elasticsearch/sgconfig/roles_mapping.yml
@@ -25,6 +25,7 @@ sg_role_admin:
 sg_role_prometheus:
   backendroles:
     - 'prometheus'
+    - 'admin_reader'
 
 sg_role_jaeger:
   users:


jcantrill:
@ewolinetz i think this would be a good ref to move these files into the EO so they can be properly tested and reconciled.  This is just the easy fix

alanconway:
/hold
stop retesting till fixed.

ewolinetz:
@jcantrill it is an easy change.. but is it really necessary? i don't think we ever had a use case where this was possible...

jcantrill:
> @jcantrill it is an easy change.. but is it really necessary? i don't think we ever had a use case where this was possible...

This is totally a guess for me as a solution.  I will defer to you if you think we should CLOSED WONTFIX.  There is probably limited value in the admin grabbing the metrics anyway given you really want prometheus to put them into a form that is more digestible by humans.

jcantrill:
On second thought I'll close it and await justification

PR #1982, opened 2020-09-08
fix deploy-logging to always use CLO make deployment

jcantrill:
This PR fixes the smoke test:

* Updates deploy-logging.sh to rely on CLO make deploy target

### DIFF
--- a/hack/deploy-logging.sh
+++ b/hack/deploy-logging.sh
@@ -170,7 +170,7 @@ get_cluster_version_maj_min() {
 
 get_cluster_version_maj_min
 # what numeric version does master correspond to?
-MASTER_VERSION=${MASTER_VERSION:-${CLUSTER_MAJ_MIN:-4.4}}
+MASTER_VERSION=${MASTER_VERSION:-${CLUSTER_MAJ_MIN:-4.6}}
 # what namespace to use for operator images?
 EXTERNAL_REGISTRY=${EXTERNAL_REGISTRY:-registry.svc.ci.openshift.org}
 EXT_REG_IMAGE_NS=${EXT_REG_IMAGE_NS:-origin}
@@ -208,7 +208,7 @@ construct_image_name() {
 
 update_images_in_clo_yaml() {
     local yamlfile=$1
-    local clo_img=$2
+    local clo_img=${2:-} #unused.  leaving for compatibility
     local version=${3:-latest}
     local filearg
     if [ "$yamlfile" = "-" ] ; then
@@ -226,8 +226,6 @@ update_images_in_clo_yaml() {
         -e "/name: CURATOR_IMAGE/,/value:/s,value:.*\$,value: ${c_img}," \
         -e "/name: FLUENTD_IMAGE/,/value:/s,value:.*\$,value: ${f_img}," \
         -e "/name: OAUTH_PROXY_IMAGE/,/value:/s,value:.*\$,value: ${op_img}," \
-        -e "s, image:.*cluster-logging-operator.*\$, image: ${clo_img}," \
-        -e "s, containerImage:.*cluster-logging-operator.*\$, containerImage: ${clo_img}," \
         $filearg
 }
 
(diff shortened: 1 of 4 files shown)


hack/deploy-logging.sh
```diff
@@ -331,14 +329,10 @@ deploy_logging_using_olm() {
...
-    # make deploy-no-build
-    EO_IMAGE=${EO_IMAGE:-$( construct_image_name elasticsearch-operator latest )}
-    CLO_IMAGE=${CLO_IMAGE:-$( construct_image_name cluster-logging-operator latest )}
     pushd $CLO_DIR > /dev/null
     cp manifests/${MASTER_VERSION}/cluster-logging.*.clusterserviceversion.yaml manifests/${MASTER_VERSION}/cluster-logging.*.clusterserviceversion.yaml.orig
-    update_images_in_clo_yaml manifests/${MASTER_VERSION}/cluster-logging.*.clusterserviceversion.yaml $CLO_IMAGE
-    REMOTE_CLUSTER=true REMOTE_REGISTRY=true NAMESPACE=$LOGGING_NS \
-        IMAGE_OVERRIDE="$CLO_IMAGE" EO_IMAGE_OVERRIDE="$EO_IMAGE" make deploy-no-build
+    update_images_in_clo_yaml manifests/${MASTER_VERSION}/cluster-logging.*.clusterserviceversion.yaml 
+    make deploy
```
vimalk78, line 335: we can capture the `$?` value and throw error in case we ever remove/change `make deploy` target in CLO Makefile

jcantrill: wont we know this by the fact the operation fails?

hack/deploy-logging.sh
```diff
@@ -420,26 +410,26 @@ else
 fi
 
 if [ "${LOGGING_DEPLOY_MODE:-install}" = install ] ; then
-    CLO_DIR=${CLO_DIR:-$GOPATH/src/github.com/openshift/cluster-logging-operator}
-    if [ ! -d $CLO_DIR ] ; then
-        CLO_DIR=$ARTIFACT_DIR/clo
-        get_operator_files $CLO_DIR cluster-logging-operator ${CLO_REPO:-openshift} ${CLO_BRANCH:-master}
+    CLO_DIR=$GOPATH/src/github.com/openshift/cluster-logging-operator
+    EO_DIR=$GOPATH/src/github.com/openshift/elasticsearch-operator
+    if [ ! -d $GOPATH/src/github.com/openshift ] ; then
```
blockloop, line 415: unnecessary `if` check. `mkdir` will noop if the dir already exists

openshift/ci-operator/build-image/cr.yaml
```diff
@@ -14,10 +14,10 @@ spec:
       redundancyPolicy: "ZeroRedundancy"
       resources:
         limits:
-          memory: 3Gi
+          memory: 2Gi
         requests:
-          cpu: 500m
-          memory: 3Gi
+          cpu: 100m
+          memory: 2Gi
```
blockloop, line 20: Can you comment on the reason for these changes? Are they arbitrary?

hack/deploy-logging.sh
vimalk78: mkdir will fail if dir already exists. `$?` is non zero
the script sets `set -euo pipefail` which makes script to exit if any operation fails.

blockloop: Not when you use the `-p` flag

```
$ mkdir -p $PWD
$ echo $?
0
```

openshift/ci-operator/build-image/cr.yaml
jcantrill: We have no specific needs for these values.  I lowered them to ensure we could land the ES nodes without having to bump the OCP sizes.

blockloop:
/lgtm
my nits can be addressed later

jcantrill:
overriding bz label as this is a change to address broken tests

PR #1985, opened 2020-09-11
Workaround COPY behavior in OCP4

jupierce:
The COPY behavior observed on OCP4 clusters is causing an error:
```
STEP 10: ADD extra-jvm.options install-es.sh ci-env.sh /var/tmp
STEP 11: RUN mkdir /artifacts
STEP 12: COPY README.md artifacts/* /artifacts
error: build error: error building at STEP "COPY README.md artifacts/* /artifacts": no files found matching "/tmp/build/inputs/artifacts/*": no such file or directory
```

This PR aims to ensure that *something* is available for the COPY.


jcantrill:
overriding bug valid label for ART

PR #1988, opened 2020-09-16
Bug 1866490: Redirect user on 401 response

jcantrill:
This PR:
* introduces an openshift-logging kibana plugin
* intercepts responses to evaluate for 401 auth failure and redirects user to /oauth/sign_off
* is dependent upon https://github.com/openshift/elasticsearch-proxy/pull/47 to function correctly

Tested with:
* setting oauth token expiration to 1min

There is some interplay between session expiration and oauth token expiration that may still require deleting of oauth cookie until such time as those 4.7 enhancement lands

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1866490


jcantrill:
cc @ewolinetz @blockloop @periklis

kibana/lib/openshift_logging_plugin/public/hack.js
```diff
@@ -0,0 +1,67 @@
+import chrome from 'ui/chrome';
+import { uiModules } from 'ui/modules';
+const app = uiModules.get('apps/openshift_logging_plugin/index');
+
+function redirectOnSessionTimeout($window) {
+    document.cookie = "name=_oauth_proxy; expires=Thu, 01 Jan 1970 00:00:00 GMT; path=/;"
```
periklis, line 6: Why does this function need to override the cookies here?

kibana/lib/openshift_logging_plugin/public/hack.js
```diff
@@ -0,0 +1,67 @@
...
+});
+
+/**
+ * Setup a wrapper around fetch so that we can
+ * handle session timeouts on ajax calls made
+ * by the kfetch component
+ * @param $window
+ */
+function setupResponseErrorHandler($window) {
+    if (!window.fetch) {
```
periklis, line 42: Without being a pro here in AngularJS, what is the difference between using the native `window` object vs. `$window`. Or is this a typo?

kibana/lib/openshift_logging_plugin/public/hack.js
```diff
@@ -0,0 +1,67 @@
...
+        return;
+    }
+
+    const nativeFetch = window.fetch;
+    window.fetch = (url, config) => {
+        return nativeFetch(url, config)
+            .then(async(result) => {
```
periklis, line 49: Do JS Promises not follow the `reject` path if they fail?

kibana/lib/openshift_logging_plugin/public/hack.js
jcantrill: In testing, the experience I saw when the oauth token was set really log (i.e. 1m) and left the cookie as is (24h) the user would be redirected as desired.  If i tried to refresh, however, screen goes blank with a 401.  This was only resolvable by removing the cookie.  Here I'm attempting to expire the cookie to avoid that experience though I'm still testing to confirm it is satisfactory.

kibana/lib/openshift_logging_plugin/public/hack.js
jcantrill: This is a total cut/paste error from some bits I grabbed from OD.  I'm actually not certain why it doesnt fail here other then maybe they are the same thing from a javascript perspective.  Note i believe that certain objects like `document` and `window` are always available.  I will correct them for consistency

kibana/lib/openshift_logging_plugin/public/hack.js
jcantrill: :man_shrugging: copy / paste here

kibana/lib/openshift_logging_plugin/public/hack.js
periklis: Yes, `document` and `window` are always available on the browser. However, since browsers don't provide the full compatibility with all methods for `document` and `window`, JS framework makers always provide wrappers/shims for these objects. If we can you the `$window` from Angular (if it is the same as `window) then we should do so, because you never know what old browser are used out there.

kibana/lib/openshift_logging_plugin/public/hack.js
periklis: Except from https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API:

ewolinetz:
/hold
/lgtm
waiting on your testing of cookie expiration for removing the hold

kibana/lib/openshift_logging_plugin/public/hack.js
jcantrill: further info.. trying to investigate if it effects the behavior i see https://stackoverflow.com/questions/28906180/difference-between-window-and-window-in-ionicframework#:~:text=window%20is%20a%20browser%20object,document.&text=where%20%24window%20is%20a%20service,make%20use%20of%20its%20APIs.

jcantrill:
My findings:

oauth token expires before cookie session:
* autoreload will trigger a 401 which causes the page to redirect to `oauth/sign_off` and then back to the page
* clicking links will trigger a 401 response causes the page to redirect to `oauth/sign_off` and then back to the page
* Page refresh in this condition results in: `{"statusCode":401,"error":"Unauthorized","message":"Authentication Exception"}`
  * User must explicitly delete their cookie.  Unverified if once the cookie expires they are allowed back in

Session expires before token:
* autoreload tries to trigger a page reload from a 401 but fails from CORs with message to user of "unknown error"
* refresh after token expires allows the page to be accessed or prompts the user to re-enter login info

Continuing to investigate if there is a better user experience

kibana/lib/openshift_logging_plugin/public/hack.js
jcantrill: Discussion with the console this does nothing because its an http cookie.  removing

jcantrill:
> My findings:
> 
> oauth token expires before cookie session:
> 
> * autoreload will trigger a 401 which causes the page to redirect to `oauth/sign_off` and then back to the page
> * clicking links will trigger a 401 response causes the page to redirect to `oauth/sign_off` and then back to the page
> * Page refresh in this condition results in: `{"statusCode":401,"error":"Unauthorized","message":"Authentication Exception"}`
>   
>   * User must explicitly delete their cookie.  Unverified if once the cookie expires they are allowed back in
> 
> Session expires before token:
> 
> * autoreload tries to trigger a page reload from a 401 but fails from CORs with message to user of "unknown error"
> * refresh after token expires allows the page to be accessed or prompts the user to re-enter login info
> 
> Continuing to investigate if there is a better user experience

@ewolinetz the lastes changes to the PR and various testing of token expiration versus cookie, we should strive to have the coolie expiration sooner then the token.  This scenerio IMO presents the best experience because:

* user doesnt have to figure out how to delete a cookie
* The page will either refresh on auth error or a user can refresh the browser and the situation will be resolved.

Note:  This change still does not address actual "logoff" so a new user can reuse the browser.  That will require changes outside the scope of logging

ewolinetz:
> we should strive to have the cookie expiration sooner then the token

but there is no sure way to do this, we can set the expire time to be less than the expire time of the console but that doesn't ensure the cookie will expire first

kibana/lib/openshift_logging_plugin/index.js
```diff
@@ -0,0 +1,28 @@
...
+  return new kibana.Plugin({
+    name: 'openshift_logging',
+    id: 'openshift_logging',
+    require: ['elasticsearch'],
+
+    uiExports: {
+      
+      hacks: [
+        //the path here is misleading as it matches the 'name' not actual path
+        'plugins/openshift_logging/hack'
```
ewolinetz, line 13: can we call it something other than `hack` ?

jcantrill: interceptors is the name I chose

jcantrill:
overriding bz valid because of visibility of bot to private comment

PR #1998, opened 2020-10-06
Bug 1885481: Rely on Koji artifacts

jcantrill:
This PR makes it so the build relies on Koji to get maven dependencies

https://bugzilla.redhat.com/show_bug.cgi?id=1885481


jcantrill:
Merging manually to address ART blocker

PR #2008, opened 2020-10-21
[release-4.2] Bug 1885481: Fix more deps in Dockerfile

joepvd:
This seems to be the last bit in the puzzle. A test with these changes
resulted in a succesful [build](https://brewweb.engineering.redhat.com/brew/buildinfo?buildID=1360533)


jcantrill (commented):
/approve
/lgtm

PR #2005, opened 2020-10-20
Bug 1886856: Add typhoeus for keep-alives in elasticsearch plugin

periklis:
### Description
This PR add a dependency namely `typhoeus` for fluentd-plugin-elasticsearch to enable keep-alive connections
 
/cc @blockloop 
/assign @ewolinetz 

### Links:
- Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1886856

PR #2010, opened 2020-10-21
Bug 1886856: Add typhoeus for keep-alives in elasticsearch plugin

jcantrill:
This PR add a dependency namely typhoeus for fluentd-plugin-elasticsearch to enable keep-alive connections

ref: https://bugzilla.redhat.com/show_bug.cgi?id=1886856

cc @periklis 

Replaces https://github.com/openshift/origin-aggregated-logging/pull/2005


ewolinetz:
this may spin until https://github.com/openshift/origin-aggregated-logging/pull/2012 is merged in

PR #2013, opened 2020-10-22
Fix Makefile targets to build individual components

periklis:
### Description
This PR addresses a long-standing issue of this repository to build individual images per component. It addresses the following:
- Pin all targets to `RELEASE_STREAM=prod` except elasticsearch pinned to `RELEASE_STREAM=origin` because we don't have access to koji-provided resources.
- Pin elasticsearch local builds to also use UBI8
- Remove obsolete `elasticsearch/Dockerfile` and `kibana/Dockerfile` because they are superseded by `elasticsearch/Dockerfile.rhel8` and `kibana/Dockerfile.rhel8`

/cc @jcantrill 
/assign @ewolinetz


ewolinetz:
@periklis i think a cleaner fix for this is instead to add an exception in `hack/build-component-images.sh` to update the dockerfile suffix for es... e.g.
```
docker_suffix=''
if [ "${fullimagename}" = "openshift/origin-logging-elasticsearch6" ]; then
  docker_suffix='.origin'
fi
```

elasticsearch/Dockerfile.origin
```diff
@@ -1,4 +1,4 @@
-FROM ubi7:latest
+FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-base-openshift-4.6
```
ewolinetz, line 1: :-1: why are we moving our origin source to be something that requires auth?

periklis: Until ART fixes the general public availability for our base images, we need to comply with this limitation. The tenet is here to continue building locally with UBI8 images and catch any build errors at development time.

ewolinetz: im not sure i entirely agree with this thought... i can build using the `ubi7:latest` image locally now

periklis: It is not about what you are able to build now, because it is possible to build with UBI7 and UBI8 today. It is more about using UBI8 as it will constantly get sync'ed from ART into our CI from 4.6 on. Thus, we should work on that version from now on to discover locally if anything breaks with ART updates to UBI8.

jcantrill (commented):
/approve

elasticsearch/Dockerfile.origin
ewolinetz: going back to my original question -- why are we updating our origin image to use a base image that requires authentication?
why not just make this `ubi8:latest` if we're concerned with building with ubi8?

periklis: IC, now I get your point. You meant `ubi8:latest`, I understood `ubi7:latest`.

elasticsearch/Dockerfile.origin
```diff
@@ -25,10 +25,12 @@ ENV ES_PATH_CONF=/etc/elasticsearch/ \
 
 ARG MAVEN_REPO_URL=https://repo1.maven.org/maven2/
 
+RUN rm /etc/yum.repos.d/*
+ADD art.repo /etc/yum.repos.d
+
 RUN  yum install -y --setopt=tsflags=nodocs --nogpgcheck \
          https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm && \
      packages="java-${JAVA_VER}-openjdk-headless \
-              PyYAML  \
```
ewolinetz, line 31: why are we removing this?

elasticsearch/Dockerfile.origin
```diff
@@ -25,10 +25,12 @@ ENV ES_PATH_CONF=/etc/elasticsearch/ \
 
 ARG MAVEN_REPO_URL=https://repo1.maven.org/maven2/
 
+RUN rm /etc/yum.repos.d/*
+ADD art.repo /etc/yum.repos.d
```
ewolinetz, line 29: are we sure we would be able to pull from the pulp repos in origin?

ewolinetz (changes requested):
more origin based comments

elasticsearch/Dockerfile.origin
periklis: This is a fat-finger delete. The intention is to replace with `python36 python3-pyyaml` as provided by `UBI8:latest`.

elasticsearch/Dockerfile.origin
periklis: This is obsolete right now for `UBI8:latest`.

jcantrill (commented):
/lgtm

PR #2012, opened 2020-10-21
fix fluent unit test image

jcantrill:
This PR fixes the unit test image that runs fluent unit tests

PR #2019, opened 2020-10-28
fix smoke test to use correct operator branches

jcantrill:
This PR:
* fixes the smoke job to utilized the correct operator branches for 4.6 release
* backports the fix for https://github.com/openshift/origin-aggregated-logging/pull/2012


jcantrill:
merging manually as this resolves testing issues:
* images job
* smoke job

PR #2021, opened 2020-11-09
Add indices:admin/mappings/get permission to READ group

lukas-vlcek:
### Description

The READ group already has the `indices:admin/mappings/fields/get*` permission but it does not have the generic `indices:admin/mappings/get` rule – and this PR adds it. It also enable cluster admin to pull index mappings in the dev console.

Right now the cluster admin can get mappings of individual index fields:
```
GET /infra-000001/_mapping/field/kubernetes.namespace_name
```

But is unable to get mapping of whole index:
```
GET /infra-000001/_mapping/
```
Instead it yields:
```JSON
{
  "error": {
    "root_cause": [
      {
        "type": "security_exception",
        "reason": "no permissions for [indices:admin/mappings/get] and User [name=kube:admin, roles=[admin_reader], requestedTenant=__user__]"
      }
    ],
    "type": "security_exception",
    "reason": "no permissions for [indices:admin/mappings/get] and User [name=kube:admin, roles=[admin_reader], requestedTenant=__user__]"
  },
  "status": 403
}
```

/cc @ewolinetz 
/assign @jcantrill

### DIFF
--- a/elasticsearch/sgconfig/roles.yml
+++ b/elasticsearch/sgconfig/roles.yml
@@ -121,13 +121,11 @@ sg_project_operations:
     '?operations?*':
       '*':
         - READ
-        - indices:admin/mappings/fields/get*
         - indices:admin/validate/query*
         - indices:admin/get*
     '*?*?*':
       '*':
         - READ
-        - indices:admin/mappings/fields/get*
         - indices:admin/validate/query*
         - indices:admin/get*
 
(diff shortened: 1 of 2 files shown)


elasticsearch/sgconfig/roles.yml
```diff
@@ -121,13 +121,11 @@ sg_project_operations:
     '?operations?*':
       '*':
         - READ
-        - indices:admin/mappings/fields/get*
```
lukas-vlcek, line 124: Removing these because they are already part of the `READ` group.

ewolinetz:
@lukas-vlcek is there a specific issue/bug that this addresses? what was your use case when running into this?

lukas-vlcek:
@ewolinetz Not sure if there is existing ticket but in general it makes no sense to not enable pulling index mappings (while we enable pulling mappings for individual index fields). This is making it harder to debug and explore existing indices (for example when investigating JSON use case).

PR #2032, opened 2020-12-15
Bug 1895581: Replace DLS rule script with more performant terms filter

lukas-vlcek:
### Description

Use terms query instead of painless scripting. This is much more performant
and uses a lot less resources.

Relevant ES docs:

- [Bool Query Filter](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/query-dsl-bool-query.html)
- [Terms Query](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/query-dsl-terms-query.html)

This PR assume that es-proxy provide list of projects in comma-delimited form
where every item is quoted.

Watch for associated PR in https://github.com/openshift/elasticsearch-proxy
before merging.

/cc @vimalk78 
/assign @jcantrill 

/cherry-pick release-4.5

### Links

- Depending on PR(s): https://github.com/openshift/elasticsearch-proxy/pull/68
- Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1895581

### DIFF
--- a/elasticsearch/sgconfig/roles.yml
+++ b/elasticsearch/sgconfig/roles.yml
@@ -139,7 +139,7 @@ project_user:
     app:
       '*':
         - READ
-      _dls_: "{\"bool\":{\"filter\":{\"script\":{\"script\":{\"lang\":\"painless\",\"params\":{\"param1\":\"${attr.proxy.ns}\"},\"source\":\"String namespace = doc['kubernetes.namespace_name'][0];StringTokenizer st = new StringTokenizer(params.param1,\\\",\\\");while (st.hasMoreTokens()){if (st.nextToken().equalsIgnoreCase(namespace)){return true;}}return false;\"}}}}}"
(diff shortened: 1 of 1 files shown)


vimalk78 (commented):
can you add some test cases which checks various conditions of the auth mechanism?
1. a user has different projects than the doc's kubernetes.namespace_name
2. user has same projects as the doc's kubernetes.namespace_name

it will serve as a documentation about how the mechanism works

lukas-vlcek:
BTW, it should be pointed out that project names are used for filtering as is without any modifications (precisely, they are [indexed as keywords](https://github.com/openshift/origin-aggregated-logging/blob/master/elasticsearch/index_templates/com.redhat.viaq-openshift-project.template.json#L413) which means no text analysis is run on them before they are put into index). I.e. "Dummy" and "dummy" are treated as two different projects. The original script was lowercasing the values.

ewolinetz:
this is going to be dependent on https://github.com/openshift/elasticsearch-proxy/pull/68 being merged first

PR #2051, opened 2021-01-22
Update fluentd README to include hack instructions

jcantrill:
### Description
This PR updates the fluentd README to add hack instructions

cc @vimalk78 @pmoogi-redhat @alanconway

### DIFF
--- a/fluentd/README.md
+++ b/fluentd/README.md
@@ -1,55 +1,9 @@
 # Fluentd
-[Fluentd](https://www.fluentd.org/) is the log collector that resides on each Openshift node to gather application and node logs
-
+[Fluentd](https://www.fluentd.org/) is the log collector that resides on each OpenShift node to gather application and node logs
 ## Configuration
-Following are the environment variables that can be modified to adjust the configuration:
-
-| Environment Variable | Description |Example|
-|----------------------|-------------|---|
-| `ENABLE_PROMETHEUS_ENDPOINT`| Enable or diable the prometheus endpoint (Default: true)| `ENABLE_PROMETHEUS_ENDPOINT=false`|
-| `MERGE_JSON_LOG`     | **DEPRECATED** Parse JSON log messages and merge them into the JSON payload to be indexed to Elasticsearch. **Default:** True | `MERGE_JSON_LOG=true`|
-| `OCP_OPERATIONS_PROJECTS`| The list of project or patterns for which messages will be sent to the operations indices|`OCP_OPERATIONS_PROJECTS="default openshift openshift- kube-*"`|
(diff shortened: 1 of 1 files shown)


fluentd/README.md
```diff
@@ -64,12 +18,49 @@ file chunk, perform a limited set of operations to confirm message validity. Use
...
+#### Make lib Changes
+* May need to install additional dependencies from `lib/<gemdir>` using `bundle install`
+* Make any needed changes to the files in lib
+* Test change `bundle exec rake test`
+* Commit change and rebuild image
+
+### Building the Container Image
+Build the container image from the fluentd directory like:
+```
+$ podman build -t fluentd:test .
```
eranra, line 65: @jcantrill it would be helpful to explain that you need access to registry.svc.ci.openshift.org inorder to build the image (i.e. make sure you have the needed credentials under ~/.docker/config.json)

eranra: @jcantrill 

I used docker instead of podman (maybe you can add that as option in the text above)

Can you add something like this:

### Test Container Image with custom fluentd configuration

run the following command to mount custom local fluentd configuration against the container and start fluentd executable:
```
docker run -v $(pwd)/fluent-test-tail.conf:/etc/fluent/fluent-test-tail.conf -it fluentd:test fluentd -v -c /etc/fluent/fluent-test-tail.conf
```

jcantrill: > @jcantrill it would be helpful to explain that you need access to registry.svc.ci.openshift.org in order to build the image (i.e. make sure you have the needed credentials under ~/.docker/config.json)

I will add a note but making the repo generally buildable outside of being an openshift dev is a longer term goal.  I don't believe this is the place to be defining larger needs around "how to hack logging"

jcantrill: > @jcantrill
> 
> I used docker instead of podman (maybe you can add that as option in the text above)
> 
> (optionally) docker build -t fluentd:test .
> 
> Can you add something like this:
> 
> ### Test Container Image with custom fluentd configuration
> run the following command to mount custom local fluentd configuration (e.g ~/fluent-test-tail.conf) against the container and start fluentd executable:
> 
> ```
> docker run -v $(pwd)/fluent-test-tail.conf:/etc/fluent/fluent-test-tail.conf -it fluentd:test fluentd -v -c /etc/fluent/fluent-test-tail.conf
> ```
> 
What am i testing here? Why would I do this?

> @jcantrill Also if you can add this section to explain how to push the custom image back to OCP image registry::

This is part of generally hacking on logging, unrelated to fluentd. I would prefer to leave any of this info to any doc about "how do I interact and develop against OCP"
>

alanconway: /lgtm
I see there are outstanding requests, but bhe changes as they stand are an improvement. We can make additional changes later.

PR #2056, opened 2021-02-04
Bug 1925081: Updating kibana logs to censor use access token and email address

ewolinetz:
### Description
This censors sensitive information like the user's email address and access token, replacing it with `X`s to look like:
```
{... ,"x-forwarded-access-token":"XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX","x-forwarded-email":"XXXXXXXXXXXXXXXXXXXXXXXX", ...}
```

We are censoring rather than removing so that if necessary we can verify that the headers are being passed as part of requests.

/cc @blockloop 

/cherry-pick release-4.6

### Links
- Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1925081

### DIFF
--- a/kibana/kibana.yml
+++ b/kibana/kibana.yml
@@ -106,6 +106,9 @@ pid.file: ${HOME}/kibana.pid
 # if this is set to true and enabled we are unable to see warning messages such as ones indicating the migration of a kibana index failed
 #logging.quiet: true
 
+logging.filter.x-forwarded-access-token: censor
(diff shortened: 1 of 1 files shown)


ewolinetz:
/override ci/prow/cluster-logging-operator-e2e

flaking test is still cluster log forwarder test which is not impacted by this change at all, overriding failing test to move this along

PR #2059, opened 2021-02-10
Update CI builds of eventrouter to address nil pointer

jcantrill:
…eventrouter from different repo

### Description
This PR fixes CI but pulling in the latest eventrouter code to address a nil pointer.  Eventrouter is built for production elsewhere.

ref: originally fixed in https://bugzilla.redhat.com/show_bug.cgi?id=1701495


alanconway:
Looks good to me apart from one nit about package names.

jcantrill:
Manually merging as this addresses the nil pointer in eventrouter test.  The test addtionally looks to be timing out prematurely which is addressed https://github.com/openshift/cluster-logging-operator/pull/888.  Merging here to unblock

PR #2060, opened 2021-02-15
addition of new metric

pmoogi-redhat:
### Description
Currently in_tail plugin doesn't support publishing of inbound logloss - i.e. difference between total bytes written to disk (logfile) and total bytes collected or read by fluentd. This PR got changes in fluentd/lib/fluent/plugin/in_tail.rb, and fluent-plugin-prometheus/lib/fluent/plugin/in_prometheus_tail_monitor.rb plugins to enable publishing of the below parameters 
1. totalbytes_logged_in_disk (written to disk at each unique inode level per container process)
2. totalbytes_collected_from_disk (read or collected by fluentd at each unique inode level per container process)

/cc @alanconway @jcantrill 
/assign @alanconway 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-1032
- Enhancement proposal: many rotations getting missed by fluentd, next enhancement proposal is to get fluentd know about all actual rotations done by CRIO/conmon process by reading extra meta data such as  <timestamp containerid maxsizeoflogfile rotation-counter>. Based on what rotations fluentd could track and which one got missed computing log-loss as [no-of-rotations-those-missed * maxsizeoflogfile + sum over all tracked rotations by fluentd as [totalbytes_logged_in_disk - totalbytes_collected_from_disk]

### DIFF
--- a/fluentd/vendored_gem_src/fluent-plugin-prometheus/lib/fluent/plugin/in_prometheus_tail_monitor.rb
+++ b/fluentd/vendored_gem_src/fluent-plugin-prometheus/lib/fluent/plugin/in_prometheus_tail_monitor.rb
@@ -9,7 +9,7 @@ module Fluent::Plugin
 
     helpers :timer
 
-    config_param :interval, :time, default: 5
+    config_param :interval, :time, default: 1
     attr_reader :registry
 
     MONITOR_IVARS = [
@@ -38,24 +38,31 @@ module Fluent::Plugin
         @base_labels[key] = expander.expand(value)
       end
 
-      if defined?(Fluent::Plugin) && defined?(Fluent::Plugin::MonitorAgentInput)
-        # from v0.14.6
-        @monitor_agent = Fluent::Plugin::MonitorAgentInput.new
-      else
-        @monitor_agent = Fluent::MonitorAgentInput.new
-      end
+      @monitor_agent = Fluent::Plugin::MonitorAgentInput.new
     end
 
     def start
       super
 
       @metrics = {
-        position: @registry.gauge(
+        position: get_gauge(
           :fluentd_tail_file_position,
           'Current position of file.'),
-        inode: @registry.gauge(
+        inode: get_gauge(
           :fluentd_tail_file_inode,
           'Current inode of file.'),
+        maxfsize: get_gauge(
+          :maxfsize,
+          'Current max fsize of file on rotation event'),
+        countonrotate: get_gauge(
+          :countonrotate,
+          'No of rotation noticed by fluentd'),
+        totalbytesread: get_gauge(
+          :totalbytesread,
+          'totalbytes read by fluentd IOHandler'),
+        totalbytesavailable: get_gauge(
+          :totalbytesavailable,
+          'totalbytes available at each instance of rotation - to be read by fluentd IOHandler'),
       }
       timer_execute(:in_prometheus_tail_monitor, @interval, &method(:update_monitor_info))
     end
@@ -77,9 +84,18 @@ module Fluent::Plugin
           # Access to internal variable of internal class...
           # Very fragile implementation
           pe = watcher.instance_variable_get(:@pe)
+          totalbytesread = watcher.instance_variable_get(:@totalbytesread)
+          totalbytesavailable = watcher.instance_variable_get(:@totalbytesavailable)
+          maxfsize = watcher.instance_variable_get(:@maxfsize)
+          countonrotate = watcher.instance_variable_get(:@countonrotate)
           label = labels(info, watcher.path)
           @metrics[:inode].set(label, pe.read_inode)
           @metrics[:position].set(label, pe.read_pos)
+          @metrics[:maxfsize].set(label, maxfsize)
+          @metrics[:countonrotate].set(label, countonrotate)
+          @metrics[:totalbytesread].set(label, totalbytesread)
+          @metrics[:totalbytesavailable].set(label, totalbytesavailable)
+          #@log.info "IN PROMETHEUS PLUGIN pr.read_inode and pe.read_pos #{pe.read_inode} #{pe.read_pos} maxfsize #{maxfsize} countonrotate #{countonrotate} totalbytesread #{totalbytesread} totalbytesavailable #{totalbytesavailable} "
         end
       end
     end
@@ -90,6 +106,14 @@ module Fluent::Plugin
         type: plugin_info["type"],
         path: path,
       )
+    end
+
+    def get_gauge(name, docstring)
+      if @registry.exist?(name)
+        @registry.get(name)
+      else
+        @registry.gauge(name, docstring)
+      end
     end
   end
 end
--- a/fluentd/vendored_gem_src/fluentd/lib/fluent/plugin/in_tail.rb
+++ b/fluentd/vendored_gem_src/fluentd/lib/fluent/plugin/in_tail.rb
@@ -510,6 +510,13 @@ module Fluent::Plugin
         @from_encoding = from_encoding
         @encoding = encoding
         @open_on_every_update = open_on_every_update
+        @totalbytesread=0
+        @totalbytesavailable=0
+        @inodelastfsize_map={}
+        @inodereadfsize_map={}
+        @countonrotate=0
+        @maxfsize=0
+        @logloss=0
       end
 
       attr_reader :path
@@ -519,6 +526,11 @@ module Fluent::Plugin
       attr_accessor :timer_trigger
       attr_accessor :line_buffer, :line_buffer_timer_flusher
       attr_accessor :unwatched  # This is used for removing position entry from PositionFile
+      attr_accessor :totalbytesavailable  # This is used for removing position entry from PositionFile
+      attr_accessor :totalbytesread  # This is used for removing position entry from PositionFile
+      attr_accessor :maxfsize  # This is used for removing position entry from PositionFile
+      attr_accessor :countonrotate  # This is used for removing position entry from PositionFile
+      attr_accessor :logloss  # This is used for removing position entry from PositionFile
 
       def tag
         @parsed_tag ||= @path.tr('/', '.').gsub(/\.+/, '.').gsub(/^\./, '')
@@ -535,7 +547,7 @@ module Fluent::Plugin
 
       def detach
         yield self
-        @io_handler.on_notify if @io_handler
+        @io_handler.on_notify(@inodereadfsize_map) if @io_handler
       end
 
       def close
@@ -553,12 +565,28 @@ module Fluent::Plugin
           stat = nil
         end
 
-        @rotate_handler.on_notify(stat) if @rotate_handler
+        @totalbytesavailable=@rotate_handler.on_notify(stat,@inodelastfsize_map) if @rotate_handler
         @line_buffer_timer_flusher.on_notify(self) if @line_buffer_timer_flusher
-        @io_handler.on_notify if @io_handler
+        @io_handler.on_notify(@inodereadfsize_map) if @io_handler
+
+        @logloss=0
+        if @io_handler && @rotate_handler
+          @totalbytesread=0
+          @countonrotate=0
+          @inodereadfsize_map.each do |k, v|
+            @totalbytesread+=v
+            @countonrotate+=1
+          end
+
+        #@log.info "inodelastfsize_map #{@inodelastfsize_map} inodereadfsize_map #{@inodereadfsize_map}"
+        @logloss=@totalbytesavailable-@totalbytesread
+        #@log.info "totalbytesavailable #{@totalbytesavailable} totalbytesread #{@totalbytesread} logloss #{@logloss} countonrotate #{@countonrotate}"
+        end  
+
       end
 
       def on_rotate(stat)
+        @maxfsize=@pe.read_pos
         if @io_handler.nil?
           if stat
             # first time
@@ -591,6 +619,7 @@ module Fluent::Plugin
             @io_handler = IOHandler.new(self, &method(:wrap_receive_lines))
           else
             @io_handler = NullIOHandler.new
+            @totalbytesread=0
           end
         else
           watcher_needs_update = false
@@ -605,7 +634,7 @@ module Fluent::Plugin
             else # file is rotated and new file found
               watcher_needs_update = true
               # Handle the old log file before renewing TailWatcher [fluentd#1055]
-              @io_handler.on_notify
+              @io_handler.on_notify(@inodereadfsize_map)
             end
           else # file is rotated and new file not found
             # Clear RotateHandler to avoid duplicated file watch in same path.
@@ -732,21 +761,28 @@ module Fluent::Plugin
           @io = nil
           @notify_mutex = Mutex.new
           @watcher.log.info "following tail of #{@watcher.path}"
+          @totalbytesread=0
+          @statatopen=nil
         end
 
-        def on_notify
-          @notify_mutex.synchronize { handle_notify }
+        def on_notify(inodereadfsize_map)
+          @notify_mutex.synchronize { handle_notify(inodereadfsize_map) }
         end
 
-        def handle_notify
+        def handle_notify(inodereadfsize_map)
           with_io do |io|
+          stat = Fluent::FileWrapper.stat(@watcher.path)
+          @statatopen=stat
             begin
               read_more = false
 
               if !io.nil? && @lines.empty?
+                ino=@statatopen.ino
                 begin
                   while true
-                    @fifo << io.readpartial(8192, @iobuf)
+                    buf = io.readpartial(8192, @iobuf)
+                    @totalbytesread+=buf.bytesize
+                    @fifo << buf
                     @fifo.read_lines(@lines)
                     if @lines.size >= @watcher.read_lines_limit
                       # not to use too much memory in case the file is very large
@@ -756,6 +792,7 @@ module Fluent::Plugin
                   end
                 rescue EOFError
                 end
+                inodereadfsize_map[ino]=@totalbytesread
               end
 
               unless @lines.empty?
@@ -784,7 +821,7 @@ module Fluent::Plugin
         def open
           io = Fluent::FileWrapper.open(@watcher.path)
           io.seek(@watcher.pe.read_pos + @fifo.bytesize)
-          io
+          return io
         rescue RangeError
           io.close if io
           raise WatcherSetupError, "seek error with #{@watcher.path}: file position = #{@watcher.pe.read_pos.to_s(16)}, reading bytesize = #{@fifo.bytesize.to_s(16)}"
@@ -840,9 +877,10 @@ module Fluent::Plugin
           @inode = nil
           @fsize = -1  # first
           @on_rotate = on_rotate
+          @totalbytesavailable=0
         end
 
-        def on_notify(stat)
+        def on_notify(stat,inodelastfsize_map)
           if stat.nil?
             inode = nil
             fsize = 0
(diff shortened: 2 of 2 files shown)


fluentd/vendored_gem_src/fluent-plugin-prometheus/lib/fluent/plugin/in_prometheus_tail_monitor.rb
```diff
@@ -38,24 +38,31 @@ def configure(conf)
...
-        inode: @registry.gauge(
+        inode: get_gauge(
           :fluentd_tail_file_inode,
           'Current inode of file.'),
+        maxfsize: get_gauge(
+          :maxfsize,
+          'Current max fsize of file on rotation event'),
+        countonrotate: get_gauge(
+          :countonrotate,
+          'No of rotation noticed by fluentd'),
```
eranra, line 59: s/No/Number

eranra, line 59: s/rotation/rotations

fluentd/vendored_gem_src/fluent-plugin-prometheus/lib/fluent/plugin/in_prometheus_tail_monitor.rb
```diff
@@ -38,24 +38,31 @@ def configure(conf)
...
+        position: get_gauge(
           :fluentd_tail_file_position,
           'Current position of file.'),
-        inode: @registry.gauge(
+        inode: get_gauge(
           :fluentd_tail_file_inode,
           'Current inode of file.'),
+        maxfsize: get_gauge(
+          :maxfsize,
+          'Current max fsize of file on rotation event'),
```
eranra, line 56: Maybe you can drop the `current`? if this is the max fsize than it is enought to state that. Current might not be "user facing" value,.

fluentd/vendored_gem_src/fluent-plugin-prometheus/lib/fluent/plugin/in_prometheus_tail_monitor.rb
```diff
@@ -38,24 +38,31 @@ def configure(conf)
...
+        countonrotate: get_gauge(
+          :countonrotate,
+          'No of rotation noticed by fluentd'),
+        totalbytesread: get_gauge(
+          :totalbytesread,
+          'totalbytes read by fluentd IOHandler'),
+        totalbytesavailable: get_gauge(
+          :totalbytesavailable,
+          'totalbytes available at each instance of rotation - to be read by fluentd IOHandler'),
```
eranra, line 65: I am not sure that I understand the term ... maybe something with the words

fluentd/vendored_gem_src/fluent-plugin-prometheus/lib/fluent/plugin/in_prometheus_tail_monitor.rb
```diff
@@ -77,9 +84,18 @@ def update_monitor_info
...
+          maxfsize = watcher.instance_variable_get(:@maxfsize)
+          countonrotate = watcher.instance_variable_get(:@countonrotate)
           label = labels(info, watcher.path)
           @metrics[:inode].set(label, pe.read_inode)
           @metrics[:position].set(label, pe.read_pos)
+          @metrics[:maxfsize].set(label, maxfsize)
+          @metrics[:countonrotate].set(label, countonrotate)
+          @metrics[:totalbytesread].set(label, totalbytesread)
+          @metrics[:totalbytesavailable].set(label, totalbytesavailable)
+          #@log.info "IN PROMETHEUS PLUGIN pr.read_inode and pe.read_pos #{pe.read_inode} #{pe.read_pos} maxfsize #{maxfsize} countonrotate #{countonrotate} totalbytesread #{totalbytesread} totalbytesavailable #{totalbytesavailable} "
```
eranra, line 98: Consider change to log.debug and remove comment

fluentd/vendored_gem_src/fluent-plugin-prometheus/lib/fluent/plugin/in_prometheus_tail_monitor.rb
```diff
@@ -38,24 +38,31 @@ def configure(conf)
...
-      else
-        @monitor_agent = Fluent::MonitorAgentInput.new
-      end
+      @monitor_agent = Fluent::Plugin::MonitorAgentInput.new
     end
 
     def start
       super
 
       @metrics = {
```
eranra, line 47: Missing: the high level value which is numebr of Log lines dropped is missing ... can you advice on the formula to calculate this from bellow metrics ?

eranra: I saw this in @logloss but this is not exposed as metric (why?)

fluentd/vendored_gem_src/fluentd/lib/fluent/plugin/in_tail.rb
```diff
@@ -553,12 +565,28 @@ def on_notify
...
+
+        #@log.info "inodelastfsize_map #{@inodelastfsize_map} inodereadfsize_map #{@inodereadfsize_map}"
+        @logloss=@totalbytesavailable-@totalbytesread
+        #@log.info "totalbytesavailable #{@totalbytesavailable} totalbytesread #{@totalbytesread} logloss #{@logloss} countonrotate #{@countonrotate}"
+        end  
+
       end
 
       def on_rotate(stat)
+        @maxfsize=@pe.read_pos
```
eranra, line 589: Do you want to MAX on all rotations so you get to more accurate value over time ? (assuming this is coming from fixed value set by conmon)

fluentd/vendored_gem_src/fluentd/lib/fluent/plugin/in_tail.rb
```diff
@@ -519,6 +526,11 @@ def initialize(path, rotate_wait, pe, log, read_from_head, enable_watch_timer, e
       attr_accessor :timer_trigger
       attr_accessor :line_buffer, :line_buffer_timer_flusher
       attr_accessor :unwatched  # This is used for removing position entry from PositionFile
+      attr_accessor :totalbytesavailable  # This is used for removing position entry from PositionFile
```
eranra, line 529: Fix the remark (to be fixed in all variables)

fluentd/vendored_gem_src/fluentd/lib/fluent/plugin/in_tail.rb
```diff
@@ -510,6 +510,13 @@ def initialize(path, rotate_wait, pe, log, read_from_head, enable_watch_timer, e
         @from_encoding = from_encoding
         @encoding = encoding
         @open_on_every_update = open_on_every_update
+        @totalbytesread=0
+        @totalbytesavailable=0
+        @inodelastfsize_map={}
+        @inodereadfsize_map={}
```
eranra, line 516: As rotations continue forever this map is gowing and growing ... I do not know how much ... but this feels like a memory leak to me ... need to see how large this gets in real-life and consider limiting this somehow in time

fluentd/vendored_gem_src/fluentd/lib/fluent/plugin/in_tail.rb
```diff
@@ -510,6 +510,13 @@ def initialize(path, rotate_wait, pe, log, read_from_head, enable_watch_timer, e
         @from_encoding = from_encoding
         @encoding = encoding
         @open_on_every_update = open_on_every_update
+        @totalbytesread=0
+        @totalbytesavailable=0
+        @inodelastfsize_map={}
```
eranra, line 515: Same as next comment

fluentd/vendored_gem_src/fluentd/lib/fluent/plugin/in_tail.rb
```diff
@@ -553,12 +565,28 @@ def on_notify
...
+        @totalbytesavailable=@rotate_handler.on_notify(stat,@inodelastfsize_map) if @rotate_handler
         @line_buffer_timer_flusher.on_notify(self) if @line_buffer_timer_flusher
-        @io_handler.on_notify if @io_handler
+        @io_handler.on_notify(@inodereadfsize_map) if @io_handler
+
+        @logloss=0
+        if @io_handler && @rotate_handler
+          @totalbytesread=0
+          @countonrotate=0
+          @inodereadfsize_map.each do |k, v|
```
eranra, line 576: Might be expensive loop after some time if the number of rotations is significant (maybe you can consolidate the info after some time to a variable and remove a lot of the data in that map)

fluentd/vendored_gem_src/fluentd/lib/fluent/plugin/in_tail.rb
```diff
@@ -854,9 +892,18 @@ def on_notify(stat)
           begin
             if @inode != inode || fsize < @fsize
               @on_rotate.call(stat)
+              inodelastfsize_map[@inode]=@fsize
+              @totalbytesavailable=@fsize
```
eranra, line 896: What is the purpose of maxfsize if you are using the fsize for logloss computation?

fluentd/vendored_gem_src/fluent-plugin-prometheus/lib/fluent/plugin/in_prometheus_tail_monitor.rb
```diff
@@ -9,7 +9,7 @@ class PrometheusTailMonitorInput < Fluent::Plugin::Input
 
     helpers :timer
 
-    config_param :interval, :time, default: 5
+    config_param :interval, :time, default: 1
```
eranra, line 12: Can you explain why we need to move to more frequent interval ?

jcantrill: This if needs to be changed should be made in the config generated for the plug-in, not here

fluentd/vendored_gem_src/fluent-plugin-prometheus/lib/fluent/plugin/in_prometheus_tail_monitor.rb
jcantrill: It's not exposed as this is a raw value we should be able to calculate from exposing what is possible to collect and what is actually collected.

jcantrill (changes requested):
These changes will need to be made against the upstream repo. If we intend to carry them here until those changes land then both these PRs must be moved to lib

fluentd/vendored_gem_src/fluentd/lib/fluent/plugin/in_tail.rb
eranra: @jcantrill Maybe aggrigation of infomration from this map every set time for oldest inodes ... this can be second phase and not part of this PR from my PoV ... but we should not forget to do that.

fluentd/vendored_gem_src/fluentd/lib/fluent/plugin/in_tail.rb
eranra: @jcantrill I think we want need one value ... prodinf a metric per file is not very useful and not very scalable. But, maybe the accomolation can be done in 'streaming' like manner and not doing the entire calculation every time

fluentd/vendored_gem_src/fluentd/lib/fluent/plugin/in_tail.rb
alanconway: I don't think this map is necessary, and the logic can be simpler. See https://github.com/alanconway/file-metric/blob/b0cb7946c3e9dd6b607ef6f5e02218e5c1b33e94/cmd/file-metric/main.go#L69 which is in Go, but the logic can be reproduced in ruby. In fact we don't need the inode, the only thing that matters is when the file size *drops*, that means we have started a new rotation.

fluentd/vendored_gem_src/fluent-plugin-prometheus/lib/fluent/plugin/in_prometheus_tail_monitor.rb
alanconway: I changed the terms in the JIRA, @eranra  if you agree these are more expressive we should update the code:

    bytes_logged (counter) bytes written to log files by containers.
    bytes_collected (counter) bytes read by the collector for forwarding. 

Log loss can be computed by a prometheus query expression: (bytes_logged - bytes_collected)

@pmoogi-redhat note the metrics should be _counters_ not _guages_. They always increase.

fluentd/vendored_gem_src/fluent-plugin-prometheus/lib/fluent/plugin/in_prometheus_tail_monitor.rb
alanconway: We don't need to track file size. The interesting metrics are bytes_logged/bytes_collected. max file size is an implementation detail that could change in future (e.g. if we start streaming logs from crio directly via pipes)

fluentd/vendored_gem_src/fluentd/lib/fluent/plugin/in_tail.rb
alanconway: I've been learning about prometheus and here's what I think: we have 2 metrics, counters, bytes_loggged and bytes_collected. However those metrics have labels that allow the user to refine the set of log streams they care about. In Go this is called a CounterVec, not sure what it is in fluentd. See https://github.com/alanconway/file-metric/blob/b0cb7946c3e9dd6b607ef6f5e02218e5c1b33e94/cmd/file-metric/main.go
I need to do some research to figure out what the right set of labels is. In my go example it's just the file name, but we need to look at other observatorium label schemes and consider our own data model to find a set that makes sense for queries. The cardinality is always going to be per-log-file, but the log file name isn't a very obvious way for users to query.

fluentd/vendored_gem_src/fluentd/lib/fluent/plugin/in_tail.rb
alanconway: We don't care about max file size. We only care about bytes_logged and bytes_collected. I think this code can be simplified to accumulate directly into the prometheus client metrics (like the Go code at https://github.com/alanconway/file-metric/blob/master/cmd/file-metric/main.go) so that the embedded prom client has an up-to-date view whenever there is a pull from prometheus.

alanconway (changes requested):
Along with the embedded comments, the code should be separated into its own class so we have minimal changes to the fluentd in_tail plugin. That will make it easier to upgrade to new versions of in_tail in future.

fluentd/vendored_gem_src/fluent-plugin-prometheus/lib/fluent/plugin/in_prometheus_tail_monitor.rb
eranra: @alanconway yes ... those terms are simple and make sense :-)

fluentd/vendored_gem_src/fluent-plugin-prometheus/lib/fluent/plugin/in_prometheus_tail_monitor.rb
pmoogi-redhat: restored back all changes done vendored_gem_src as it is not supposed to be changed.

pmoogi-redhat: restored back all changes done in vendored_gem_src as it is not supposed to be changed.

fluentd/vendored_gem_src/fluent-plugin-prometheus/lib/fluent/plugin/in_prometheus_tail_monitor.rb
pmoogi-redhat: Restored back all changes done in vendored_gem_src as it is not supposed to be changed. Final changes don't have maxfsize and countonrotate parameters getting published.

fluentd/vendored_gem_src/fluent-plugin-prometheus/lib/fluent/plugin/in_prometheus_tail_monitor.rb
pmoogi-redhat: changed parameter names from totalbytesavailable and totalbytesread to  totalbytes_logged_in_disk
 &  totalbytes_collected_from_disk respectively. Got a separate class implementation to monitor stat for getting totalbytes_logged_in_disk

fluentd/vendored_gem_src/fluent-plugin-prometheus/lib/fluent/plugin/in_prometheus_tail_monitor.rb
pmoogi-redhat: restored back all changes done in vendored_gem_src as that is not supposed to be changed.

fluentd/vendored_gem_src/fluent-plugin-prometheus/lib/fluent/plugin/in_prometheus_tail_monitor.rb
pmoogi-redhat: restored back all changes done in vendored_gem_src as that is not supposed to be changed.

fluentd/vendored_gem_src/fluentd/lib/fluent/plugin/in_tail.rb
pmoogi-redhat: Taken out this change as maxfsize need not be monitored as a metric.

pmoogi-redhat: @eranra MAX over all rotation is not required as rotation always happen when fsize > maxsize of log file.

fluentd/vendored_gem_src/fluentd/lib/fluent/plugin/in_tail.rb
pmoogi-redhat: resolved. please see my earlier comments.

fluentd/vendored_gem_src/fluentd/lib/fluent/plugin/in_tail.rb
pmoogi-redhat: Hash map based implementation is taken out to make implementation more performance efficient. A new write_watcher.rb is introduced to measure totalbytes_logged_in_disk.

fluentd/vendored_gem_src/fluentd/lib/fluent/plugin/in_tail.rb
pmoogi-redhat: @eranra @alanconway resolved this in the new implementation.

fluentd/vendored_gem_src/fluentd/lib/fluent/plugin/in_tail.rb
pmoogi-redhat: maxfsize is taken out now as it is not required to be monitored.

fluentd/vendored_gem_src/fluent-plugin-prometheus/lib/fluent/plugin/in_prometheus_tail_monitor.rb
pmoogi-redhat: all changes done in the vendored_gem_src are restored back as it is not supposed to be changed for creating custom plugin. now all changes sit in fluentd/lib/ folder.

pmoogi-redhat:
closing this one as new PR is raised with lot many revamp of this implementation and placement of code.

PR #2064, opened 2021-02-19
Bug 1930360: fluentd is unable to send logs to Elasticsearch because it can't authenticate to the es-proxy

syedriko:
### Description

Switch the base image for the fluentd image to ubi8 in order to pick up newer libcurl and its dependencies.

/cc @jcantrill 
/assign @ewolinetz 

### Links
- Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1930360


jcantrill:
/approve

@syedriko  looks like we are missing failing on the curator build which seems odd. needs investigation

jcantrill:
Note this change will only fix local dev workflow.  Need to work with art and make a change against "ocp-build-data" to fix downstream builds

jcantrill:
labeling "valid" bug as this issue only affects 4.5 and pulls in changes that already exist in later releases

derekwaynecarr:
(Patch-Manager) approve

PR #2070, opened 2021-02-25
LOG-1213: Metric for inbound log data loss at the collector

pmoogi-redhat:
### Description
Currently in_tail plugin doesn't support publishing of inbound logloss - i.e. difference between total bytes written to disk (logfile) and total bytes collected or read by fluentd. This PR got changes in fluentd/lib/fluent/plugin/in_tail.rb, and fluent-plugin-prometheus/lib/fluent/plugin/in_prometheus_tail_monitor.rb plugins to enable publishing of the below parameters 
1. totalbytes_logged_in_disk (written to disk per container process)
2. totalbytes_collected_from_disk (read or collected by fluentd per container process)

/cc @alanconway @jcantrill 
/assign @alanconway 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-1213

- Enhancement proposal: many rotations getting missed by fluentd, next enhancement proposal is to get fluentd know about all actual rotations done by CRIO/conmon process by reading extra meta data such as  <timestamp containerid maxsizeoflogfile rotation-counter>. Based on what rotations fluentd could track and which one got missed computing log-loss as [no-of-rotations-those-missed * maxsizeoflogfile + sum over all tracked rotations by fluentd as [totalbytes_logged_in_disk - totalbytes_collected_from_disk]


fluentd/lib/fluent-plugin-prometheus/in_prometheus_tail_monitor.rb
```diff
@@ -0,0 +1,108 @@
+require 'fluent/plugin/input'
+require 'fluent/plugin/in_monitor_agent'
+require 'fluent/plugin/prometheus'
+
+module Fluent::Plugin
+  class PrometheusTailMonitorInput < Fluent::Plugin::Input
+    Fluent::Plugin.register_input('prometheus_tail_monitor', self)
```
alanconway, line 7: If we use this plugin, change the name to avoid confusion with the fluentd built-in.
Can we just add this logic to the other plugin, so we only have one? 
We're always going to want the metric functionality so I don't see much benefit in splitting them

fluentd/lib/fluent-plugin-prometheus/in_prometheus_tail_monitor.rb
```diff
@@ -0,0 +1,108 @@
...
+    include Fluent::Plugin::PrometheusLabelParser
+
+    helpers :timer
+
+    config_param :interval, :time, default: 1
```
alanconway, line 12: Why did you change the default from 5s to 1s?

alanconway: This is hard to review because there's so much code that is copied from the original. Can you subclass the original and  override just what we need to change? That would make it easier to review and maintain.

fluentd/lib/in_tail_metric/write_watcher.rb
```diff
@@ -0,0 +1,76 @@
...
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+require 'fluent/plugin/input'
+
+module Fluent::Plugin
+  class TailInput < Fluent::Plugin::Input
```
alanconway, line 20: Need to rename class TailInputMetric to avoid clash with original plugin class.

fluentd/lib/in_tail_metric/write_watcher.rb
```diff
@@ -0,0 +1,76 @@
...
+        if !stat || info.ino != stat.ino || stat.size < info.size
+          info.sum_bytes_logged_in_disk += info.size
+          info.size = 0
+        end
+        info.ino, info.size = stat.ino, stat.size if stat
+
+        @notify_count += 1
+        if @notify_count> 100
+          @notify_count = 0
+          @log.error "FIXME #{path} sum of bytes available in disk size #{info.sum_bytes_logged_in_disk + info.size}"
```
alanconway, line 52: Remove .error FIXME logs. Any logs you want to keep should be .debug logs.

fluentd/lib/in_tail_metric/write_watcher.rb
```diff
@@ -0,0 +1,76 @@
...
+        end
+        # Accumulate last size if file changed (no stat, new inode or truncated)
+        if !stat || info.ino != stat.ino || stat.size < info.size
+          info.sum_bytes_logged_in_disk += info.size
+          info.size = 0
+        end
+        info.ino, info.size = stat.ino, stat.size if stat
+
+        @notify_count += 1
+        if @notify_count> 100
```
alanconway, line 50: Remove this if entirely, I just put it in there for debugging.

alanconway (changes requested):
I suggest metric names:
- fluentd_input_status_total_bytes_logged 
- fluentd_input_status_total_bytes_collected 
This fits the pattern of other fluentd metrics. The "on_disk", seems redundant.

The code is hard to review because so much of it is copied from fluentd. Can you subclass the original plugin classes and just override what you need instead of copying the code?

fluentd/lib/fluent-plugin-prometheus/in_prometheus_tail_monitor.rb
pmoogi-redhat: resolved

fluentd/lib/in_tail_metric/write_watcher.rb
pmoogi-redhat: need to further discuss with you tonight.

fluentd/lib/in_tail_metric/write_watcher.rb
pmoogi-redhat: resolved

fluentd/lib/in_tail_metric/write_watcher.rb
pmoogi-redhat: resolved

fluentd/lib/fluent-plugin-prometheus/in_prometheus_tail_monitor.rb
pmoogi-redhat: need further discussions

pmoogi-redhat: class PrometheusTailMontiorInput is renamed to  PrometheusTailMontiorInputMetric

fluentd/lib/in_tail_metric/write_watcher.rb
pmoogi-redhat: parent class class TailInput < Fluent::Plugin::Input not required so keeping only module Fluent::Plugin
  class WriteWatcher < Fluent::Plugin::Input

fluentd/lib/fluent-plugin-prometheus/in_prometheus_tail_monitor.rb
pmoogi-redhat: resolved. have rolled back this change.

fluentd/lib/in_tail_metric/write_watcher.rb
pmoogi-redhat: parent class class TailInput < Fluent::Plugin::Input not required so keeping only module Fluent::Plugin
class WriteWatcher < Fluent::Plugin::Input

fluentd/lib/fluent-plugin-prometheus/in_prometheus_tail_monitor.rb
```diff
@@ -51,11 +51,11 @@ def start
         inode: get_gauge(
           :fluentd_tail_file_inode,
           'Current inode of file.'),
-        totalbytes_logged_in_disk: get_gauge(
-          :fluentd_totalbytes_logged_in_disk,
+        totalbytes_logged: get_gauge(
```
alanconway, line 54: Drop the bytes-logged metric, it isn't useful to generate it in fluentd. The same rotations are missed for both metrics, so we'll never see an accurate difference between logged and collected. We'll need to move that to Go.

fluentd/lib/fluent-plugin-prometheus/in_prometheus_tail_monitor.rb
```diff
@@ -51,11 +51,11 @@ def start
...
+          :fluentd_totalbytes_logged,
           'totalbytes logged in disk.'),
-        totalbytes_collected_from_disk: get_gauge(
-          :fluentd_totalbytes_collected_from_disk,
+        totalbytes_collected: get_gauge(
```
alanconway, line 57: Consistent underscores: total_bytes_collected
The metrics should be counters, not gauges. The total bytes collected always increases. When fluentd restarts the counter will reset to 0. Prometheus will handle this correctly for a counter metric (continue to accumulate the total) but for a guage metric the value at prometheus will drop to 0. Guages are intended for values that can go up and down.

fluentd/lib/in_tail_metric/in_tail_metric.rb
```diff
@@ -566,6 +566,7 @@ def on_notify
         @line_buffer_timer_flusher.on_notify(self) if @line_buffer_timer_flusher
         @io_handler.on_notify if @io_handler
         @totalbytes_collected=@write_watcher.update_totalbytes_collected(@path)
+        @log.debug "TOTALBYTES LOGGED #{@totalbytes_logged} TOTALBYTESCOLLECTED #{@totalbytes_collected}"
```
alanconway, line 569: Remove the @log - this is a sensitive path. I'm not sure what the overhead of disabled logs statements is but best not take a chance.  It works now, so we can get the values via metrics :)

fluentd/lib/in_tail_metric/write_watcher.rb
```diff
@@ -17,9 +17,8 @@
 require 'fluent/plugin/input'
 
 module Fluent::Plugin
-  class TailInput < Fluent::Plugin::Input
```
alanconway: We can drop this watcher since it' doesn't work correctly. The Go process will generate these metrics.

alanconway (changes requested):
Some simple clean up comments.
We will need a test to prove that this works before we merge it.
I think it would be sufficient to test it on a stand-alone fluentd with a very simple dummy configuration, and use the Go HTTP client to scrape the metrics and validate them.
Check out tests/functional and check with @jcantrill on using his functional framework, it's designed to deploy a stand-alone fluentd without setting up the whole logging system.

fluentd/lib/in_tail_metric/in_tail_metric.rb
pmoogi-redhat: resolved

pmoogi-redhat:
get_gauge_or_counter is not working as per expectations, throwing some errors hence leaving it to get_gauge type for total_bytes_collected metric. require some help there. have tested the functionality by looking at prometheus dashboard if total_bytes_collected getting published correctly. it is working fine as per my stand-alone test environment.

fluentd/lib/fluent-plugin-prometheus/in_prometheus_tail_monitor.rb
```diff
@@ -0,0 +1,114 @@
...
+      if @registry.exist?(name)
+        @registry.get(name)
+      else
+        @registry.gauge(name, docstring)
+      end
+    end
+
+
+
+    def get_gauge_or_counter(name, docstring)
```
alanconway, line 105: get_gauge_or_counter looks like dead code, it refers to an attribute @guage_or_counter that isn't defined anywhere. 
Delete it and replace with similar `get_counter()` which calls `@registry.counter()` to create a counter metric.

fluentd/lib/fluent-plugin-prometheus/in_prometheus_tail_monitor.rb
```diff
@@ -0,0 +1,114 @@
...
+      super
+
+      @metrics = {
+        position: get_gauge(
+          :fluentd_tail_file_position,
+          'Current position of file.'),
+        inode: get_gauge(
+          :fluentd_tail_file_inode,
+          'Current inode of file.'),
+        total_bytes_collected: get_gauge(
```
alanconway, line 54: Should be a counter, see below.

fluentd/lib/in_tail_metric/write_watcher.rb
```diff
@@ -0,0 +1,72 @@
...
+
+      # :ino inode number of current file
+      # :size of current file
+      # :bytes_logged bytes_logged size of previous files for this path
+      Entry = Struct.new(:ino, :size,:sum_bytes_collected,:sum_bytes_logged)
+
+      def on_notify(path, stat)
+        info = @map[path]
+        if !info
+          @log.debug "GETTING TOTALBYTES_LOGGED METRICS FOR #{path} new file"
```
alanconway, line 38: Remove the debug log, sensitive path.

alanconway (changes requested):
Almost there!

fluentd/lib/in_tail_metric/write_watcher.rb
```diff
@@ -0,0 +1,72 @@
...
+     def update_total_bytes_logged(path)
+       info=@map[path]
+       info.sum_bytes_logged
+     end
+     
+     def update_total_bytes_collected(path)
+       info=@map[path]
+       info.sum_bytes_collected
+     end
+
```
alanconway, line 70:
```suggestion
     def forget(path)
       @map.delete(path)
     end
```

fluentd/lib/in_tail_metric/write_watcher.rb
```diff
@@ -0,0 +1,72 @@
+#
```
alanconway, line 1: WriteWatcher is no longer needed, delete this file and remove all references to it.

alanconway (changes requested):
A couple of points to clean up, otherwise LGTM.

alanconway:
Final names for the metrics:

    log_logged_bytes_total: total number of bytes written to a single log file path, accounting for rotations.
    log_collected_bytes_total: total number of bytes collected from a single log file.
    Both metrics have the label "path" which is the absolute path to the log file.

fluentd/lib/fluent-plugin-prometheus/in_prometheus_tail_monitor.rb
```diff
@@ -0,0 +1,120 @@
+require 'fluent/plugin/input'
```
jcantrill, line 1: This file should be either removed entirely, appear as a patch to vendored_src, are a new file to vendored_src since you cant patch it

fluentd/lib/in_tail/in_tail.rb
```diff
@@ -0,0 +1,1048 @@
+#
```
jcantrill, line 1: This file should be either removed entirely, appear as a patch to vendored_src, are a new file to vendored_src since you cant patch it

fluentd/lib/in_tail/write_watcher.rb
```diff
@@ -0,0 +1,72 @@
+#
```
jcantrill, line 1: This file should be either removed entirely, appear as a patch to vendored_src, are a new file to vendored_src since you cant patch it

alanconway: All the files except write_watcher.rb should be patches to existing vendored files. They were renamed when I thought we wer creating a separate plugin, if we're patching we should have the minimum diff from the vendored files.

write_watcher is misnamed, all it contains now is mertric exporter code. Probably it should be merged into the vendored fluentd exporter, wherever that is.

fluentd/lib/fluent-plugin-prometheus/in_prometheus_tail_monitor.rb
pmoogi-redhat: @jcantrill  i would like to know where we can keep this file version controlled as a single piece of truth to all changes that are used to create patch over vendored_src baseline implementation of this plugin.

fluentd/lib/in_tail/write_watcher.rb
pmoogi-redhat: @jcantrill  i would like to know where we can keep this file version controlled as a single piece of truth to all changes that are used to create patch over vendored_src baseline implementation of this plugin.

alanconway: @pmoogi-redhat the patch file goes in the /fluentd directory, patches are installed during the image build by https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/install-gems.sh 
See the comment at the top about naming the patch file.

alanconway (approved):
/lgtm

fluentd/lib/fluent-plugin-prometheus/in_prometheus_tail_monitor.rb
```diff
@@ -0,0 +1,131 @@
...
+      @monitor_agent = Fluent::Plugin::MonitorAgentInput.new
+    end
+
+    def start
+      super
+
+      @metrics = {
+        position: get_gauge(
+          :fluentd_tail_file_position,
+          'Current position of file.'),
```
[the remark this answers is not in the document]
pmoogi-redhat: This is part of previous baseline version of fluent-plugin-prometheus/in_prometheus_tail_monitor.rb. I added support for publishing log_collected_bytes_total metric

pmoogi-redhat (commented):
@alanconway @jcantrill kindly review - added k8 regex based filename parsing.

fluentd/lib/fluent-plugin-prometheus/in_prometheus_tail_monitor.rb
```diff
@@ -0,0 +1,131 @@
...
+          end
+          @metrics[:total_bytes_collected].increment(label, total_bytes_collected - @prev_total_bytes_collected[label])
+          @prev_total_bytes_collected[label]=total_bytes_collected
+        end
+      end
+    end
+
+    def labels(plugin_info, path)
+      #taking out dirname and .log from the full pathname e.g. /var/log/containers/xxx.log --> xxx
+      tmp = path[20..-5]
```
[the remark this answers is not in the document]
pmoogi-redhat: Added support for regex based parsing of filepath name

fluentd/lib/fluent-plugin-prometheus/in_prometheus_tail_monitor.rb
```diff
@@ -0,0 +1,116 @@
+require 'fluent/plugin/input'
+require 'fluent/plugin/in_monitor_agent'
+require 'fluent/plugin/prometheus'
+
+module Fluent::Plugin
+  class PrometheusTailMonitorInputMetric < Fluent::Plugin::Input
+    Fluent::Plugin.register_input('prometheus_tail_monitor', self)
```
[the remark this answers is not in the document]
pmoogi-redhat: renamed the plugin files

alanconway:
@jcantrill @pmoogi-redhat I think we can re-factor this code into a single, fully independent plug-in belonging to us, with no patches to fluentd. That will solve the labelling problems.
The idea is to follow the model of the fluentd prometheus-tail-monitor: 
https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/vendored_gem_src/fluent-plugin-prometheus/lib/fluent/plugin/in_prometheus_tail_monitor.rb#L1
That plugin scans for registered in-tail plugins and uses inside knowledge of the in-tail implementation to grab the read-position and inode values collected by in-tail. It samples on a timer, so it doesn't see every read - but that doesn't matter.  In-tail keeps watching the old inode for a grace period of several seconds after the file rolls over, which gives us time to register the last read position -so we should get a full count.
I've started writing this out, but I'm not sure we'll have time to get it done and tested for 5.1. I'll try to finish toda so @pmoogi-redhat can review & test it tomorrow.

fluentd/lib/fluent-plugin-prometheus/in_prometheus_tail_monitor.rb
```diff
@@ -0,0 +1,131 @@
...
+      }
+      timer_execute(:in_prometheus_tail_monitor, @interval, &method(:update_monitor_info))
+    end
+
+    def update_monitor_info
+      opts = {
+        ivars: MONITOR_IVARS,
+      }
+
+      agent_info = @monitor_agent.plugins_info_all(opts).select {|info|
```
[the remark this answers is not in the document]
pmoogi-redhat: {} used in the baseline version, i have not introduced it.

pmoogi-redhat:
Separate PR is raised to enable this new metric as a new plugin - new_plugin_Metric-for-inbound-log-data-loss-at-the-collector-JIRA-ticket-LOG-1032. Hence closing this PR which reflects in_tail and fluent-prometheus-plugin changes.

PR #2075, opened 2021-03-05
Update hack/deploy-logging.sh to use CI env vars

jcantrill:
### Description
rebaesd 5.0 backport of https://github.com/openshift/origin-aggregated-logging/pull/2074 to fix test to use CI pipeline vars

cc @vimalk78 @alanconway

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,3 @@
-OPERATOR_LOGGING_IMAGE_STREAM?=stable
 
 FLUENTD_IMAGE?="openshift/origin-logging-fluentd"
 KIBANA_IMAGE?="openshift/origin-logging-kibana6"
@@ -10,7 +9,7 @@ build-images: build-all-images
 .PHONY: build-images
 
 test:
-	OPERATOR_LOGGING_IMAGE_STREAM=$(OPERATOR_LOGGING_IMAGE_STREAM) MASTER_VERSION=5.0 CLO_BRANCH=release-5.0 EO_BRANCH=release-5.0 ./hack/test-e2e.sh
+	LOGGING_VERSION=5.0 ./hack/test-e2e.sh
 .PHONY: test
 
 .PHONY: test-pre-upgrade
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -10,7 +9,7 @@ build-images: build-all-images
 .PHONY: build-images
 
 test:
-	OPERATOR_LOGGING_IMAGE_STREAM=$(OPERATOR_LOGGING_IMAGE_STREAM) ./hack/test-e2e.sh
+	LOGGING_VERSION=5.0 ./hack/test-e2e.sh
```
alanconway, line 12: Should that be `LOGGING_VERSION=${LOGGING_VERSION:-5.0}` too allow override?

hack/deploy-logging.sh
```diff
@@ -168,23 +168,26 @@ get_cluster_version_maj_min() {
...
 
-CLO_BRANCH="release-${MASTER_VERSION}"
-EO_BRANCH="release-${MASTER_VERSION}"
-
-if [ "${MASTER_VERSION}" == "${MASTER_RELEASE_VERSION}" ]; then
-  CLO_BRANCH="master"
-  EO_BRANCH="master"
-fi
+LOGGING_IS=${LOGGING_IS:-logging}
+LOGGING_VERSION=${LOGGING_VERSION:-5.1}
```
alanconway, line 182: The Makefile sets LOGGING_VERSION=5.0, should this be 5.0?

hack/deploy-logging.sh
```diff
@@ -218,23 +220,19 @@ construct_image_name() {
 update_images_in_clo_yaml() {
     local yamlfile=$1
     local clo_img=${2:-} #unused.  leaving for compatibility
-    local version=${3:-latest}
+    local version=${3:-latest} #unused. leaving for compatibility
```
alanconway, line 223: Why leaving for compatibility? You're ignoring the argument, why create an unused local variable?

alanconway (changes requested):
Marked a couple of nits, otherwise LGTM

Makefile
jcantrill: not here I don't believe.  This is an explicit change to 5.0 branch so I don't believe we would desire for it to be anything else

PR #2077, opened 2021-03-05
Converting JKS format to PKCS for elasticsearch

Red-GV:
Second Part of this PR: https://github.com/openshift/elasticsearch-operator/pull/669

/cc @periklis 
/assign @ewolinetz 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-95


elasticsearch/utils/logging
```diff
@@ -161,52 +113,6 @@ wait_for_port_open() {
     exit 1
 }
 
-build_jks_from_pem() {
```
ewolinetz, line 164: we're still going to need to keep some of this because we're converting the `pem` files into `pkcs12` in here. we just don't need the latter half where we convert `pkcs12` into `jks`

Red-GV: thanks for clarifying. I've restored the functions and removed out the jks portions.

periklis (approved):
/lgtm

elasticsearch/utils/logging
```diff
@@ -113,6 +161,41 @@ wait_for_port_open() {
...
+
+  openssl                                   \
+    pkcs12                                  \
+    -export                                 \
+    -in $secret_dir/$cert_name              \
+    -inkey $secret_dir/$key_name            \
+    -out $secret_dir/$p12_name.p12          \
+    -passout pass:kspass
+
+  keytool                                   \
```
ewolinetz, line 179: won't using keytool create a .jks file?

Red-GV:
All keystores and truststores now have PKCS#7 data. I also removed the hack files because they are not being used anymore in this repo. My understanding is that the other repos are creating the secrets now and they have not been touched in years.

Would be dependent on #2094 to now use python3 properly.

elasticsearch/utils/logging
```diff
@@ -80,11 +80,11 @@ check_index_exists() {
   [[ $response_code -eq 200 ]] || exit 1
 }
 
-# Pull in the certs provided in our secret and generate our necessary jks and truststore files
-build_jks_truststores() {
```
ewolinetz, line 84: i think we need to keep this function for now and then we can remove it afterwards once EO is only configuring clusters with .p12

PR #2086, opened 2021-03-23
LOG-660: Added fluent-plugin-label-router to support Forwarder Label Selector

k-keiichi-rh:
### Description
This PR vendors [the label router plugin](https://github.com/banzaicloud/fluent-plugin-label-router) into the fluentd image.

/cc @alanconway @jcantrill
/assign

### Links
- Github issue: https://github.com/openshift/cluster-logging-operator/issues/787
- JIRA: https://issues.redhat.com/browse/LOG-660
- Enhancement proposal: https://github.com/openshift/enhancements/blob/master/enhancements/cluster-logging/forwarder-label-selector.md


fluentd/Gemfile.lock
```diff
@@ -59,6 +59,8 @@ GEM
       fluentd (>= 0.10.58, < 2)
```
alanconway, line 59: I think this should be replaced by the fluentd requirement below.
```suggestion
      fluentd (>= 0.14.10, < 2)      
```

fluentd/Gemfile.lock
```diff
@@ -59,6 +59,8 @@ GEM
...
       ltsv
       ruby-kafka (>= 0.7.8, < 2)
+    fluent-plugin-label-router (0.2.4)
+      fluentd (>= 0.14.10, < 2)
```
alanconway, line 63:
```suggestion
```

alanconway (changes requested):
One minor change, otherwise looks good.

fluentd/Gemfile.lock
k-keiichi-rh: Thanks. I will fix it. 
Just confirmation. fluent-plugin-label-router requires ">= 0.14.10". So minimum version for other components should be also raised. Is my understanding correct? 
If yes, do I need to raise the lower version of other components(0.14.0 and 0.14.2)?
Or is it no problem if 0.14.x is same for the compatibility of fluentd?

k-keiichi-rh: I updated my patch. If there is no problem, please merge it.

PR #2094, opened 2021-03-26
Setting python alternatives to python3

Red-GV:
Setting the alternatives to python3 so that local elasticsearch images can run correctly. Without this, the image will break down in `es_util_env` because the python command cannot be run. The rhe8 Dockerfile already has this command in it.

/cc @periklis 
/assign @jcantrill


jcantrill (commented):
/lgtm

PR #2099, opened 2021-04-08
Bug 1940563: Fix Dockerfile to be buildable by ART

jcantrill:
### Description
This PR updates the Dockerfile to consume the Elasticsearch and be buildable by ART

### Links
https://bugzilla.redhat.com/show_bug.cgi?id=1940563


hack/build-images.sh
```diff
@@ -304,17 +304,16 @@ if [ "${PUSH_ONLY:-false}" = false ] ; then
...
     echo "-                                                                                                              -"
     echo "Building image $img - this may take a few minutes until you see any output..."
     echo "-                                                                                                              -"
     echo "----------------------------------------------------------------------------------------------------------------"
-    $IMAGE_BUILDER build $IMAGE_BUILDER_OPTS $mountarg -f $dfpath$suffix -t "$fullimagename" $dir
+    if [ "$dir" = "elasticsearch" ] ; then
+        $(pwd)/hack/build-component-image.sh elasticsearch $fullimagename
+    else
+      $IMAGE_BUILDER build $IMAGE_BUILDER_OPTS $mountarg -f $dfpath -t "$fullimagename" $dir
+    fi
```
ewolinetz, line 316: why not make all of these just use the build-component-image.sh script?

jcantrill: I was doing as little as possible to modify what is here while still converging on single source of truth for the Dockerfile.  I don't know its worth the effort to fix in 4.6 but agree we should have that as a goal for 5.x

mfojtik:
[patch-manager] :rocket: Approved for z-stream by score: 1.50

PR #2133, opened 2021-06-24
LOG-1379: Updated viaq data model

sasagarw:
### Description

This PR:
- Removes `org.ovirt.viaq-collectd.asciidoc` from docs
- Removes `org.ovirt.viaq-collectd.template.json` from `elasticsearch/index_templates`
- Updates the index_templates and index_pattern to new changes from [PR](https://github.com/ViaQ/elasticsearch-templates/pull/129)
- Updates `/elasticsearch/init/0200-seed-index-templates` to delete deprecated templates from clusters that are upgrading

/cc @lukas-vlcek 
/assign @ewolinetz 

### Links

- Depending on PR(s): https://github.com/ViaQ/elasticsearch-templates/pull/129
- JIRA: https://issues.redhat.com/browse/LOG-1379


lukas-vlcek (approved):
/lgtm

I am thinking if we need to take care of removal of deprecated index templates from ES cluster too. Right now when users upgrade all present index templates in ES will be carried over including those that were deprecated here.

elasticsearch/init/0200-seed-index-templates
```diff
@@ -17,6 +17,25 @@
 
 source "logging"
 
+info Deleting deprecated index templates
+shopt -s failglob
+for deprecated_file in ${ES_HOME}/index_templates/deprecated/*.json
```
ewolinetz, line 22: didn't we talk about just deleting any templates that match our naming pattern that we don't provide in the es image anymore?

```
templates = getListOfTemplates()
for template in templates
  if template !matchProvidedTemplates
    delete template
  fi
done
```

sasagarw: @ewolinetz I tried that but it was making the script more complicated with multiple loops. Also, we have to embed python to parse the output of `_cat/templates`. I feel this would be a better way of handling by putting in deprecated folder and then deleting. This also reduces little overhead and code looks cleaner to me.

lukas-vlcek: @ewolinetz Do we not allow CUs to push in their own index templates?

elasticsearch/init/0200-seed-index-templates
```diff
@@ -17,6 +17,25 @@
...
+do
+    template=`basename $deprecated_file`
+    # Check if index template already exists
+    response_code=$(es_util --query=_template/$template \
+        ${DEBUG:+-v} -s \
+        --request HEAD --head --output /dev/null \
+        -w '%{response_code}')
+    if [ "${response_code}" == "200" ]; then
+        info "Index template '$template' found in the cluster, deleting it"
```
lukas-vlcek, line 31: Just a small nit, but maybe we can add the word "deprecated" into the log message.
> Deprecated index templates .... found ... blah, deleting it.
It makes is explicit if only a small snippet of the log output is shared ...

elasticsearch/init/0200-seed-index-templates
ewolinetz: > @ewolinetz Do we not allow CUs to push in their own index templates?

we can, but we can just advise them to not match our template naming convention

> @ewolinetz I tried that but it was making the script more complicated with multiple loops. Also, we have to embed python to parse the output of `_cat/templates`.

Why do we need python to parse the output? the `_cat` end points print new line delimited strings which feeds into a loop easily.

ewolinetz: also, it doesn't make sense to maintain a list of deprecated templates simply to delete them from the cluster. for the sanity of the repo we should remove ones that we don't provide that match our naming convention.

lukas-vlcek: @sasagarw I think what @ewolinetz suggests is the following.

Get two lists:

1) **First list** (what comes in the image)
- pull index templates files from provided folder (where files from https://github.com/openshift/origin-aggregated-logging/tree/master/elasticsearch/index_templates are copied)
- filter out all files that do not match `common.settings.*.template.json` or `com.redhat.*.template.json`
2) **Second list** (what is currently in ES cluster)
- do the same, except pull index templates from ES, ideally from `_cat/templates` (you can [tell it to output only a single column](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/cat.html#headers) with the name of the index template - which is the same as the file name it was created from)
- again filter out all items that do not match the above mentioned patters

Delete from ES all index templates that are found in list 2) but not in list 1).

ewolinetz: yes, except we don't need to filter for the first list :)

sasagarw: @ewolinetz @lukas-vlcek both the ways we will be deleting custom templates created by CUs. Also, we create `ocp-gen-app` and `ocp-gen-infra` templates which will not match the pattern and will be deleted wrongly. 
For `ocp-gen-(app|infra)` I can filter using pattern `ocp-gen-*` but the custom templates of CUs will be wrongly deleted.

sasagarw:
The only thing is that we need to ensure that CUs create custom templates without `.json` extension.

sasagarw:
/assign @igor-karpukhin
need relabels on this

lukas-vlcek (approved):
/lgtm
Thanks!

PR #2144, opened 2021-06-30
Updated ingest plugin path

vimalk78:
### Description
Fixed ingest plugin path in ci-env.sh
This is the path from which ingest plugin will be installed into ES image

/cc @lukas-vlcek 
/assign @jcantrill 

/cherry-pick master

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA:
- Enhancement proposal:


lukas-vlcek (commented):
/approve

PR #2146, opened 2021-07-06
LOG-1499: Vendor fluent-plugin-elasticsearch v5.0.5 to drop messages

jcantrill:
### Description
This PR:
* Vendors a new version of the elasticsearch plugin which drops log entries the server will never be capable of consuming
* The plugin will unblock a particular chunk and allow the remaining messages to be evaluated and submitted

/assign @alanconway 

### Links
https://issues.redhat.com/browse/LOG-1499


alanconway (commented):
/lgtm

PR #2147, opened 2021-07-07
fix storage class to match GCP

jcantrill:
### Description
This PR fixes the smoke test to use a storage class that is valid for GCP

cc @ewolinetz

### DIFF
--- a/openshift/ci-operator/build-image/cr.yaml
+++ b/openshift/ci-operator/build-image/cr.yaml
@@ -9,7 +9,6 @@ spec:
     elasticsearch:
       nodeCount: 3
       storage:
-        storageClassName: gp2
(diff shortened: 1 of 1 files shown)


openshift/ci-operator/build-image/cr.yaml
```diff
@@ -9,7 +9,7 @@ spec:
     elasticsearch:
       nodeCount: 3
       storage:
-        storageClassName: gp2
+        storageClassName: standard
```
ewolinetz, line 12: if you leave this field off it should just use the OCP cluster default

jcantrill: lets try... removed the class

PR #2171, opened 2021-08-27
Fix multitenancy check of project users to infra indices

periklis:
### Description
The following change set address a hindsight in checking access for project users to infra indices through Kibana. Project users are not allowed to access these indices. The fix changes the test to use the assert denied access against the new `infra*` indices from the old data model `.operations.*` indices.

/cc @jcantrill 
/assign @jcantrill


igor-karpukhin (approved):
/lgtm

PR #2178, opened 2021-09-09
LOG-1199: Allow log level for record to be set from message

jcantrill:
### Description
This PR is the first part to fixing normalization of log level. It
* Allows for a configurable set of patterns to extract the level from the log message
* For backwards compatibility: apply the patterns after first checking if the message is JSON and extracting it from the "level" field
* Adds the missing Gemfile.lock file to source control

This will additionally require a change in CLO collector config generation to consume the changes

/cc @alanconway @vimalk78 

### Links
* https://issues.redhat.com/browse/LOG-1199


fluentd/lib/fluent-plugin-viaq_data_model/lib/fluent/plugin/filter_viaq_data_model.rb
```diff
@@ -181,6 +196,19 @@ def configure(conf)
...
       end
       @undefined_dot_replace_char = nil if @undefined_dot_replace_char == DOT_REPLACE_CHAR_UNUSED
+      
+      if @levels
+        buffer = []
+        @levels.each_with_index do |level,i|
+          if i < 10 
+            buffer.append("?<l#{i}_#{level.name}>#{level.match}")
+          else
+            log.info("Levels supports a maximum of 10 matches. Ignoring: #{level}")
```
alanconway, line 206: Why 10? Why a fixed limit at all?

fluentd/lib/fluent-plugin-viaq_data_model/lib/fluent/plugin/viaq_data_model_log_level_normalizer.rb
```diff
@@ -0,0 +1,98 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+module ViaqDataModel
+# https://github.com/ViaQ/elasticsearch-templates/blob/master/namespaces/_default_.yml#L63
+    NORMAL_LEVELS = {
```
alanconway, line 20: This could be a Set rather than an identity Hash. Not a big deal, it's just surprising to see a Hash with value == key for all entries.

alanconway (approved):
Nits only, feel free to /unhold if you want to carry on.
/hold
/lgtm

fluentd/lib/fluent-plugin-viaq_data_model/lib/fluent/plugin/filter_viaq_data_model.rb
jcantrill: only to make it easier to do the substring of the prefix.  It was a convenient way to not need to figure out the logic

fluentd/lib/fluent-plugin-viaq_data_model/lib/fluent/plugin/viaq_data_model_log_level_normalizer.rb
jcantrill: this is just a carry over from what already existed

vimalk78:
> /cherrypick release-5.2

@jcantrill do we need cherry-pick now? downstream all the components are using same image now

jcantrill:
@vimalk78 we will not need the cherrypick but we do need to solve the consistent multi-tenant test failure (cc @igor-karpukhin @periklis )

/retest

Issue #2181, opened 2021-09-17
Elasticsearch pods CrashLoop on cgroup v2


m-yosefpor:
**Describe the bug**
Elasticsearch pods can not start on systems with cgroup v2 enabled, probably due to this line: https://github.com/openshift/origin-aggregated-logging/blob/master/elasticsearch/run.sh#L56

**Environment**
- Openshift 4.7, cgroupv2 enabled

**Logs**

```
W
[2021-09-17 18:13:19,547][INFO ][container.run            ] Begin Elasticsearch startup script
[2021-09-17 18:13:19,555][INFO ][container.run            ] Comparing the specified RAM to the maximum recommended for
Elasticsearch...
[2021-09-17 18:13:19,558][INFO ][container.run            ] Inspecting the maximum RAM available...
[2021-09-17 18:13:19,561][ERROR][container.run            ] 2021-09-17T18:13:19.562198180Z Unable to determine the maxi
mum allowable RAM for this host in order to configure Elasticsearch
```

**Expected behavior**
The pods should be able to run on cgroupv2.

**Actual behavior**
A clear and concise description of what actually happened.

**To Reproduce**
1. Enable cgroup V2 with MachineConfig on Openshift
2. Create a ClusterLogging Instance
3. The elasticsearch pods will stuck in crashLoopBackOff

**Additional context**
Hierarchy in cgroup v2 is different, and the path for memory limit is at `/sys/fs/cgroup/memory.max`

PR #2182, opened 2021-09-17
fix: memory detection for ES on cgroupv2

m-yosefpor:
### Description

Elasticsearch pods can not start on systems with cgroup v2 enabled, since its hierarchy is different from cgroup v1 and the memory limit file is at `/sys/fs/cgroup/memory.max` instead of `/sys/fs/cgroup/memory/memory.limit_in_bytes`.

For full description of the bug see this issue: https://github.com/openshift/origin-aggregated-logging/issues/2181

/cc @lukas-vlcek @jcantrill  
/assign @jcantrill 

/cherry-pick 

### Links

- Github issue: Closes https://github.com/openshift/origin-aggregated-logging/issues/2181


periklis (commented):
Looks so far good to me.

PR #2229, opened 2022-01-10
LOG-2119: Add required permission to sg_role_curator for delete by query

sasagarw:
### Description

This PR:
- adds `indices:data/write/bulk*` and `indices:data/read/scroll*` permission to `sg_role_curator`

/cc @shwetaap 
/assign @jcantrill 

### Links

- JIRA: https://issues.redhat.com/browse/LOG-2119

### DIFF
--- a/elasticsearch/sgconfig/action_groups.yml
+++ b/elasticsearch/sgconfig/action_groups.yml
@@ -45,6 +45,8 @@ READ:
 
 DELETE:
   - "indices:data/write/delete*"
+  - "indices:data/write/bulk*"
+  - "indices:data/read/scroll*"
 
 CRUD:
   - READ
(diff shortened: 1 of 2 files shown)


elasticsearch/sgconfig/action_groups.yml
```diff
@@ -45,6 +45,8 @@ READ:
 
 DELETE:
   - "indices:data/write/delete*"
+  - "indices:data/write/bulk*"
+  - "indices:data/read/scroll*"
```
Red-GV, line 49: Should these permissions just be given to the specific `sg_curator` role or to anyone adopting this action? I see that `DELETE` is a part `INDEX_KIBANA`.

sasagarw: We should give the permission to this group since delete by query can be run from kibana as well like this https://github.com/openshift/elasticsearch-operator/blob/master/docs/troubleshooting.md#delete-logs-for-specific-namespace

Red-GV: I see. Thanks!

PR #2246, opened 2022-03-16
Bug 2064718: Use 8.2 eus base image

joepvd:
### Description
The curator5 image is currently based on a s2i python image. For 8.2 eus, this causes alerts in the grading. This PR proposes to use the default rhel8 base image as a base instead.

/cc @sosiouxme @periklis 
/assign @jcantrill 

### Links
- Depending on PR(s): (ocp-build-data PR incoming)
- Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=2064718

### DIFF
--- a/curator/Dockerfile
+++ b/curator/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel8.2.els.python.36
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-base-openshift-4.6
 
 MAINTAINER OpenShift Development <dev@lists.openshift.redhat.com>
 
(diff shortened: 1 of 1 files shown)


curator/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel8.2.els.python.36
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-base-openshift-4.6
```
vimalk78, line 1: can't we use `registry.redhat.io/ubi8/python-38:latest` ? then we wont have to install python

joepvd: OCP4.6 is based on RHEL 8.2 EUS. By choosing ubi8/python* we'd deviate from the requirement.

joepvd:
/hold

Maybe the base image should become `openshift-enterprise-base`

PR #2259, opened 2022-07-29
update admin cluster permissions

btaani:
### Description
This PR solves the problem of creating new empty indices after rollover. The cluster permission `indices:admin/template/get` was missing for the user `admin`

### Links
- JIRA: [LOG-2823](https://issues.redhat.com/browse/LOG-2823)


periklis (commented):
/approve

